PR #4, opened 2024-05-31
feat: explain command

a-frantz:
All the functionality changes here are quite straightforward. I added a new dependency, `colored`, in order to get some "prettier" explain output. Maybe there's more we can do in that direction?

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,8 @@
 
 ## 🎨 Features
 
-* **`sprocket lint`.** Lint Workflow Description Language files.
+* **`sprocket lint`** Lint Workflow Description Language files.
+* **`sprocket explain`** Explain lint warnings.
 
 ## Guiding Principles
 
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@
 ## 🎨 Features
 
 * **`sprocket lint`** Lint Workflow Description Language files.
-* **`sprocket explain`** Explain lint warnings.
+* **`sprocket explain`** Explain lint rules.
 
 ## Guiding Principles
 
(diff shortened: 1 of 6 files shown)


src/commands/explain.rs
```diff
@@ -0,0 +1,45 @@
...
+pub struct Args {
+    /// The rule name to explain.
+    #[arg(required = true)]
+    pub rule_name: String,
+}
+
+pub fn explain(args: Args) -> anyhow::Result<()> {
+    let name = args.rule_name;
+
+    let rules = grammar_lint::rules()
```
adthrasher: It seems odd to me that we're iterating over both lists twice. Can this be streamlined?

peterhuene (approved):

adthrasher (approved):
Looks good, though I'm hoping Peter has a clever solution to avoid the two `match` statements.

peterhuene:
I think we can live with the two `match` statements for now, as I don't think we can chain the two iterators together since they're different concrete types due to the `E` type parameter.

The solution will be to have a single list of lint rules moving forward in the new implementation, where everything is based off the AST.

I'm hoping that we might even put them in their own crate, such as `wdl-lint`.

README.md
```diff
@@ -34,7 +34,8 @@
 
 ## 🎨 Features
 
-* **`sprocket lint`.** Lint Workflow Description Language files.
+* **`sprocket lint`** Lint Workflow Description Language files.
+* **`sprocket explain`** Explain lint warnings.
```
adthrasher: Should this be `rules` instead of `warnings`?

a-frantz: Nice catch, you are correct

PR #6, opened 2024-06-18
[WIP] upgrade to new wdl parser

a-frantz:

### DIFF
--- a/src/commands/explain.rs
+++ b/src/commands/explain.rs
@@ -1,49 +1,45 @@
 use clap::Parser;
 use colored::Colorize;
-use wdl::ast::v1::lint as ast_lint;
-use wdl::core::concern::lint::Rule;
-use wdl::grammar::v1::lint as grammar_lint;
+use wdl::lint;
 
 /// Arguments for the `explain` subcommand.
 #[derive(Parser, Debug)]
 #[command(author, version, about)]
 pub struct Args {
-    /// The name or code of the rule to explain.
+    /// The name of the rule to explain.
     #[arg(required = true)]
     pub rule_identifier: String,
 }
 
-pub fn pretty_print_rule<E>(rule: &dyn Rule<E>) {
-    println!("{}", rule.name().bold().underline());
-    println!("{}", format!("{}::{}", rule.code(), rule.tags(),).yellow());
+pub fn list_all_rules() {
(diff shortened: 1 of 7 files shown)


src/commands/explain.rs
```diff
@@ -1,49 +1,49 @@
...
 
 pub fn explain(args: Args) -> anyhow::Result<()> {
-    let ident = args.rule_identifier;
+    let name = args.rule_name;
 
-    let rule = grammar_lint::rules()
+    let rule = lint::v1::rules()
         .into_iter()
-        .find(|rule| rule.name() == ident || rule.code().to_string() == ident);
+        .find(|rule| rule.id().to_lowercase() == name.to_lowercase());
```
peterhuene: nit: can we move `name.to_lowercase()` to outside of the `find`?

src/file.rs
```diff
@@ -6,19 +6,12 @@ use codespan_reporting::files::SimpleFiles;
 use codespan_reporting::term::termcolor::StandardStream;
 use codespan_reporting::term::Config;
 use indexmap::IndexMap;
-use wdl::core::Concern;
 
 use crate::report::Reporter;
 
 /// A filesystem error.
 #[derive(Debug)]
 pub enum Error {
```
peterhuene, line 14: Not related to this PR, but for the future I think we can eliminate error enumerations in CLI tools since we're using `anyhow` anyway; the `anyhow::Context` trait gives us what we need to provide context on errors and that should make the code in the CLI tool a little cleaner.

This can be addressed in a future PR, though.

src/report.rs
```diff
@@ -1,130 +1,45 @@
...
-                    diagnostic = diagnostic.with_notes(vec![format!("fix: {}", fix)]);
-                }
-
-                diagnostic
-            }
-        };
+    /// Reports diagnostics to the terminal.
+    pub(crate) fn emit_diagnostics(
+        &mut self,
+        file: SimpleFile<String, String>,
```
peterhuene: Can this be by reference since we're just passing it to `emit`? Cloning `SimpleFile` might be expensive as it contains the source that was parsed.

```suggestion
        file: &SimpleFile<String, String>,
```

peterhuene (commented):
Looks great, just a few comments!

peterhuene (approved):

PR #14, opened 2024-08-12
docs: adds the initial documentation site

claymcleod:
I wanted to add a few sentences describing the philosophy and goals that led to the creation of Sprocket. Though everything on the page may not be true (yet) today, it will be a good goal-post to work towards while providing the community some sense of where we are going.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,14 +44,14 @@
 * Provide a **high-performance** workflow execution engine capable of
   orchestrating massive bioinformatics workloads (the stated target is 20,000+
   concurrent jobs).
-* Remain **simple and accessible user experience** when complexity isn't
-  warranted.
 * Develop a suite of **modern development tools** that brings bioinformatics
   development on par with other modern languages.
 * Maintain an **community-focused codebase** that enables a diverse set of
   contributors from academic, non-profit, and commercial organizations.
 * Build on an **open, domain-tailored standard** to ensure the toolset remains
   singularly focused on unencumbered innovation within bioinformatics.
+* Retain a **simple and accessible user experience** when complexity isn't
+  warranted.
 
 ## 📚 Getting Started
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
   </p>
 
   <p align="center">
-    A package manager for Workflow Description Language files.
+    A bioinformatics workflow orchestration engine and package manager built on top of the Workflow Description Language (WDL).
     <br />
     <br />
     <a href="https://github.com/stjude-rust-labs/sprocket/issues/new?assignees=&title=Descriptive%20Title&labels=enhancement">Request Feature</a>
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -41,18 +41,24 @@
 
 ## Guiding Principles
 
-* **Modern, reliable foundation for everyday bioinformatics analysis—written in Rust.** `sprocket` aims to package together a fairly comprehensive set of tools and for developing bioinformatics tasks and workflows using the [Workflow Description Language](http://openwdl.org/). It is built with modern, multi-core systems in mind and written in Rust.
-* **WDL specification focused.** We aim to implement the various versions of the [OpenWDL specification](https://github.com/openwdl/wdl) to the letter. In other words, `sprocket` aims to be workflow engine independent. In the future, we plan to make `sprocket` extendable for workflow engine teams.
+* Provide a **high-performance** workflow execution engine capable of
+  orchestrating massive bioinformatics workloads (the stated target is 20,000+
```
a-frantz, line 45: Just thoughts/questions that don't necessarily need to be answered _in_ the document:

Where does the number 20,000 come from? And how is a "job" defined? Asking because I noticed in `cromwell server` that a limiting factor of performance wasn't "number of concurrent workflows" but "number of concurrent tasks". e.g. our `rnaseq-standard` workflow is relatively linear, where QC is highly parallel. Cromwell could handle many more `rnaseq-standard` jobs than it could QC jobs.

docs/overview.md
```diff
@@ -0,0 +1,135 @@
...
+  development of the project, backends are intended to be developed and
+  maintained by the backend providers themselves. When that isn't possible,
+  custom backends should be able to be configured by end-users with a moderate
+  amount of effort.
+
+### Non-goals
+
+The following are non-goals of the project.
+
+* **Supporting multiple workflow languages or standards.** Ironically, workflow
```
a-frantz:
```suggestion
* **Supporting multiple workflow languages or standards.** Workflow
```

docs/overview.md
```diff
@@ -0,0 +1,135 @@
...
+  languages are ideally a simple means to an end: to run large-scale
+  bioinformatics analyses with as little development and operational friction as
+  possible. We stand unconvinced that there is a strong technical argument for
+  multiple workflow languages existing within bioinformatics.
```
a-frantz, line 133: I think this begs the question, "if there should only be one workflow language, **why WDL?**"

Do we need a paragraph justifying our backing of WDL over Nextflow, CWL, etc?

README.md
claymcleod: Good question. To me it means "workflows in progress". So that is, perhaps, even more ambitious than tasks. Note that a workflow that is "pending" still counts in my book (it's really just the practice of 'submit and forget' and you're confident that these will eventually run to completion).

The 20,000 number came from my head. I'd like to pick a concrete number to shoot for, and this seemed like a high number that was also in the realm of achievable. For now, I'd like to stick with that.

docs/overview.md
claymcleod: At least for me, when combining the pieces about who our target audience is and the need to develop based on an open, community-driven standard, the choice is clear (though it might could just be communicated succinctly this way).

docs/overview.md
```diff
@@ -0,0 +1,135 @@
...
+([source](https://github.com/stjude-rust-labs/sprocket-vscode)).
+
+## Project Goals
+
+### High-Performance Orchestration Engine
+
+Most fundamentally, the Sprocket project was created due to the lack of workflow
+engines that were powerful enough to orchestrate tens of thousands or hundreds
+of thousands of concurrent jobs. Enabling the next generation of large-scale,
+open science will always remain the top-level objective for the project.
```
adthrasher: I'd prefer that we give this some more thought before putting it out there. What exactly does this look like? There are several execution engines (Cromwell, Miniwdl, dx-compiler, AWS HealthOmics). These have a variety of different target environments and I think it'd be best to define what this looks like and what we want to support.

docs/overview.md
```diff
@@ -0,0 +1,135 @@
...
+means **bioinformaticians** and **bioinformatics software engineers** working at
+a moderate to large scale, but Sprocket should work reasonably well for many
+other users too. In particular, we aim to keep Sprocket as simple as we possibly
+can—only introducing complexity when needed to achieve the project's other
+stated goals.
+
+### Modern Development Tools
+
+A suite of development tools are included alongside the execution engine and
+package manager. We believe that, when you _do_ need to dust off your code
```
adthrasher: I think the `overview` is missing a description of the package manager component.

docs/overview.md
claymcleod: This was discussed extensively in Slack and I've updated the text here to
reflect that discussion.

docs/overview.md
claymcleod: I put a sentence here about WDL. I think it would be good to add a future page regarding the details of why we chose WDL.

docs/overview.md
```diff
@@ -33,29 +32,41 @@ code that drives Sprocket is split across the [`wdl`] family of crates, the
...
+### High-Performance Workflow Execution Engine
 
-Most fundamentally, the Sprocket project was created due to the lack of workflow
-engines that were powerful enough to orchestrate tens of thousands or hundreds
-of thousands of concurrent jobs. Enabling the next generation of large-scale,
-open science will always remain the top-level objective for the project.
+Fundamentally, the Sprocket project was created to address the lack of
+bioinformatics workflow execution engines that can reliably handle tens of
+thousands to hundreds of thousands of concurrently executing workflows across a
+variety of execution environments.
```
adthrasher: I'm not sure we necessarily need an answer to this, but I would point out that I'm not sure we have an infrastructure capable of running 10k - 100k workflows concurrently. Maybe if we were able to leverage multiple backends.

claymcleod: For the public record, this was my response to Andrew's thought during an internal conversation.

> For me, when I say "handling tens of thousands to hundreds of thousands of concurrently executing workflows", I don't necessarily mean that every workflow is executing at every time point. Instead, I'm referring to the experience where one can "submit and forget" hundreds of thousands of workflows to an orchestrator, and those workflows will execute as quickly as possible (similar to common HPC schedulers today). 
>
> In terms of the concurrent execution of workflows within a particular runtime, that's going to be dependent on a ton of factors, and we'll just have to gather metrics to see what is actually possible.

adthrasher (approved):

PR #24, opened 2024-10-16
feat: add the `sprocket format` command.

peterhuene:
This commit adds the `format` command to `sprocket`.

The command will format a single WDL document.

Also updated the existing commands to the latest WDL API and corrected the environment variable name for publishing crates.

Included the analysis rules in `sprocket explain`; the list of rules is now sorted by id.

### DIFF
--- a/src/commands/format.rs
+++ b/src/commands/format.rs
@@ -1,5 +1,4 @@
 //! Implementation of the format command.
-//! Implementation of the analyzer command.
 
 use std::fs;
 use std::io::Read;
--- a/src/commands/format.rs
+++ b/src/commands/format.rs
@@ -1,5 +1,6 @@
 //! Implementation of the format command.
 
+use std::ffi::OsStr;
 use std::fs;
 use std::io::Read;
 use std::num::NonZeroUsize;
@@ -13,8 +14,11 @@ use anyhow::bail;
 use clap::Parser;
 use codespan_reporting::files::SimpleFile;
 use codespan_reporting::term::emit;
+use colored::Colorize;
+use walkdir::WalkDir;
 use wdl::ast::Document;
 use wdl::ast::Node;
+use wdl::format::Config;
 use wdl::format::Formatter;
 use wdl::format::config::Builder;
 use wdl::format::config::Indent;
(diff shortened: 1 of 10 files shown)


peterhuene:
Once the `wdl` crate publishes, I'll update this to reference it from crates.io.

src/commands/format.rs
```diff
@@ -0,0 +1,112 @@
+//! Implementation of the format command.
+//! Implementation of the analyzer command.
```
```suggestion
```

src/commands/format.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// Use tabs instead of spaces for indentation.
+    #[arg(long)]
+    pub with_tabs: bool,
+
+    /// The number of spaces that represents an indentation level.
+    #[arg(
+        long,
+        value_name = "SIZE",
+        default_value = "4",
+        conflicts_with = "with_tabs"
```
a-frantz: I'm fine with this, but if we're drawing this line in the sand (that when using tabs, only one should be used for each "level"), we should probably update the `wdl-format` `Config` to match.

a-frantz: We should also consider capping the number of spaces possible. Imagine someone supplied `MAX_INT` here. Nothing bad would happen but the file would be garbage. Maybe cap at `12`? `16`?

a-frantz (commented):
This is different from how I imagined it to be implemented. I think we want it to replace the formatted file, and to work on an entire directory, like how `cargo fmt` works, right? Or is your thinking that it is a potentially destructive and unexpected behavior? Maybe we need to gate it behind an `press Y to continue` interface?

src/commands/format.rs
a-frantz: Tangent, I've had a similar thought about how `max_line_width` could be set. There should be a minimum for that, but IDK what the magic number is. `60`? `40`? Why would you ever want a file narrower than that?

peterhuene: It'd be weird to me to indent with multiple tabs, but can't hurt to be more flexible. I'll make the change along with capping it at a reasonable value.

a-frantz:
> I'll update it to replace the file; perhaps add a `--print` option to print it to stdout instead?

I was looking forward to running `sprocket format .` from the root of the `workflows` repo and timing it 😅 but I can see a new user running a similar command (without source control) and being really pissed at us for reformatting every WDL file. I'm conflicted on the right approach.

peterhuene:
How about we default to printing, which can only be done with a single file (it errors if you pass it a path that's a directory).

We add a `--overwrite` option that overwrites the file instead of printing it; if that option is supplied, we allow it to operate on a path that is a directory.

a-frantz:
> How about we default to printing, which can only be done with a single file (it errors if you pass it a path that's a directory).
> 
> We add a `--overwrite` option that overwrites the file instead of printing it; if that option is supplied, we allow it to operate on a path that is a directory.

Yea I like that. I think that's reasonable.

src/main.rs
```diff
@@ -17,7 +17,7 @@ git_testament!(TESTAMENT);
 #[derive(Subcommand)]
 #[allow(clippy::large_enum_variant)]
 enum Commands {
-    /// Checks a WDL file (or a directory containing WDL files) and reports
+    /// Checks a WDL document (or a directory containing WDL files) and reports
```
peterhuene:
```suggestion
    /// Checks a WDL document (or a directory containing WDL documents) and reports
```

src/commands/format.rs
```diff
@@ -0,0 +1,217 @@
...
+/// The maximum acceptable indentation size.
+const MAX_INDENT_SIZE: usize = 16;
+
+/// The default number of tabs to use for indentation.
+const DEFAULT_TAB_INDENT_SIZE: usize = 1;
+
+/// The default number of spaces to use for indentation.
+const DEFAULT_SPACE_IDENT_SIZE: usize = 4;
+
+/// Arguments for the `analyzer` subcommand.
```
```suggestion
/// Arguments for the `format` subcommand.
```

a-frantz (approved):

src/commands/format.rs
peterhuene: Copy and paste, I trusted you!

PR #46, opened 2024-12-17
chore: update sprocket to latest WDL crate

### LINKED ISSUE #39: option to only print diagnostics from files specified at CL
Doing some development on `workflows` and using sprocket. It's annoying when I run `sprocket lint <small workflow file>.wdl` and am faced with a huge number of printed diagnostics from files imported by my workflow. In this situation, I only care about diagnostics originating from `<small workflow file>.wdl`.

### LINKED ISSUE #45: feat: limit the printed diagnostics when there are many
The problem: for a large repo, running `sprocket lint` can produce an overwhelmingly large number of diagnostics. Particularly when the default "full diagnostic" view is being used.

Two potential solutions that aren't mutually exclusive:
1. If there are more than `n` diagnostics to print, dynamically switch to the one-line view (with a new flag for `--always-full` or something)
2. only print the first `m` diagnostics, and if there are more `bail!` with a message to the effect of `failing due to X number of additional diagnostics` (and again, a new CL param that can be passed like `--show-all`)

Note that `n` and `m` are probably going to have different "sweet spots" for this.

a-frantz:
This also adds the ability to check/lint URLs

closes #39 
sort of closes #45 by providing `--local-only` and `--single-document` args which narrow the scope of printed diagnostics. It was decided over slack not to pursue either of the 2 fixes presented in #45 as they were likely undesirable UXs.


CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+* Updated WDL crate to latest (by pointing to `wdl:main`
+  as the latest crates haven't been released yet).
```
adthrasher: Should this be the default for `sprocket:main` and the release process for `sprocket` will require pinning the `wdl` crate versions?

a-frantz: That's a good question. I think the our current procedure is to really only touch this repo directly before/after/during a `wdl` crate release, so we've always had `sprocket:main` point to official releases. As the `sprocket` binary and `wdl` crates mature, we may want to have a more synchronous development flow.

So I'm not sure. @peterhuene I'm curious if you have any thoughts on this Q.

a-frantz: This was addressed by @claymcleod over Slack. The decision made was that `sprocket:main` should always point to official `wdl` releases.

a-frantz:
This PR was put back into draft state and will remain as a draft until the `wdl` crates make an official release. In the meantime, I'll use this branch as the basis for the development of the other sprocket PRs I have planned.

This PR will have to be reviewed+merged before those other ones.

CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+* Updated WDL crate to latest. This adds support for
```
adthrasher, line 12: Didn't Clay want to keep `sprocket` pinned to a release?

src/commands/check.rs
```diff
@@ -79,19 +105,19 @@ pub struct LintArgs {
 
 /// Checks WDL source files for diagnostics.
 pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
-    if !args.lint && !args.common.except.is_empty() {
-        bail!("cannot specify `--except` without `--lint`");
+    if args.common.shellcheck && !args.lint {
+        bail!("`--shellcheck` requires `--lint` to be enabled");
```
adthrasher, line 98: Is this the experience we want or should `--shellcheck` add `--lint` automatically (with a warning)?

src/commands/check.rs
```diff
@@ -182,17 +257,17 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
     } else if args.common.deny_warnings && warning_count > 0 {
         bail!(
-            "aborting due to previous {warning_count} warning{s} (`--deny-warnings` was specified)",
+            "failing due to {warning_count} warning{s} (`--deny-warnings` was specified)",
             s = if warning_count == 1 { "" } else { "s" }
         );
     } else if args.common.deny_notes && note_count > 0 {
         bail!(
-            "aborting due to previous {note_count} note{s} (`--deny-notes` was specified)",
+            "failing due to {note_count} note{s} (`--deny-notes` was specified)",
```
adthrasher, line 204: I'm not sure about this language change. I'm not sure the new proposed language makes it clear that `sprocket` is stopping because it has encountered too many errors/notes/warning.

CHANGELOG.md
a-frantz: Yes, but that release doesn't exist yet. This PR will be updated to pull an official release once one exists.

I think the language here is fine? Maybe it should specify the version of "latest" once that is out?

src/commands/check.rs
a-frantz: I could go either way I suppose. I don't think I feel strongly on this either way. Could it confuse users if they specify `--shellcheck` without `--lint` and then they see non-shellcheck lint diagnostics?

src/commands/check.rs
a-frantz: I think that this language is an improvement, but maybe it can be improved further. I don't like the `aborting` language because in my mind it implies that `sprocket` was going to do something else but it's quitting early. This really just controls whether it's a non-zero exit code or not.

CHANGELOG.md
adthrasher: OK, I would swap `latest` with the version when it is known.

src/commands/check.rs
adthrasher: Can you clarify the behavior here? `sprocket` is computing the full set of diagnostics, reporting all of them, and then setting the exit code if any of the requested subtypes has content?

src/commands/check.rs
```diff
@@ -29,12 +31,14 @@ const PROGRESS_BAR_DELAY: Duration = Duration::from_secs(2);
 #[derive(Parser, Debug)]
 #[command(author, version, about)]
 pub struct Common {
-    /// The files or directories to check.
+    /// The file, URL, or directory to check.
     #[arg(required = true)]
-    pub paths: Vec<PathBuf>,
+    #[clap(value_name = "PATH or URL")]
+    pub file: String,
```
a-frantz, line 26: A potential nit @adthrasher : is `file` the right name for this arg? I couldn't think of a better name...

src/commands/check.rs
a-frantz: You are correct. The `bail!` macro exits the program with a non-zero exit code (IDK what code specifically) and this block of checks is the last thing `sprocket` is doing in `check/lint` mode. So there is no further computation that _might_ occur if there aren't problems detected. This is the end no matter what.

src/commands/check.rs
adthrasher: I considered commenting on that, but couldn't come up with an alternative I preferred. You could verbosely call it `path_or_url`.

src/commands/check.rs
adthrasher: OK, I initially thought this was related to limiting the number of diagnostics emitted to the first `N`. I know that was discussed, but can't recall what the decision was.

Now that I understand it, I agree that the language is an improvement.

src/commands/check.rs
```diff
@@ -182,17 +257,17 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
 
     if error_count > 0 {
         bail!(
-            "aborting due to previous {error_count} error{s}",
+            "failing due to {error_count} error{s}",
```
adthrasher, line 194:
```suggestion
            "failing with {error_count} error{s}",
```
How do you feel about `with` instead? I'm not certain that's an improvement.

src/commands/check.rs
a-frantz: Eh. I think I prefer just plain `path` to `path_or_url`, but don't love either. I'll leave this comment unresolved in case someone else reviewing has a better alternative.

src/commands/check.rs
a-frantz: We did discuss that over slack, but it was decided that limiting to the first `N` would likely lead to a bad UX. So we always report the full set of diagnostics, with `--local-only` and `--single-document` controlling the scope of what's reported (but everything is always computed regardless)

adthrasher: Thanks for the refresher, the after-effects of the holiday break are quite apparent today.

src/commands/check.rs
a-frantz: Hm. Kinda feels like half-dozen of one, 6 of the other to me. I don't see how it clarifies anything, but it doesn't seem worse in any way. I can make that change if you want. I'm fine with it.

a-frantz:
Closing in favor of having one PR (#48 )

PR #47, opened 2024-12-18
feat: run subcommand

a-frantz:


src/commands/run.rs
```diff
@@ -0,0 +1,263 @@
...
+
+use crate::Mode;
+use crate::analyze;
+use crate::get_display_config;
+
+/// Arguments for the run command.
+#[derive(Parser, Debug)]
+#[command(author, version, about)]
+pub struct RunArgs {
+    /// The path or URL to the WDL document containing the task to run.
```
adthrasher, line 27: Do we want to write these as `task`? Eventually, it will support workflows, so should the comments reflect that or just the current functionality?

src/commands/run.rs
```diff
@@ -0,0 +1,263 @@
...
+    ///
+    /// Required if no `inputs` file is provided.
+    #[arg(short, long, value_name = "NAME", conflicts_with = "inputs")]
+    pub name: Option<String>,
+
+    /// The output directory; defaults to the task name.
+    ///
+    /// If no output directory is provided, a default nested directory is
+    /// created based on the task name and the current time in the form
+    /// `sprocket_runs/<task_name>/<timestamp>/`.
```
adthrasher:
```suggestion
    /// `sprocket_runs/<execution_name>/<timestamp>/`.
```
I think this one should definitely be written as the more generic executable, since we're establishing a directory structure and it will need to support task and workflow execution.

src/commands/run.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    Ok(output_dir)
+}
+
+/// Runs a task.
+pub async fn run(args: RunArgs) -> Result<()> {
+    eprintln!(
+        "the `run` command is in alpha testing and does not currently support workflows or using \
+         containers."
+    );
```
adthrasher, line 111: With the warning being printed, I would suggest we use something like `workload`, `executable`, etc. instead of `task` to represent that this will eventually support larger executions.

src/commands/run.rs
```diff
@@ -0,0 +1,263 @@
...
+        path_to_uri(&file).expect("file should be a local path")
+    };
+
+    let result = results
+        .iter()
+        .find(|r| **r.document().uri() == uri)
+        .context("failed to find document in analysis results")?;
+    let document = result.document();
+
+    let mut engine = Engine::new(LocalTaskExecutionBackend::new());
```
adthrasher: Do we have a concept for how the execution backend will be chosen in the future?

src/commands/run.rs
```diff
@@ -0,0 +1,263 @@
...
+            format!(
+                "failed to determine the absolute path of `{path}`",
+                path = path.display()
+            )
+        })?;
+        match Inputs::parse(document, &abs_path)? {
+            Some((name, inputs)) => (Some(path), name, inputs),
+            None => bail!(
+                "inputs file `{path}` is empty; use the `--name` option to specify the name of \
+                 the task or workflow to run",
```
adthrasher: SInce there can be only one `workflow` in a file, is it safe to make an assumption to run that if `name` isn't specified? That's what the other runners do correct?

src/commands/run.rs
```diff
@@ -0,0 +1,263 @@
...
+                path = path.display()
+            ),
+        }
+    } else if let Some(name) = args.name {
+        if document.task_by_name(&name).is_some() {
+            (None, name, Inputs::Task(Default::default()))
+        } else if document.workflow().is_some() {
```
adthrasher: This doesn't actually check that `workflow` matches `name` does it? You could specify any `name` value and it would run whatever `workflow` was in the file.

src/commands/run.rs
```diff
@@ -0,0 +1,263 @@
...
+    };
+
+    let output_dir = create_output_dir(args.output, &name, args.overwrite)?;
+
+    match inputs {
+        Inputs::Task(mut inputs) => {
+            // Make any paths specified in the inputs absolute
+            let task = document
+                .task_by_name(&name)
+                .ok_or_else(|| anyhow!("document does not contain a task named `{name}`"))?;
```
adthrasher: This comment doesn't match the code

src/commands/run.rs
adthrasher: It does get checked below though. L248-250

src/commands/run.rs
a-frantz: I think we should reflect the current functionality, i.e. specify that we only look at tasks for now. I wouldn't be opposed to making the language more agnostic as a future-proof, so long as it doesn't imply anything we don't currently support.

src/commands/run.rs
a-frantz: Agree, that's a good catch.

src/commands/run.rs
a-frantz: Ah, you mean in the doc comment? I was confused for a second there. Yea that makes sense to me, can update the comment.

src/commands/run.rs
a-frantz: I don't have much of a vision for that yet, but I imagine Clay and Peter have thoughts. I think it's safe to assume that without advanced configuration/options, we will default to the local backend.

src/commands/run.rs
a-frantz: Right now, `wdl-engine` expects the name of the task _or_ workflow to be specified in the JSON (in the form `"<name of task or workflow>.<name of parameter>": <value>`). I played around with trying to loosen that requirement within the `sprocket` bin, but I think it will require some tweaks upstream.

So if no input JSON is supplied, we currently rely on `--name` to select what task/workflow to run. I suppose it would be fine if both are omitted to run any found workflow by default. I can make that change.

a-frantz: However, I don't think it's really a common use-case to omit the input JSON, so I'm not certain how important this is. The input JSON must specify the name of the task or workflow, so supplying `--name` is redundant or leads to contradictions, so isn't allowed at the same time. There aren't many tasks or workflows "in the wild" that you can run without specifying any parameters (which currently must be done via JSON).

src/commands/run.rs
a-frantz: Actually, on second thought I think this should reflect the current capabilities. I don't think this comment should change. Same reasoning as explained higher up in another thread.

src/commands/run.rs
a-frantz: See other thread.

src/commands/run.rs
a-frantz: To clarify: the only time a user can or should be using the `--name` parameter is if they want to run a task (or workflow) without any arguments. This is a pretty niche use case.

However in a future update, we will probably want to allow CL specified parameters, and when we add that we will want to use this `--name` flag to enable that behavior.

I think it's fine to require specificity from users and supply the workflow name if they want to use this feature.

a-frantz:
Closing in favor of having one PR (#48 )

PR #48, opened 2024-12-19
feat: update to latest wdl crate

### LINKED ISSUE #39: option to only print diagnostics from files specified at CL

### LINKED ISSUE #45: feat: limit the printed diagnostics when there are many

a-frantz:
This adds the ability to check/lint URLs, the `validate-inputs` subcommand, and various other features/fixes/improvements.

closes #39 
sort of closes #45 by providing `--local-only` and `--single-document` args which narrow the scope of printed diagnostics. It was decided over slack not to pursue either of the 2 fixes presented in #45 as they were likely undesirable UXs.


src/commands/validate.rs
```diff
@@ -0,0 +1,126 @@
...
+    #[arg(required = true)]
+    #[clap(value_name = "PATH or URL")]
+    pub document: String,
+
+    /// The path to the input JSON file.
+    #[arg(short, long, value_name = "JSON")]
+    pub inputs: PathBuf,
+}
+
+/// Validates the inputs for a task or workflow.
```
adthrasher: I think this should more precisely state what is being checked.

.gitignore
```diff
@@ -4,4 +4,6 @@ target/
 .vscode/
 **/Cargo.lock
 
-rustc-ice*
\ No newline at end of file
+rustc-ice*
+
+sprocket_run*/
```
claymcleod: My personal opinion is that the directory should just be called `runs/`.

claymcleod (approved):
Otherwise looks pretty good to me. Keep in mind that this is sort of a temporary release, so I'm not thinking long-term in terms of the user experience. In other words, some of this UX I will advocate to change in the future.

.gitignore
adthrasher: I was about to ping this in Slack, but I don't think the discussion on this was a thread that included Clay.

a-frantz: @claymcleod me, Andrew, and Peter discussed over slack and decided the run directory structure should be `./sprocket_runs/<workflow or task name>/<timestamp>/`.

However as we're not exposing the `run` command in this release, I've removed this line from the `.gitignore`

claymcleod: Okay, I'm fine with that for now.

peterhuene (approved):

Issue #29, opened 2024-10-17
feature: add a `--check` option to `sprocket format`


peterhuene:
This option would be useful in CI where it would:

* Format a WDL document or directory containing WDL documents
* Instead of overwriting the files, it would diff the format output with the original output and treat any difference as an error; it should print out a nice git-like diff of the changes (the `pretty-assertions` crate has a very nice implementation of this).
* It should exit with a non-zero status if any file had a diff.

a-frantz:
How much of the logic for `--check` should be in this repo? I think that if the answer is "all of it", this is going to be the first case of "business logic" living in the sprocket repo. Not exactly a problem, but something that should be done intentionally.

I wonder if this would make more sense to live in one of the `wdl-` crates (though which one isn't obvious to me).

peterhuene:
I think it'd be fine to expose the "check" functionality from `wdl-format`.

a-frantz:
I've flipflopped. The "check" functionality isn't really anything other than `input == formatted`. I think it's fine to house that logic within `sprocket` just to keep things simple. Consider my comment revoked.

PR #49, opened 2025-01-08
feat: add check option to format command

Scott8440:
* Adds `--check` option to `format` command. 
* If a file is not formatted, prints out the resulting diff - from `pretty_assertions` [crate](https://crates.io/crates/pretty_assertions)
* Exits with non-zero exit code if any file had a diff
* Addresses #29 

Test:
* `cargo build --release` succeeds
* `cargo run --release format --check ~/wdl/wdl-format/tests/format/ ` successfully checks all example files
* Sample output:
image


src/commands/format.rs
```diff
@@ -172,7 +188,7 @@ pub fn format(args: FormatArgs) -> Result<()> {
 
     let mut diagnostics = 0;
     if args.path.to_str() != Some("-") && args.path.is_dir() {
-        if !args.overwrite {
+        if !args.overwrite && !args.check {
             bail!("formatting a directory requires the `--overwrite` option");
```
```suggestion
            bail!("formatting a directory requires the `--overwrite` or `--check` option");
```

src/commands/format.rs
```diff
@@ -140,6 +147,15 @@ fn format_document(
     let formatter = Formatter::new(config);
     let formatted = formatter.format(&document)?;
 
+    if check {
+        if formatted != source {
+            print!("{}", StrComparison::new(&source, &formatted));
+            return Ok(1);
+        }
+        println!("ok");
```
a-frantz: Can you update this print statement to include the path of what's "ok"? `{path} is formatted correctly`

src/commands/format.rs
```diff
@@ -72,6 +73,10 @@ pub struct FormatArgs {
     /// Overwrite the WDL documents with the formatted versions
     #[arg(long)]
     pub overwrite: bool,
+
+    /// Check if files are formatted correctly and print diff if not
+    #[arg(long)]
```
a-frantz: `--check` needs to be mutually exclusive with `--overwrite`

a-frantz (changes requested):
Hi @Scott8440 ! Thanks for this contribution! Looks pretty good, but requesting some small tweaks.

Also, we prefer folks to comment on issues to "claim" them before submitting a PR. This is to prevent two people from working on the same issue at the same time, but fortunately no one was working on this one; no harm, no foul :)

claymcleod:
@a-frantz it also begs the question: "what is the intended behavior of `sprocket fmt` without `--check` or `--overwrite`"? I believe `cargo fmt` assumes you want to overwrite things and only reverts to checking with the `--check` flag.

a-frantz:
> @a-frantz it also begs the question: "what is the intended behavior of `sprocket fmt` without `--check` or `--overwrite`"? I believe `cargo fmt` assumes you want to overwrite things and only reverts to checking with the `--check` flag.

Yea, we considered that behavior back when first introducing the feature. We thought that it was unwise to perform something potentially destructive by default, which is why the default behavior is to just print to STDOUT. Granted, that means the default behavior isn't very useful, but it also won't anger any users.

claymcleod:
I see. Perhaps in the long run it's worth doing what `cargo fmt` does, which I believe is check if the files are checked into `git` before formatting them.

adthrasher:
> > @a-frantz it also begs the question: "what is the intended behavior of `sprocket fmt` without `--check` or `--overwrite`"? I believe `cargo fmt` assumes you want to overwrite things and only reverts to checking with the `--check` flag.
> 
> Yea, we considered that behavior back when first introducing the feature. We thought that it was unwise to perform something potentially destructive by default, which is why the default behavior is to just print to STDOUT. Granted, that means the default behavior isn't very useful, but it also won't anger any users.

I'd be OK with implementing it such that a user is required to specify either `--check` or `--overwrite`. Omitting both options would produce an error and the help text. To me, that's the safest option, we simply force the users to make the choice. Otherwise, I'd suggest the `--check` behavior should be the default.

a-frantz:
> I'd be OK with implementing it such that a user is required to specify either `--check` or `--overwrite`. Omitting both options would produce an error and the help text. To me, that's the safest option, we simply force the users to make the choice.

I like this approach. However this conversation is a bit of a tangent from the PR we're having it on. I'll open an issue for tracking and further discussion.

Scott8440:
Thanks for the feedback. I've addressed the comments.

a-frantz (approved):
LGTM! @Scott8440 we generally require two approvals from core team members before merging. So this has my approval, but we won't merge till someone else gets around to the review. Should be sometime today!

This should be included in the next sprocket release we're (hopefully) pushing out later today!

adthrasher (approved):

PR #51, opened 2025-01-10
feat: require one of check or overwrite options in format command

Scott8440:
Requires that one of `--overwrite` or `--check` be provided and removes the default behavior of just printing the formatted files to stdout.

The `--overwrite` flag is now just a safety check so the caller knows that what they are doing is destructive, as writing the file is now "default" behavior, with `--check` being the modifier.

Test:
```
❯ cargo run --release format --check --overwrite ~/wdl/wdl-format/tests/format/optional_structs/
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/sprocket format --check --overwrite /home/scott/wdl/wdl-format/tests/format/optional_structs/`
error: the argument '--check' cannot be used with '--overwrite'

Usage: sprocket format <--overwrite|--check> <PATH>

For more information, try '--help'.
```

```
❯ cargo run --release format ~/wdl/wdl-format/tests/format/optional_structs/ 
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/sprocket format /home/scott/wdl/wdl-format/tests/format/optional_structs/`
error: the following required arguments were not provided:
  <--overwrite|--check>

Usage: sprocket format <--overwrite|--check> <PATH>

For more information, try '--help'.
```

### DIFF
--- a/src/commands/format.rs
+++ b/src/commands/format.rs
@@ -43,9 +43,10 @@ const DEFAULT_SPACE_IDENT_SIZE: usize = 4;
     author,
     version,
     about,
-    after_help = "Use the `--overwrite` option to replace a WDL document, or a directory \
-                  containing WDL documents, with the formatted source. Use the `--check` option \
-                  to print the diff or none if the source is already formatted."
+    after_help = "Use the `--overwrite` option to replace a WDL document or a directory \
+                  containing WDL documents with the formatted source.\nUse the `--check` option \
+                  to verify that a document or a directory containing WDL documents is already formatted \
+                  and print the diff if not."
 )]
 pub struct FormatArgs {
     /// The path to the WDL document to format (`-` for STDIN); the path may be
@@ -107,8 +108,8 @@ fn read_source(path: &Path) -> Result<String> {
 
 /// Formats a document.
 ///
-/// Checks if the document is formatted correctly and prints the diff if not
-/// if check_only is true.
+/// If `check_only` is true, checks if the document is formatted correctly and prints the diff if not
+/// then exits. Else will format and overwrite the document.
 ///
 /// If the document failed to parse, this emits the diagnostics and returns
 /// `Ok(count)` of the diagnostics to the caller.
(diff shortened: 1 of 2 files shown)


src/commands/format.rs
```diff
@@ -188,7 +196,7 @@ pub fn format(args: FormatArgs) -> Result<()> {
 
     let mut diagnostics = 0;
     if args.path.to_str() != Some("-") && args.path.is_dir() {
-        if !args.overwrite && !args.check {
+        if !args.mode.overwrite && !args.mode.check {
             bail!("formatting a directory requires the `--overwrite` or `--check` option");
         }
```
adthrasher: Now that these are required inputs, do we still need this check?

a-frantz: Agree with Andrew. This check can be removed.

src/commands/format.rs
```diff
@@ -98,20 +107,22 @@ fn read_source(path: &Path) -> Result<String> {
 
 /// Formats a document.
 ///
+/// Checks if the document is formatted correctly and prints the diff if not
+/// if check_only is true.
```
```suggestion
/// If `check_only` is true, checks if the document is formatted correctly and prints the diff if not
/// then exits. Else will format and overwrite the document.
```
Just rewrote this in Github. Will require a run of `cargo +nightly fmt` after committing.

src/commands/format.rs
```diff
@@ -43,9 +43,9 @@ const DEFAULT_SPACE_IDENT_SIZE: usize = 4;
     author,
     version,
     about,
-    after_help = "By default, the `format` command will print a single formatted WDL \
-                  document.\n\nUse the `--overwrite` option to replace a WDL document, or a \
-                  directory containing WDL documents, with the formatted source."
+    after_help = "Use the `--overwrite` option to replace a WDL document, or a directory \
+                  containing WDL documents, with the formatted source. Use the `--check` option \
+                  to print the diff or none if the source is already formatted."
```
```suggestion
    after_help = "Use the `--overwrite` option to replace a WDL document or a directory \
                  containing WDL documents with the formatted source.\nUse the `--check` option \
                  to verify that a document or a directory containing WDL documents is already formatted \
                  and print the diff if not."
```

a-frantz (commented):
Some nits over word smithing. Please commit my suggestions and then run `cargo +nightly fmt` after.

src/commands/format.rs
a-frantz:
```suggestion
```

a-frantz (approved):

peterhuene (approved):

Issue #43, opened 2024-11-18
feature request: Docker container for running sprocket


a-frantz:
This should be built and published to ghcr.io as part of CD. As far as I'm aware, there aren't any benefits to being published on multiple registries. DockerHub has annoying rate limits we want to avoid, which is the reason for my preference for ghcr.io.

PR #56, opened 2025-01-21
feat: Docker container

adthrasher:
Adds a Dockerfile and release automation to address #43.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,9 @@ on:
     tags:
       - v*
 
+env:
+  REGISTRY: ghcr.io
+
 jobs:
   release:
     runs-on: ubuntu-22.04
@@ -23,6 +26,18 @@ jobs:
       with:
         generate_release_notes: true
         make_latest: true
+    - name: Log in to the Container registry
+      uses: docker/login-action@v3
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - uses: docker/setup-qemu-action@v3
+    - uses: docker/setup-buildx-action@v3
+    - uses: docker/build-push-action@v6
+      with:
+        push: true
+        tags: ${{ env.REGISTRY }}/${{ github.repository_owner }}/sprocket:${{ github.ref_name }}
   build_artifacts:
     runs-on: ${{ matrix.os }}
     strategy:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,18 @@
-FROM rust:1.82
+FROM rust:1.82 AS builder
 
-RUN cargo install sprocket
+WORKDIR /tmp/sprocket
 
-ENTRYPOINT ["sprocket"]
+COPY .git/ .git/
+COPY Cargo.lock Cargo.toml ./
+COPY src/ src/
+
+RUN cargo build --release
+
+FROM debian:bookworm
+
+COPY --from=builder /tmp/sprocket/target/release/sprocket /opt/sprocket/bin/sprocket
+
+ENV PATH=/opt/sprocket/bin:$PATH
+
+ENTRYPOINT ["/opt/sprocket/bin/sprocket"]
 CMD ["--help"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,6 @@ FROM rust:1.82 AS builder
 
 WORKDIR /tmp/sprocket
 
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -23,6 +26,18 @@ jobs:
...
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - uses: docker/setup-qemu-action@v3
+    - uses: docker/setup-buildx-action@v3
+    - uses: docker/build-push-action@v6
+      with:
+        push: true
+        tags: ${{ env.REGISTRY }}/${{ github.repository_owner }}/sprocket:${{ github.ref_name }}
```
a-frantz, line 40: Just checking, Is `github.ref_name` going to be the `v*` tag?

adthrasher: Yes, I believe so. I modeled it after: https://github.com/stjude-rust-labs/sprocket/blob/3760b07e894f31f15fd623a05ec6685c29105683/.github/workflows/release.yml#L70. I'm open to whether the `v` prefix should be included, but since it is used in the release and in the artifacts, I opted to include it here.

a-frantz: LGTM

.github/workflows/release.yml
```diff
@@ -23,6 +26,18 @@ jobs:
...
+    - name: Log in to the Container registry
+      uses: docker/login-action@v3
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - uses: docker/setup-qemu-action@v3
+    - uses: docker/setup-buildx-action@v3
+    - uses: docker/build-push-action@v6
+      with:
```
a-frantz, line 38: This needs to run after the `cargo publish` step. I think as written it might install the wrong version.

adthrasher: I thought these were run sequentially.

a-frantz: O yes. I think you're right, I misread this. Nevermind then.

adthrasher: This could also be a separate job with a `needs: release` to ensure that it runs after the `release` job completes.

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM rust:1.82
+
+RUN cargo install sprocket
```
claymcleod: We should probably use `rust:1.82` as a build container but only pull out the `sprocket` binary in a final image that doesn't contain all of Rust.

adthrasher: Do you want the binary in a `scratch` image? Or just copy the binary into a clean base image that contains libc?

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM rust:1.82 AS builder
+
+WORKDIR /tmp/sprocket
+
+COPY .git/ .git/
```
a-frantz: Why is the git folder needed? Does `cargo` use it for something?

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+RUN cargo build --release
+
+FROM debian:bookworm
+
+COPY --from=builder /tmp/sprocket/target/release/sprocket /opt/sprocket/bin/sprocket
+
+ENV PATH=/opt/sprocket/bin:$PATH
+
+ENTRYPOINT ["/opt/sprocket/bin/sprocket"]
```
a-frantz:
```suggestion
ENTRYPOINT ["sprocket"]
```
nit: this looks a little cleaner and should be equivalent.

Dockerfile
adthrasher: I'm not sure. I did a little copy/paste from https://github.com/zaeleus/squab/blob/master/Dockerfile

Dockerfile
claymcleod: Yeah I think the latter. Just enough to get it running, but cut out all the Rust-related stuff.

adthrasher: OK, that's what I have now. Build is in a Rust container, then the binary is copied to a debian container.

Dockerfile
claymcleod: I think we can try without this and, if it still builds, we can remove it.

Dockerfile
claymcleod: Yeah this is much better. Thanks

claymcleod (approved):
LGTM after addressing the outstanding comments.

Issue #54, opened 2025-01-21
make `--local-only` the default behavior


a-frantz:
I think most users are going to expect the "local only" behavior by default. At least when I am personally using `sprocket check` I pretty much always specify `--local-only` or `--single-document`.

If `--local-only` becomes the default behavior, we should make a new flag called `--show-remote-diagnostics` which would enable the _current_ behavior.

PR #59, opened 2025-01-23
feat(check): improvements to default behavior

a-frantz:

Addresses #54 and also surfaces diagnostics with `Severity::Error` regardless of CL options that would otherwise suppress the diagnostic.

### DIFF
--- a/src/commands/check.rs
+++ b/src/commands/check.rs
@@ -160,29 +160,11 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
             None => result.document().diagnostics(),
         };
 
-        // If any errors occurred but this document is meant to be suppressed, only
-        // display the errors.
-        if suppress {
-            let errors = diagnostics
-                .iter()
-                .filter(|d| d.severity() == Severity::Error)
-                .cloned()
-                .collect::<Vec<_>>();
-            if !errors.is_empty() {
-                emit_diagnostics(
-                    &errors,
-                    &uri,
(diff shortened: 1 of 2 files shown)


src/commands/check.rs
```diff
@@ -160,6 +160,26 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
         };
 
+        // If any errors occurred but this document is meant to be suppressed, only
+        // display the errors.
+        if suppress {
+            let errors = diagnostics
+                .iter()
+                .filter(|d| d.severity() == Severity::Error)
+                .cloned()
+                .collect::<Vec<_>>();
```
peterhuene: I think we should probably refactor `emit_diagnostics` to take a generic iterator rather than a slice.

This would produce less code here as we could simply apply a filter to the iterator when `suppress` is true and have the singular `emit_diagnostics` call.

src/commands/check.rs
```diff
@@ -160,6 +160,26 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
+            if !errors.is_empty() {
+                emit_diagnostics(
+                    &errors,
+                    &uri,
+                    &result.document().node().syntax().text().to_string(),
+                    args.common.report_mode,
+                    args.common.no_color,
+                );
+            }
+            continue;
```
peterhuene: Also, we're not updating the error count here; another reason a singular diagnostics handling clause would be useful.

peterhuene (approved):

PR #64, opened 2025-01-30
ci: adds ARM build to Docker image builder

claymcleod:
This PR adds ARM builds to the Docker image builder.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,10 +34,13 @@ jobs:
         password: ${{ secrets.GITHUB_TOKEN }}
     - uses: docker/setup-qemu-action@v3
     - uses: docker/setup-buildx-action@v3
+      with:
+        platforms: linux/amd64,linux/arm64
     - uses: docker/build-push-action@v6
       with:
         push: true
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -34,6 +34,8 @@ jobs:
         password: ${{ secrets.GITHUB_TOKEN }}
     - uses: docker/setup-qemu-action@v3
     - uses: docker/setup-buildx-action@v3
+      with:
+        platforms: linux/amd64,linux/arm64
```
adthrasher, line 38: Do you also need to add a `platforms` argument to `build-push-action`? My assumption is that as-is you're just adding build support to the node and the build action needs to be instructed to build for those platforms.

adthrasher: I just looked at https://docs.docker.com/build/ci/github-actions/multi-platform/ and my takeaway from that is that  you do need a `platforms` argument for the `build-push-action` step.

claymcleod: Good catch!

adthrasher (approved):

PR #66, opened 2025-03-06
feat: introduces creation of template JSON files

### LINKED ISSUE #40: feature: create blank/template input JSONs
The workflow for using this feature would look something like:
```bash
sprocket inputs quality-check.wdl > qc-inputs.json
vim qc-inputs.json
sprocket run --input qc-inputs.json quality-check.wdl
```
There are some considerations to make before implementing. Are the output JSONs filled in with anything? Or blank? How to handle nested inputs? Inputs with default values? etc.

fahdfady:
resolves #40 
adding a CLI command `sprocket inputs <input_path> --output <output_path>` that extracts the inputs of a workflow inside a `wdl` file in to a `json` file.
- the output file is optional, if not provided the JSON will be outputted in the terminal
- the syntax of the command is different from the one mentioned in the issue.
- I'm not sure how to handle "default values" like an input like `String s = "hello"` but i added it as a boolean input from the CLI user and has no functionality right now, so i need guidance on how it should be working.

### examples:
```bash
$ sprocket inputs --help

Usage: sprocket inputs [OPTIONS] <input path>

Arguments:
  <input path>  

Options:
  -o, --output <output path>  
  -v, --verbose...            Increase logging verbosity
  -q, --quiet...              Decrease logging verbosity
  -h, --help                  Print help
  -V, --version               Print version
```

example.wdl
```wdl
version 1.2

task say_hello {
    input {
        String greeting
        String name
    }

    command <<<
        echo "~{greeting}, ~{name}!"
    >>>

    output {
        String message = read_string(stdout())
    }

    requirements {
        container: "ubuntu:latest"
    }
}

workflow main {
    input {
        String name
        Boolean is_pirate = false
    }

    Array[String] greetings = select_all([
        "Hello",
        "Hallo",
        "Hej",
        (if is_pirate then "Ahoy" else None),
    ])

    scatter (greeting in greetings) {
        call say_hello {
            greeting,
            name,
        }
    }

    output {
        Array[String] messages = say_hello.message
    }
}
```
output.
```json
{
  "is_pirate": false,
  "name": null
}
```


src/commands/inputs.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub output: Option<PathBuf>,
+
+    #[arg(long)]
+    #[clap(value_name = "include defaults")]
+    pub include_defaults: bool,
+}
+
+pub async fn generate_inputs(args: InputsArgs) -> Result<()> {
+    let source = std::fs::read_to_string(&args.document)?;
+    let (document, diagnostics) = Document::parse(&source);
```
a-frantz: Instead of using `wdl::ast::Document::parse`, you should use [`wdl::cli::parse_inputs`](https://github.com/stjude-rust-labs/wdl/blob/main/wdl/src/cli.rs#L134). See [here](https://github.com/stjude-rust-labs/wdl/blob/main/wdl/src/cli.rs#L196-L207) for an example of that function being called

src/commands/inputs.rs
```diff
@@ -0,0 +1,96 @@
...
+        // let ty = decl.ty();
+        // Create a default expression if none is provided
+        let value = if let Some(expr) = decl.expr() {
+            expr_to_json(&expr)
+        } else {
+            Value::Null
+        };
+
+        template.insert(name, value);
+    }
```
a-frantz: @adthrasher we need to think through exactly what we want here. My thinking is that for required inputs, the JSON value should be a String with the type of the input? That seems the most user friendly way to render the output JSON.

But I'm a little stumped on the best way to handle inputs with default values. Going to walk through each case:
1. An optional input without a default (or `= None`)
    - Seems safe to render these as a JSON `null`?
2. An input with a _literal_ value
    - Just rendering the literal value, whatever it is (could be a complex object or struct, but `serde` should handle that fine without too much work on our end)
3. An input with an arbitrarily complex `Expr`
    - I'm stumped here. What do you think?

src/commands/inputs.rs
```diff
@@ -0,0 +1,96 @@
...
+            LiteralExpr, LiteralNone, Type,
+        },
+    },
+    doc,
+    grammar::SyntaxTree,
+};
+
+#[derive(Parser, Debug)]
+#[command(about = "Generate input JSON from a WDL document", version, about)]
+pub struct InputsArgs {
```
a-frantz, line 24: We want to support being able to construct an input JSON for both WDL workflows and tasks. This will require accepting a `name: Option<String>` argument. If the provided name doesn't match a task or workflow in the provided document, we should error. If no `name` is provided, we should assume the user wants to create inputs for the workflow in the document if it's present. If no workflow is in the document, we then assume there will be _one_ task defined in the entire document, and construct inputs for that task. If more than one task is present, we can error with a message instructing the user that a `name` must be provided for this WDL document.

src/commands/inputs.rs
a-frantz: I think maybe there should be multiple behavior modifiers here. Further brainstorming, tell me what you think:
- without any flags, `sprocket inputs` will just render required inputs
- `--literal-defaults` will additionally render anything that can be expressed as a simple literal (`null` or a JSON value)
- `--override-expressions` will render any inputs with `Expr` defaults similarly to how I suggested required inputs are rendered? i.e. the JSON value will be a string with the expected `Type`

a-frantz: Additionally, there needs to be `--nested-inputs` flag which will surface task level inputs for workflows (and do nothing in the case we're rendering inputs for a task)

adthrasher: > @adthrasher we need to think through exactly what we want here. My thinking is that for required inputs, the JSON value should be a String with the type of the input? That seems the most user friendly way to render the output JSON.

I think that's reasonable. It also matches the output of WOMtool, so users are likely familiar with it. The only downside is that it will look strange for non-String inputs. Engines should be able to coerce String to other types assuming the contained value is valid

adthrasher: > 1. An optional input without a default (or `= None`)
>    
>    * Seems safe to render these as a JSON `Null`?

I would render these with the literal `None` so the user doesn't have to make any changes. I haven't actually tried passing values like this to any of the engines, so I would caveat it with "it depends on how the engine handles `None` and `Null`".

a-frantz: So
```wdl
workflow test {
    input {
        File bam
        Int read_depth
        Float some_fraction
        String? read_group
        String strandedness = "Stranded-Reverse"
        Boolean complicated = if read_depth > 30 then true else false
    }
}
```
`sprocket inputs --name test --literal-defaults --override-expressions` would generate
```json
{
    "test.bam": "File",
    "test.read_depth": "Int",
    "test.some_fraction": "Float",
    "test.read_group": null,
    "test.strandedness": "Stranded-Reverse",
    "test.complicated": "Boolean"
}
```
?

a-frantz: > > 1. An optional input without a default (or `= None`)
> >    
> >    * Seems safe to render these as a JSON `Null`?
> 
> I would render these with the literal `None` so the user doesn't have to make any changes. I haven't actually tried passing values like this to any of the engines, so I would caveat it with "it depends on how the engine handles `None` and `Null`".

I believe that engines should be coercing JSON `null` into WDL `None` automatically.

adthrasher: > 2. An input with a _literal_ value
>    
>    * Just rendering the literal value, whatever it is (could be a complex object or struct, but `serde` should handle that fine without too much work on our end)

I agree with this. Just render the literal.

> 3. An input with an arbitrarily complex `Expr`
>    
>    * I'm stumped here. What do you think?

I don't think I've ever tried supplying expressions in a JSON inputs file. If that is supported by the engines, I'd just render the expression. If not how about something like 
```
{
    "wf.input.a": "<default: expression>"
}
```

With the expression rendered in the `<>` block. If the user wants the default, it's easy to delete the whole line. If they want to override, it's a little more annoying to delete, but still not too difficult.

adthrasher: > I think maybe there should be multiple behavior modifiers here. Further brainstorming, tell me what you think:
> 
> * without any flags, `sprocket inputs` will just render required inputs
> * `--literal-defaults` will additionally render anything that can be expressed as a simple literal (`Null` or a JSON value)
> * `--override-expressions` will render any inputs with `Expr` defaults similarly to how I suggested required inputs are rendered? i.e. the JSON value will be a string with the expected `Type`

I think the *default* behavior needs to be rendering all of the inputs somehow. I think users will be surprised if the default command doesn't include some subset of inputs. If we deviate from that, then the subcommand should produce some very clear output to indicate things were skipped, like:

```
Rendered X required inputs.
Skipped Y inputs with literal defaults.
Skipped Z inputs with default expressions.
```

adthrasher: > Additionally, there needs to be `--nested-inputs` flag which will surface task level inputs for workflows (and do nothing in the case we're rendering inputs for a task)

Agreed. I think that is missing in the other solutions that exist. We may want to look at some examples though as we have some workflows that will have a **very** large list.

a-frantz: > > 2. An input with a _literal_ value
> >    
> >    * Just rendering the literal value, whatever it is (could be a complex object or struct, but `serde` should handle that fine without too much work on our end)
> 
> I agree with this. Just render the literal.
> 
> > 3. An input with an arbitrarily complex `Expr`
> >    
> >    * I'm stumped here. What do you think?
> 
> I don't think I've ever tried supplying expressions in a JSON inputs file. If that is supported by the engines, I'd just render the expression. If not how about something like
> 
> ```
> {
>     "wf.input.a": "<default: expression>"
> }
> ```
> 
> With the expression rendered in the `<>` block. If the user wants the default, it's easy to delete the whole line. If they want to override, it's a little more annoying to delete, but still not too difficult.

@peterhuene I am fairly certain that you can't supply arbitrary WDL expressions in an input JSON. Can you confirm? It must be a literal, right?

a-frantz: > > I think maybe there should be multiple behavior modifiers here. Further brainstorming, tell me what you think:
> > 
> > * without any flags, `sprocket inputs` will just render required inputs
> > * `--literal-defaults` will additionally render anything that can be expressed as a simple literal (`Null` or a JSON value)
> > * `--override-expressions` will render any inputs with `Expr` defaults similarly to how I suggested required inputs are rendered? i.e. the JSON value will be a string with the expected `Type`
> 
> I think the _default_ behavior needs to be rendering all of the inputs somehow. I think users will be surprised if the default command doesn't include some subset of inputs. If we deviate from that, then the subcommand should produce some very clear output to indicate things were skipped, like:
> 
> ```
> Rendered X required inputs.
> Skipped Y inputs with literal defaults.
> Skipped Z inputs with default expressions.
> ```

I think including a summary of what was rendered/skipped is good. but can you elaborate a bit on the above?

adthrasher: > The inputs for a workflow invocation may be specified as a single JSON object that contains one member for each top-level workflow input. The name of the object member is the [fully-qualified name](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#fully-qualified-names--namespaced-identifiers) of the input parameter, and the value is the [serialized form](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#appendix-a-wdl-value-serialization-and-deserialization) of the WDL value.

Reading the spec, it does appear that only literals can be specified in JSON.

adthrasher: > > > 1. An optional input without a default (or `= None`)
> > >    
> > >    * Seems safe to render these as a JSON `Null`?
> > 
> > 
> > I would render these with the literal `None` so the user doesn't have to make any changes. I haven't actually tried passing values like this to any of the engines, so I would caveat it with "it depends on how the engine handles `None` and `Null`".
> 
> I believe that engines should be coercing JSON `null` into WDL `None` automatically.

I should have looked at the spec beforehand. It references using `null`, so that is a requirement for the engines.

adthrasher: > > > I think maybe there should be multiple behavior modifiers here. Further brainstorming, tell me what you think:
> > > 
> > > * without any flags, `sprocket inputs` will just render required inputs
> > > * `--literal-defaults` will additionally render anything that can be expressed as a simple literal (`Null` or a JSON value)
> > > * `--override-expressions` will render any inputs with `Expr` defaults similarly to how I suggested required inputs are rendered? i.e. the JSON value will be a string with the expected `Type`
> > 
> > 
> > I think the _default_ behavior needs to be rendering all of the inputs somehow. I think users will be surprised if the default command doesn't include some subset of inputs. If we deviate from that, then the subcommand should produce some very clear output to indicate things were skipped, like:
> > ```
> > Rendered X required inputs.
> > Skipped Y inputs with literal defaults.
> > Skipped Z inputs with default expressions.
> > ```
> 
> I think including a summary of what was rendered/skipped is good. but can you elaborate a bit on the above?

I wouldn't have flags for `--literal-defaults` and `--override-expressions`. By default, I would suggest we render all of those. We could have flags to skip those, but I'd make that opt-in behavior.

So something like 
```
workflow wf {
...
input {
    File foo
    Int disk = 2
    Int mem = disk * 2
}
...
```

Would, by default, render as: 
```
{
   "wf.foo": "File",
   "wf.disk": 2,
    "wf.mem": "Int <default: disk * 2>"
}
```

If we do this, then we should think about sorting the inputs before writing the JSON. Anything that isn't a valid JSON entry (e.g. `wf.foo` and `wf.mem` above) would be first to draw the user's attention.

a-frantz: Ahh. That makes sense. But here's where it gets tricky for me: if we're defaulting to rendering even inputs with default values (whether a literal or an expr), does that carry over when `--nested-inputs` is supplied? If so, the output JSON can be _ginormous_.

Maybe these are the flags we should have:
- default as you described
- `--hide-expressions` (don't love the name) hides any input which isn't defaulted to a literal or required
- `--hide-defaults` only render required inputs

Still thinking over what the behavior should be when `--nested-inputs` is supplied

src/commands/inputs.rs
```diff
@@ -0,0 +1,96 @@
...
+        template.insert(name, value);
+    }
+
+    let json_output = serde_json::to_string_pretty(&template)?;
+
+    if let Some(output_path) = args.output {
+        std::fs::write(output_path, json_output)?;
+    } else {
+        // ? output in the console if no output path is provided is that good?
+        println!("{}", json_output);
```
a-frantz: Yes, I think this is the desired behavior!

src/commands/inputs.rs
a-frantz: I suppose that `--nested-inputs` should just carry over the default behavior when no additional flags are provided. Anything else would be confusing. So that means to be most useful, users will _probably_ want to call `--nested-inputs` in conjunction with `--hide-expressions` (or whatever we call it), or else they'll have a huge JSON to parse through and edit. But I think that's consistent and sane, even if it requires extra flags for the most common use case 🤷

adthrasher: > I suppose that `--nested-inputs` should just carry over the default behavior when no additional flags are provided. Anything else would be confusing. So that means to be most useful, users will _probably_ want to call `--nested-inputs` in conjunction with `--hide-expressions` (or whatever we call it), or else they'll have a huge JSON to parse through and edit. But I think that's consistent and sane, even if it requires extra flags for the most common use case 🤷

That would be my thought as well. The `--nested-inputs` help text should probably suggest users include one of the other flags.

src/commands/inputs.rs
fahdfady: I'm kind of stuck on how to use it. I tried my best but still the parse_input function panic.
```bash
InputsArgs { document: "example/main.wdl", output: None, include_defaults: false }
error: failed to parse input file `/home/fahd/softwarecoding/sprocket/example/main.wdl`

Caused by:
    expected value at line 1 column 1
 ```
 
 you can see the code [here](https://github.com/stjude-rust-labs/sprocket/pull/66/commits/885d913b30e7d0ea223428e7345b6195995f6319) .. 
snippet:
 ```rust
 
pub async fn generate_inputs(args: InputsArgs) -> Result<()> {
    println!("{:?}", args);

    let results: Vec<wdl::analysis::AnalysisResult> =
        analyze(args.document.as_str(), vec![], false, false).await?;

    // println!("{:?}", results);

    let uri: Url = Url::parse(args.document.as_str()).unwrap_or_else(|_| {
        path_to_uri(args.document.as_str()).expect("file should be a local path")
    });

    // println!("{:?}", uri);

    let result = results
        .iter()
        .find(|r| **r.document().uri() == uri)
        .context("failed to find document in analysis results")?;

    // println!("result: {:?}", result);

    let document = result.document();

    // println!("doc: {:?}", document);

    let doc_name = Path::new(&args.document)
        .file_name()
        .and_then(|n| n.to_str())
        .unwrap_or(&args.document);

    let (_path, name, inputs) = wdl::cli::parse_inputs(
        document,
        Some(doc_name),
        Some(Path::new(args.document.as_str())),
    )?;

    println!("{:?},{:?}, {:?}", _path, name, inputs);
 ```

Comment (a-frantz) [reply src/commands/inputs.rs]: You're getting close! You'll also need to implement the changes I talk about here - https://github.com/stjude-rust-labs/sprocket/pull/66/files#r1983275592

Then, what you'll pass into `parse_inputs()` would be something like
```rust
parse_inputs(
    result.document, // This is of type wdl_analysis::document::Document
    args.name, // This is an Option<&str>
    None // There is no "inputs file" to pass in, because we're working on creating one!
)
```

adthrasher:
> @adthrasher @a-frantz will it be a good idea to generate comments beside the json inputs? something like :
> 
> ```json
> {
>   "name": null,  // Required: String (e.g., "sample1")
>   "is_pirate": false  // Optional: Boolean, defaults to false
> }
> ```
> 
> mainly to increase readability, especially to non-programmers that use sprocket in validating their inputs in WDL. I do think it's a kind of a privilege right now, but i really like the idea. so, it can be opened as an issue and be resolved later if you see that it's useful.

Unfortunately, comments are not valid JSON.

a-frantz:
> > @adthrasher @a-frantz will it be a good idea to generate comments beside the json inputs? something like :
> > ```json
> > {
> >   "name": null,  // Required: String (e.g., "sample1")
> >   "is_pirate": false  // Optional: Boolean, defaults to false
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > mainly to increase readability, especially to non-programmers that use sprocket in validating their inputs in WDL. I do think it's a kind of a privilege right now, but i really like the idea. so, it can be opened as an issue and be resolved later if you see that it's useful.
> 
> Unfortunately, comments are not valid JSON.

Yea, comments aren't allowed in JSON, but your idea is good! We have discussed wanting to put comments in files like this before, but unfortunately the SPEC prohibits it. Good instinct though!

src/commands/inputs.rs
fahdfady: I'm really stuck on this one 😅 it does not return anything: 
```rust
let (_path, name, inputs) = wdl::cli::parse_inputs(result.document(), Some("main"), None)?;

println!("name: {:?}    {:?}", name, inputs);
```
```bash
name: "main"    Workflow(WorkflowInputs { inputs: {}, calls: {} })
```

I've reverted with [commit](https://github.com/stjude-rust-labs/sprocket/pull/66/commits/014ea6c321a5d819857b2c3800d18f5450264e1f) into implementing with the `wdl::analysis` and it's working right now 
i tried to solve the problem using `wdl::cli::parse_inputs` but i couldn't, can you tell me what i might be missing? i tried to read the docs and the source code but it didn't really help that much.

src/commands/inputs.rs
```diff
@@ -0,0 +1,167 @@
...
+
+#[derive(Parser, Debug)]
+#[command(about = "Generate input JSON from a WDL document", version, about)]
+pub struct InputsArgs {
+    #[arg(required = true)]
+    #[clap(value_name = "input path")]
+    pub document: String,
+
+    #[arg(short, long)]
+    #[clap(value_name = "workflow name")]
```
a-frantz:
```suggestion
    #[clap(value_name = "workflow or task name")]
```

src/commands/inputs.rs
```diff
@@ -0,0 +1,167 @@
...
+    let document = result.document();
+
+    let diagnostics = document.diagnostics();
+    if !diagnostics.is_empty() {
+        for diagnostic in diagnostics {
+            anyhow::bail!("Failed to parse WDL document: {:?}", diagnostic);
+        }
+    }
+
+    let (_path, name, inputs) = wdl::cli::parse_inputs(document, Some("main"), None)?;
```
a-frantz:
```suggestion
    let (_path, name, inputs) = wdl::cli::parse_inputs(document, args.name, None)?;
```

src/commands/inputs.rs
```diff
@@ -0,0 +1,167 @@
...
+            println!("tasks: {:?}", tasks);
+            match tasks.len() {
+                0 => anyhow::bail!("No workflow or tasks found in document"),
+                1 => tasks[0].inputs(),
+                _ => anyhow::bail!(
+                    "Multiple tasks found in document but no name specified. Please provide a name using --name"
+                ),
+            }
+        }
+    };
```
a-frantz: this should all be handled internally by `cli::parse_inputs()`, so you shouldn't have to do any of this

src/commands/inputs.rs
a-frantz: Not sure how you got that output, as supplying `Some("main")` should have resulted in an error when `main` couldn't be found in your document. Unless your test document has a workflow or task called `main`? But regardless, the second arg needs to be the `args.name` from the CL

a-frantz:
please format future commits with `cargo +nightly fmt`, and tell us what WDL documents you are using for testing

src/commands/inputs.rs
```diff
@@ -0,0 +1,167 @@
...
+    #[clap(value_name = "output path")]
+    pub output: Option<PathBuf>,
+
+    #[arg(short, long)]
+    #[clap(value_name = "literal defaults")]
+    pub literal_defaults: bool,
+
+    #[arg(short, long)]
+    #[clap(value_name = "override expressions")]
+    pub override_expressions: bool,
```
a-frantz: Sorry if our thread got confusing, I did at one point suggest these two flags, but me and Andrew T at the end of the discussion decided the behavior flags should be `--hide-expressions`, `--hide-defaults`, and `--nested-inputs`. See the comment thread for what these flags should do.

src/commands/inputs.rs
fahdfady: yes,I was testing on a workflow named `main` i didn't want to complicate the code since i only tested on this one to make sure it's working
```wdl
workflow main {
    input {
        String name
        Boolean is_pirate = false
    }
```
I think the problem occurs in [this code of `wld::cli::parse_inputs`](https://github.com/stjude-rust-labs/wdl/blob/3c69f90a09f1bb43439abdfe7fec5463b8f0fb56/wdl/src/cli.rs#L160C1-L164C79) where our condition is met ( not a task, is a workflow, the name from the params matches the name in the document ) so it returns `Ok((None, name.to_string(), Inputs::Workflow(Default::default())))` instead of the [desired behavior](https://github.com/stjude-rust-labs/wdl/blob/3c69f90a09f1bb43439abdfe7fec5463b8f0fb56/wdl/src/cli.rs#L190C1-L190C33) `Ok((None, name, inputs))`

a-frantz: O yes! I gave you some bad direction here. The `parse_inputs()` function actually isn't what you should be using for this PR. Apologies, I misremembered what `parse_inputs()` actually does. The documentation should be improved. Can you please open a new issue on the `wdl` repo asking for better documentation of `cli::parse_inputs()`? I can take a closer look later, and don't want to forget!

You started down the right track by switching from the `Document` type from `wdl_ast` to the `Document` type offered by `wdl_analysis`. Keep going with this implementation!

The `wdl_analysis::document::Document` type will provide you with fully resolved WDL type analysis which is what you need for this PR. (`wdl_ast::Document` is what you initially committed and I pointed out is incorrect, but then I sent you down a wild goose chase with `parse_inputs()` and I'm sorry for that!)

src/commands/inputs.rs
a-frantz: This actually looks correct! I should have been paying closer attention earlier. Ignore my advice to use `parse_inputs()`, you were doing it right!

fahdfady:
> and tell us what WDL documents you are using for testing

those are the wdl docs i use for testing
https://gist.github.com/fahdfady/0e14fa1d56802a9f82ff1d7a37504e1b#file-conditional-wdl-L3

fahdfady:
> those are the wdl docs i use for testing

by the way, when i use `sprocket inputs` on a file that has unused inputs (e.g. `plot.wdl` provided in the gist) the program errors and refuses to continue, even though it says the severity is "warning"

```bash
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.52s
     Running `target/debug/sprocket inputs example/plot.wdl`
error: Failed to parse WDL document: Diagnostic { rule: Some("UnusedInput"), severity: Warning, message: "unused input `csv_file`", fix: None, labels: [Label { message: "", span: Span { start: 59, end: 67 } }] }
```

[the file](https://gist.github.com/fahdfady/0e14fa1d56802a9f82ff1d7a37504e1b#file-plot-wdl)

src/commands/inputs.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    let result = results
+        .iter()
+        .find(|r| **r.document().uri() == uri)
+        .context("failed to find document in analysis results")?;
+
+    let document = result.document();
+
+    let diagnostics = document.diagnostics();
+    if !diagnostics.is_empty() {
```
a-frantz: This should check for diagnostics of `Severity::Error` instead of just any diagnostic.

fahdfady:
@peterhuene @a-frantz @adthrasher the functionality is ready right now

### example usage
```bash
# Basic usage - outputs JSON to console
sprocket inputs workflow.wdl

# Save output to a file
sprocket inputs workflow.wdl -o inputs.json

# Get inputs for a specific workflow
sprocket inputs workflow.wdl --name MyWorkflow

# Hide inputs with default values
sprocket inputs workflow.wdl --hide-defaults

# Hide inputs that don't have a literal default
sprocket inputs workflow.wdl --hide-expressions

# Include inputs from all tasks and workflows
sprocket inputs workflow.wdl --nested-inputs

```

- `--hide-defaults` flag hides all inputs with any default
- `--hide-expressions` hides inputs that don't have a literal default (meaning they either have no default or have a complex expression as default).
- `--nested inputs` shows inputs from tasks and workflows together

a-frantz:
Before making future pushes, make sure that your branch is passing all the tests and lints locally first. You're most recent push is failing the lint check (`cargo clippy --all-features`)

src/commands/inputs.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+type InputDefaultMap = IndexMap<String, bool>; // input_name -> has_default
+
+struct InputVisitor {
+    inputs: InputDefaultMap,
+    literal_defaults: IndexMap<String, bool>, // input_name -> has_literal_default
+}
+
+impl Visitor for InputVisitor {
```
a-frantz: This is an interesting approach. It could work with some tweaking, unfortunately it is pretty inefficient. I don't think we want to implement this with a `Visitor` approach.

src/commands/inputs.rs
```diff
@@ -0,0 +1,279 @@
...
+            }
+        } else if let Some(workflow) = document.workflow() {
+            if workflow.name() == name {
+                // Add workflow inputs
+                for (input_name, input) in workflow.inputs() {
+                    result.push((workflow.name(), input_name.as_str(), input));
+                }
+
+                // If nested_inputs is true, add all called task inputs
+                if args.nested_inputs {
```
a-frantz, line 322: This is not the intended behavior for `--nested-inputs`. "nested inputs" refer to this section of the WDL spec - https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#computing-call-inputs

> Any optional inputs that are not specified in the call may be set by the user at runtime if the execution engine supports the allow_nested_inputs hint and it is set to true in the workflow's hints section.

To support nested inputs in this command, you will need to iterate through a workflows "calls" and extract the relevant inputs from the called tasks and subworkflows.

src/commands/inputs.rs
```diff
@@ -0,0 +1,279 @@
...
+            wdl::analysis::types::PrimitiveType::Directory => {
+                Value::String("Directory".to_string())
+            }
+        },
+
+        _ => Value::Null,
+    }
+}
+
+fn has_default(document: &Document, document_path: &str, input_name: &str) -> bool {
```
a-frantz: We don't only care if the input has a default, but also what that default is. You will need to update this to extract the expression as well so that we can render that in the output JSON.

src/commands/inputs.rs
```diff
@@ -0,0 +1,279 @@
...
+        inputs: IndexMap::new(),
+        literal_defaults: IndexMap::new(),
+    };
+
+    ast_doc.visit(&mut (), &mut visitor);
+
+    visitor.inputs.get(input_name).cloned().unwrap_or(false)
+}
+
+fn has_literal_default(document: &Document, document_path: &str, input_name: &str) -> bool {
```
a-frantz: Same here. We need to actually extract the literal value and insert it into the out JSON

a-frantz (changes requested):
Seems there was some confusion about the intended behavior for this command. If my comments aren't clear enough and you need additional guidance, please reach me on Slack!

src/commands/inputs.rs
fahdfady: > unfortunately it is pretty inefficient 

I see. I think the problem is we're visiting the AST multiple times based on the number of inputs. I want to propse another approach that might be interesting aswell. https://github.com/stjude-rust-labs/sprocket/pull/66/commits/3a6ff469304159ecfd6df6e5ea0f9a02f6a58a82
with this approach we:
1. collect all input info from a single pass through the AST.
2. store the info in 2 IndexMaps.
3. reuse the cached information when filtering inputs.
This might be a lot more efficient that the previous approach. If this is still not enough, I'd be happy to hear from you, I can go through a completely new approach

a-frantz: The new changes are more efficient than the prior commits! But just using an AST Visitor at all is inefficient, as the AST has already been traversed as part of analysis, and you are now doing it again to parse out information that's already been parsed as part of that first pass. Instead you should be using the methods available to you on the `wdl::analysis::document::Document`. That will have all the information you need.

src/commands/inputs.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub only_required: bool,
+}
+
+/// Generate input JSON from a WDL document
+pub async fn generate_inputs(args: InputsArgs) -> Result<()> {
+    let results: Vec<wdl::analysis::AnalysisResult> =
+        analyze(args.document.as_str(), vec![], false, false).await?;
+
+    let uri: Url = Url::parse(args.document.as_str()).unwrap_or_else(|_| {
+        path_to_uri(args.document.as_str()).expect("file should be a local path")
```
a-frantz: We should support remote files here as well

src/commands/inputs.rs
```diff
@@ -0,0 +1,349 @@
...
+    let mut template = serde_json::Map::new();
+
+    // Collect inputs and their parent information
+    let inputs_with_parents = collect_inputs_with_parents(&args, document)?;
+
+    for (parent_name, name, input) in inputs_with_parents {
+        let v: &wdl::analysis::types::Type = input.ty();
+        let key = format!("{}.{}", parent_name, name);
+        
+        let is_required = !input.ty().is_optional();
```
a-frantz, line 154: this isn't correct. Let me share an example:
```
input {
    File required
    File? not_required
    Int not_optional_but_also_not_required = 42
}
```
So in addition to checking `is_optional()`, you also need to see if a default has been provided. The required inputs to a task or workflow are the set that are neither optional nor have a default value.

src/commands/inputs.rs
```diff
@@ -0,0 +1,349 @@
...
+        } else if let Some(workflow) = document.workflow() {
+            if workflow.name() == name {
+                // Add workflow inputs
+                for (input_name, input) in workflow.inputs() {
+                    result.push((workflow.name(), input_name.as_str(), input));
+                }
+
+                // If nested_inputs is true, add all called task inputs
+                if args.nested_inputs {
+                    for task in document.tasks() {
```
a-frantz: this is also incorrect. Instead of iterating through the _document's tasks_, we want to iterate through the _workflow's calls_. https://docs.rs/wdl-analysis/latest/wdl_analysis/document/struct.Workflow.html#method.calls

a-frantz (changes requested):
This is shaping up! Nice work. Unfortunately some of the behavior is still not quite right. I've pointed out some of the big picture changes needed.

Please also make sure that `cargo +nightly fmt` and `cargo clippy --all-features` pass to get the CI green! :)

src/commands/inputs.rs
fahdfady: am i on the right path here?
```rs
for (input_name, input) in called_task.inputs() {
  // 1. marked as optional
  // 2. has a default value
  let is_optional_type = input.ty().is_optional();
  let has_default = task_input_has_default(&ast_doc, called_task.name(), input_name);
  
  // An input is NOT required if it's either optional or has a default
  let is_not_required: bool = is_optional_type || has_default;
```
I'm iterating over the inputs of the called tasks.

a-frantz: That logic seems ok to me 👍 think you're headed in the right direction.

Although I think you should be able to pull most of this information from the `wdl-analysis` API. I think you're using the `wdl-ast` API for some things that would be easier pulled from the analysis side of things.

fahdfady:
@a-frantz I'm ready for a re-review on This

a-frantz:
> @a-frantz I'm ready for a re-review on This

@fahdfady it looks like there haven't been any new commits since I last reviewed. Per our new contributing policies (https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md), we aren't going to review further tilll the CI is turned all green. Please address the failures reported by the CI

fahdfady:
> > @a-frantz I'm ready for a re-review on This
> 
> @fahdfady it looks like there haven't been any new commits since I last reviewed. Per our new contributing policies (https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md), we aren't going to review further tilll the CI is turned all green. Please address the failures reported by the CI and resolve any merge conflicts.

My bad, i think the problem was that the PR was closed and it couldn't see changes i made on my forked branch? anyways i just pushed a commit now and it should pass the CI. sorry for any confusion

PR #69, opened 2025-03-08
feat: YAML input support

### LINKED ISSUE #68: feature request: accept YAML as an input format
The WDL SPEC requires engines support JSON as a WDL input format, but YAML is (subjectively) a nicer format and I'd like us to support it.

`run` and `validate-inputs` would need to be updated to handle YAML _in addition_ to JSON. Being able to output YAML from `inputs` (#66 ) should also be supported.

Gyan-max:
## Add YAML input support for validate-inputs command

This PR implements YAML support for the `validate-inputs` command, as requested in [issue stjude-rust-labs/wdl#68](https://github.com/stjude-rust-labs/sprocket/issues/68).

The WDL SPEC requires engines to support JSON as a WDL input format, but YAML is often more user-friendly and readable. This implementation adds the ability for users to provide input files in YAML format alongside the existing JSON support.

Key changes:
- Added robust content-based file format detection (tries to parse as both JSON and YAML)
- Created utility functions to automatically convert YAML to JSON internally
- Updated the `validate-inputs` command to work seamlessly with both formats
- Uses serde's built-in YAML<->JSON conversion capabilities for reliability
- Added appropriate logging to track the detection and conversion process

Fixes stjude-rust-labs/wdl#68


src/utils.rs
```diff
@@ -0,0 +1,94 @@
+//! Utility functions for the Sprocket command line tool.
+
+use std::path::PathBuf;
+use std::fs;
+use std::io::Read;
+
+use anyhow::{Result, Context};
+
+/// Determines if a file is a YAML file based on its extension.
+pub fn is_yaml_file(path: &PathBuf) -> bool {
```
a-frantz: Haven't read through the serde docs, but can this be done more robustly by peeking file contents? File extensions aren't very reliable. I'd rather try to parse as YAML or JSON and return which is error free (and just fail if neither can be parsed)

src/utils.rs
```diff
@@ -0,0 +1,94 @@
...
+    if let Some(extension) = path.extension() {
+        let ext = extension.to_string_lossy().to_lowercase();
+        return ext == "yaml" || ext == "yml";
+    }
+    false
+}
+
+/// Converts YAML content to JSON.
+pub fn convert_yaml_to_json(yaml_path: &PathBuf) -> Result<String> {
```
a-frantz: Are you sure serde (or another crate) doesn't provide this functionality? YAML to JSON seems a common enough conversion I'd expect you to be able to pull a solution someone else wrote.

a-frantz:
Please re-write your PR description according to our template

Gyan-max:
@a-frantz I've addressed both review comments:

1. Improved file format detection to use content-based parsing rather than extensions
2. Utilized serde's built-in conversion capabilities for more robust YAML to JSON conversion

I've also added comprehensive tests and a CHANGELOG entry for this feature.

src/commands/validate.rs
```diff
@@ -37,8 +39,14 @@ pub struct ValidateInputsArgs {
 /// * Every supplied input is correctly typed.
 /// * No extraneous inputs are provided.
 /// * Any provided `File` or `Directory` inputs exist.
+///
+/// This command supports both JSON and YAML input files.
 pub async fn validate_inputs(args: ValidateInputsArgs) -> Result<()> {
-    if let Some(diagnostic) = wdl_validate_inputs(&args.document, &args.inputs).await? {
+    // Convert input file to JSON if necessary
+    info!("Checking input file format");
```
a-frantz: This log message isn't helpful to end users, so please delete it.

src/utils.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    // If neither worked, return an error
+    bail!("File is neither valid JSON nor valid YAML: {}", file_path.display())
+}
+
+/// Converts input file to JSON if necessary.
+///
+/// If the file is already JSON, returns the original path.
+/// If the file is YAML, converts it to JSON and returns the path to the temporary JSON file.
+pub fn get_json_path(input_path: &PathBuf) -> Result<PathBuf> {
```
a-frantz: Instead of taking a reference and potentially cloning the input path, this function should take ownership of `input_path` and return it directly

a-frantz (commented):
Looking good!

Some suggestions: you are parsing each input file two times if it's JSON and three times if it's YAML. This should be done in one parse for both formats.

Additionally, this should be possible without creating a new file on disk. Just read the YAML and convert to JSON in memory, without writing back to disk.

Gyan-max:
Hi @a-frantz,

Thank you for the review feedback! I've made the following improvements:

1. **More efficient parsing**: I've completely redesigned the approach to parse the input file only once, extracting and storing both the format information and the parsed JSON value in a single pass. This eliminates the multiple parsing issues.

2. **Better function design**: I've changed `get_json_path` to take ownership of the input path rather than a reference, and created a new `get_json_string` function that works directly with the parsed content.

3. **Removed unhelpful logging**: Removed the "Checking input file format" log message as suggested.

The new implementation is more efficient and cleaner:
- For JSON files: We parse once and use the result directly
- For YAML files: We parse and convert to JSON in a single function call

I still need to create a temporary file at the end because the underlying `wdl_validate_inputs` function requires a file path rather than in-memory content. If the underlying API is updated in the future to accept string content directly, we could eliminate the file I/O entirely.

All tests are passing with these changes. Let me know if you'd like any further adjustments!

a-frantz:
> The new implementation is more efficient and cleaner:
> 
> * For JSON files: We parse once and use the result directly
> * For YAML files: We parse and convert to JSON in a single function call
> 
> I still need to create a temporary file at the end because the underlying `wdl_validate_inputs` function requires a file path rather than in-memory content. If the underlying API is updated in the future to accept string content directly, we could eliminate the file I/O entirely.

Ah you're right. Can you open an issue on the `wdl` repo suggesting an API change for `wdl::cli::validate_inputs`?

In the meantime, you can avoid writing the temp file in the case of a JSON input. Just return `args.input` directly without any modifications if it parses as JSON correctly. The data flow for YAML looks good as is though 👍

Gyan-max:
Hi @a-frantz,

Thank you for the suggestion! You're absolutely right  for JSON inputs, we can avoid the unnecessary file I/O entirely. I've made this optimization:

1. Added a new `get_json_file_path` function that returns the original file path for JSON inputs and only creates a temporary file for YAML inputs
2. Updated the validation code to use this function

I'll also create an issue in the WDL repository to suggest an API enhancement for `wdl::cli::validate_inputs` to accept input content directly rather than requiring a file path. This would allow even more efficient handling in the future.

a-frantz:
Looking good for the most part, but your Rust isn't formatted correctly. Run `cargo +nightly fmt` from the root. Also `git rm` everything in the `test_files/` directory, as we don't want that committed to `main`.

Gyan-max:
Hi @a-frantz,

Thank you for the feedback! I've made the requested changes:

1. Fixed the Rust formatting by running `cargo +nightly fmt` from the root
2. Removed the test_files directory from the repository and added it to .gitignore to prevent it from being added in the future


Let me know if there's anything else you'd like me to adjust!

.gitignore
```diff
@@ -3,5 +3,6 @@
 target/
 .vscode/
 **/Cargo.lock
+test_files/
```
a-frantz: Please remove this addition.

src/utils.rs
a-frantz, line 1: Everything in this file is related to input processing, so IMO it would be better to make this an `input` module than a generic "util" module. This additionally makes sense as we have plans for further input processing features in Sprocket, so an `input` module will likely be extended in the near future.

src/utils.rs
```diff
@@ -0,0 +1,232 @@
...
+    // If neither worked, return an error
+    bail!(
+        "File is neither valid JSON nor valid YAML: {}",
+        path.display()
+    )
+}
+
+/// Gets a JSON string representation of the input file.
+///
+/// Takes ownership of the input path and parses the file only once.
```
a-frantz:
```suggestion
```
This information is a matter of the internal implementation, that a user of this func doesn't need to concern themselves with.

src/utils.rs
```diff
@@ -0,0 +1,232 @@
...
+        }
+        ParsedInput::Yaml(json) => {
+            debug!("Using YAML input converted to JSON");
+            serde_json::to_string_pretty(&json)
+                .context("Failed to convert YAML-derived JSON value to string")
+        }
+    }
+}
+
+/// Gets a path to a JSON file for use with APIs that require a file path.
```
a-frantz:
```suggestion
/// Gets a path to a JSON file from an input file which may be in either JSON or YAML format.
```

src/utils.rs
```diff
@@ -0,0 +1,232 @@
...
+pub fn create_json_file_from_input(input_path: PathBuf) -> Result<PathBuf> {
+    let json_string = get_json_string(input_path)?;
+
+    let temp_dir = std::env::temp_dir();
+    let temp_file = temp_dir.join("sprocket_temp_input.json");
+
+    fs::write(&temp_file, json_string).context("Failed to write temporary JSON file")?;
+
+    Ok(temp_file)
+}
```
a-frantz: This is redundant with `get_json_file_path()`. Just delete this function instead of having both

src/input.rs
```diff
@@ -0,0 +1,216 @@
...
+
+use std::fs;
+use std::path::PathBuf;
+
+use anyhow::Context;
+use anyhow::Result;
+use anyhow::bail;
+use tracing::debug;
+
+/// Parse result containing either JSON or YAML data
```
a-frantz:
```suggestion
/// Parse result containing JSON data.
```
This always holds JSON data, so it shouldn't say "containing... YAML data"

src/input.rs
```diff
@@ -0,0 +1,216 @@
...
+#[derive(Debug)]
+pub enum ParsedInput {
+    /// JSON data
+    Json(serde_json::Value),
+    /// YAML data converted to JSON
+    Yaml(serde_json::Value),
+}
```
a-frantz: I'm not convinced this enum should hold a JSON `Value`. It's important to think about ownership of data, and this `enum` seems to be doing two unrelated tasks at once.
1. It is tracking what format the input file was in
2. It is owning the JSON value we want to pass into `validate_inputs()`

IMO those two tasks should be done separately.

src/utils.rs
a-frantz: Please delete this file now that you've created `input.rs`

Gyan-max:
@a-frantz Can you please clearify me if I delete that input.rs file, how can we handle the input file processing? I even deleted the utilis.rs previously

a-frantz:
> @a-frantz Can you please clearify me if I delete that input.rs file, how can we handle the input file processing? I even deleted the utilis.rs previously

Apologies for not being clear! In the HEAD of this branch, you have _both_ `input.rs` and `util.rs`. Please remove `util.rs` from git. We want `input.rs` to get merged in.

Gyan-max:
@a-frantz,

I've now explicitly removed the `utils.rs` file from git tracking with `git rm src/utils.rs`. It was a bit confusing because I had deleted it locally in a previous commit, but it seems git was still tracking it. 

The situation is now fixed:
- `input.rs` is properly created and committed (with the improved structure separating format detection from data ownership)
- `utils.rs` is completely removed from the repository

a-frantz:
@Gyan-max can you move the contents of the `input` module into a `input::yaml` submodule? So you would move most of the contents of `src/input.rs` into a new `src/input/yaml.rs` file. Once you push that change, @DeepakSilaych you should fork this and write an `input::command_line` submodule. That should minimize `git` merge conflicts if we all do it right!

src/input.rs
a-frantz: @Gyan-max this has not been resolved. It should look something like:
```suggestion
pub enum InputFormat {
    /// JSON
    Json,
    /// YAML
    Yaml,
}
```

Gyan-max:
@a-frantz Sorry for the delay 
I have moved the contents of the input module into a yaml submodule as per your feedback.

- Created a new directory structire: `src/input/` with a `yaml.rs` file 
- Moved all the code from `src/input.rs` to `src/input/yaml.rs`
- created a `src/input/mode.rs` file that re-exports the functionality from the yaml submodule
- removed the old `src/input.rs` file
- verified all tests pass with new structure and formatted the code according to rust's style guidelines.

do let me know if any other changes are needed.

src/input/mod.rs
a-frantz, line 1: This is just a style nit: we prefer to structure our rust crates without `mod.rs` files. So please run `git mv src/input/mod.rs src/input.rs` and everything should work the same

src/input/yaml.rs
```diff
@@ -0,0 +1,211 @@
...
+            serde_json::to_string_pretty(&json_value)
+                .context("Failed to convert JSON value to string")
+        }
+        InputFormat::Yaml => {
+            debug!("Using YAML input converted to JSON");
+            serde_json::to_string_pretty(&json_value)
+                .context("Failed to convert YAML-derived JSON value to string")
+        }
+    }
+}
```
a-frantz: This isn't being used, correct? Please delete it. We don't like to have "dead code" but because you made this `pub` it isn't being flagged.

src/input/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! Input processing functionality for Sprocket.
+
+pub mod yaml;
+
+// Re-export the main functionality from the yaml module
+pub use yaml::InputFormat;
+pub use yaml::get_json_file_path;
+pub use yaml::get_json_string;
+pub use yaml::parse_input_file;
```
a-frantz:
```suggestion
pub use yaml::get_json_file_path;
```
Just export what is being used.

a-frantz (commented):
This is looking awesome! We're very close to merging this 🎉

Gyan-max:
@a-frantz addressed all your feedback.

src/input.rs
```diff
@@ -0,0 +1,9 @@
+//! Input processing functionality for Sprocket.
+
+pub mod yaml;
+
+// Re-export the main functionality from the yaml module
+pub use yaml::InputFormat;
+pub use yaml::get_json_file_path;
+// Removed unused export: pub use yaml::get_json_string;
+pub use yaml::parse_input_file;
```
a-frantz, line 9:
```suggestion
pub use yaml::get_json_file_path;
```
please commit this exactly

Gyan-max: `pub use yaml::get_json_file_path` is already exported @a-frantz .
I'm sorry to bother you but can you clarify it

a-frantz (approved):

Cargo.toml
```diff
@@ -30,3 +30,8 @@ clap-verbosity-flag = "2.2.2"
 pretty_assertions = "1.4.1"
 url = "2.5.4"
 chrono = "0.4.39"
+serde_yaml = "0.9.32"
```
peterhuene, line 33: Unfortunately, `serde-yaml` is no longer maintained and we shouldn't be introducing a dependency on it.

It appears that [`serde-yaml-ng`](https://crates.io/crates/serde_yaml_ng) is a fork from an earnest contributor seeking to maintain the original `serde-yaml`.

src/input/yaml.rs
```diff
@@ -0,0 +1,159 @@
...
+    Json,
+    /// YAML format
+    Yaml,
+}
+
+/// Parses an input file as either JSON or YAML.
+///
+/// Tries to parse as JSON first, then YAML if JSON parsing fails.
+/// Returns the parsed data as a JSON value along with the detected format.
+pub fn parse_input_file(path: &PathBuf) -> Result<(InputFormat, serde_json::Value)> {
```
peterhuene, line 24: While this approach does work, it makes it harder to discover errors in the input file and requires several different parses of the input file in addition to creation of a translated file on disk when the input file is YAML.

An alternative to this approach would be:

* Determine the input file type by extension, supporting `json`, `yaml`, and `yml`, as those are the most popular extensions for these file types; doing it this way allows us to parse the inputs file exactly once and report errors based on the expected format. 
* Call `Inputs::parse_json` or `Inputs::parse_yaml` depending on the file extension; error for unsupported extension.
* In `wdl-engine`, rename `Inputs::parse` to `Inputs::parse_json`.
* In `wdl-engine`, implement a new method `Inputs::parse_yaml` that uses `serde-yaml-ng` to parse the file into a `serde_yaml_ng::Mapping` and calls `Inputs::parse_object`, much like `parse_json` does.
* Change `Inputs::parse_object` (and other methods it calls) to take a generic type parameter with a `: Map` bounds (see trait definition below) for abstracting the common operations needed.

It can be tricky to abstract the common functionality to a trait, but looking at `wdl-engine/src/inputs.rs`, generally we need the following:

```rust
trait Map {
    type Value: TryIntoValue;

    // These methods return `Result` items as YAML map keys aren't always string, unlike JSON
    // In the YAML implementation, if a key isn't a string, we return an error.
    fn iter(&self) -> impl Iterator<Item = Result<(&str, &Self::Value)>>;
    fn into_iter(self) -> impl Iterator<Item = Result<(String, Self::Value)>>;
}

impl Map for serde_json::Map<String, serde_json::Value> {
    type Value = serde_json::Value;

    // Implementation here
}

impl Map for serde_yaml_ng::Mapping {
    type Value = serde_yaml_ng::Value;

    // Implementation here
}

trait TryIntoValue {
    fn try_into(self) -> Result<wdl_engine::Value>;
}

impl TryIntoValue for serde_json::Value {
    // Implementation here using `serde_json::from_value`.
}

impl TryIntoValue for serde_yaml_ng::Value {
    // Implementation here using `serde_yaml_ng::from_value`
}
```

peterhuene (changes requested):
Hi @Gyan-max. Thank you very much for the contribution!

While I think this is a workable approach, I added some comments that I think we should discuss before proceeding with accepting this PR.

src/input/yaml.rs
a-frantz: @peterhuene I thought that we didn't want to add YAML support to the `wdl` crates? I advised this approach to @Gyan-max with that in mind.

peterhuene: I am the one to blame for that as I believe I advised you that this would be better implemented in Sprocket rather than in `wdl-engine` based solely on JSON being the only "official" input format in the WDL specification.

However, seeing the burden that decision places on the implementation in Sprocket (especially having to parse files multiple times and secretly converting YAML to JSON on disk to feed it into `wdl-engine`, coupled with the potential to swallow more meaningful errors in the input files), I'm left with the belief that we should elevate YAML to first class input support in `wdl-engine`.

I apologize if this caused undue work as I should have thought ahead as to what it meant for the current `wdl-engine` API to remain unchanged.

peterhuene:
@Gyan-max if we're all in agreement that YAML support is best added to `wdl-engine`, then my advice would be to leave this PR open, implement the changes in `wdl-engine`, and then come back to this PR and update it to take a `git` dependency on `wdl-engine` (which we'll update once we publish a new `wdl-engine` version to crates.io) so you add the file extension logic to call `parse_json` or `parse_yaml` into Sprocket.

src/input/yaml.rs
a-frantz: Peter, can you check out https://github.com/stjude-rust-labs/wdl/issues/347 ?

I made some suggestions there based on the premise we didn't want to add YAML support.

Gyan-max:
> k

sure.
so what should I do next?

Gyan-max:
> @Gyan-max if we're all in agreement that YAML support is best added to `wdl-engine`, then my advice would be to leave this PR open, implement the changes in `wdl-engine`, and then, once those changes have merged into `wdl-engine`, come back to this PR and update it to take a `git` dependency on `wdl-engine` (which we'll update once we publish a new `wdl-engine` version to crates.io) so you add the file extension logic to call `parse_json` or `parse_yaml` into Sprocket.

sure.
so what should I do next?

peterhuene:
> so what should I do next?

I would start by changing the `Inputs` type in `wdl-engine` as described above and then submitting a PR to the `wdl` repository. You can reach out to me on Slack if you run into any issues.

Gyan-max:
> > so what should I do next?
> 
> I would start by changing the `Inputs` type in `wdl-engine` as described above and then submitting a PR to the `wdl` repository. You can reach out to me on Slack if you run into any issues.

Got it, @peterhuene! I'll start working on modifying the Inputs type in wdl-engine as you suggested, implementing the trait-based approach for handling both JSON and YAML formats.
I'll create a PR to the wdl repository once I've completed the changes. Looking forward to implementing this properly at the right layer of the architecture.
Thanks for the guidance. I'll reach out on Slack if I run into any challenges during implementation.

peterhuene:
I think we can now re-implement this PR in terms of updating the option text for `--inputs` now that `wdl-engine` supports YAML input files automatically.

Gyan-max:
> I think we can now re-implement this PR in terms of updating the option text for `--inputs` now that `wdl-engine` supports YAML input files automatically.

could you clarify more.

Gyan-max:
> This PR is being superseded by #79
> 
> @Gyan-max appreciate the contribution, but the vast majority of these existing changes are rendered unnecessary by [stjude-rust-labs/wdl#352](https://github.com/stjude-rust-labs/wdl/pull/352) so I've just pushed up the necessary tweaks in a fresh PR

okey
let me know if i can contribute other ways.

PR #74, opened 2025-03-15
Resolves #77 improved the get_display_config function

### LINKED ISSUE #77: perf: each call to `emit_diagnostics()` allocates a `Config` struct
Currently, the function get_display_confg creates a new Config struct every time its called, even though the configuration remains unchanged. leads to unnecessary allocations whenever get_display_config() is invoked.

vishwajeet1729:
### Introduce lazy_static! for Global Diagnostic Configurations to Ensure Consistency and Avoid Redundant Call/Allocations
files changed : lib.rs, cargo.toml

this pr resolves #77 
### Changes in this PR:
1)Introduced lazy_static! for defining FULL_CONFIG and ONE_LINE_CONFIG.
2)Updated get_display_config to use these predefined configurations.
3)fixed missing dependency by adding lazy_static to Cargo.toml.

previous code :
   fn get_display_config(report_mode: Mode, no_color: bool) -> (Config, StandardStream) {
    let display_style = match report_mode {
        Mode::Full => DisplayStyle::Rich,
        Mode::OneLine => DisplayStyle::Short,
    };

    let config = Config {
        display_style,
        ..Default::default()
    };

    let color_choice = if no_color {
        ColorChoice::Never
    } else {
        ColorChoice::Always
    };

    let writer = StandardStream::stderr(color_choice);

    (config, writer)
}

changes made by me :
use lazy_static::lazy_static;

lazy_static! {
    static ref FULL_CONFIG: Config = Config {
        display_style: DisplayStyle::Rich,
        ..Default::default()
    };
    static ref ONE_LINE_CONFIG: Config = Config {
        display_style: DisplayStyle::Short,
        ..Default::default()
    };
}

fn get_display_config(report_mode: Mode, no_color: bool) -> (Config, StandardStream) {
    let config = match report_mode {
        Mode::Full => FULL_CONFIG.clone(),
        Mode::OneLine => ONE_LINE_CONFIG.clone(),
    };

    let color_choice = if no_color {
        ColorChoice::Never
    } else {
        ColorChoice::Always
    };

    let writer = StandardStream::stderr(color_choice);

    (config, writer)
}

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,10 +34,13 @@ jobs:
         password: ${{ secrets.GITHUB_TOKEN }}
     - uses: docker/setup-qemu-action@v3
     - uses: docker/setup-buildx-action@v3
+      with:
+        platforms: linux/amd64,linux/arm64
     - uses: docker/build-push-action@v6
       with:
         push: true
         tags: ${{ env.REGISTRY }}/${{ github.repository_owner }}/sprocket:${{ github.ref_name }}
+        platforms: linux/amd64,linux/arm64
   build_artifacts:
     runs-on: ${{ matrix.os }}
     strategy:


vishwajeet1729:
@claymcleod can you please review this ?

a-frantz:
@vishwajeet1729 what compilation errors is this solving? I'm not aware of any issues that this resolves.

vishwajeet1729:
@a-frantz  , I was referring to importing lazy_static or adding it to Cargo.toml, but that was mistake from my side .

This PR fixes redundant calls in get_display_config(). Currently, the function creates a new Config struct every time its called, even though the configuration remains unchanged.  leads to unnecessary allocations whenever get_display_config() is invoked.

https://www.reddit.com/r/learnrust/comments/knd3j7/global_config_best_practice/
here i found information about it .
is i am wrong ?

a-frantz:
> @a-frantz  , I was referring to importing lazy_static or adding it to Cargo.toml, but that was mistake from my side,sorry .
> 
> This PR fixes redundant calls in get_display_config(). Currently, the function creates a new Config struct every time its called, even though the configuration remains unchanged.  leads to unnecessary allocations whenever get_display_config() is invoked.
> 
> https://www.reddit.com/r/learnrust/comments/knd3j7/global_config_best_practice/
> here i found information about it .
> is i am wrong ? 

This would be a good change _if_ we were repeatedly calling `get_display_config()`. However, I just reviewed the codebase, and `get_display_config()` will be invoked a maximum of one time per run of the executable. This will not lead to any performance benefits for us and increases the size of the executable by pulling in a new dependency.

a-frantz:
> This would be a good change _if_ we were repeatedly calling `get_display_config()`. However, I just reviewed the codebase, and `get_display_config()` will be invoked a maximum of one time per run of the executable. This will not lead to any performance benefits for us and increases the size of the executable by pulling in a new dependency.

Apologies, I did the above search of the codebase from mobile and misread our source. `get_display_config()` is in fact called multiple times in both the `format` and `check/lint` commands, because we call `emit_diagnostics` once per source file and those commands operate on multiple source files.

However, I'm not sure introducing a new dependency is the optimal solution here. We may instead just want to refactor `emit_diagnostics()` to avoid the redundant allocations you point out without pulling in a new crate.

Can you please open an issue for us to discuss this in more detail? Maybe we will use your fix here, but I want to think through it a bit more first.

In the future, please create an issue for discussion prior to implementing a new PR.

PR #78, opened 2025-03-17
perf: use static `Config` instead of new allocations

### LINKED ISSUE #77: perf: each call to `emit_diagnostics()` allocates a `Config` struct

vishwajeet1729:
**This PR refactors the diagnostic configuration in the Sprocket CLI tool by introducing static, lazily initialized Config instances. Instead of creating a new Config on every function call, we now reference pre-initialized static variables**
 
[Resolves #77 ](https://github.com/stjude-rust-labs/sprocket/issues/77)    
file changed :lib.rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ readme = "README.md"
 anyhow = "1.0.91"
 clap = { version = "4.5.20", features = ["derive", "string"] }
 codespan-reporting = "0.11.1"
+once_cell = "1.19"
 git-testament = "0.2.5"
 indexmap = "2.6.0"
 nonempty = "0.10.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ readme = "README.md"
 anyhow = "1.0.91"
 clap = { version = "4.5.20", features = ["derive", "string"] }
 codespan-reporting = "0.11.1"
-once_cell = "1.19"
 git-testament = "0.2.5"
 indexmap = "2.6.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ readme = "README.md"
 anyhow = "1.0.91"
 clap = { version = "4.5.20", features = ["derive", "string"] }
 codespan-reporting = "0.11.1"
+once_cell = "1.19"
```
a-frantz: This can be achieved using `std::sync::LazyLock` without adding a new dependency.

a-frantz:
this is still failing the format check. The easiest way to fix this is to just run `cargo +nightly fmt` from the root of the repo.

vishwajeet1729:
thank you , i was following the documentation ,i think need to update int documentation for Tests in sprocket

src/lib.rs
```diff
@@ -61,7 +70,7 @@ fn get_display_config(report_mode: Mode, no_color: bool) -> (Config, StandardStr
     (config, writer)
 }
 
-/// Emits the given diagnostics to the terminal.
+///  Emits the given diagnostics to the terminal.
```
a-frantz:
```suggestion
/// Emits the given diagnostics to the terminal.
```

peterhuene (approved):
Looks great, thanks!

PR #86, opened 2025-04-08
fix: ignore failure to `publish` to crates.io

a-frantz:

continue on publish error.

Using `|| true` feels a bit hacky here, but there aren't great alternatives in the GH workflow syntax. We could alternatively use `if: always()` or similar in subsequent steps, but this seems mildly preferable.

Also of note, there's no flag to `cargo publish` that can be used to achieve this, which seems like a missing feature if you ask me

GH refs:
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions
- https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Update lock file
       run: cargo update
     - name: Publish to crates.io
-      run: cargo publish
+      run: cargo publish || true # Ignore errors if the version is already published
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
     - name: Create a GH release


a-frantz:
making a separate comment for visibility (because I know we sometimes miss things in PR descriptions)

A more involved alternative would be to do a similar check as what we have in the `wdl` crates - https://github.com/stjude-rust-labs/wdl/blob/main/ci/src/main.rs#L336

This would be querying crates.io prior to running `cargo publish` and checking for the current version, and then skipping `cargo publish` and continuing with the rest of the release. This would technically be more robust to other potential errors that might slip through by my `|| true` use. Seems like more work for a near identical functionality, but willing to write it up if that's what we want 🤷‍♀️

peterhuene:
I think I'd prefer to duplicate what we have in the `wdl` repository for robustness, but I'm OK with this change as a temporary solution.

peterhuene (approved):
Approved under consideration of my previous comment.

a-frantz:
issue logged #87

Issue #38, opened 2024-10-23
feature request: warn users about an unknown `--except` rule


a-frantz:
Current behavior is essentially ignoring bad values to `--except`. They are assumed to be valid and never checked. The result of passing an unknown rule to `--except` is the same as just not specifying the arg at all. Is this desired behavior?

a-frantz:
Thoughts on this @adthrasher @peterhuene ?

I think my and Peter talked about this back when the `--except` flag was first added, but I'm not sure we made the right call anymore.

adthrasher:
I'd probably go with a `warning`. It's worth noting to the user that a specified rule doesn't exist, but I don't think it's necessarily an error.

a-frantz:
> I'd probably go with a `warning`. It's worth noting to the user that a specified rule doesn't exist, but I don't think it's necessarily an error.

Should it just be prepended to the list of printed diagnostics? I assume that if we don't consider it an error, we'll continue with the full lint analysis. That would be a special case, as I don't think sprocket currently emits any diagnostics of its own.

a-frantz:
Desired behavior for this: a diagnostic of severity `warning` should be at the top of the list of emitted diagnostics, and the `check`/`lint` should continue as otherwise.

dead8309:
can i work on this?
correct order is warning->check->lint?

a-frantz:
> can i work on this?
> correct order is warning->check->lint?

Yes to the first question, not quite to the second!

The ordering of the `check` or `lint` diagnostics is controlled by the `wdl` crates (more specifically, the `Diagnostic` struct used by the crates has sorting logic attached to it).

The warning that we should emit for a bad rule name to `except` is going to originate completely from `sprocket`, and doesn't need to use the `Diagnostic` struct from the `wdl` crates. It shouldn't be sorted along with the rest of the diagnostics, and should just always come first.

I'm not sure the exact right way to do it, but there should be a way to create a diagnostic from the `codespan-reporting` crate we use under the hood (so that the new diagnostic you're going to create looks similar to the rest of the diagnostics emitted).

(Typing from my phone, not checking the codespan API right now. But if you need more help, start a thread in slack!)

Issue #91, opened 2025-04-09
bug: `ShellCheck` rule is excluded from `explain`


a-frantz:
I don't think we actually need to push a fix for this, because as part of https://github.com/stjude-rust-labs/wdl/pull/341 the "optional_rules()" (that only contains `ShellCheck`) is being removed. #341 is _hopefully_ going to get some attention from me and merge before the next release.

PR #94, opened 2025-04-11
feat: warn users about unknown --except rule

dead8309:
closes: #38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Added
+### Changed
 
 * Unknown `--except` rules will be shown at the top of the list ([#94](https://github.com/stjude-rust-labs/sprocket/pull/94))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-* Unknown `--except` rules will be shown at the top of the list ([#94](https://github.com/stjude-rust-labs/sprocket/pull/94))
+* Unknown `--except` rules will now emit a warning instead of being silently ignored ([#94](https://github.com/stjude-rust-labs/sprocket/pull/94))
 
 ## 0.11.0 - 04-01-2025
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* Unknown `--except` rules will be shown at the top of the list ([#94](https://github.com/stjude-rust-labs/sprocket/pull/94))
+
 ## 0.11.0 - 04-01-2025
 
 ### Added
--- a/src/commands/check.rs
+++ b/src/commands/check.rs
@@ -166,7 +166,6 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
             let warning = codespan_reporting::diagnostic::Diagnostic::warning()
                 .with_message(message)
                 .with_notes(vec![
-                    // "rule names are case-insensitive".to_string(),
                     "run `sprocket explain --help` to see available rules".to_string(),
                 ]);
 
--- a/src/commands/check.rs
+++ b/src/commands/check.rs
@@ -1,18 +1,24 @@
 //! Implementation of the check and lint commands.
 
+use std::collections::HashSet;
 use std::fs;
 
 use anyhow::Context;
 use anyhow::bail;
 use clap::Parser;
+use codespan_reporting::files::SimpleFiles;
+use codespan_reporting::term;
 use url::Url;
+use wdl::analysis;
 use wdl::ast::AstNode;
 use wdl::ast::Diagnostic;
 use wdl::ast::Severity;
 use wdl::cli::analyze;
+use wdl::lint;
 
 use crate::Mode;
 use crate::emit_diagnostics;
+use crate::get_display_config;
 
 /// Common arguments for the `check` and `lint` subcommands.
 #[derive(Parser, Debug)]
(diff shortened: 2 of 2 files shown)


src/commands/check.rs
```diff
@@ -119,9 +124,61 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
+    if !exceptions.is_empty() {
+        let analysis_rules: HashSet<String> = analysis::rules()
+            .iter()
+            .map(|r| r.id().to_lowercase())
+            .collect();
+        let lint_rules: HashSet<String> = lint::rules()
+            .iter()
+            .map(|r| r.id().to_lowercase())
+            .collect();
+        let all_rules: HashSet<_> = analysis_rules.union(&lint_rules).collect();
```
a-frantz, line 137: This misses `ShellCheck`! Then again, as stated in #91 the `optional_rules()` is going away in the next release, so it might be ok to ignore for now.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* Unknown `--except` rules will be shown at the top of the list ([#94](https://github.com/stjude-rust-labs/sprocket/pull/94))
```
```suggestion
### Changed

* Unknown `--except` rules will now emit a warning instead of being silently ignored ([#94](https://github.com/stjude-rust-labs/sprocket/pull/94))
```

src/commands/check.rs
```diff
@@ -166,28 +223,35 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
+                    &uri,
+                    &result.document().root().inner().text().to_string(),
+                    args.common.report_mode,
+                    args.common.no_color,
+                );
+
+                for diagnostic in diagnostics.iter() {
+                    match diagnostic.severity() {
+                        Severity::Error => error_count += 1,
+                        Severity::Warning if !suppress => warning_count += 1,
```
a-frantz, line 248: I'm not sure, but I wonder if the "unknown except" warning should be counted here? Probably not, in which case we should add a comment here explicitly saying we don't add it to the count

src/commands/check.rs
```diff
@@ -119,9 +124,61 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
+                        .iter()
+                        .map(|r| format!("`{r}`"))
+                        .collect::<Vec<_>>()
+                        .join(", ")
+                )
+            };
+
+            let warning = codespan_reporting::diagnostic::Diagnostic::warning()
+                .with_message(message)
+                .with_code("UnknownExceptionRule")
```
a-frantz: I think this should be removed. A user might see this and expect to be able to `--except UnknownExceptionRule` which won't work.

So no `with_code()` here is my thinking.

a-frantz (commented):
just some small nits. This looks good 👍

a-frantz:
> skeptical if this is correct wdl, but didn't got any diagnostics

the `command` won't work as you probably intended, but the WDL is all good 👍 
Haven't actually run it, but as I understand: in the `command` `${name}` will be interpreted as a Bash variable that doesn't exist. But since you didn't `set -eu`, it won't error, it will just silently substitute nothing, resulting in an STDOUT of `Hello, !`

src/commands/check.rs
```diff
@@ -119,9 +124,60 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
+            } else {
+                format!(
+                    "ignoring unknown rules provided via --except: '{}'",
+                    unknown_exceptions
+                        .iter()
+                        .map(|r| format!("`{r}`"))
+                        .collect::<Vec<_>>()
+                        .join(", ")
+                )
+            };
```
a-frantz: can you collapse these into one `format!` call? something like
```rust
format!("ignoring unknown rule{s} provided via --except: '{rules}'"`, s=..., rules=...)
```

src/commands/check.rs
```diff
@@ -119,9 +124,60 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
+
+            let warning = codespan_reporting::diagnostic::Diagnostic::warning()
+                .with_message(message)
+                .with_notes(vec![
+                    "Rule names are case-insensitive".to_string(),
```
a-frantz: IMO this doesn't  belong as part of the diagnostic. I'd remove this (but is this specified in any of the doc strings? We should make that clear in the `clap` doc strings if it isn't)

dead8309: I thought since `explain` can take case insensitive rules, so check could also take one

a-frantz: just checked with my local `sprocket` install, and it appears we do require exact capitalization for `--except` args, although I think you're right that we should probably relax that. I'll create an issue

a-frantz: https://github.com/stjude-rust-labs/sprocket/issues/95

a-frantz: @dead8309 remove this line about case insensitivity (which isn't true)

src/commands/check.rs
```diff
@@ -119,9 +124,58 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
+            .collect();
+        let lint_rules: HashSet<String> = lint::rules()
+            .iter()
+            .map(|r| r.id().to_lowercase())
+            .collect();
+        let all_rules: HashSet<_> = analysis_rules.union(&lint_rules).collect();
+
+        let unknown_exceptions: Vec<&str> = exceptions
+            .iter()
+            .filter(|rule| !all_rules.contains(&rule.to_lowercase()))
```
```suggestion
            .filter(|rule| !all_rules.contains(&rule))
```

dead8309: but `all_rules` are in lowercase

```rust
        let analysis_rules: HashSet<String> = analysis::rules()
            .iter()
            .map(|r| r.id().to_lowercase())
            .collect();
        let lint_rules: HashSet<String> = lint::rules()
            .iter()
            .map(|r| r.id().to_lowercase())
            .collect();
        let all_rules: HashSet<_> = analysis_rules.union(&lint_rules).collect();

        let unknown_exceptions: Vec<&str> = exceptions
            .iter()
            .filter(|rule| !all_rules.contains(&rule.to_lowercase()))
            .map(|rule| rule.as_str())
            .collect();
```

a-frantz: ah I missed that. Please remove those calls to `to_lowercase()` as well. We can add them back when we make the next `wdl` crate release.

src/commands/check.rs
```diff
@@ -119,9 +124,58 @@ pub async fn check(args: CheckArgs) -> anyhow::Result<()> {
...
+                    .map(|r| format!("`{r}`"))
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            );
+
+            // TODO: revisit case-insensitive rule handling when wdl crates are updated
+            let warning = codespan_reporting::diagnostic::Diagnostic::warning()
+                .with_message(message)
+                .with_notes(vec![
+                    // "rule names are case-insensitive".to_string(),
```
```suggestion
```

a-frantz (approved):

claymcleod:
@a-frantz can this be merged?

PR #96, opened 2025-04-14
ci: add stale PR closure

adthrasher:
There are a number of stale PRs. This copies the same logic we use in the `wdl` repo.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,40 @@
+name: Close stale issues
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,40 @@
+name: Close stale issues
```
a-frantz:
```suggestion
name: Close stale PRs
```

adthrasher: Good catch, I also updated the `gh` calls from `issue` to `pr` below. It doesn't seem to matter as GitHub handles that seamlessly, but might as well be correct.

PR #98, opened 2025-04-14
ci: fixes multi-platform Docker builds

### LINKED ISSUE #97: Multi-platform builds
The current Sprocket build logic is broken. It builds Docker images for multiple platforms, but doesn't reconcile the manifest. So the last image pushed is the only one that this available in GHCR. Right now, that means there is only an arm image.

claymcleod:
This PR adds manifest merging across the multi-platform Docker builds (as described [here](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners)) to fix the issue described in #97. You can see a successful build [here](https://github.com/claymcleod/sprocket/actions/runs/14450087515).

Closes #97.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,36 +38,113 @@ jobs:
     - uses: actions/checkout@v4
     - name: Update Rust
       run: rustup update stable && rustup default stable
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/sprocket
     - name: Log in to the Container registry
       uses: docker/login-action@v3
       with:
         registry: ${{ env.REGISTRY }}
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-    - uses: docker/setup-buildx-action@v3
-    - uses: docker/build-push-action@v6
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+    - name: Build and push by digest
+      id: build
+      uses: docker/build-push-action@v6
       with:
-        push: true
-        tags: ${{ env.REGISTRY }}/${{ github.repository_owner }}/sprocket:${{ github.ref_name }}
         platforms: linux/amd64
+        labels: ${{ steps.meta.outputs.labels }}
+        tags: ${{ env.REGISTRY }}/${{ github.repository_owner }}/sprocket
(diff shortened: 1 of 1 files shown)


adthrasher (approved):
This should be good for now. I think we'll want to revisit this later and get the multi-platform builds running in parallel using native build images.

claymcleod:
@adthrasher I updated my approach to just merge the manifests. Would you mind taking another look?

.github/workflows/release.yml
```diff
@@ -38,36 +38,113 @@ jobs:
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/sprocket
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
```
adthrasher: Do we need these tags? This only runs for a tag push, so these won't trigger, right?

.github/workflows/release.yml
```diff
@@ -38,36 +38,113 @@ jobs:
...
+            type=semver,pattern=v{{version}}
```
adthrasher: Do we want to add any labels as part of the build? Something like this, possibly with additional fields?

```suggestion
            type=semver,pattern=v{{version}}
        labels: |
            org.opencontainers.image.source=https://github.com/${{ github.repository }}
```

claymcleod: I've added this.

.github/workflows/release.yml
claymcleod: Looks like I just forgot to remove them from this one.

.github/workflows/release.yml
```diff
@@ -38,36 +38,113 @@ jobs:
...
+        uses: docker/setup-buildx-action@v3
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/sprocket
+          tags: type=semver,pattern=v{{version}}
+          labels: |
+            org.opencontainers.image.source=https://github.com/${{ github.repository }}
+            oorg.opencontainers.image.version=v{{version}}
```
adthrasher:
```suggestion
            org.opencontainers.image.version=v{{version}}
```

adthrasher (approved):

PR #102, opened 2025-04-26
feat: introduces incremental command line parsing

claymcleod:
This commit is the second half of the changes needed to land incremental command line parsing into `sprocket`. The first half is at https://github.com/stjude-rust-labs/wdl/pull/430.

As indicated in the changes there, this commit also has the side effect of removing the `wdl` command line tool and putting all of the subcommands in `sprocket` instead. As such, you'll see many commands be modified to use the code created in the new `wdl-cli` package _OR_ brand new subcommands (like `run`).

For external contributors:

For all contributors:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,21 +12,30 @@ readme = "README.md"
 
 [dependencies]
 anyhow = "1.0.91"
-clap = { version = "4.5.20", features = ["derive", "string"] }
-codespan-reporting = "0.11.1"
+chrono = "0.4.41"
+clap = { version = "4.5.37", features = ["derive", "string"] }
+clap-verbosity-flag = { version = "3.0.2", features = ["tracing"] }
+codespan-reporting = "0.12.0"
+colored = "3.0.0"
+futures = "0.3.31"
 git-testament = "0.2.5"
-indexmap = "2.6.0"
-nonempty = "0.10.0"
-pest = { version = "2.7.14", features = ["pretty-print"] }
-tracing = "0.1.40"
-tracing-subscriber = "0.3.18"
-wdl = { version = "0.12.0", features = ["cli", "lsp"] }
-walkdir = "2.5.0"
-colored = "2.1.0"
-tokio = { version = "1.41.0", features = ["full"] }
-tracing-log = "0.2.0"
+indexmap = "2.9.0"
 indicatif = "0.17.8"
-clap-verbosity-flag = "2.2.2"
+nonempty = "0.11.0"
+pest = { version = "2.7.14", features = ["pretty-print"] }
 pretty_assertions = "1.4.1"
+serde_json = "1.0.140"
+tokio = { version = "1.41.0", features = ["full"] }
+tokio-util = "0.7.15"
+toml = "0.8.22"
+tracing = "0.1.40"
+tracing-indicatif = "0.3.9"
+tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 url = "2.5.4"
-chrono = "0.4.39"
+walkdir = "2.5.0"
+wdl = { version = "0.13.1", features = [
+    "cli",
+    "codespan",
+    "engine",
+    "lsp",
+] }
--- a/src/commands/check.rs
+++ b/src/commands/check.rs
@@ -1,20 +1,18 @@
-//! Implementation of the check and lint commands.
+//! Implementation of the `check` and `lint` subcommands.
 
 use std::collections::HashSet;
-use std::fs;
 
 use anyhow::Context;
 use anyhow::bail;
 use clap::Parser;
+use codespan_reporting::diagnostic::Diagnostic;
 use codespan_reporting::files::SimpleFiles;
-use codespan_reporting::term;
-use url::Url;
-use wdl::analysis;
+use tracing::info;
 use wdl::ast::AstNode;
-use wdl::ast::Diagnostic;
 use wdl::ast::Severity;
-use wdl::cli::analyze;
-use wdl::lint;
+use wdl::cli::Analysis;
+use wdl::cli::analysis::Source;
+use wdl::lint::find_nearest_rule;
 
 use crate::Mode;
 use crate::emit_diagnostics;
@@ -24,16 +22,19 @@ use crate::get_display_config;
 #[derive(Parser, Debug)]
 #[command(author, version, about)]
 pub struct Common {
-    /// The file, URL, or directory to check.
-    #[arg(required = true)]
+    /// A set of source documents as files, directories, or URLs.
     #[clap(value_name = "PATH or URL")]
-    pub file: String,
+    pub sources: Vec<Source>,
 
-    /// A single rule ID to except from running.
+    /// Excepts (ignores) an analysis or lint rule.
     ///
-    /// Can be specified multiple times.
-    #[arg(short, long, value_name = "RULE")]
+    /// Repeat the flag multiple times to except multiple rules.
+    #[clap(short, long, value_name = "RULE")]
     pub except: Vec<String>,
+
+    /// Enable `shellcheck` lints.
+    #[clap(long, requires = "lint")]
+    pub shellcheck: bool,
 
     /// Causes the command to fail if warnings were reported.
     #[clap(long)]
@@ -43,12 +44,12 @@ pub struct Common {
     #[clap(long)]
     pub deny_notes: bool,
 
-    /// Supress diagnostics from imported documents.
+    /// Supress diagnostics from documents that were not explicitly provided in
+    /// the sources list (i.e., were imported from a provided source).
     ///
-    /// This will only display diagnostics for the document specified by `file`.
-    /// If specified with a directory, an error will be raised.
+    /// If the sources list contains a directory, an error will be raised.
     #[arg(long)]
-    pub single_document: bool,
+    pub suppress_imports: bool,
 
     /// Show diagnostics for remote documents.
     ///
@@ -57,14 +58,6 @@ pub struct Common {
     /// This flag has no effect when checking a remote document.
     #[arg(long)]
     pub show_remote_diagnostics: bool,
-
-    /// Run the `shellcheck` program on command sections.
-    ///
-    /// Requires linting to be enabled. This feature is experimental.
-    /// False positives may be reported.
-    /// If `shellcheck` is not installed, an error will be raised.
-    #[arg(long)]
-    pub shellcheck: bool,
 
     /// Hide diagnostics with `note` severity.
     #[arg(long)]
@@ -87,7 +80,7 @@ pub struct CheckArgs {
     #[command(flatten)]
     pub common: Common,
 
-    /// Perform lint checks in addition to checking for errors.
+    /// Enable lint checks in addition to validation errors.
     #[arg(short, long)]
     pub lint: bool,
 }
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -12,21 +12,32 @@ readme = "README.md"
...
+tracing-indicatif = "0.3.9"
 tracing-log = "0.2.0"
-indicatif = "0.17.8"
-clap-verbosity-flag = "2.2.2"
-pretty_assertions = "1.4.1"
+tracing-subscriber = "0.3.18"
 url = "2.5.4"
-chrono = "0.4.39"
+walkdir = "2.5.0"
+wdl = { git = "https://github.com/claymcleod/wdl-crates.git", features = [
```
claymcleod: This needs updating.

src/commands/check.rs
```diff
@@ -1,40 +1,34 @@
...
-use crate::get_display_config;
 
 /// Common arguments for the `check` and `lint` subcommands.
 #[derive(Parser, Debug)]
 #[command(author, version, about)]
 pub struct Common {
-    /// The file, URL, or directory to check.
-    #[arg(required = true)]
+    /// A set of source WDL files (files or URLs).
     #[clap(value_name = "PATH or URL")]
```
a-frantz: should specify directories are supported

src/commands/check.rs
```diff
@@ -1,40 +1,34 @@
...
-    pub file: String,
+    pub sources: Vec<Source>,
 
-    /// A single rule ID to except from running.
-    ///
-    /// Can be specified multiple times.
-    #[arg(short, long, value_name = "RULE")]
+    /// Excepts (ignores) an analysis or lint rule.
```
a-frantz, line 29:
```suggestion
    /// Excepts (ignores) an analysis or lint rule.
    ///
    /// Repeat argument to except multiple rules.
```
this clarification was added to make it clear the syntax is `-e Foo -e Bar` instead of `-e Foo Bar`.

src/commands/check.rs
```diff
@@ -1,40 +1,34 @@
...
+    #[clap(long)]
```
```suggestion
    #[clap(short, long)]
```

src/commands/check.rs
```diff
@@ -43,12 +37,12 @@ pub struct Common {
...
-    /// Supress diagnostics from imported documents.
+    /// Supress diagnostics from documents that were not explicitly provided in
+    /// the sources list.
     ///
-    /// This will only display diagnostics for the document specified by `file`.
-    /// If specified with a directory, an error will be raised.
+    /// If the sources list contains a directory, an error will be raised.
     #[arg(long)]
-    pub single_document: bool,
+    pub explicitly_scoped: bool,
```
a-frantz: not a huge fan of `explicitly_scoped`. Seems overly technical for a CL arg? Maybe `suppress_imports`?

src/commands/check.rs
```diff
@@ -101,180 +87,132 @@ pub struct LintArgs {
...
-                .to_string_lossy()
-                .to_string(),
-            _ => {
-                if !remote_file && !args.common.show_remote_diagnostics {
-                    suppress = true;
+
+        match uri.scheme() {
+            "file" => {}
+            "http" | "https" => {
+                if !args.common.show_remote_diagnostics {
```
a-frantz: this will lead to confusing behavior if the `Source` is a remote file. Every diagnostic will be hidden by default, which will appear to be the "no diagnostic" case.

The old behavior should be maintained. (Check up front if a remote file is provided, in which case ignore the value of `show_show_remote_diagnostics`)

src/commands/check.rs
```diff
@@ -101,180 +87,132 @@ pub struct LintArgs {
...
-    let mut note_count = 0;
-    for result in &results {
-        let mut suppress = false;
+    for result in results {
+        let uri = &**result.document().uri();
 
-        // Attempt to strip the CWD from the result path
-        let uri = result.document().uri();
-        if args.common.single_document && !uri.as_str().contains(file) {
+        if args.common.explicitly_scoped && !source_uris.contains(uri) {
```
a-frantz: I think this will "hide" errors from imported documents, which we don't want to do. `check` should always show all `error` level diagnostics regardless of `explicitly_scoped`

src/commands/explain.rs
```diff
@@ -169,14 +192,14 @@ pub fn explain(args: Args) -> anyhow::Result<()> {
...
+                        bail!("no rule found with the name `{rule_name}`");
                     }
                 }
             }
         }
 
         Ok(())
     } else {
-        bail!("Invalid arguments: either a rule_name, a --tag, or --definitions must be provided");
+        bail!("invalid arguments: either a rule_name, a --tag, or --definitions must be provided");
```
a-frantz: I think this will `unreachable!()`? If not, the message needs to be updated to include `list_all_[rules,tags]`

src/commands/run.rs
```diff
@@ -0,0 +1,346 @@
...
+    /// If inputs are provided, this will be attempted to be inferred from the
+    /// prefixed names of the inputs (e.g, `<name>.<input-name>`).
+    ///
+    /// If no inputs are provided and this argument is not provided, it will be
+    /// assumed you're trying to run the workflow present in the specified WDL
+    /// document.
+    #[clap(short, long, value_name = "NAME")]
+    pub name: Option<String>,
+
+    /// The task execution output directory; defaults to the task name.
```
a-frantz: Is this the default behavior we want? Would something like `$PWD/sprocket/run/[timestamp]_$name/` be preferred?

src/commands/check.rs
claymcleod: I would consider `PATH` to include both files and directories.

src/commands/check.rs
claymcleod: What if, as a slightly better case, we simply error if the user provides a remote source when providing this flag? It seems this explicit complaining is probably better than implicitly overriding a user-provided argument, and I can't see a case where a user would intentionally want the behavior where their choice is overridden.

src/commands/check.rs
a-frantz: "Files or URLs" in the doc string implies to me that directories aren't supported

src/commands/check.rs
claymcleod: That sounds counter-intuitive to me. If I'm providing a flag called `--suppress-imports`, I'm counting on only seeing diagnostics for the documents I'm immediately providing on the command line, effectively ignoring everything outside that universe. If we want to have the behavior you specify here, we should call the flag something different (`--supress-import-notes` and/or `--suppress-import-warnings`).

Personally, I don't really understand why we have the flag in the first place, as this seems like a really niche area. Perhaps we should just remove the flag?

src/commands/run.rs
claymcleod: For now, I've kept the default of what Peter had in `wdl`. I think that's fine for the very first version, and we can obviously piecewise improve the process as we go along.

src/commands/check.rs
a-frantz: The default behavior on `main` is informed by usage in the `workflows` repo. Where all local files are maintained by the immediate user of `sprocket`, and remote files are maintained by someone else (SEA-Seq for example). This would also be the case if we started an `open-wdl/modules` repo. We don't want imports of external sources not maintained by the user to show diagnostics by default, as there's nothing they can do to fix it (aside from log an issue or fork it).

However when someone is linting a remote file, there's no (good) way to distinguish between a file the user cares about VS external imports, so we just ignore this flag and show all remote diagnostics.

src/commands/check.rs
a-frantz: I personally use it when I'm writing a workflow that imports a bunch of stuff. I don't want to see diagnostics from the files I'm not currently working on, so I pass `--single-document` to only show me lints from the workflow I'm writing.

Similarly, it's used in `sprocket-action` to more explicitly group diagnostics from the same file/filter out specific files the user doesn't want to see diagnostics from.

The reason to still show errors is that an error means that a `run` command is going to fail due to a bad import. I would be confused if `sprocket check` passes cleanly and then `sprocket run` failed.

a-frantz:
I think between #430 and this PR, we've lost the `doc` command. `doc` clearly isn't ready for "prime time", but we do need some CLI interface for development.

I was thinking about the above, and also how the `analyzer` command probably shouldn't be featured so prominently considering it's not useful to most end users, and came up with the idea to add a `sprocket dev` group of commands. So `sprocket dev doc/analyzer/<anything else?>`. This would also be where any "prototype" commands would go before we're ready to bring them into the "core functionality". "Advanced users" could still use them, but there's an implicit acknowledgment by gating behind `dev` that things might not be totally polished.

What do folks think of this?

src/commands/check.rs
claymcleod: Oh, okay the docstring. Got it. I was looking at the value name for some reason.

src/commands/check.rs
claymcleod: Alright, I still don't love it but I suppose it is reasonable.

src/commands/check.rs
claymcleod: Alright, I've updated this to now work as you outline.

claymcleod:
> I think between #430 and this PR, we've lost the `doc` command. `doc` clearly isn't ready for "prime time", but we do need some CLI interface for development.
> 
> I was thinking about the above, and also how the `analyzer` command probably shouldn't be featured so prominently considering it's not useful to most end users, and came up with the idea to add a `sprocket dev` group of commands. So `sprocket dev doc/analyzer/<anything else?>`. This would also be where any "prototype" commands would go before we're ready to bring them into the "core functionality". "Advanced users" could still use them, but there's an implicit acknowledgment by gating behind `dev` that things might not be totally polished.
> 
> What do folks think of this?

I'm okay with the idea for (a) having a `dev`/`experimental` namespace and (b) putting `docs` in it. I don't like the idea of hiding `analyzer` beneath it, as analyzer is totally ready for primetime and is one of our most commonly used commands right now (given that LSP support is one of our biggest strengths in the current version of Sprocket).

claymcleod:
I would recommend that, after these changes land, you add the changes underneath whatever name for the namespace we prefer. I like `experimental` myself—though a bit longer, it's more explicit that the subcommand is truly could just be totally dropped in the future, and I suspect shell autocompletion is going to mean it's not a lot of time wasted typing it out.

a-frantz:
> > I think between #430 and this PR, we've lost the `doc` command. `doc` clearly isn't ready for "prime time", but we do need some CLI interface for development.
> > 
> > I was thinking about the above, and also how the `analyzer` command probably shouldn't be featured so prominently considering it's not useful to most end users, and came up with the idea to add a `sprocket dev` group of commands. So `sprocket dev doc/analyzer/<anything else?>`. This would also be where any "prototype" commands would go before we're ready to bring them into the "core functionality". "Advanced users" could still use them, but there's an implicit acknowledgment by gating behind `dev` that things might not be totally polished.
> > 
> > What do folks think of this?
> 
> I'm okay with the idea for (a) having a `dev`/`experimental` namespace and (b) putting `docs` in it. I don't like the idea of hiding `analyzer` beneath it, as analyzer is totally ready for primetime and is one of our most commonly used commands right now (given that LSP support is one of our biggest strengths in the current version of Sprocket).

I was thinking we might prefer `dev` as it's broader in scope. For example, I was recently using the old `wdl parse` command for debugging something in my analysis refactor. We might want to keep some of those older/basic commands for debug purposes, in which case the `experimental` namespace doesn't make sense. Not a strong opinion, but I do like the extra latitude with a generic `dev` namespace.

claymcleod:
What do you think about potentially both? After some thought, I think there are distinct goals for commands that are used as part of development/are currently in development vs. commands we want to try out with a broader audience but don't yet want to fully commit to.

a-frantz:
There's a behavior regression from PR #94

Arguments to `-e` are no longer being checked

a-frantz:
> What do you think about potentially both? After some thought, I think there are distinct goals for commands that are used as part of development/are currently in development vs. commands we want to try out with a broader audience but don't yet want to fully commit to.

This makes sense to me. `dev` and `experimental` have different connotations. But I think `dev` covers the meaning of `experimental` well enough that we don't need both namespaces.

I could go either way. I'm curious if anyone else on the team has feelings about the distinction.

src/lib.rs
```diff
@@ -52,37 +66,73 @@ impl std::fmt::Display for Mode {
...
-fn get_display_config(report_mode: Mode, no_color: bool) -> (&'static Config, StandardStream) {
+/// Emits the given diagnostics to the terminal.
+fn emit_diagnostics<'a>(
+    path: &str,
+    source: String,
+    diagnostics: impl IntoIterator<Item = &'a Diagnostic>,
+    backtrace: &[CallLocation],
+    report_mode: Mode,
+    no_color: bool,
+) -> Result<usize> {
```
a-frantz: is this result used anywhere?

claymcleod:
> There's a behavior regression from PR #94
> 
> Arguments to `-e` are no longer being checked

Good catch.

claymcleod:
Remove Spellcheck args.

claymcleod:
https://github.com/stjude-rust-labs/wdl/pull/430#discussion_r2069198833

src/commands/check.rs
```diff
@@ -101,180 +90,158 @@ pub struct LintArgs {
...
-            file = file
-        );
+    if args.common.suppress_imports {
+        for source in args.common.sources.iter() {
+            if let Source::Directory(dir) = source {
+                return anyhow::__private::Err(anyhow::__anyhow!(
+                    "`--explicitly-scoped` was specified, but sources contained the directory \
+                     `{dir}` was specified",
+                    dir = dir.display()
+                ));
```
a-frantz: this error message needs updating

src/commands/run.rs
```diff
@@ -0,0 +1,351 @@
...
+
+/// Arguments to the `run` subcommand.
+#[derive(Parser, Debug)]
+#[clap(disable_version_flag = true)]
+pub struct Args {
+    /// A source WDL file or URL.
+    #[clap(value_name = "PATH or URL")]
+    pub source: Source,
+
+    /// The inputs passed in on the command line.
```
a-frantz: this should be more descriptive. Something like:
```suggestion
    /// The inputs for the task or workflow. Can be file paths or key value pairs.
```

src/commands/validate.rs
```diff
@@ -1,26 +1,38 @@
...
+    /// If no inputs are provided and this argument is not provided, it will be
+    /// assumed you're trying to validate the workflow present in the specified
+    /// document.
+    #[clap(short, long, value_name = "NAME")]
+    pub name: Option<String>,
 
-    /// The path to the input JSON or YAML file.
-    #[arg(short, long, value_name = "INPUTS")]
-    pub inputs: PathBuf,
+    /// The inputs passed in on the command line.
```
a-frantz: should be the same as the `run` inputs doc comment

src/commands/check.rs
```diff
@@ -101,184 +94,226 @@ pub struct LintArgs {
...
+    let mut counts = Counts::default();
 
-    let cwd = std::env::current_dir().ok();
-    let mut error_count = 0;
-    let mut warning_count = 0;
-    let mut note_count = 0;
-    for result in &results {
-        let mut suppress = false;
+    for result in results {
+        let uri = &**result.document().uri();
```
peterhuene: Nit: could we perhaps remove the `&**` and use `&` where we're passing it as a `&Url`?

src/commands/validate.rs
```diff
@@ -31,24 +46,73 @@ pub struct ValidateInputsArgs {
...
+    // above.
+    let document = results.filter(&[&args.source]).next().unwrap().document();
+
+    let inferred = Inputs::coalesce(args.inputs)?.into_engine_inputs(document)?;
+
+    let (name, inputs, _) = if let Some(inputs) = inferred {
+        inputs
+    } else {
+        // SAFETY: we expect that the current working directory is always
+        // available for the platforms that `wdl` will run within.
```
peterhuene: Nit: as we're returning `Result<()>` anyway, could we just do a "std::env::current_dir().context("failed to get current directory")?` instead of the `unwrap`?

src/commands/run.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// The task execution output directory; defaults to the task name if
+    /// provided, otherwise, `output`.
+    #[clap(short, long, value_name = "OUTPUT_DIR")]
+    pub output: Option<PathBuf>,
+
+    /// The path to the engine TOML configuration file.
+    #[clap(short, long, value_name = "CONFIG")]
+    pub config: Option<PathBuf>,
+
+    /// Overwrites the task execution output directory if it exists.
```
peterhuene:
```suggestion

    /// Overwrites the execution output directory if it exists.
```

src/commands/run.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// If inputs are provided, this will be attempted to be inferred from the
+    /// prefixed names of the inputs (e.g, `<name>.<input-name>`).
+    ///
+    /// If no inputs are provided and this argument is not provided, it will be
+    /// assumed you're trying to run the workflow present in the specified WDL
+    /// document.
+    #[clap(short, long, value_name = "NAME")]
+    pub name: Option<String>,
+
+    /// The task execution output directory; defaults to the task name if
```
peterhuene:
```suggestion
    /// The execution output directory; defaults to the task name if
```

src/commands/run.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+/// A callback for updating state based on engine events.
+fn progress(kind: ProgressKind<'_>, pb: &tracing::Span, state: &Mutex<State>) {
+    pb.pb_start();
+
+    let message = {
+        let mut state = state.lock().expect("failed to lock progress mutex");
+        match kind {
+            ProgressKind::TaskStarted { .. } => {
```
peterhuene:
```suggestion
            ProgressKind::TaskStarted { .. } | ProgressKind::TaskRetried { .. } => {
```

This was a bug fixed in my `tes` branch which didn't make it into the `wdl` executable since it had been deleted.

When a task was retried, we were removing it from the executing set, but it was showing the task as not ready while it was waiting in the queue to run on the TES server. These two events should just bump the ready count.

src/commands/run.rs
```diff
@@ -0,0 +1,354 @@
...
+        .unwrap_or_default();
+
+    let inferred = Inputs::coalesce(args.inputs)?.into_engine_inputs(document)?;
+
+    let (name, inputs, origins) = if let Some(inputs) = inferred {
+        inputs
+    } else {
+        // SAFETY: we expect that the current working directory is always
+        // available for the platforms that `wdl` will run within.
+        let origins = OriginPaths::from(std::env::current_dir().unwrap());
```
peterhuene: Same comment as above regarding just using `context(...)?` instead of `unwrap`.

peterhuene (approved):

a-frantz (approved):

PR #104, opened 2025-04-29
feat: configuration file support

adthrasher:
This adds support for reading configuration files containing options specified by the user.

For external contributors:

For all contributors:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Unknown `--except` rules will now emit a warning instead of being silently ignored ([#94](https://github.com/stjude-rust-labs/sprocket/pull/94))
 
+### Added
+
+* Added configuration file support ([#104](https://github.com/stjude-rust-labs/sprocket/pull/104)).
+
 ## 0.11.0 - 04-01-2025
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,8 @@ clap-verbosity-flag = { version = "3.0.2", features = ["tracing"] }
 clap_complete = "4.5.48"
 codespan-reporting = "0.12.0"
 colored = "3.0.0"
+dirs = "6.0.0"
+figment = { version = "0.10.19", features = ["toml"] }
 futures = "0.3.31"
 git-testament = "0.2.5"
 indexmap = "2.9.0"
@@ -25,6 +27,7 @@ indicatif = "0.17.8"
 nonempty = "0.11.0"
 pest = { version = "2.7.14", features = ["pretty-print"] }
 pretty_assertions = "1.4.1"
+serde = "1.0.219"
 serde_json = "1.0.140"
 tokio = { version = "1.41.0", features = ["full"] }
 tokio-util = "0.7.15"
@@ -33,8 +36,5 @@ tracing = "0.1.40"
 tracing-indicatif = "0.3.9"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 url = "2.5.4"
-figment = { version = "0.10.19", features = ["toml"] }
-serde = "1.0.219"
-dirs = "6.0.0"
 walkdir = "2.5.0"
 wdl = { version = "0.13.2", features = ["cli", "codespan", "engine", "lsp"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,3 +30,7 @@ clap-verbosity-flag = "2.2.2"
 pretty_assertions = "1.4.1"
 url = "2.5.4"
 chrono = "0.4.39"
+figment = { version = "0.10.19", features = ["toml"] }
+toml = "0.8.21"
+serde = "1.0.219"
+dirs = "6.0.0"
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@
 
 - **`sprocket analyzer`** Run Sprocket as a LSP server for IDE integration.
 - **`sprocket check`** Performs static analysis on WDL documents.
+- **`sprocket config`** Print effective configuration.
 - **`sprocket explain`** Explain lint rules.
 - **`sprocket format`** Formats WDL documents.
 - **`sprocket lint`** Performs static analysis on WDL documents with additional
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
 
 - **`sprocket analyzer`** runs Sprocket as a LSP server, which is useful for IDE integration.
 - **`sprocket check`** performs static analysis on a document or directory of documents.
-- - **`sprocket config`** prints effective configuration.
+- **`sprocket config`** prints effective configuration.
 - **`sprocket explain`** explains validation and lint rules supported by Sprocket.
 - **`sprocket format`** formats a document.
 - **`sprocket lint`** performs static analysis on a document or directory of documents with additional linting rules enabled (effectively a shortcut for `check --lint`).
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@
 - **`sprocket check`** performs static analysis on a document or directory of documents.
 - **`sprocket config`** prints configuration values.
 - **`sprocket explain`** explains validation and lint rules supported by Sprocket.
-- **`sprocket format`** formats a document.
+- **`sprocket format`** formats a document or directory of documents.
 - **`sprocket lint`** performs static analysis on a document or directory of documents with additional linting rules enabled (effectively a shortcut for `check --lint`).
 - **`sprocket run`** runs a task or workflow.
 - **`sprocket validate`** validates a set of inputs read from files or on the command line against a task or workflow.
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
 
 - **`sprocket analyzer`** runs Sprocket as a LSP server, which is useful for IDE integration.
 - **`sprocket check`** performs static analysis on a document or directory of documents.
-- **`sprocket config`** prints effective configuration.
+- **`sprocket config`** prints configuration values.
 - **`sprocket explain`** explains validation and lint rules supported by Sprocket.
 - **`sprocket format`** formats a document.
 - **`sprocket lint`** performs static analysis on a document or directory of documents with additional linting rules enabled (effectively a shortcut for `check --lint`).
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -29,3 +29,37 @@ The platform-specific locations can also be found [here](https://docs.rs/dirs/la
 ## Configuration Values
 
 Running the command `sprocket config resolve` will print the effective configuration. The default configuration can be written out using the `sprocket config init` argument.
+
+## Configuration Resolution
+
+Atomic configuration values are overwritten by higher priority configuration files. List values are appended.
+
+For example, if you specify the following in a configuration file in the current working directory.
+
+```
+[format]
+indentation_size = 5
+
+[check]
+except = ['ContainerUri']
+```
+
+And the following in a configuration file pointed to by `SPROCKET_CONFIG`.
+
+```
+[format]
+indentation_size = 3
+
+[check]
+except = ['SnakeCase']
+```
+
+The final, effective configuration will be:
+
+```
+[format]
+indentation_size = 3
+
+[check]
+except = ['ContainerUri', 'SnakeCase']
+```
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -18,7 +18,7 @@ Sprocket will look for a `sprocket.toml` in the current working directory when t
 
 ### XDG_CONFIG_HOME
 
-Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html).
+Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). On MacOS, Sprocket will attempt to read the configuration from `HOME/.config/sprocket/sprocket.toml`.
 
 ## Configuration Values
 
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -22,4 +22,4 @@ Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocke
 
 ## Configuration Values
 
-Running the command `sprocket config` will print the effective configuration. The default configuration can be written out using the `--generate` argument.
+Running the command `sprocket config resolve` will print the effective configuration. The default configuration can be written out using the `sprocket config init` argument.
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -22,4 +22,4 @@ Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocke
 
 ## Configuration Values
 
-Running the command `sprocket config` will print the effective configuration.
+Running the command `sprocket config` will print the effective configuration. The default configuration can be written out using the `--generate` argument.
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -18,7 +18,13 @@ Sprocket will look for a `sprocket.toml` in the current working directory when t
 
 ### XDG_CONFIG_HOME
 
-Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). On MacOS, Sprocket will attempt to read the configuration from `$HOME/.config/sprocket/sprocket.toml`.
+On systems that support the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/), such as many Linux distributions, Sprocket will attempt to read configuration from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. Please note that the location of `XDG_CONFIG_HOME` is operating system dependent.
+
+On MacOS, Sprocket will attempt to read the configuration from `$HOME/.config/sprocket/sprocket.toml`.
+
+On Windows, Sprocket will attempt to read configuration from `%USERPROFILE%\AppData\Roaming`.
+
+The platform-specific locations can also be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). 
 
 ## Configuration Values
 
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -63,3 +63,5 @@ indentation_size = 3
 [check]
 except = ['ContainerUri', 'SnakeCase']
 ```
+
+Configuration resolution can be disabled by passing the `--skip-config-search` option on the command line. This will disable searching for and loading configuration files. The only configuration loaded will be that (if) specified by the `--config` command line argument.
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -1,6 +1,6 @@
 # Configuration Files
 
-The Sprocket configuration loader sources from command-line options, environment variables, and configuration files, in that priority order.
+The Sprocket configuration loader sources and merges values from a location specified by a command-line option, an environment variable, and common configuration locations, in that priority order.
 
 ## Configuration Options
 
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -1,0 +1,25 @@
+# Configuration Files
+
+The Sprocket configuration loader sources from command-line options, environment variables, and configuration files, in that priority order.
+
+## Configuration Options
+
+### Command-line configuration
+
+A configuration file can be specified at runtime on the command line using the `--config` argument.
+
+### Environment Variable
+
+The path to a configuration file can be specified via the environment variable `SPROCKET_CONFIG`.
+
+### Current Working Directory
+
+Sprocket will look for a `sprocket.toml` in the current working directory when the `sprocket` command runs.
+
+### XDG_CONFIG_HOME
+
+Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html).
+
+## Configuration Values
+
+Running the command `sprocket config` will print the effective configuration.
--- a/docs/cli/config.md
+++ b/docs/cli/config.md
@@ -18,7 +18,7 @@ Sprocket will look for a `sprocket.toml` in the current working directory when t
 
 ### XDG_CONFIG_HOME
 
-Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). On MacOS, Sprocket will attempt to read the configuration from `HOME/.config/sprocket/sprocket.toml`.
+Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). On MacOS, Sprocket will attempt to read the configuration from `$HOME/.config/sprocket/sprocket.toml`.
 
 ## Configuration Values
 
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -30,7 +30,7 @@ pub enum Commands {
     /// Explains linting and validation rules.
     Explain(explain::Args),
 
-    /// Formats a document.
+    /// Formats a document or a directory containing documents.
     #[clap(alias = "fmt")]
     Format(format::Args),
 
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -15,37 +15,38 @@ pub mod validate;
 #[derive(Subcommand, Debug)]
 #[allow(clippy::large_enum_variant)]
 pub enum Commands {
-    /// Checks a WDL document (or a directory containing WDL documents) and
-    /// reports diagnostics.
+    /// Runs the Language Server Protocol (LSP) server.
+    Analyzer(analyzer::Args),
+
+    /// Checks a document or a directory containing documents.
     Check(check::CheckArgs),
 
-    /// Lints Workflow Description Language files.
-    Lint(check::LintArgs),
-
-    /// Explains a rule.
+    /// Explains linting and validation rules.
     Explain(explain::Args),
 
-    /// Runs the analyzer LSP server.
-    Analyzer(analyzer::AnalyzerArgs),
-
-    /// Formats a WDL document.
+    /// Formats a document.
     #[clap(alias = "fmt")]
-    Format(format::FormatArgs),
+    Format(format::Args),
 
-    /// Validates an input JSON or YAML file against a task or workflow input
-    /// schema.
+    /// Lints a document or a directory containing documents.
+    Lint(check::LintArgs),
+
+    /// Runs a task or workflow.
+    Run(run::Args),
+
+    /// Validate a set of inputs against a task or workflow.
     ///
     /// This ensures that every required input is supplied, every supplied input
     /// is correctly typed, that no extraneous inputs are provided, and that any
     /// provided `File` or `Directory` inputs exist.
     ///
-    /// It will not catch potential runtime errors that
-    /// may occur when running the task or workflow.
-    ValidateInputs(validate::ValidateInputsArgs),
+    /// It will not catch potential runtime errors that may occur when running
+    /// the task or workflow.
+    Validate(validate::Args),
 
     /// Display the effective configuration.
-    Config(config::ConfigArgs),
+    Config(config::Args),
 
     /// Generates shell completions.
-    Completions(commands::completions::Args),
+    Completions(completions::Args),
 }
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -1,8 +1,46 @@
 //! Implementation of sprocket CLI commands.
 
+use clap::Subcommand;
+
 pub mod analyzer;
 pub mod check;
 pub mod config;
 pub mod explain;
 pub mod format;
 pub mod validate;
+
+/// Represents the available commands for the Sprocket CLI.
+#[derive(Subcommand, Debug)]
+#[allow(clippy::large_enum_variant)]
+pub enum Commands {
+    /// Checks a WDL document (or a directory containing WDL documents) and
+    /// reports diagnostics.
+    Check(check::CheckArgs),
+
+    /// Lints Workflow Description Language files.
+    Lint(check::LintArgs),
+
+    /// Explains a rule.
+    Explain(explain::Args),
+
+    /// Runs the analyzer LSP server.
+    Analyzer(analyzer::AnalyzerArgs),
+
+    /// Formats a WDL document.
+    #[clap(alias = "fmt")]
+    Format(format::FormatArgs),
+
+    /// Validates an input JSON or YAML file against a task or workflow input
+    /// schema.
+    ///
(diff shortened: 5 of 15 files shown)


src/main.rs
```diff
@@ -54,11 +63,47 @@ struct Cli {
...
 
 pub async fn inner() -> anyhow::Result<()> {
     let cli = Cli::parse();
 
+    // Start a new Figment instance with default values
+    let mut figment =
+        Figment::new().admerge(Serialized::from(config::Config::default(), "default"));
+
+    // If provided, check config file from environment
+    if let Ok(config_file) = env::var("SPROCKET_CONFIG") {
```
a-frantz: I think the ENV config should take precedence over the HOME config(s)

src/main.rs
```diff
@@ -54,11 +63,47 @@ struct Cli {
...
+    if let Ok(xdg_config_home) = env::var("XDG_CONFIG_HOME") {
+        figment = figment.admerge(Toml::file(
+            Path::new(&xdg_config_home.as_str()).join("sprocket.toml"),
+        ));
+    }
+
+    // Check HOME for a config file
+    if let Some(home) = home_dir() {
+        figment = figment.admerge(Toml::file(home.join(".config").join("sprocket.toml")));
+    }
```
a-frantz: I don't think we need both of these, but I don't know which one we should keep.

src/main.rs
```diff
@@ -68,13 +81,64 @@ pub async fn inner() -> anyhow::Result<()> {
...
+    if let Ok(xdg_config_home) = env::var("XDG_CONFIG_HOME") {
+        tracing::info!(
+            "Reading configuration from XDG_CONFIG_HOME: {xdg_config_home:?}/.config/sprocket.toml"
+        );
+        figment = figment.admerge(Toml::file(
+            Path::new(&xdg_config_home.as_str())
+                .join(".config")
+                .join("sprocket.toml"),
+        ));
+    }
```
a-frantz: looks like I don't have this ENV var set in any of my set ups, but isn't it supposed to be set `~/.config`? So joining it with `.config` would lead to `~/.config/.config/sprocket.toml`?

src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Disables color output.
+    pub no_color: bool,
+    /// The report mode.
+    pub report_mode: Option<Mode>,
+}
+
+/// Represents the configuration for the Sprocket `validate` command.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct ValidateInputsConfig {
```
a-frantz: Said this over slack, but I think `no_color` and `report_mode` should probably be moved to the root of the config, and then we can drop the `ValidateInputsConfig` entirely

src/main.rs
adthrasher: Looking at this further, the [spec](https://specifications.freedesktop.org/basedir-spec/latest/) says the following:

> $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

So I'll plan to combine these so that `XDG_CONFIG_HOME` is preferred, but if it is unset, then `$HOME/.config` will be used.

a-frantz (commented):
oops, I misread. I think `SPROCKET_CONFIG=... sprocket ...` should also take precedence over the PWD. So the full list of priority in my mind (and curious if anyone disagrees):
1. default config
2. home config (I like the current logic for this where we first check `XDG_CONFIG_HOME` and fall back to `~/.config`)
3. PWD
4. File specified by `SPROCKET_CONFIG` in the ENV
5. File specified by `--config` flag

src/config.rs
```diff
@@ -0,0 +1,191 @@
...
+/// Represents the configuration for the Sprocket `format` command.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct FormatConfig {
+    /// Disables color output.
+    pub no_color: bool,
+    /// Enable or disable color output.
+    pub color: bool,
+    /// The report mode.
+    pub report_mode: Option<Mode>,
```
a-frantz: this should be removed, right? if they are in the `CommonConfig` they shouldn't be here as well IMO

src/config.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /// Write a configuration to the specified path.
+    pub fn write_config(&self, path: &str) -> Result<()> {
+        let data = toml::to_string(self).context("Failed to serialize config")?;
+        std::fs::write(path, data).context("Failed to write config file")
+    }
+
+    /// Merge the current configuration with commandline arguments.
+    pub fn merge_args(&self, args: crate::commands::Commands) -> crate::commands::Commands {
```
a-frantz: can you add `use crate::commands::Commands` so this is easier to read?

src/config.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Rule IDs to except from running.
+    pub except: Vec<String>,
+    /// Causes the command to fail if any warnings are reported.
+    pub deny_warnings: bool,
+    /// Causes the command to fail if any notes are reported.
+    pub deny_notes: bool,
+    /// Suppress diagnostics from imported documents.
+    pub single_document: bool,
+    /// Show diagnostics for remote documents.
+    pub show_remote_diagnostics: bool,
```
a-frantz: should this be in the config? They seem kind of borderline to me on whether you'd want to "set and forget" these. It might be best to keep these as CL only?

adthrasher: I'm conflicted, but don't have a strong opinion. I could see someone never wanting to see diagnostics for anything other than the document they're working on. Similarly, I can see another user always wanting to see everything, even remote file diagnostics that they can't change.

a-frantz: I think for the sake of preventing "config sprawl" we can remove these. If we hear from users that they want this in the config file we can add them back. I think my preference is to err on the side of not offering config values, but that isn't a strongly held conviction

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Unknown `--except` rules will now emit a warning instead of being silently ignored ([#94](https://github.com/stjude-rust-labs/sprocket/pull/94))
 
+### Added
+
+* Added configuration file support ([#104](https://github.com/stjude-rust-labs/sprocket/pull/104)).
```
a-frantz, line 44: I'd add some more here. Considering that we're going to use the CHANGELOG as the basis for release notes in the future, I'm ok being verbose. I'd mention the new `config` subcommand at least

adthrasher: I was trying to figure out where else to add documentation. Currently the sprocket documentation is in rough shape.

a-frantz: Yea I see what you mean. sprocket.bio is lacking much about the actual `sprocket` bin.

@claymcleod before this next release of Sprocket (with `run`), we probably want to add some more to the docs site.

src/main.rs
```diff
@@ -68,13 +50,61 @@ pub async fn inner() -> anyhow::Result<()> {
         .finish();
     tracing::subscriber::set_global_default(subscriber)?;
 
+    // Start a new Figment instance with default values
+    let mut figment =
+        Figment::new().admerge(Serialized::from(config::Config::default(), "default"));
+
+    // Check XDG_CONFIG_HOME for a config file
+    // If XDG_CONFIG_HOME is not set, check HOME for a config file
```
peterhuene: For this, I've generally preferred using the `dirs` crate, as it supports Windows as well.

peterhuene: Oh never mind, I missed that you're already using `dirs`. I think `dirs` already checks for `XDG_CONFIG_HOME`, but let me look.

peterhuene: I think we can replace this all with `dirs::config_dir` (which on Windows uses the roaming directory for users joined to a domain). I would also recommend that we create a `sprocket` subdirectory in there, so something like `<config_dir>/sprocket/sprocket.toml`.

src/config.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /// Write a configuration to the specified path.
+    pub fn write_config(&self, path: &str) -> Result<()> {
+        let data = toml::to_string(self).context("Failed to serialize config")?;
+        std::fs::write(path, data).context("Failed to write config file")
+    }
+
+    /// Merge the current configuration with commandline arguments.
+    pub fn merge_args(&self, args: Commands) -> Commands {
```
a-frantz: @peterhuene curious if you know of any way to do this that's a bit cleaner. This works, and with the pattern matching we should be getting compile errors if any of the underlying structs change without being updated here. But it's a pretty ugly function.

src/main.rs
a-frantz: > I would also recommend that we create a `sprocket` subdirectory in there, so something like `<config_dir>/sprocket/sprocket.toml`.

Are we sure we want/need a whole directory in the config dir? I think the vision here is that `sprocket.toml` (where ever that's located) will be comprehensive, and multiple config files should be avoided.

peterhuene: For these platform-specific locations, generally applications stick to creating their own subdirectory to contain their files, even when it's a single file.

src/config.rs
peterhuene: My only suggestion is that we implement a `fn merge(self, other: Self) -> Self` on each of these args types; I think that would clean things up a bit.

src/config.rs
```diff
@@ -0,0 +1,181 @@
...
+use crate::commands::check;
+use crate::commands::format;
+use crate::commands::validate;
+/// Represents the configuration for the Sprocket CLI tool.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct Config {
+    /// Configuration for the `format` command.
+    #[serde(rename = "format")]
+    pub format_config: FormatConfig,
```
peterhuene: Nit: I think `_config` is a bit redundant given the type name, so perhaps just `format`, `check`, etc. for field names?

src/config.rs
adthrasher: Wouldn't that require the configuration options to exactly mirror the command line arguments? Right now, we omit some values from the configuration. We also invert at least one argument in the configuration (I think).

src/main.rs
```diff
@@ -68,13 +50,61 @@ pub async fn inner() -> anyhow::Result<()> {
...
+    if let Some(ref cli) = cli.config {
+        tracing::info!("Reading configuration from --config: {cli:?}");
+        figment = figment.admerge(Toml::file(cli));
+    }
+
+    // Get the configuration from the Figment
+    let config: Config = figment.extract().expect("Failed to extract config");
+
+    // Write effective configuration to the log
+    tracing::info!(
```
peterhuene: I'd prefer not to have such a potentially verbose entry at the `info` level; perhaps `debug`?

src/main.rs
```diff
@@ -68,13 +50,61 @@ pub async fn inner() -> anyhow::Result<()> {
...
+        "Effective configuration:\n{}",
```
peterhuene: Nit: we generally forgo sentence capitalization in our log messages and output (applies here and other logged messages).

```suggestion
        "effective configuration:\n{}",
```

src/commands/config.rs
```diff
@@ -0,0 +1,16 @@
...
+/// Arguments for the `config` subcommand.
+#[derive(Parser, Debug)]
+pub struct ConfigArgs;
+
+/// Runs the `config` command.
+pub fn config(_args: ConfigArgs, config: crate::config::Config) -> anyhow::Result<()> {
+    println!(
+        "Effective configuration:\n{}",
+        toml::to_string(&config).unwrap_or_default()
+    );
```
peterhuene: Nit: I think just printing out the configuration is enough and we can eliminate this header; that way it's easier to pipe to a file if desired.

```suggestion
    println!(
        "{}",
        toml::to_string_pretty(&config).unwrap_or_default()
    );
```

src/main.rs
```diff
@@ -68,13 +50,61 @@ pub async fn inner() -> anyhow::Result<()> {
...
+    }
+
+    // If provided, check command line config file
+    if let Some(ref cli) = cli.config {
+        tracing::info!("Reading configuration from --config: {cli:?}");
+        figment = figment.admerge(Toml::file(cli));
+    }
+
+    // Get the configuration from the Figment
+    let config: Config = figment.extract().expect("Failed to extract config");
```
peterhuene: I feel like the above config loading code could be extracted to its own function in the library rather than in `main.rs`.

src/config.rs
peterhuene: Oh perhaps just `fn apply(&mut self, config: T)` where `T` is the relevant config type, then?

src/main.rs
adthrasher: Is it going to be intuitive that the MacOS implementation for `dirs::config_dir` points to `$HOME/Library/Application Support`? We can document that, but it doesn't feel "right" to me.

peterhuene: I mean, it _is_ the correct directory for macOS, just a historically bad name given by Apple. We already use `~/Library/Cache` for our download cache, so I personally expect _something_ under `Library` (even if it's a directory with a space in it, ugh).

We'll definitely want to document where the file is expected to be on each platform.

a-frantz: > I mean, it _is_ the correct directory for macOS, just a historically bad name given by Apple. We already use `~/Library/Cache` for our download cache, so I personally expect _something_ under `Library` (even if it's a directory with a space in it, ugh).
> 
> We'll definitely want to document where the file is expected to be on each platform.

I don't love this, even if it is _technically_ correct. I just did some poking around my own `~/Library` directory, and it's got _a lot_ of junk in it that I'm scared to touch, because I didn't put it there and I don't really know what macOS is doing with it behind the scenes. Whereas my `~/.config` is much sparser and contains all things I'm familiar with.

My two cents as a user is that `~/.config` is a friendlier place to put our config, even if it's technically "incorrect" for macOS.

CHANGELOG.md
adthrasher: @a-frantz - I added a markdown page to the docs with information. It's currently a minimalist document.

src/main.rs
adthrasher: Looking at the spec (https://specifications.freedesktop.org/basedir-spec/latest/), I'm confused about the `dirs` behavior on Mac. Relevant spec:

> $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

On my machine `XDG_CONFIG_HOME` is empty, yet `dirs` points to `$HOME/Library/Application Support`.

adthrasher: Look at the apple [docs](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW28), I'm not sold that `Application Support` is the right location. Specifically, the part that states "that should remain hidden from the user". These are specifically user-generated configurations.

> Put app-created support files in the Library/Application support/ directory. In general, this directory includes files that the app uses to run but that should remain hidden from the user. This directory can also include data files, configuration files, templates and modified versions of resources loaded from the app bundle.

peterhuene: I'm fine with perhaps `~/.config/sprocket/sprocket.toml` for Linux and macOS, but roaming `...\AppData\Roaming\sprocket\sprocket.toml` for Windows.

Perhaps just:

```rust
#[cfg(target_os = "macos")]
{
    // join `.config/sprocket/sprocket.toml` to `dirs::home_dir()`
}
#[cfg(not(target_os = "macos"))]
{
    // join `sprocket/sprocket.toml` to `dirs::config_dir()`
}
```

Cargo.toml
```diff
@@ -33,5 +33,8 @@ tracing = "0.1.40"
 tracing-indicatif = "0.3.9"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 url = "2.5.4"
+figment = { version = "0.10.19", features = ["toml"] }
```
claymcleod: Can we keep this list alphabetically sorted?

docs/cli/config.md
```diff
@@ -0,0 +1,25 @@
...
+
+The path to a configuration file can be specified via the environment variable `SPROCKET_CONFIG`.
+
+### Current Working Directory
+
+Sprocket will look for a `sprocket.toml` in the current working directory when the `sprocket` command runs.
+
+### XDG_CONFIG_HOME
+
+Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). On MacOS, Sprocket will attempt to read the configuration from `HOME/.config/sprocket/sprocket.toml`.
```
claymcleod:
```suggestion
Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). On MacOS, Sprocket will attempt to read the configuration from `$HOME/.config/sprocket/sprocket.toml`.
```

docs/cli/config.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Configuration Values
+
+Running the command `sprocket config` will print the effective configuration. The default configuration can be written out using the `--generate` argument.
```
claymcleod: I would personally view the flag name `--generate` confusing here: it's "generating" a configuration either way, just one is the current resolved configuration and one is the default configuration.

I almost am thinking a required subsubcommand of `resolve` or `init` might be good here. That might at least make it more concrete that there are two different lenses for this subcommand.

src/commands.rs
```diff
@@ -1,9 +1,52 @@
...
+    ///
+    /// This ensures that every required input is supplied, every supplied input
+    /// is correctly typed, that no extraneous inputs are provided, and that any
+    /// provided `File` or `Directory` inputs exist.
+    ///
+    /// It will not catch potential runtime errors that may occur when running
+    /// the task or workflow.
+    Validate(validate::Args),
+
+    /// Display the effective configuration.
```
claymcleod: We should put these in alphabetical order.

src/commands/config.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct Args {
+    /// Generates a default configuration file.
+    #[arg(long)]
+    generate: bool,
+}
+
+/// Runs the `config` command.
+pub fn config(args: Args, config: crate::config::Config) -> anyhow::Result<()> {
+    if args.generate {
+        tracing::info!("Generating default configuration file...");
```
claymcleod: I don't think this log message adds much, we can just remove it.

src/config.rs
```diff
@@ -0,0 +1,178 @@
...
+        // Check for a config file in the current directory
+        // Start a new Figment instance with default values
+        let mut figment = Figment::new().admerge(Serialized::from(Config::default(), "default"));
+
+        // Check XDG_CONFIG_HOME for a config file
+        // On MacOS, check HOME for a config file
+        #[cfg(target_os = "macos")]
+        {
+            if let Some(home) = dirs::home_dir() {
+                tracing::info!(
```
claymcleod: These are probably `trace!` statements, only in the rarest of occasions are you going to want this level of information IMO.

src/config.rs
```diff
@@ -0,0 +1,178 @@
...
+                     {xdg_config_home:?}/sprocket/sprocket.toml"
+                );
+                figment = figment.admerge(Toml::file(
+                    xdg_config_home.join("sprocket").join("sprocket.toml"),
+                ));
+            }
+        }
+
+        // Check PWD for a config file
+        if Path::exists(Path::new("sprocket.toml")) {
```
claymcleod: Big question: should it be `sprocket.toml` or `Sprocket.toml`. I think the latter personally in all cases.

src/config.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        // If provided, check config file from environment
+        if let Ok(config_file) = env::var("SPROCKET_CONFIG") {
+            tracing::info!("reading configuration from SPROCKET_CONFIG: {config_file:?}");
+            figment = figment.admerge(Toml::file(config_file));
+        }
+
+        // If provided, check command line config file
+        if let Some(ref cli) = path {
+            tracing::info!("reading configuration from --config: {cli:?}");
```
claymcleod:
```suggestion
            tracing::info!("reading configuration from provided configuration file: {cli:?}");
```

I think this is more future proof rather than calling out specific command line arguments.

src/config.rs
```diff
@@ -0,0 +1,178 @@
...
+            figment = figment.admerge(Toml::file(cli));
+        }
+
+        // Get the configuration from the Figment
```
claymcleod, line 147: This seems like a broken sentence.

src/main.rs
```diff
@@ -93,18 +63,27 @@ pub async fn inner() -> anyhow::Result<()> {
         }
     };
 
+    let config = Config::new(cli.config);
+
+    // Write effective configuration to the log
+    tracing::debug!(
```
claymcleod: This is almost certainly `trace!` as well IMO.

src/config.rs
adthrasher: I disagree. I do not like capitalized file names generally. Without doing any investigation, I feel like most configuration files are lowercased (with `Cargo.toml` being a notable exception).

src/config.rs
adthrasher: Can you elaborate? The block is taking the Figment data structure and getting a configuration out of it. I suppose I could write `Extract` instead, but that feels very redundant with just reading the function call.

docs/cli/config.md
a-frantz, line 1: this page needs to specify how different config files are merged. Have we finalized that? Because I'm not 100% clear on the specifics (but about to dig into that part of the code)

docs/cli/config.md
```diff
@@ -0,0 +1,25 @@
...
+
+The path to a configuration file can be specified via the environment variable `SPROCKET_CONFIG`.
+
+### Current Working Directory
+
+Sprocket will look for a `sprocket.toml` in the current working directory when the `sprocket` command runs.
+
+### XDG_CONFIG_HOME
+
+Sprocket will attempt to read a configuration file from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. The location of `XDG_CONFIG_HOME` is operating system dependent. The platform-specific values can be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). On MacOS, Sprocket will attempt to read the configuration from `$HOME/.config/sprocket/sprocket.toml`.
```
peterhuene:
```suggestion
On systems that support the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/), such as many Linux distributions, Sprocket will attempt to read configuration from `XDG_CONFIG_HOME/sprocket/sprocket.toml`. Please note that the location of `XDG_CONFIG_HOME` is operating system dependent.

On MacOS, Sprocket will attempt to read the configuration from `$HOME/.config/sprocket/sprocket.toml`.

On Windows, Sprocket will attempt to read configuration from `%USERPROFILE%\AppData\Roaming`.

The platform-specific locations can also be found [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). 
```

src/config.rs
peterhuene: I could go either way, with a slight preference for `sprocket.toml`.

Fwiw, Diesel, a popular Rust ORM, uses lowercase `diesel.toml`.

peterhuene (approved):
Looks good; approving with a nit comment on the docs.

src/commands/run.rs
```diff
@@ -82,8 +82,30 @@ pub struct Args {
...
+
+impl Args {
+    /// Applies the configuration to the arguments.
+    pub fn apply(mut self, config: crate::config::Config) -> Self {
+        self.config = match self.config {
+            Some(path) => Some(path),
+            None => {
+                if config.run.config.exists() {
+                    Some(config.run.config)
+                } else {
```
a-frantz: If a user specifies a config path that doesn't exist, we should log a warning (and maybe bail?) instead of silently continuing with no config

src/config.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// Hide diagnostics with `note` severity.
+    pub hide_notes: bool,
+}
+
+/// Represents the configuration for the Sprocket `run` command.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct RunConfig {
+    /// Path to the engine configuration file.
+    pub config: PathBuf,
```
a-frantz, line 96: It can be done in a follow-up PR, but we should enable putting the engine config within `sprocket.toml`

src/config.rs
a-frantz: > I could go either way, with a slight preference for `sprocket.toml`.

same here

docs/cli/config.md
adthrasher: I think that is final, unless there are any new suggestions. I'll add it here.

adthrasher: Ah, wait. The first line describes the order:

> The Sprocket configuration loader sources from command-line options, environment variables, and configuration files, in that priority order.

src/config.rs
claymcleod: Okay, fair enough then, let's do `sprocket.toml`.

docs/cli/config.md
a-frantz: Have we considered searching and using "highest priority found" instead of merging? it's much easier to explain and communicate to users, but does somewhat limit flexibility.

Having a `config resolve` command does make the complexity easier to deal with, so maybe the current approach is worth it.

src/config.rs
claymcleod: In reading it again, I'm not sure what I thought was broken. Perhaps I annotated this on the wrong line. Anyways, this seems fine.

docs/cli/config.md
a-frantz: > Ah, wait. The first line describes the order:
> 
> > The Sprocket configuration loader sources from command-line options, environment variables, and configuration files, in that priority order.

The merge logic still needs to be communicated. Something like "atomic values are overwritten by higher priority configs. Lists are appended."

src/commands/run.rs
adthrasher: To avoid having an `Option` in the `RunConfig`, this can be omitted, but gets the default value of `""`. So maybe the check needs to be: exists, then raise an error only if the `path` is not the empty string.

docs/cli/config.md
adthrasher: I added a new section that describes the merge and provides a small example.

adthrasher: > Have we considered searching and using "highest priority found" instead of merging? it's much easier to explain and communicate to users, but does somewhat limit flexibility.
> 
> Having a `config resolve` command does make the complexity easier to deal with, so maybe the current approach is worth it.

That's certainly a change we could make. Personally, I think there is some value in the combination approach. For example, I could see a user deciding they never want to see a particular lint rule and keeping that in a config (like `$HOME`). Then per-repository having a `sprocket.toml` that ignores lints particular to that project. Maybe we're overcomplicating it, but it does seem useful (in theory). That said, a user can always just have a single config file. 

Maybe we add an additional flag like `--only-config` (or a better name) that takes a file path and ignores any other found configs. Or it could be something like `--ignore-other-configs` that gets specified with the `--config` argument to stop loading alternative configuration files.

claymcleod: I agree that the paradigm Andrew describes in the first paragraph is desirable. I think adding more flags, as suggested in the second paragraph, is probably not a great idea, as it risks creating a mess of different configuration options and parameters (in my mind, adding new command line flags has an ever-rising cost of increasing the mental burden overall to work with the tool).

a-frantz: That ties into another idea I think I shared on Slack. Some flag that blocks looking for configs entirely (maybe `--no-config`). Maybe a user could write `sprocket lint --no-config --config ../path/sprocket.toml` to load a specific config (and only that config). Or just `sprocket lint --no-config` to get the default "out of the box" behavior

ETA: Clay commented seconds before me saying we should simplify things, so maybe scrap the above concept 😅

adthrasher: Another thought, I believe the Figment crate tracks metadata. So the `resolve` command could potentially annotate the output configuration to explain where the values originated.

adthrasher: > Another thought, I believe the Figment crate tracks metadata. So the `resolve` command could potentially annotate the output configuration to explain where the values originated.

This works, but only for simple types. For array types (e.g. `check.except`), it only stores the original source location (e.g. when `default()` was loaded).

src/config.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// The maximum line length (default is 90).
+    pub max_line_length: usize,
+}
+
+impl Default for FormatConfig {
+    fn default() -> Self {
+        Self {
+            with_tabs: false,
+            indentation_size: 4,
+            max_line_length: 90,
```
a-frantz: can this instead parse the defaults from the struct defined in `wdl-format`? I wouldn't want to update there and forget to update here (not that I intend any of these values to change, but)

src/config.rs
```diff
@@ -0,0 +1,177 @@
...
+            if let Some(home) = dirs::home_dir() {
+                trace!("reading configuration from: {home:?}/.config/sprocket/sprocket.toml");
+                figment = figment.admerge(Toml::file(
+                    home.join(".config").join("sprocket").join("sprocket.toml"),
+                ));
+            }
+        }
+        #[cfg(not(target_os = "macos"))]
+        {
+            if let Some(xdg_config_home) = dirs::config_dir() {
```
a-frantz: is this right? I haven't looked at the `dirs` docs, but is `config_dir()` really the same thing as checking `XDG_CONFIG_HOME`? If so, what are we gaining by outsourcing to this crate?

docs/cli/config.md
a-frantz: I still think we should have `--no-config` flag (although not tied to that name) which will skip reading any config files and just use all default values for each sub command. I understand @claymcleod 's argument and I agree in the principle, but I think there's value to providing a simple way to get an "out of the box" invocation of `sprocket`. And the mental complexity/burden for this flag is fairly straightforward (as opposed to some of the other ideas discussed).

README.md
```diff
@@ -32,6 +32,7 @@
 
 - **`sprocket analyzer`** runs Sprocket as a LSP server, which is useful for IDE integration.
 - **`sprocket check`** performs static analysis on a document or directory of documents.
+- **`sprocket config`** prints effective configuration.
```
a-frantz: This doesn't account for the `init` option. Maybe this should be more general to future proof? Something like "prints configuration values"?

README.md
```diff
@@ -32,6 +32,7 @@
...
 - **`sprocket explain`** explains validation and lint rules supported by Sprocket.
 - **`sprocket format`** formats a document.
```
a-frantz:
```suggestion
- **`sprocket format`** formats a document or directory of documents.
```

docs/cli/config.md
```diff
@@ -0,0 +1,65 @@
+# Configuration Files
+
+The Sprocket configuration loader sources from command-line options, environment variables, and configuration files, in that priority order.
```
a-frantz:
```suggestion
The Sprocket configuration loader sources and merges values from a location specified by a command-line option, an environment variable, and common configuration locations, in that priority order.
```
my edit could potentially use some more word smithing but I think is more accurate/clear

src/commands.rs
```diff
@@ -1,9 +1,52 @@
...
+    /// Generates shell completions.
+    Completions(completions::Args),
+
+    /// Display the effective configuration.
+    Config(config::Args),
+
+    /// Explains linting and validation rules.
+    Explain(explain::Args),
+
+    /// Formats a document.
```
a-frantz:
```suggestion
    /// Formats a document or a directory containing documents.
```

src/config.rs
adthrasher: The biggest advantage is that on Windows it correctly finds the roaming app data folder. On Linux, it also correctly handles checking `$DXG_CONFIG_HOME` and falls back to `$HOME/.config` if unset. https://docs.rs/dirs/latest/dirs/fn.config_dir.html

a-frantz: In that case I'd suggest this nit:
```suggestion
            if let Some(config_home) = dirs::config_dir() {
```

it's not strictly the value of `XGD_CONFIG` but as currently written that's what is implied.

a-frantz (approved):
Looks great. Awesome feature to have 🚀

docs/cli/config.md
adthrasher: Discussed via Slack, but I implemented a `--skip-config-search` argument that disables the config search.

src/config.rs
adthrasher: Added an issue to track: https://github.com/stjude-rust-labs/sprocket/issues/115

PR #107, opened 2025-05-05
feat: dev doc command

a-frantz:

This PR introduces a beta version of the `doc` command under the `dev` namespace.

For external contributors:

For all contributors:


src/commands/doc.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Output directory for the generated documentation.
+    /// If not specified, the documentation will be generated in
+    /// `<workspace>/docs`.
+    #[arg(long, value_name = "DIR")]
+    pub output: Option<PathBuf>,
+    /// Whether to overwrite any existing documentation.
+    ///
+    /// If specified, any existing files in the output directory will be
+    /// deleted. Otherwise, the command will ignore existing files.
+    /// **Regadless** of this flag, the command will overwrite any existing
```
a-frantz:
```suggestion
    /// Regardless of this flag, the command will overwrite any existing
```

src/commands/doc.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub output: Option<PathBuf>,
+    /// Whether to overwrite any existing documentation.
+    ///
+    /// If specified, any existing files in the output directory will be
+    /// deleted. Otherwise, the command will ignore existing files.
+    /// Regardless of this flag, the command will overwrite any existing
+    /// files which conflict with the generated documentation.
+    #[arg(long)]
+    pub overwrite: bool,
+    /// Whether to open the generated documentation in the default web browser.
```
a-frantz:
```suggestion
    /// Open the generated documentation in the default web browser.
```

sprocket-doc.md
```diff
@@ -0,0 +1,63 @@
...
+
+## General structure
+
+The generated documentation directory (named `docs` by default) is completely self contained and can be moved, zipped, and shared without any of the raw WDL files it documents. Please let us know if you run into any issues while sharing your documentation!
+
+The `docs` directory will always contain a `style.css` file, an `index.js` file, an `index.html` file, and an `assets/` directory filled with SVG icons and logos. The rest of the contents are dynamically created based on the structure of the WDL workspace being documented. Each WDL file found in the workspace will be represented as a directory within `docs`. The relative path from the root of the workspace to each WDL file is copied in the `docs` directory to keep your documentation organized.
```
adthrasher: `copied` feels like the wrong word here. Maybe "reproduced"?

sprocket-doc.md
```diff
@@ -0,0 +1,63 @@
...
+
+The directory for each WDL file will have an `index.html` file and one HTML file per struct, task, or workflow defined in the WDL.
+
+## Custom themes
```
adthrasher: I'm not sure we need this. I think it gets left undocumented. Maybe a short blurb along the lines of "We provide a single styling that should be widely useful. If you want to customize the styling, please let us know.".

sprocket-doc.md
```diff
@@ -0,0 +1,63 @@
...
+### Inputs
+
+Each entry in the `input` section of a task or workflow is expected to have a corresponding entry in the `parameter_meta` section. To get the most out of `sprocket doc`, it's recommended that each input entry be an object. That object should have at least a `description` key. There is special handling for the following sub-keys:
+
+- `group`: all inputs sharing the same `String` value will be rendered together in a dedicated table
+    - required inputs are _always_ rendered under the "Required Inputs" table and thus should _not_ have a `group` key (it will be ignored if present)
+    - The `Common` group of inputs will always come after the required inputs
+    - Inputs without a `group` will be rendered under "Other Inputs" which will be the last input table
+    - The `Resource` group of inputs will immediately precede the "Other Inputs" table
+    - All other groups will render alphabetically between tables which get special treatment (`Common`, `Resource` and `Other Inputs`)
```
adthrasher: "Between"? How does that work since these four aren't in alphabetical order?

sprocket-doc.md
a-frantz: That makes some sense, but I think we should leave it in for now. I intend to invite some beta testers from the openwdl slack this week, and I imagine some of them might want to do this explicitly, so I think its worth mentioning here. Once we hit wider distribution with an official release, we can revisit this section and maybe remove.

sprocket-doc.md
```diff
@@ -0,0 +1,63 @@
...
+- `help`: will render at the top of the "Additional Meta" cell of the table
+
+If an input has a `String` value for its parameter meta entry instead of a meta object, that string value will be treated as if it were the `description` key of a meta object.
+
+### Outputs
+
+Outputs should be documented in one of two places: either in the task/workflow `meta` section under an `outputs` key or at the root of the `parameter_meta` section.
+
+Similar to inputs, each output should either be documented with an object which has a `description` key with a `String` value, or a `String` value directly.
```
adthrasher: We should probably note what happens if a WDL output has an object with keys. I'm assuming other keys are ignored.

sprocket-doc.md
```diff
@@ -0,0 +1,63 @@
...
+
+TODO not yet implemented
+
+## Parameters (inputs and outputs)
+
+Each input and output to a workflow or task should be documented, but there is some flexibility in the specifics.
+
+### Inputs
+
+Each entry in the `input` section of a task or workflow is expected to have a corresponding entry in the `parameter_meta` section. To get the most out of `sprocket doc`, it's recommended that each input entry be an object. That object should have at least a `description` key. There is special handling for the following sub-keys:
```
adthrasher: Probably should note that other keys are ignored.

sprocket-doc.md
```diff
@@ -0,0 +1,63 @@
...
+
+Outputs should be documented in one of two places: either in the task/workflow `meta` section under an `outputs` key or at the root of the `parameter_meta` section.
+
+Similar to inputs, each output should either be documented with an object which has a `description` key with a `String` value, or a `String` value directly. 
+
+## Meta entries with special handling
+
+Every struct, task, and workflow `meta` section should have a `description` key with a `String` value. This description string can have Markdown formatting. The `description` string should be less than 140 characters or it will be clipped in some contexts.
+
+The `help` key should only have a `String` value, but can be of any length. It's best practice to keep `description`s short, and put any additional text needed under the `help` key. Help strings can also be styled with Markdown!
```
adthrasher:
```suggestion
The `help` key should only have a `String` value, but can be of any length. It's best practice to keep `description` short, and put any additional text needed under the `help` key. Help strings can also be styled with Markdown!
```

sprocket-doc.md
```diff
@@ -0,0 +1,63 @@
...
+
+If an input has a `String` value for its parameter meta entry instead of a meta object, that string value will be treated as if it were the `description` key of a meta object.
+
+### Outputs
+
+Outputs should be documented in one of two places: either in the task/workflow `meta` section under an `outputs` key or at the root of the `parameter_meta` section.
+
+Similar to inputs, each output should either be documented with an object which has a `description` key with a `String` value, or a `String` value directly. 
+
+## Meta entries with special handling
```
adthrasher: Ah, maybe ignore the above comments about `keys` and just document here that all other keys are ignored.

sprocket-doc.md
a-frantz: yea... it's a little awkward. Rust is here - https://github.com/stjude-rust-labs/wdl/blob/feat/wdl-doc-styling/wdl-doc/src/callable.rs#L49

In English:
- Always put `Common` first
- Always put `Resource` at end of inputs _with a group_
    - (and `Other Inputs` without a group follow this)
- Default Rust compare (alphabetical) otherwise

sprocket-doc.md
a-frantz: They actually aren't ignored! Anything that isn't given special treatment (mentioned in this doc) gets rendered in an `Additional Meta` column of the table (for I/Os) or under the `description` and `help` for tasks/workflows

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
...
+
+The directory for each WDL file will have an `index.html` file and one HTML file per struct, task, or workflow defined in the WDL.
+
+## Custom themes
+
+While it is technically possible to supply your own custom CSS styling, this capability is currently undocumented. We recommend you stick with the default styling at this point in time, but do let us know what kinds of customization you would like to see in future releases! 
+
+## Per WDL file `index.html`
+
+At a minimum, the `index.html` associated with each WDL file will have a table of contents with links to each struct, task, or workflow documentation page in that directory. If the WDL file has a "preamble", that will be rendered as Markdown text above the table of contents.
```
claymcleod:
```suggestion
At a minimum, the `index.html` associated with each WDL file will have a table of contents with links to each struct, task, or workflow documentation page in that directory. If the WDL file has a preamble comment (a comment block at the very top of the file where each line starts with `##`), that will be rendered as Markdown text above the table of contents.
```

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Preamble
```
claymcleod: I would prefer going with my sentence change above and deleting this section. I don't think there's enough content to stand on its own.

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
...
+- `help`: will render at the top of the "Additional Meta" cell of the table
+
+If an input has a `String` value for its parameter meta entry instead of a meta object, that string value will be treated as if it were the `description` key of a meta object.
+
+### Outputs
+
+Outputs should be documented in one of two places: either in the task/workflow `meta` section under an `outputs` key or at the root of the `parameter_meta` section.
+
+Similar to inputs, each output should either be documented with an object which has a `description` key with a `String` value, or a `String` value directly. 
+
```
claymcleod: You might want to point users to a task in our workflows repository for an example.

claymcleod: Sorry I thought this was the end of the output section, but it's not. Maybe this needs to be a section at the end of the document.

sprocket-doc.md
claymcleod, line 1: I didn't see any details about the tags at the top of some of these workflows like how allow nested inputs is rendered. Just pointing that out in case you want to add details on what all is pulled out at the top.

Cargo.toml
```diff
@@ -38,4 +40,4 @@ tracing-indicatif = "0.3.9"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 url = "2.5.4"
 walkdir = "2.5.0"
-wdl = { version = "0.14.0", features = ["cli", "codespan", "engine", "lsp"], git = "https://github.com/stjude-rust-labs/wdl", rev = "ccc8850" }
+wdl = { version = "0.14.0", features = ["cli", "codespan", "engine", "lsp", "doc"], path = "../wdl/wdl" }
```
acfoltzer: I'm guessing this is a temporary change while the branch is under development? I recommend using a [`[patch]` section](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) for this in a `.cargo/config.toml` file somewhere. That should let you override it locally to develop other crates in tandem without having to deal with temporary changes in checked-in files.

sprocket-doc.md
```diff
@@ -0,0 +1,77 @@
+# How to make the most of Sprocket doc
+
+Sprocket is capable of rendering rich HTML documentation for any WDL workspace! However to make the most of this tool, there are certain documentation conventions which should be followed. Of course you are free to ignore this document, and Sprocket should do a passable job of documenting your WDL, but there may be some awkward rendering or unexpected behavior if you aren't conforming to what the tool expects.
```
acfoltzer: I'd like to see this document start with a clear statement of purpose that gives the reader an idea of whether it's relevant to their work. They don't need our permission to ignore it; they will do that themselves regardless if they decide it's not worth their time 😅 

Here's a strawman attempt to illustrate what I mean, but you should feel free to make a version that uses your intended voice:

---

# Documenting WDL workflows with `sprocket dev doc`

The `sprocket dev doc` command produces rich HTML documentation for any valid WDL workflow. This page describes several practices that are optional, but that enhance the quality of the resulting documentation.

Cargo.toml
a-frantz: I'll look into it! I've seen Peter use `[patch]` before, but have considered it magic I haven't investigated haha. Thanks for the idea :)

sprocket-doc.md
```diff
@@ -0,0 +1,77 @@
...
+
+## General structure
+
+The generated documentation directory (named `docs` by default) is completely self contained and can be moved, zipped, and shared without any of the raw WDL files it documents. Please let us know if you run into any issues while sharing your documentation!
+
+The `docs` directory will always contain a `style.css` file, an `index.js` file, an `index.html` file, and an `assets/` directory filled with SVG icons and logos. The rest of the contents are dynamically created based on the structure of the WDL workspace being documented. Each WDL file found in the workspace will be represented as a directory within `docs`. The relative path from the root of the workspace to each WDL file is reproduced in the `docs` directory to keep your documentation organized.
+
+The directory for each WDL file will have an `index.html` file and one HTML file per struct, task, or workflow defined in the WDL.
```
acfoltzer: Outside of a reference document (as opposed to a how-to/guide/tutorial), try to keep each section actionable and make it clear why the reader should care. For this one, I think the actionable core would be better described as "Distributing generated documentation".

It's not clear what knowing the details of the directory structure allows a user to do; I would consider either making that explicit, or including the details in a separate reference document rather than a how-to.

sprocket-doc.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Custom themes
+
+While it is technically possible to supply your own custom CSS styling, this capability is currently undocumented. We recommend you stick with the default styling at this point in time, but do let us know what kinds of customization you would like to see in future releases! 
+
+## Per WDL file `index.html`
+
+At a minimum, the `index.html` associated with each WDL file will have a table of contents with links to each struct, task, or workflow documentation page in that directory. If the WDL file has a "preamble" (a comment block before the version statement where each line starts with `##`), that will be rendered as Markdown text above the table of contents.
+
+Example preamble:
```
acfoltzer: Another worked example of focusing a section on "what can I do and how?"
```suggestion
## Using preamble comments for file-level documentation

To provide top-level documentation for a file, add a comment block before the `version` statement where each line starts with `##`. These preamble comments will be rendered as Markdown above the generated table of contents on that file's dedicated page. For example:
```

sprocket-doc.md
```diff
@@ -0,0 +1,77 @@
...
+
+Structs defined in v1.2 WDL _will_ eventually get rich HTML documentation similar to tasks and workflows. However this has not yet been implemented, so they are given the same treatment as pre-v1.2 structs.
+
+## Parameters (inputs and outputs)
+
+Each input and output to a workflow or task should be documented, but there is some flexibility in the specifics.
+
+### Inputs
+
+Each entry in the `input` section of a task or workflow is expected to have a corresponding entry in the `parameter_meta` section. To get the most out of `sprocket doc`, it's recommended that each input entry be an object. That object should have at least a `description` key. There is special handling for the following sub-keys:
```
acfoltzer: The recommendation for using objects is confusing here given the [🗑️ next to `Object` in the spec](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#-object); is this right?

sprocket-doc.md
```diff
@@ -0,0 +1,77 @@
...
+## This is a paragraph with **bolding**, _italics_, and `code` formatting
+
+version 1.2
+
+workflow foo {}
+```
+
+## v1.0 and v1.1 structs
+
+The WDL specification does not offer a way to document structs prior to WDL v1.2, so the pages for them are rather limited. The pages for these structs will have a copy of the raw WDL definition.
```
acfoltzer: Is there something we can recommend to counter this limitation like including comments in the `struct` WDL even if they're not recognized as structured like a `meta` section would be? (I assume `meta` for structs is what's not supported before 1.2?)

sprocket-doc.md
```diff
@@ -0,0 +1,77 @@
...
+
+Each entry in the `input` section of a task or workflow is expected to have a corresponding entry in the `parameter_meta` section. To get the most out of `sprocket doc`, it's recommended that each input entry be an object. That object should have at least a `description` key. There is special handling for the following sub-keys:
+
+- `group`: all inputs sharing the same `String` value will be rendered together in a dedicated table
+    - required inputs are _always_ rendered under the "Required Inputs" table and thus should _not_ have a `group` key (it will be ignored if present)
+    - The `Common` group of inputs will always come after the required inputs
+    - Inputs without a `group` will be rendered under "Other Inputs" which will be the last input table
+    - The `Resource` group of inputs will immediately precede the "Other Inputs" table
+    - All other groups will render alphabetically between the `Common` table and the `Resource` table.
+- `help`: will render at the top of the "Additional Meta" cell of the table
```
acfoltzer: An example would be helpful here, perhaps with a screenshot or link to example rendered output resulting from a particular configuration of groups.

sprocket-doc.md
acfoltzer: > That object should have at least a `description` key. There is special handling for the following sub-keys:

The use of `key` in the first sentence and `sub-keys` in the second is confusing and makes me wonder if `group` and `help` are meant to be nested under `description`, rather than being siblings in a single object.

sprocket-doc.md
```diff
@@ -0,0 +1,77 @@
...
+
+If an input has a `String` value for its parameter meta entry instead of a meta object, that string value will be treated as if it were the `description` key of a meta object.
+
+### Outputs
+
+Outputs should be documented in one of two places: either in the task/workflow `meta` section under an `outputs` key or at the root of the `parameter_meta` section.
```
acfoltzer: A how-to vs. reference moment: if the former, we should recommend one or the other; if the latter, we should describe differences between the two choices and what happens if both are present.

sprocket-doc.md
```diff
@@ -0,0 +1,77 @@
...
+
+Every struct, task, and workflow `meta` section should have a `description` key with a `String` value. This description string can have Markdown formatting. The `description` string should be less than 140 characters or it will be clipped in some contexts.
+
+The `help` key should only have a `String` value, but can be of any length. It's best practice to keep `description` short, and put any additional text needed under the `help` key. Help strings can also be styled with Markdown!
+
+The `external_help` key should have a URL as its value, and will be rendered as a button which will open a new tab or window visiting the link.
+
+If a `warning` key has a `String` value, it will be rendered in a special "warning box" to draw the attention of users.
+
+Workflows can have a `category` key which will group workflow pages on the left sidebar.
```
acfoltzer: This definitely feels more like reference content than best-practices, and might work well as a table, e.g.,

| Key | Allowed contexts | Type | Description |
| --- | --------------- | ---- | ----------- |
| `help` | Structs, tasks, workflows | `String` | Short description (< 140 characters) that can include Markdown |

A best-practices presentation should consist primarily of an example with a link to the details for those interested.

sprocket-doc.md
acfoltzer, line 1: A couple top-level thoughts on this document, which should be taken with a grain of salt as they're ultimately subjective, and might not actually be relevant depending on what this document aims to achieve.

- The content alternates between how-to/guide/tutorial recommendations and reference material. Both certainly have a place, but the purposes of both are better served when they are presented separately with cross-references. On the whole, this feels more like a guide, so most of my specific comments below assume a commitment to that direction.
- We haven't adopted a style guide for docs on this project (yet?), but I recommend checking out the pages from the Google developer documentation style guide on [tone](https://developers.google.com/style/tone), [tense](https://developers.google.com/style/tense), and [voice](https://developers.google.com/style/voice).

Cargo.toml
acfoltzer: Cool, let me know if you'd like to pair on it sometime!

sprocket-doc.md
acfoltzer: Adding a link to the Google guide's page on [Prescriptive documentation](https://developers.google.com/style/prescriptive-documentation), which is a better term than my "how-to/guide/tutorial" mouthful

a-frantz: I think you're right that the current draft tries to walk the line between a how-to guide and a reference document, and we'd be better served splitting that into two separate documents (both of which should exist).

More immediately we need a how-to guide with the target audience of beta testers (who are volunteering to test drive experimental code, so are likely a more technical audience than our general user base), so going to incorporate your feedback from that angle.

I should write a proper reference document for this feature as there are a lot of tiny details about this implementation that need to be documented in a more comprehensive way than scattered throughout the Rust code. Some of this content could be reworked into that document.

sprocket-doc.md
a-frantz: Great question! This is definitely a detail I took for granted, but is not clear at all 😅 

I am _not_ talking about the `Object` WDL type you linked, but rather a "meta object". The WDL meta sections (both `meta` and `parameter_meta`) are made up of key-value pairs, where the value can be a variety of types, including a nested "object" with more key-values underneath it.

So the recommendation here can be illustrated with this WDL snippet:
```wdl
task task1 {
    input {
        Int foo
        Int bar
    }

    parameter_meta {
        foo: "the `foo` key does _not_ have a meta object value, but has a `String` value instead"
        bar: {
            description: "the `bar` key has a meta object value",
            help: "with the keys `description` and `help` which each have `String` values",
            arbitrary_nesting: {
                something: "this can get messy with arbitrary nesting",
                other_field: [1, 2, 42, 64],
            }
        }
    }
}
```

sprocket-doc.md
a-frantz: > Is there something we can recommend to counter this limitation like including comments in the struct WDL even if they're not recognized as structured like a meta section would be?

I'm hesitant to. It's a tricky problem. We (mostly me and Andrew but others have joined in at points) have discussed what to do about this on multiple occasions without arriving at any satisfactory decisions. I can elaborate on the pros and cons of any particular approach if you want to get into the weeds, but in short I'd like to avoid the Sprocket team making an official stance on "how to document structs prior to WDL v1.2" (which is where both `meta` and `parameter_meta` sections were introduced for structs. And yes, it is still called "parameter" meta even though a struct doesn't have parameters. The keys of the `parameter_meta` section of a task are expected to match the member names)

It is worth noting that with the current implementation of the `doc` command, any comments _within the struct definition_ will get rendered here. So that does suggest users should document their structs with comments within the definition (as opposed to above it), but I'm still a little iffy on pointing that out.

sprocket-doc.md
a-frantz: Instead of a table, I went with a definitions list. Probably still a little too reference-y, but not sure how to fix that while still conveying this kind of dense information in a guide style 🤔

sprocket-doc.md
a-frantz: True 🤔 however with things still in a prototype phase not sure how to accomplish an example at this point in time.

Once we have the `workflows` docs rendering at a public site, we can link there with references to the raw WDL which generated them.

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
+# How to make the most of Sprocket doc
+
+Sprocket is capable of rendering rich HTML documentation for any WDL workspace! However to make the most of this tool, there are certain documentation conventions which should be followed. This page describes practices that are optional, but that enhance the quality of the resulting documentation.
+
+If you've already generated documentation for your workspace and have found any awkward rendering or unexpected behavior, you should refer here to see if there's a convention offered to better address your specific use case.
```
adthrasher:
```suggestion
If you find any awkward rendering or unexpected behavior, refer to this document and see if a convention better addresses your specific use case.
```

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Homepage
+
+We encourage you to customize the experience of your users by writing a custom Markdown document that will be embedded at the root of your generated documentation. Every page is going to contain links back to the homepage. If you don't provide anything for the homepage, your users will be faced with an empty screen which says "There's nothing to see on this page".
```
adthrasher:
```suggestion
We encourage you to customize the experience of your user documentation by writing a custom Markdown document that will be embedded at the root of your generated documentation. Every page contains links back to the homepage. If no homepage is provided, your users will be faced with an empty screen stating "There's nothing to see on this page".
```

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
...
+workflow foo {}
+```
+
+## v1.0 and v1.1 structs
+
+The WDL specification does not offer a way to document structs prior to WDL v1.2, so the pages for them are rather limited. The pages for these structs will have a copy of the raw WDL definition.
+
+## v1.2 structs
+
+Structs defined in v1.2 WDL _will_ eventually get rich HTML documentation similar to tasks and workflows. However this has not yet been implemented, so they are given the same treatment as pre-v1.2 structs.
```
adthrasher:
```suggestion
HTML documentation for structs defined in v1.2 WDL has not yet been implemented, so they are given the same treatment as pre-v1.2 structs.
```

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Meta entries with special handling
+
+All meta entries will render in the final HTML documentation, but there are some special cases we introduce. Each of the below keys are expected to have a WDL `String` value.
```
adthrasher:
```suggestion
All meta entries will render in the final HTML documentation, but there are some special conventions we introduce. Each key below is expected to have a WDL `String` value.
```

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
...
+
+`description`
+: Every struct, task, and workflow `meta` section should have a `description` key. This description string can have Markdown formatting. The `description` string should be less than 140 characters or it will be clipped in some contexts.
+
+`help`
+: This text can be of any length. It's best practice to keep `description` short, and put any additional text needed under the `help` key. Help strings can also be styled with Markdown.
```
adthrasher:
```suggestion
: This text can be of any length. It is best practice to keep `description` short and put any additional text needed under the `help` key. Help strings can also be styled with Markdown.
```

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
...
+
+`external_help`
+: This key should have a URL as its value (i.e. a valid hyperlink represented as a WDL `String`), and will be rendered as a button which will open a new tab or window visiting the link.
+
+`warning`
+: This text will be rendered in a special "warning box" to draw the attention of users.
+
+## Parameters (inputs and outputs)
+
+Each input and output to a workflow or task should be documented, but there is some flexibility in the specifics. To get the most out of `sprocket doc`, it's recommended that each instance of parameter documentation be a meta object. That object should have at least a `description` key. If a parameter has a `String` value for its meta entry instead of a meta object, that string value will be treated as if it were the `description` key of a meta object with no other entries.
```
adthrasher:
```suggestion
Each input and output to a workflow or task should be documented, but there is some flexibility in the specifics. To get the most out of `sprocket doc`, it is recommended that each instance of parameter documentation be a meta object. That object should have at least a `description` key. If a parameter has a `String` value for its meta entry instead of a meta object, that string value will be treated as if it were the `description` key of a meta object with no other entries.
```

adthrasher: Does this need an example? Or is "meta object" clear enough?

adthrasher: My edit above is personal preference. I try to avoid contractions in formal writing, such as documentation.

a-frantz: > Does this need an example? Or is "meta object" clear enough?

I think it's ok for now. The group of testers I intend to share with is small enough that if any of them are unclear on what this means, I figure they will reach out directly in some way.

I said in a reply to Adam somewhere else in this PR that by the time we're ready for show time and an actual release, we should be hosting the `workflows` docs publically and then we can link to that with references back to the raw WDL for examples.

But for now, I think adding examples we'll have to update or remove is a hassle I don't think is worth it lol

adthrasher: Fair enough

sprocket-doc.md
adthrasher: > Once we have the `workflows` docs rendering at a public site, we can link there with references to the raw WDL which generated them.

In your opinion, what remains to get to that point? I'm not opposed to generating docs from this PR and hosting those, with the knowledge that those could change in the future.

a-frantz: There's not much harm in hooking up the prototype docs, and if you want to write the github action(s) needed to get that rolling, I think you should go for it!

The biggest obstacle at the moment is that we're re-implementing all the tables (which is the bulk of the content). But I think the CL API is relatively stable, I don't have plans to radically change anything in the build system (which should be irrelevant for users (e.g. the workflows repo), but that would be the most likely source of friction).

a-frantz: To get the current prototype running you'd have to build off my branches, which may not be trivial in an action. So not saying "jump on it **right now**" as it might be a better use of effort to wait till _something_ hits release.

I intend to make a `sprocket` bin release that exposes `doc` under the `dev` command-space to indicate it's prototype status before we go for a full release and announcement

adthrasher: I think we still have a CD action that uses `wdldoc`. I think we could 1. release the command under `dev`, 2. build and deploy `workflows` docs, and 3. send the announcement to the beta group. If we do that, I think it'd be relatively safe to go ahead and embed links to where the `workflows` docs will be in this PR.

sprocket-doc.md
```diff
@@ -0,0 +1,79 @@
+# How to make the most of Sprocket doc
+
+Sprocket is capable of rendering rich HTML documentation for any WDL workspace! However to make the most of this tool, there are certain documentation conventions which should be followed. This page describes practices that are optional, but that enhance the quality of the resulting documentation.
+
+If you find any awkward rendering or unexpected behavior, refer to this document and see if a convention better addresses your specific use case.
+
+## Homepage
+
+We encourage you to customize the experience of your user documentation by writing a custom Markdown document that will be embedded at the root of your generated documentation. Every page contains links back to the homepage. If no homepage is provided, your users will be faced with an empty screen stating "There's nothing to see on this page".
```
adthrasher: This should probably include guidance on what to name the custom markdown document.

a-frantz: O, the file can be named anything. But it has to be specified at the CL via the `--homepage <path>` flag.

I didn't include any information about the CL API in this doc, but maybe I should?

adthrasher: Ah, yeah, I found that flag after commenting. Maybe: 

```suggestion
We encourage you to customize the experience of your user documentation by writing a custom Markdown document which can be embedded at the root of your generated documentation by specifying it on the command line when generating the documentation. Every page contains links back to the homepage. If no homepage is provided, your users will be faced with an empty screen stating "There's nothing to see on this page".
```

sprocket-doc.md
acfoltzer: Oh, I think this part, whether as a table or a definition list, is great as a reference. Once there's enough fleshed out for separate reference and best practice docs, that's when the crosslinking would be beneficial.

sprocket-doc.md
acfoltzer: Ahh, I hadn't seen this note in the spec:

>Note that, unlike the WDL Object type, metadata objects are not deprecated and will continue to be supported in future versions.

sprocket-doc.md
acfoltzer: Is the hesitance to avoid biasing people one way or another, or to avoid appearing to endorse any solution that's not "upgrade to 1.2"?

There's already somewhat of an implication with "The pages for these structs will have a copy of the raw WDL definition" that comments will be included, and while it doesn't rise to an endorsement it does make a reader ask "why was this included?"

sprocket-doc.md
acfoltzer: If a screenshot of generated docs would be easier as an interim solution, I think that would do the job too.

Cargo.toml
a-frantz: @acfoltzer getting this branch ready for review and tried to use `[patch]`, but got the following error:
```
patch for `wdl` in `https://github.com/stjude-rust-labs/wdl` points to the same source, but patches must point to different sources
```
so I just changed the `revision` key to `branch` and pointed to my dev branch 🤷‍♀️ is there a better way to do this?

docs/cli/doc.md
```diff
@@ -0,0 +1,83 @@
...
+
+The generated documentation directory (named `docs` by default) is completely self contained and can be moved, zipped, and shared without any of the raw WDL files it documents. Please let us know if you run into any issues while sharing your documentation.
+
+## Custom themes
+
+While it is technically possible to supply your own custom CSS styling, this capability is currently undocumented. We recommend you stick with the default styling at this point in time, but do let us know what kinds of customization you would like to see in future releases! 
+
+## Using preamble comments for file-level documentation
+
+To provide top-level documentation for a file, add a comment block before the `version` statement where each line starts with `##`. These preamble comments will be rendered as Markdown above the generated table of contents on that file's dedicated page. For example:
```
adthrasher:
```suggestion
To provide top-level documentation for a file, add a comment block before the `version` statement where each line starts with a double pound sign (i.e., `##`, which we term a "preamble comment"). These preamble comments will be rendered as Markdown above the generated table of contents on that file's dedicated page. For example:
```

src/commands/doc.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    if args.watch {
+        if let Some(theme) = &args.theme {
+            let (tx, rx) = mpsc::channel::<NotifyResult<Event>>();
+            let mut watcher = recommended_watcher(tx)?;
+
+            watcher.watch(&theme.join("src"), RecursiveMode::Recursive)?;
+
+            println!("watching for changes in theme directory...");
+            println!("press Ctrl+C to stop watching");
```
adthrasher: Are these meant to be temporary? Or should they use the logging functionality, rather than `println!`?

a-frantz: I'm not sure. As stated in the doc comment (lines 47-49) this is really only meant to be "seen" by developers of the `wdl` crates. It would probably be best to use the actual logging functionality, so I can make that change now.

src/commands/doc.rs
```diff
@@ -0,0 +1,112 @@
...
+    .with_context(|| {
+        format!(
+            "failed to generate documentation for workspace at `{}`",
+            args.workspace.display()
+        )
+    })?;
+
+    if args.open {
+        opener::open(docs_dir.join("index.html"))
+            .map_err(|e| anyhow::anyhow!("failed to open documentation: {e}"))?;
```
acfoltzer: Using `Context` is better in situations like this because it preserves backtraces.
```suggestion
        opener::open(docs_dir.join("index.html"))
            .context("failed to open documentation"))?;
```

src/commands/doc.rs
```diff
@@ -0,0 +1,112 @@
...
+    ///
+    /// This argument is meant to be used by developers of the `wdl` crates;
+    /// customizing the theme used for the generated documentation is currently
+    /// unsupported.
+    #[arg(long, value_name = "DIR")]
+    pub theme: Option<PathBuf>,
+
+    /// Install the theme if it is not already installed.
+    ///
+    /// Requires the `--theme` argument to be specified.
```
acfoltzer: Maybe worth mentioning that this is an `npm` operation in this help text. Without context, I don't know what it means to install a theme.

src/commands/doc.rs
```diff
@@ -0,0 +1,112 @@
...
+    #[arg(long)]
```
acfoltzer: I believe that an attribute like this can encode the dependency between the arguments more directly, unfortunately I wasn't able to try it out as there's some other compile errors currently on the branch
```suggestion
    #[arg(long, requires = "theme")]
```

Cargo.toml
```diff
@@ -40,7 +41,7 @@ tracing-indicatif = "0.3.9"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 url = "2.5.4"
 walkdir = "2.5.0"
-wdl = { version = "0.14.0", features = ["cli", "codespan", "engine", "lsp"], git = "https://github.com/stjude-rust-labs/wdl", rev = "ccc8850" }
+wdl = { version = "0.14.0", features = ["cli", "codespan", "engine", "lsp", "doc"], git = "https://github.com/stjude-rust-labs/wdl", branch = "a-frantz/wdl-doc" }
```
acfoltzer: I know you're already aware, but I'd like to hang a comment thread here to make sure we get back to a mainline commit before merging

acfoltzer (changes requested):
Looking good, though perhaps it's worth converting the PR back to a draft to avoid a premature merge? There's currently a tempting green button that should not be pressed...

docs/cli/doc.md
```diff
@@ -0,0 +1,83 @@
...
+## v1.2 structs
+
+HTML documentation for structs defined in v1.2 WDL has not yet been implemented, so they are given the same treatment as pre-v1.2 structs.
+
+## Meta entries with special handling
+
+All meta entries will render in the final HTML documentation, but there are some special conventions we introduce. Each key below is expected to have a WDL `String` value.
+
+`description`
+: Every struct, task, and workflow `meta` section should have a `description` key. This description string can have Markdown formatting. The `description` string should be less than 140 characters or it will be clipped in some contexts.
```
a-frantz, line 54: Note to self: ensure this `X characters` length matches the final implementation

adthrasher (approved):

Cargo.toml
acfoltzer: That's what I would expect to happen if your local `wdl` checkout matches the revision that `Cargo.toml` points to; the patch section is mostly useful for the intermediate phases when one development branch depends on another development branch. This feature is a little fiddly, so I'm happy to go over it in a chat, but I'm gonna resolve this thread for now in preparation to merge.

PR #113, opened 2025-05-07
feat: create template input JSONs

adthrasher:
Address #40. Create template JSONs for inputs. By default, it prints only top-level inputs that are required (without defaults) and inputs that have a literal default value. Arguments allow the user to include more complex expressions (`--show-expressions`), hide inputs with defaults (`--hide-defaults`), and also recurse to find nested inputs (`--nested-inputs`). Output defaults to JSON, but can be switched to YAML (`--yaml`). 

Input rendering is based on discussion in #66 as well as additional Slack discussion. Feel free to disagree with decisions that were made.

For external contributors:

For all contributors:


src/commands/input.rs
```diff
@@ -0,0 +1,439 @@
...
+use serde_json::Map;
+use serde_json::Value;
+use wdl::analysis::types::CallKind;
+use wdl::ast::AstNode;
+use wdl::ast::AstToken;
+use wdl::ast::SyntaxKind;
+use wdl::ast::v1::Expr;
+use wdl::ast::v1::InputSection;
+use wdl::ast::v1::LiteralExpr;
+use wdl::ast::v1::Type;
```
a-frantz: this uses a lot from `ast`. Is it possible to use some more from `analysis` to simplify the code?

To be clear, I don't know. Just a hunch 🤷‍♀️

src/commands/input.rs
```diff
@@ -0,0 +1,439 @@
...
+use wdl::cli::analysis::AnalysisResults;
+use wdl::cli::analysis::Source;
+
+/// Arguments for the `input` subcommand.
+#[derive(Parser, Debug)]
+pub struct InputArgs {
+    /// The path to the WDL document or a directory containing WDL documents to
+    /// validate.
+    #[arg(value_name = "PATH or URL")]
+    pub path: Source,
```
a-frantz: How would a directory work? Also this needs a `name` arg to support tasks

a-frantz: oop it's also talking about "validate" when it should be about `input`

adthrasher: Good catch. Lazy copy/past inheritance. Task support also needs to be implemented, though it should be fairly easy to add.

src/commands/input.rs
adthrasher: Agreed. If there is a way to get the relevant information from `analysis`, that'd be ideal. Admittedly, I started with `ast` and switched to `analysis` later (I haven't used it previously). It's possible I missed features in the `analysis` crate.

Cargo.toml
```diff
@@ -25,7 +25,8 @@ indicatif = "0.17.8"
 nonempty = "0.11.0"
 pest = { version = "2.7.14", features = ["pretty-print"] }
 pretty_assertions = "1.4.1"
-serde_json = "1.0.140"
+serde_json = { version = "1.0.140", features = ["preserve_order"] }
+serde_yaml = "0.9.34"
```
claymcleod: This should be `serde_yaml_ng`.

README.md
```diff
@@ -34,6 +34,7 @@
 - **`sprocket check`** performs static analysis on a document or directory of documents.
 - **`sprocket explain`** explains validation and lint rules supported by Sprocket.
 - **`sprocket format`** formats a document.
+- **`sprocket input`** writes template JSON for WDL input.
```
claymcleod:
```suggestion
- **`sprocket input`** writes template input file (JSON or YAML) for a task or workflow.
```

src/commands/input.rs
claymcleod: As I was writing a precursor to this, I recall needing a _lot_ from `wdl_ast`. Not to say that you can't use `wdl_analysis` some, but you definitely won't be able to eliminate it.

src/commands/input.rs
```diff
@@ -0,0 +1,469 @@
...
+use wdl::ast::v1::InputSection;
+use wdl::ast::v1::LiteralExpr;
+use wdl::ast::v1::Type;
+use wdl::cli::Analysis;
+use wdl::cli::analysis::AnalysisResults;
+use wdl::cli::analysis::Source;
+
+/// Arguments for the `input` subcommand.
+#[derive(Parser, Debug)]
+pub struct InputArgs {
```
claymcleod:
```suggestion
pub struct Args {
```

claymcleod: This should just be `Args` unless there are multiple `Args` in the subcommand.

src/commands/input.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// The path to the WDL document for which to generate an input template.
+    #[arg(value_name = "PATH or URL")]
+    pub path: Source,
```
claymcleod: I would unify your names and docstrings with what we did for the `run` subcommand, just so that we are consistent within our subcommands.

src/commands/input.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    /// Show inputs with non-literal default values and non-required inputs.
+    #[arg(long)]
+    pub show_expressions: bool,
+
+    /// Hide inputs with default values.
+    #[arg(long)]
+    pub hide_defaults: bool,
+
+    /// Include task-level inputs.
```
claymcleod:
```suggestion
    /// Generate inputs for all tasks called in the workflow.
    ///
    /// When using this option may want to omit `--show-expressions` to avoid 
    /// cluttering the output with task-level inputs.
```

Then you can eliminate the long help I believe.

src/commands/input.rs
```diff
@@ -0,0 +1,469 @@
...
+    #[arg(
+        long,
+        long_help = "Includes inputs from tasks called in the workflow. When using this option \
+                     may want to omit `--show-expressions` to avoid cluttering the output with \
+                     task-level inputs."
+    )]
+    pub nested_inputs: bool,
+
+    /// Output input template in YAML format.
```
claymcleod:
```suggestion
    /// Output the template as a YAML file.
```

Just reads a bit less awkward.

claymcleod:
I've suggested a great deal of updates here: https://github.com/adthrasher/sprocket/pull/1. Beyond those, I would also advocate this be called `inputs` instead of `input` (or, if we want it to remain singular, `input-file` at a minimum).

src/commands/inputs.rs
```diff
@@ -0,0 +1,484 @@
...
+    #[clap(short, long, value_name = "NAME")]
+    pub name: Option<String>,
+
+    /// Show inputs with non-literal default values and non-required inputs.
+    #[arg(long)]
+    pub show_expressions: bool,
+
+    /// Include inputs with default values.
+    #[arg(long)]
+    pub include_defaults: bool,
```
a-frantz: It's not clear to me just from these docstrings how these flags interact.
How would I show optional inputs without a default? Those are semantically equivalent to a defaulted input with `= None`, so they should appear with `--include-defaults`, but that should be made explicitly clear in the docstring

src/commands/inputs.rs
```diff
@@ -0,0 +1,484 @@
...
+        .run()
+        .await
+    {
+        Ok(results) => results,
+        Err(errors) => {
+            // SAFETY: this is a non-empty, so it must always have a first
+            // element.
+            bail!(errors.into_iter().next().unwrap())
+        }
+    };
```
a-frantz, line 450: there should be an error if the `Source` is a directory

src/commands/inputs.rs
adthrasher: I'm not sure. I merged Clay's changes, but I'm now realizing they're not correct.

adthrasher: With the way it is currently written, `--show-expressions` does nothing on its own.

adthrasher: It actually makes little sense to me how it is currently written. Outputs aren't consistently ordered between runs. Also this `"rnaseq_standard.bam_to_fastqs.fqlint.disable_validator_codes": "Array[String]",` is output for `--include-defaults`. It explicitly does not include the `default`, which is only included if you also supply `--show-expressions`: `"rnaseq_standard.bam_to_fastqs.fqlint.disable_validator_codes": "Array[String] (default = [])",`

a-frantz: I think the best behavior might be:
```suggestion
    /// Show inputs with literal default values and non-required inputs.
    #[arg(long)]
    pub show_literal_defaults: bool,

    /// Include inputs with complex expressions for defaults. Ignored if `--show-literal-defaults` not specified.
    #[arg(long)]
    pub show_expressions: bool,
```
?

adthrasher: @claymcleod - It looks like you took a much different approach to what these arguments were meant to be. Can you clarify your intent with them? It looks like your implementation of `show_expressions` only controls whether the complex expression component is included in the output. My original implementation omitted the entire input if its value was a complex expression.

a-frantz: IMO that should be:
```
"rnaseq_standard.bam_to_fastqs.fqlint.disable_validator_codes": []
```

adthrasher: @claymcleod - Also I'm curious if you know immediately why the ordering of the output varies between runs.

adthrasher: > IMO that should be:
> 
> ```
> "rnaseq_standard.bam_to_fastqs.fqlint.disable_validator_codes": []
> ```

I'm fairly certain that was the previous behavior, but I haven't rolled back the changes to check.

adthrasher: I'm tempted to shelve this until @claymcleod weighs in. It appears he has a significantly different vision for this subcommand than the one we outlined previously. I'd like to hear his reasons before attempting to recapitulate the prior behavior.

claymcleod: For the output ordering, running the following command with [the commit I started with](https://github.com/claymcleod/sprocket/commit/03db56387e164ea3cb5cc5ef9efbc051f484e515) yields inputs in different orders across runs.

```
sprocket input https://raw.githubusercontent.com/stjudecloud/workflows/refs/heads/main/workflows/rnaseq/rnaseq-standard.wdl --show-expressions --nested-inputs
```

So, though I also did not the output order when I made these changes, it seems that this behavior is not a regression. ~This makes sense too, because `serde_json::Map` does not preserve the order of inserts unless you tell it to with the `preserve_order` feature.~ EDIT: upon looking further, we _do_ actually have that feature enabled. I'm assuming then it's something else, like the order with which the analysis completes for imported files. At any rate, sorting however you like before output in the way to go IMO.

---

For the other pieces, admittedly I did totally interpret what these flags were meant to be differently. What I was aiming for was:

- `--includes-defaults` only modulates whether input parameters with a default value are shown.
- `--show-expressions` plays off of the idea of showing the more complicated expression (as defined by `decl.expr()`) when an easy conversion to a value is not known.

The confusing bit that doesn't align with the above cleanly is be that _some_ expressions are converted to literals automatically if they are straightforward enough, regardless of the value of `--show-expressions`. 

That being said, the underlying point is that I appear to have taken a stronger separation of concerns approach regarding what "including defaults" and "including expressions" means (it is true that the `--show-expressions` flag has no effect without `--include-defaults` because required inputs never have an expression i.e. they are unbound).

If I understand right, the two are you are advocating for a more practical approach, whereby

- `--show-defaults` is not just about whether or not to display a default input parameter but also around showing the default value directly rather than annotating its type.
- for `--show-expressions`... I'm not actually sure.

Take the code I referred to above that I started with. If you run this command:

```bash
sprocket input https://raw.githubusercontent.com/stjudecloud/workflows/refs/heads/main/workflows/rnaseq/rnaseq-standard.wdl --show-expressions
```

You get this output.

```json
{
  "rnaseq_standard.bam": "File",
  "rnaseq_standard.gtf": "File",
  "rnaseq_standard.star_db": "File",
  "rnaseq_standard.xenocp_aligner": "star",
  "rnaseq_standard.strandedness": "",
  "rnaseq_standard.mark_duplicates": false,
  "rnaseq_standard.cleanse_xenograft": false,
  "rnaseq_standard.validate_input": true,
  "rnaseq_standard.use_all_cores": false,
  "rnaseq_standard.subsample_n_reads": -1
}
```

But if you run it with `--show-expressions`:

```bash
sprocket input https://raw.githubusercontent.com/stjudecloud/workflows/refs/heads/main/workflows/rnaseq/rnaseq-standard.wdl --show-expressions
```

You get this output.

```json
{
  "rnaseq_standard.bam": "File",
  "rnaseq_standard.gtf": "File",
  "rnaseq_standard.star_db": "File",
  "rnaseq_standard.contaminant_db": null,
  "rnaseq_standard.prefix": "String (default = basename(bam, \".bam\"))",
  "rnaseq_standard.xenocp_aligner": "star",
  "rnaseq_standard.strandedness": "",
  "rnaseq_standard.mark_duplicates": false,
  "rnaseq_standard.cleanse_xenograft": false,
  "rnaseq_standard.validate_input": true,
  "rnaseq_standard.use_all_cores": false,
  "rnaseq_standard.subsample_n_reads": -1
}
```

The difference here is the following:

```txt
4a5,6
>   "rnaseq_standard.contaminant_db": null,
>   "rnaseq_standard.prefix": "String (default = basename(bam, \".bam\"))",
```

A couple of questions come out of that.

1. Doesn't `rnaseq_standard.contaminant_db` have a default value of `null` regardless of if `--show-expressions` is used? Why is it only showing up for `--show-expressions`? I don't see the relation between that showing up with the flag at all. 

To be sure, the spec says:

> An optional declaration has a default initialization of `None`, which indicates that it is undefined. An optional declaration may be initialized to any literal or expression of the correct type, including the special None value.

2. Perhaps similarly, `rnaseq_standard.prefix` _does_ have a default value—it's just that it's a complex expression. Why then does it only show up with `--show-expressions`? I would expect to see it in the list by default with a value of `"String"` since this implementation shows default values without `--hide-defaults`.

a-frantz: Here's the behavior I want to see _as a user_:

Without any flags, I would want just the required inputs with a type annotation as the value.

I'd like to modify that behavior with a flag that will fill in literal defaults (i.e., I don't have to edit these values to get this file to run, so they should not have type annotations. This would also include `null` for optional inputs).

Then there would be a "show me everything that can be tweaked" flag, which includes things that are calculated at runtime based on other values (complex expressions) that can't be rendered in JSON. These should have a type annotation and the runtime expression as a string.

Lastly, nested inputs should be viewable with a flag.

a-frantz: The above is the optimal behavior of the command IMO in terms of maximizing usefulness, but I would be interested in hearing any alternatives.

As for the ordering of the JSON, I think the best behavior is to probably put "you need to edit this field" values at the top of the file (so required inputs and "complex expression" values that are type annotated), but I'm ambivalent about sub-ordering beyond that (assuming it's consistent across re-runs).

a-frantz: > I'd like to modify that behavior with a flag that will fill in literal defaults (i.e., I don't have to edit these values to get this file to run, so they should not have type annotations. This would also include null for optional inputs).

After some more thought, I'd argue that should probably be the default behavior, as it's the most useful. I imagine that most times I personally use this command, I would be specifying this flag. Many of our WDL workflows as well as external workflows are really meant to have some of the defaults overridden (e.g. read2 FASTQs which are defaulted to an empty list).

That said, only outputting the required inputs is a sane default behavior and would be ok if someone strongly disagrees with me.

claymcleod: Yeah I agree with all of the above. If Andrew agrees with it, I think the flags should be tweaked in this direction. I do think it's a good idea to go ahead and print literals without requiring a flag when we can.

README.md
```diff
@@ -32,13 +32,14 @@
...
 - **`sprocket check`** performs static analysis on a document or directory of documents.
+- **`sprocket completions`** generates shell completions for Sprocket.
 - **`sprocket config`** prints configuration values.
 - **`sprocket explain`** explains validation and lint rules supported by Sprocket.
 - **`sprocket format`** formats a document or directory of documents.
+- **`sprocket inputs`** writes template input file (JSON or YAML) for a task or workflow.
 - **`sprocket lint`** performs static analysis on a document or directory of documents with additional linting rules enabled (effectively a shortcut for `check --lint`).
 - **`sprocket run`** runs a task or workflow.
 - **`sprocket validate`** validates a set of inputs read from files or on the command line against a task or workflow.
-- **`sprocket completions`** generates shell completions for Sprocket.
```
a-frantz, line 41: This was discussed somewhere I can't remember, but are we sure we want this sorted alphabetically? I'd slightly prefer this in order of "priority" (which is subjective). IMO:
- run
- analyzer
- check
- lint
- explain
- validate
- format
- config
- completions

We don't need to decide in this PR, but starting the conversation

adthrasher: I think @claymcleod asked for an alphabetized list previously. It's worth a discussion, though. I do think a prioritized list might make sense here.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Added tab completions for `sprocket` commands ([#105](https://github.com/stjude-rust-labs/sprocket/pull/105)).
 * Added `shellcheck` to Dockerfile ([#114](https://github.com/stjude-rust-labs/sprocket/pull/114)).
+* Introduced the `input` subcommand ([#113](https://github.com/stjude-rust-labs/sprocket/pull/113)).
```
```suggestion
* Introduced the `inputs` subcommand ([#113](https://github.com/stjude-rust-labs/sprocket/pull/113)).
```

src/commands/inputs.rs
```diff
@@ -0,0 +1,540 @@
+//! Implementation of the `input` command.
```
a-frantz:
```suggestion
//! Implementation of the `inputs` command.
```

src/commands/inputs.rs
```diff
@@ -0,0 +1,540 @@
...
+use wdl::ast::v1::Decl;
+use wdl::ast::v1::Expr;
+use wdl::ast::v1::InputSection;
+use wdl::ast::v1::LiteralExpr;
+use wdl::ast::v1::TaskDefinition;
+use wdl::ast::v1::Type;
+use wdl::cli::Analysis;
+use wdl::cli::analysis::Source;
+
+/// Arguments for the `input` subcommand.
```
a-frantz:
```suggestion
/// Arguments for the `inputs` subcommand.
```

src/commands/inputs.rs
```diff
@@ -0,0 +1,540 @@
...
+    #[arg(long)]
+    pub show_expressions: bool,
+
+    /// Hide inputs with default values.
+    #[arg(long, conflicts_with = "show_expressions")]
+    pub hide_defaults: bool,
+
+    /// Generate inputs for all tasks called in the workflow.  
+    #[arg(long)]
+    pub include_nested_inputs: bool,
```
a-frantz:
```suggestion
    pub nested_inputs: bool,
```
nit: I think this is currently too long to type and should be shortened

src/commands/inputs.rs
```diff
@@ -0,0 +1,540 @@
...
+                        let (k, v) = item.key_value();
+                        let key_name: String = match k {
+                            Expr::Literal(LiteralExpr::String(k)) => {
+                                if let Some(text) = k.text() {
+                                    let mut buffer = String::new();
+                                    text.unescape_to(&mut buffer);
+                                    buffer
+                                } else {
+                                    String::new()
+                                }
```
a-frantz: I believe that `text()` here always returns `Some` unless there's an interpolation. Is returning `""` the right behavior there? I'm not sure what should happen if it's something like
```
String weird_input = "this is ~{bizarre}"
Int bizarre = 42
```
This should probably be rendered under the "complex expression" case.

a-frantz: Ah wait. This is inside a map? Now I'm really unsure how to handle it. Would something like the above be a legal key?

src/commands/inputs.rs
```diff
@@ -0,0 +1,540 @@
...
+                            }
+                            _ => k.text().to_string(),
```
a-frantz, line 184: This case is the `Map` has a non-String key? Should this instead recurse maybe? Wouldn't we want an `Int` map to render as number values instead of strings?

src/commands/inputs.rs
```diff
@@ -0,0 +1,540 @@
...
+                        .filter_map(|item| {
+                            let (name, value) = item.name_value();
+                            self.expression(ty.clone(), Key::empty(), name.text(), &value)
+                        })
+                        .map(|(k, v)| (k.join().expect("key to join"), v))
+                        .collect::<Map<_, _>>();
+
+                    return Some((namespace.push(name), Value::Object(map)));
+                }
+                _ => {
```
a-frantz: what is this case? It's not clear to me

src/commands/inputs.rs
adthrasher: The WDL spec requires map keys to be of primitive type. I don't think interpolation is allowed.

src/commands/inputs.rs
adthrasher: A `LiteralExpr` has additional variants: `None`, `Hints`, `Input`, and `Output`.

src/commands/inputs.rs
a-frantz: In that case, I think this can be switched from an `If let Some()... else` to an `expect()`

src/commands/inputs.rs
a-frantz: I think `None` should be converted to `null` and then we can explicitly match on the remaining 3 variants as an error case as they aren't allowed within an input section. Does that sound correct?

src/commands/inputs.rs
adthrasher: Ideally, but serde_json's map is `Map<String, Value>`. JSON objects require the key to be a string. The strings will need to be parsed to the correct type in the consumer code.

src/commands/inputs.rs
```diff
@@ -0,0 +1,540 @@
...
+                        .collect::<Map<_, _>>();
+
+                    return Some((namespace.push(name), Value::Object(map)));
+                }
+                LiteralExpr::None(_) => {
+                    return Some((namespace.push(name), Value::Null));
+                }
+                LiteralExpr::Hints(_) | LiteralExpr::Input(_) | LiteralExpr::Output(_) => {
+                    // These are not allowed in the input section.
+                    panic!("unexpected literal expression");
```
a-frantz: I think that's it's discouraged to call `panic!` directly in Rust? This could instead call `unreachable!()` which is clearer in intention

a-frantz (approved):
Looks pretty awesome to me. Excited to get this into production 🤓 

I have some additional thoughts, but they don't need to stand in the way of a merge.

1. I made a comment about this earlier, but I expected this implementation to make more use of `wdl-analysis` than it does (as opposed to doing so much "manually" using the AST). I wonder if some of the logic needed for this could be added upstream in `wdl-analysis`? cc @peterhuene 
2. I believe that this implementation introduces at least one new "input syntax" (`"left"` and `"right"` for Pairs), and we should make sure anything introduced here is supported upstream in `wdl-engine`. Can we open tracking issues?

adthrasher:
> Looks pretty awesome to me. Excited to get this into production 🤓
> 
> I have some additional thoughts, but they don't need to stand in the way of a merge.
> 
> 1. I made a comment about this earlier, but I expected this implementation to make more use of `wdl-analysis` than it does (as opposed to doing so much "manually" using the AST). I wonder if some of the logic needed for this could be added upstream in `wdl-analysis`? cc @peterhuene
> 2. I believe that this implementation introduces at least one new "input syntax" (`"left"` and `"right"` for Pairs), and we should make sure anything introduced here is supported upstream in `wdl-engine`. Can we open tracking issues?

Pair syntax is tracked here: https://github.com/stjude-rust-labs/wdl/issues/450

peterhuene (approved):
This looks great!

peterhuene:
> I made a comment about this earlier, but I expected this implementation to make more use of wdl-analysis than it does (as opposed to doing so much "manually" using the AST). I wonder if some of the logic needed for this could be added upstream in wdl-analysis? cc @peterhuene

I think a pure AST approach is fine as it's pulling the defaults from literal expressions; for any other expressions, we wouldn't be able to statically determine a default value anyway (other than for expressions whose operands are also literals, perhaps, such as `1+2`, if we wanted to get slightly fancier, but probably not worth it).

Issue #116, opened 2025-05-15
bug: arguments to `--except` are no longer case insensitive


a-frantz:
this is a regression from prior behavior

bfrazho:
Hi Ari, 

I'm looking to be a new contributor to these projects, and this one seems to be a pretty straight forward issue to solve to get started. :) 

I think I might have a solution already, but just to clarify, this feature is only a part of the "check" command, correct?

The readme also suggested that there were tests as part of the project, but I couldn't find any existing test suite specifically for this one. For now I've added a new test folder very roughly following the wdl repository so I could test drive the solution, would that be appropriate?

Thanks! :)

bfrazho:
@a-frantz would it be alright if I took this issue? :)

a-frantz:
@bfrazho hi there! We don't assign issues to external collaborators anymore, but you are welcome to open a PR to address this!

> I think I might have a solution already, but just to clarify, this feature is only a part of the "check" command, correct?

This is correct!

> The readme also suggested that there were tests as part of the project, but I couldn't find any existing test suite specifically for this one.

Yea the tests for the `sprocket` binary are a bit light. We would welcome some new tests!

bfrazho:
@a-frantz oh okay that's good to know, thanks!

 I was going through the PR checklist, do I just leave these checks blank then?
- [ ] You have added yourself or the appropriate individual as the assignee.
- [ ] You have added at least one relevant code reviewer to the PR.

Just want to make sure I'm approaching this appropriately!

PR #118, opened 2025-05-24
Fixed case sensitivity on --except flag on check function #116 

bfrazho:
Fixed case sensitivity on --except flag on check function #116 

Internal code seems to address case insensitivity, but Clap's "PossibleValuesParser" is case sensitive by default, and was causing it to fail early. 

- set an ignore_case flag that is provided by Clap to get it to work as intended.
- added a new tests folder because it appears that there aren't any test cases directly for this project. 
- added a new test dependency "assert_cmd" in order to write tests that use the sprocket binary directly.
- wrote basic happy path test for check function
- wrote test that gives error output for a caught check
- wrote test that uses the --except flag matching case
- wrote test that uses the --except flag with mixed wrong case

For external contributors:

- [X] You have read the `wdl` crates' [CONTRIBUTING](https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md) guide in its entirety.

For all contributors:

- [X] You have added a few sentences describing the PR here.
- [X] Your code builds clean without any errors or warnings.
- [X] You have added tests (when appropriate).
- [X] You have added an entry in the CHANGELOG (when appropriate).
- [X] You have updated the README or other documentation to account for these changes (when appropriate).


src/commands/check.rs
```diff
@@ -33,6 +33,7 @@ pub struct Common {
     /// Repeat the flag multiple times to except multiple rules.
     #[clap(short, long, value_name = "RULE",
         value_parser = PossibleValuesParser::new(ALL_RULE_IDS.iter()),
+        ignore_case = true,
```
a-frantz, line 36: I think this same fix is going to be needed for the `explain` command as well. Can you add it there too?

tests/check.rs
a-frantz, line 1: This is a great start, but it looks like it's going to be tough to maintain if anything upstream changes (e.g. if the errors get their format tweaked, that will be annoying to track down and update here as it's kind of buried in the Rust and repeated in several places).

I think this would be better architected to more closely follow the `wdl` crates tests, where the test results are stored as full files (usually named `errors.txt`) which can be mass updated using a `BLESS` environment variable.

With that being said, I don't think you specifically need to get in the weeds with writing test infrastructure from scratch 😅
I think we can merge this without the tests for now and someone can do a follow-up PR that adds more thorough testing.

src/commands/check.rs
bfrazho: Yep, I think I can manage that 😁

tests/check.rs
bfrazho: That's fair, admittedly that is partially why I left the snippet that I was comparing pretty short.

I took a quick look at wdl tests, it feels pretty similar to UI snapshot sorts of tests, so definitely makes sense. I might take a swing at making a test suite closer to that example. I'll remove the tests for now though so we can close this out :)

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added tab completions for `sprocket` commands ([#105](https://github.com/stjude-rust-labs/sprocket/pull/105)).
 * Added `shellcheck` to Dockerfile ([#114](https://github.com/stjude-rust-labs/sprocket/pull/114)).
 * Introduced the `inputs` subcommand ([#113](https://github.com/stjude-rust-labs/sprocket/pull/113)).
+* Fixed check --except and explain rule not being case-insensitive ([#116](https://github.com/stjude-rust-labs/sprocket/issues/116)).
```
```suggestion
* Fixed `check --except` and `explain` rule not being case-insensitive ([#116](https://github.com/stjude-rust-labs/sprocket/issues/116)).
```

PR #120, opened 2025-06-03
feat: container lock file

adthrasher:
Add a `lock` subcommand to write the current Docker shasums for the specified tags to a lock file. Additional changes will be needed to consume this file during runs.

For external contributors:

For all contributors:

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -8,6 +8,7 @@ pub mod completions;
 pub mod config;
 pub mod explain;
 pub mod format;
+pub mod lock;
 pub mod run;
 pub mod validate;
 
@@ -36,6 +37,9 @@ pub enum Commands {
 
     /// Lints a document or a directory containing documents.
     Lint(check::LintArgs),
+
+    /// Locks Docker images to a sha256 digest.
+    Lock(lock::Args),
 
     /// Runs a task or workflow.
     Run(run::Args),
--- a/src/commands/lock.rs
+++ b/src/commands/lock.rs
@@ -1,0 +1,112 @@
+//! Implementation of the `lock` subcommand.
+
+use std::collections::HashMap;
+use std::collections::HashSet;
+
+use anyhow::Result;
+use anyhow::bail;
+use clap::Parser;
+use crankshaft_docker::Docker as crankshaft_docker;
+use wdl::ast::AstNode;
+use wdl::ast::AstToken;
+use wdl::cli::Analysis;
+use wdl::cli::analysis::Source;
+
+use crate::Mode;
+
+/// Arguments for the `lock` subcommand.
+#[derive(Parser, Debug)]
+pub struct Args {
+    /// A source WDL file or URL.
+    #[clap(value_name = "PATH or URL")]
+    pub source: Source,
+
+    /// Disables color output.
+    #[arg(long)]
+    pub no_color: bool,
+
+    /// The report mode.
+    #[arg(short = 'm', long, value_name = "MODE")]
+    pub report_mode: Option<Mode>,
+}
+
+impl Args {
+    /// Applies the configuration to the command arguments.
+    pub fn apply(mut self, config: crate::config::Config) -> Self {
+        self.no_color = self.no_color || !config.common.color;
+        self.report_mode = match self.report_mode {
+            Some(mode) => Some(mode),
+            None => Some(config.common.report_mode),
+        };
+        self
+    }
+}
+
+/// Performs the `lock` command.
+pub async fn lock(args: Args) -> Result<()> {
+    let results = match Analysis::default().add_source(args.source).run().await {
+        Ok(results) => results,
+        Err(errors) => {
+            // SAFETY: this is a non-empty, so it must always have a first
+            // element.
+            bail!(errors.into_iter().next().unwrap())
+        }
+    };
+
+    let mut images: HashSet<String> = HashSet::new();
+    for result in results {
+        let doc = result.document().root();
+        for task in result.document().tasks() {
+            doc.ast()
+                .as_v1()
+                .expect("should be a v1 document")
+                .tasks()
+                .filter(|t| t.name().text() == task.name())
+                .for_each(|t| {
+                    if let Some(runtime) = t.runtime() {
+                        if let Some(container) = runtime.container() {
+                            if let Ok(image) = container.value() {
+                                if let Ok(text) =
+                                    serde_json::from_str(image.expr().text().to_string().as_str())
+                                {
+                                    images.insert(text);
+                                }
+                            }
+                        }
+                    }
+                })
+        }
+    }
+
+    let mut map: HashMap<String, String> = HashMap::new();
+    for image in images {
+        let prefix = image.split(':').next().unwrap_or("");
+        let docker = crankshaft_docker::with_defaults()?;
+        docker
+            .ensure_image(&image)
+            .await
+            .expect("should ensure image");
+
+        let image_info = docker
+            .inner()
+            .inspect_image(image.as_str())
+            .await
+            .expect("should inspect image");
+        if let Some(digests) = image_info.repo_digests {
+            for d in digests {
+                if !d.starts_with(prefix) {
+                    continue;
+                }
+                map.insert(image.clone(), d.clone());
+            }
+        }
+    }
+
+    if !map.is_empty() {
+        let data = toml::to_string_pretty(&map)?;
+        let path = "sprocket.lock";
+        std::fs::write(path, data)?;
+    }
+
+    Ok(())
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,12 +101,12 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.7"
+version = "3.0.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca3534e77181a9cc07539ad51f2141fe32f6c3ffd4df76db8ad92346b003ae4e"
+checksum = "6680de5231bd6ee4c6191b8a1325daa282b415391ec9d3a37bd34f2060dc73fa"
 dependencies = [
  "anstyle",
- "once_cell",
+ "once_cell_polyfill",
  "windows-sys 0.59.0",
 ]
 
@@ -238,9 +238,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.9.0"
+version = "2.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c8214115b7bf84099f1309324e63141d4c5d7cc26862f97a0a857dbefe165bd"
+checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
 dependencies = [
  "serde",
 ]
@@ -269,9 +269,9 @@ dependencies = [
 
 [[package]]
 name = "bollard"
-version = "0.19.0-rc1"
+version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b30ae2237f0d8f5d2ccb412fc9e3b415aab8a19cf62d0fac494eba684aba14c"
+checksum = "af706e9dc793491dd382c99c22fde6e9934433d4cc0d6a4b34eb2cdc57a5c917"
 dependencies = [
  "base64 0.22.1",
  "bollard-stubs",
@@ -357,9 +357,9 @@ checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
 
 [[package]]
 name = "cc"
-version = "1.2.21"
+version = "1.2.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8691782945451c1c383942c4874dbe63814f61cb57ef773cda2972682b7bb3c0"
+checksum = "16595d3be041c03b09d08d0858631facccee9221e579704070e6e9e4915d3bc7"
 dependencies = [
  "shlex",
 ]
@@ -393,9 +393,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.37"
+version = "4.5.38"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eccb054f56cbd38340b380d4a8e69ef1f02f1af43db2f0cc817a4774d80ae071"
+checksum = "ed93b9805f8ba930df42c2590f05453d5ec36cbb85d018868a5b24d31f6ac000"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -403,9 +403,9 @@ dependencies = [
 
 [[package]]
 name = "clap-verbosity-flag"
-version = "3.0.2"
+version = "3.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2678fade3b77aa3a8ff3aae87e9c008d3fb00473a41c71fbf74e91c8c7b37e84"
+checksum = "eeab6a5cdfc795a05538422012f20a5496f050223c91be4e5420bfd13c641fb1"
 dependencies = [
  "clap",
  "log",
@@ -414,9 +414,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.37"
+version = "4.5.38"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "efd9466fac8543255d3b1fcad4762c5e116ffe808c8a3043d4263cd4fd4862a2"
+checksum = "379026ff283facf611b0ea629334361c4211d1b12ee01024eec1591133b04120"
 dependencies = [
  "anstream",
  "anstyle",
@@ -587,8 +587,7 @@ dependencies = [
 [[package]]
 name = "crankshaft"
 version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c4b3b078dc5c110d3e2dc3396909b2f4b516a013ea548467e889f224efd95fa5"
+source = "git+https://github.com/stjude-rust-labs/crankshaft?rev=6b17024#6b170247c00993a43d71571e56e10e251d6d89da"
 dependencies = [
  "crankshaft-config",
  "crankshaft-engine",
@@ -597,8 +596,7 @@ dependencies = [
 [[package]]
 name = "crankshaft-config"
 version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc556efa065005f792a179b37ca38b64b4cde134650d564ba09b92f4b40a0a17"
+source = "git+https://github.com/stjude-rust-labs/crankshaft?rev=6b17024#6b170247c00993a43d71571e56e10e251d6d89da"
 dependencies = [
  "bon",
  "config",
@@ -628,17 +626,34 @@ dependencies = [
 ]
 
 [[package]]
+name = "crankshaft-docker"
+version = "0.2.0"
+source = "git+https://github.com/stjude-rust-labs/crankshaft?rev=6b17024#6b170247c00993a43d71571e56e10e251d6d89da"
+dependencies = [
+ "anyhow",
+ "bollard",
+ "bon",
+ "futures",
+ "indexmap 2.9.0",
+ "serde",
+ "tar",
+ "thiserror 2.0.12",
+ "tokio",
+ "tokio-stream",
+ "tracing",
+]
+
+[[package]]
 name = "crankshaft-engine"
 version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a18269fdab1246ad7b334bdafd9d06d392436ae698df11d03ceea8512efea076"
+source = "git+https://github.com/stjude-rust-labs/crankshaft?rev=6b17024#6b170247c00993a43d71571e56e10e251d6d89da"
 dependencies = [
+ "anyhow",
  "async-trait",
  "bollard",
  "bon",
  "crankshaft-config",
- "crankshaft-docker",
- "eyre",
+ "crankshaft-docker 0.2.0 (git+https://github.com/stjude-rust-labs/crankshaft?rev=6b17024)",
  "futures",
  "growable-bloom-filter",
  "indexmap 2.9.0",
@@ -918,22 +933,12 @@ checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
 
 [[package]]
 name = "errno"
-version = "0.3.11"
+version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "976dd42dc7e85965fe702eb8164f21f450704bdde31faefd6471dba214cb594e"
+checksum = "cea14ef9355e3beab063703aa9dab15afd25f0667c341310c1e5274bb1d0da18"
 dependencies = [
  "libc",
  "windows-sys 0.59.0",
-]
-
-[[package]]
-name = "eyre"
-version = "0.6.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7cd915d99f24784cdc19fd37ef22b97e3ff0ae756c7e492e9fbfe897d61e2aec"
-dependencies = [
- "indenter",
- "once_cell",
 ]
 
 [[package]]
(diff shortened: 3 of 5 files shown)


src/commands/lock.rs
```diff
@@ -66,10 +76,12 @@ pub async fn lock(args: Args) -> Result<()> {
                     if let Some(runtime) = t.runtime() {
                         if let Some(container) = runtime.container() {
                             if let Ok(image) = container.value() {
-                                if let Ok(text) =
-                                    serde_json::from_str(image.expr().text().to_string().as_str())
-                                {
-                                    images.insert(text);
+                                if let Expr::Literal(LiteralExpr::String(s)) = image.expr() {
+                                    if let Some(text) = s.text() {
+                                        let mut buffer = String::new();
```
a-frantz: performance tip that will never make a tangible impact: you could declare the buffer outside the `for result` loop and clear it after each `insert(buffer.clone())` so that you aren't making a bunch of one-off allocations

src/commands/lock.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// Default name for the lock file.
+const LOCK_FILE: &str = "sprocket.lock";
+
+/// Arguments for the `lock` subcommand.
+#[derive(Parser, Debug)]
+pub struct Args {
+    /// A source WDL file or URL.
+    #[clap(value_name = "PATH or URL")]
+    pub source: Source,
```
a-frantz: Don't love this API; I would change it to one of these two UXs:
1. no arguments, `sprocket lock` recurses the CWD finding all WDL files and locking each image
2. A single argument for `workspace` that _must_ be a directory. `sprocket.lock` gets written to `<workspace>/`

Either way, we need to operate on directories instead of individual files.... well should you be able to "lock" a remote workflow? 🤔 now I'm not sure....

adthrasher: I see a completely different use case for this. I think the flow is:

1. User creates a directory for analysis
2. User locks a specific workflow in that directory
3. User runs N samples through the workflow

Using it for an analysis like this, it needs to support remote workflows.

a-frantz: Ah interesting. Well as written that's not how it works. As written, you are only locking the tasks in the supplied file, as opposed to the tasks _called by the workflow_

What you describe here should probably also fail if there's no workflow in the supplied source WDL

adthrasher: It operates on the set of `AnalysisResult`s, so the dependencies get locked as well. Since it depends on `wdl-analysis`, the behavior will be to lock whatever wdl-analysis returns as a result.

a-frantz: Oop, I was wrong. Although I think you are awfully close to doing what I originally suggested. You are locking _everything_ analyzed, which would include tasks which are imported but not called. I don't think that's right. It should either operate _on a workflow_ (only locking the called tasks) or operate on a workspace (lock everything). This implementation is somewhere between those, and I think that's strange.

adthrasher: > You are locking everything analyzed, which would include tasks which are imported but not called

I suppose for WDL in the wild, that is a real difference. For our WDL task files, every task uses the same image, so the implementations don't have different results.

adthrasher: I added a field to the lock file to record the workflow names and their URIs. If a user provides a directory or no argument, it will search for all WDL files (the `wdl-analysis` default). If the user provides a path to a single file, it will load it and all dependencies.

src/commands/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/// Represents the lock file structure.
+#[derive(Debug, Serialize, Deserialize)]
+struct Lock {
+    /// The time when the lock file was created.
+    #[serde(rename = "generation_time")]
+    timestamp: String,
+    /// A set of workflow names that were found in the source.
+    workflows: HashMap<String, String>,
```
a-frantz: what's the point of this? It seems to me it might just be noise

src/commands/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// A mapping of Docker image names to their sha256 digests.
+    images: HashMap<String, String>,
+}
+
+/// Performs the `lock` command.
+pub async fn lock(args: Args) -> Result<()> {
+    let s = args.source.unwrap_or(Source::Directory(PathBuf::from(
```
a-frantz, line 53:
```suggestion
    // TODO: replace with `Default` once that's upstream
    let s = args.source.unwrap_or(Source::Directory(PathBuf::from(
```

src/commands/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+                });
+        }
+    }
+
+    let time = Utc::now();
+
+    let mut map: HashMap<String, String> = HashMap::new();
+    for image in images {
+        let prefix = image.split(':').next().unwrap_or("");
+        let docker = crankshaft_docker::with_defaults()?;
```
a-frantz: this should be moved to outside the `for` loop

src/commands/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+            .expect("should inspect registry image");
+
+        // Insert the manifest digest into the map.
+        map.insert(
+            image.clone(),
+            prefix.to_owned() + "@" + &i.descriptor.digest.expect("should have a digest"),
+        );
+    }
+
+    if !map.is_empty() {
```
a-frantz: Theoretically someone passes in a WDL `Source` that results in an empty map here. What's the desired UX? I think writing an empty lock is preferable to doing nothing.

src/commands/lock.rs
```diff
@@ -0,0 +1,122 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+use wdl::ast::AstToken;
+use wdl::ast::v1::Expr;
+use wdl::ast::v1::LiteralExpr;
+use wdl::cli::Analysis;
+use wdl::cli::analysis::Source;
+
+/// Default name for the lock file.
+const LOCK_FILE: &str = "sprocket.lock";
```
a-frantz, line 21: I think we should hardcode the filename (`sprocket.lock`), but if the user is supplying a `Source` other than CWD, shouldn't the directory this gets written to be configured via CL arg?

a-frantz: I also wouldn't call it the "Default name" as we offer no way to specify something else

src/commands/lock.rs
adthrasher: First, I think it's good to record which workflows are "locked". Second, I think when we consume the lock file, we could double-check that the lock file applies to the workflow that is being run.

a-frantz: What if the user is using this to run and lock tasks? I think also tracking every task is too noisy, but I don't see enough distinction between workflows and tasks to exclude tasks here 🤔 

> we could double-check that the lock file applies to the workflow that is being run.

is this desirable?

We **_do_** need to handle the case where there's a difference between what was locked and what is being `run`, but I think just tracking the workflow names isn't the right solution. What if the only difference is the workflow name, but all the task calls are identical? I think you are proposing we fail in that case? What if the workflow name stays the same, but all the calls are different?

adthrasher: The value of the map is the file URI. If someone updates it in-place, then that's a problem that this doesn't help with. I suppose we could do something more like store the paths and checksums, then check (at runtime) that the file is in the list and hasn't been altered.

a-frantz: I think this is introducing unneeded complexity 🤔 I think we can just lock every task we find (that is lockable), and then down the line when `run` is invoked with `sprocket.lock`, if every called task is "accounted for" in the lock, we are all good ✅ and if some tasks _aren't_ accounted for in the lock, then we emit a warning or fail or something. The specifics there can be worked out on a future PR, but I'm of the opinion the lock file itself shouldn't care about these details.

adthrasher: Fair enough. I'm curious to hear if anyone else has thoughts.

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   no longer require the `--name` option if passed a WDL document containing a
   single task and no workflow ([#121](github.com/stjude-rust-labs/sprocket/pull/121)).
 
+### Added
+
+* Added `lock` subcommand to store container image checksums ([#120](https://github.com/stjude-rust-labs/sprocket/pull/120))
```
```suggestion
* Added `lock` subcommand to store container manifest checksums ([#120](https://github.com/stjude-rust-labs/sprocket/pull/120))
```
maybe? More technically accurate, but potentially also less clear?

README.md
```diff
@@ -38,6 +38,7 @@
 - **`sprocket format`** formats a document or directory of documents.
 - **`sprocket inputs`** writes template input file (JSON or YAML) for a task or workflow.
 - **`sprocket lint`** performs static analysis on a document or directory of documents with additional linting rules enabled (effectively a shortcut for `check --lint`).
+- **`sprocket lock`** write container image checksums to a lock file.
```
a-frantz:
```suggestion
- **`sprocket lock`** write container manifest checksums to a lock file.
```
not sure this is better

src/commands/lock.rs
```diff
@@ -0,0 +1,142 @@
...
+            workflows.insert(wf.name().to_owned(), result.document().uri().to_string());
+        }
+
+        for task in result.document().tasks() {
+            doc.ast()
+                .as_v1()
+                .expect("should be a v1 document")
+                .tasks()
+                .filter(|t| t.name().text() == task.name())
+                .for_each(|t| {
```
a-frantz: So you need the AST representation to get the container values? I think this should either be rewritten to operate solely on the AST (i.e. loop on `doc.ast().as_v1().expect().tasks()`) or to loop on the the analysis tasks (`result.document().tasks()`) and I think there should be a way to get the AST rep from that? Maybe a call to `inner()`? But as written this does a lot of extra compute per loop

a-frantz (commented):
Should we introduce the documentation for "locking" as part of this PR? I'm not sure if anyone is intending to jump straight into consuming it in `run`, so this might be in a weird undocumented state until that happens.

If there's plans to immediately write the run functionality, I'd be ok waiting to document till that gets implemented. But if that is still just a TODO not immediately being worked on, I'd like us to have documentation stating as much.

src/commands/lock.rs
adthrasher: Yes, the analysis representation doesn't contain the information. I suppose this could be simplified to 

```suggestion
        for task in doc.ast().as_v1().expect("should be a v1 document").tasks() {
```

My original goal was just to parse the tasks used by the workflow, back when this took a single document as input. Now it probably makes more sense to just grab every task in each document found and loop directly over those.

docs/cli/lock.md
```diff
@@ -0,0 +1,8 @@
+# Lock Files
+
+> [!CAUTION]
+> This functionality is considered experimental. There is also currently no consumer for the information that this command writes out. In the future, the `run` command will be modified to utilize this information at execution time.
+
+The Sprocket lock command searches the given input (or the current working directory if no input is provided) for WDL workflows. Each workflow is then loaded along with its dependencies. Within the workflows, each WDL `task` is checked for a `container` key in the `runtime` section. If a single literal value is found, the lock command attempts to get the detailed manifest from the container registry. For each `container` value that is found, the command gets the checksum for that image tag from the manifest. It then writes a the `sprocket.lock` file with a record of each image encountered and the corresponding checksum at the time the lock command was run. It also records a timestamp of the command invocation.
```
a-frantz: This would still work if supplied a directory with all task definitions (and no workflow definitions), so this paragraph should be rephrased. Something like?
```suggestion
The Sprocket lock command searches the given input (or the current working directory if no input is provided) for WDL documents. Each WDL document is then loaded along with any of its imported dependencies. Within the WDL documents, each WDL `task` is checked for a `container` key in the `runtime` section. If a single literal value is found, the lock command attempts to get the detailed manifest from the container registry. For each `container` value that is found, the command gets the checksum for that image tag from the manifest. It then writes a the `sprocket.lock` file with a record of each image encountered and the corresponding checksum at the time the lock command was run. It also records a timestamp of the command invocation.
```

docs/cli/lock.md
```diff
@@ -0,0 +1,8 @@
...
+
+The `sprocket.lock` file that is generated can then be consumed by the `run` subcommand. If the `sprocket.lock` file is available to the `run` command, then the execution engine will use the image map contained in the lock file to replace tagged image values with checksummed image values at runtime. The goal of this functionality is to help users comply with the guidance in the [spec](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#container) to use the most specific URI to refer to containers, typically the digest, while allowing users and developers to continue using the human-readable tag values in their source. This functionality can also help ensure a level of consistency across runs for a cohort by "locking" the container values to immutable checksums, rather than mutable tags that are commonly utilized in WDL source.
```
a-frantz: small nit
```suggestion
The `sprocket.lock` file that is generated can then be consumed by the `run` subcommand. If the `sprocket.lock` file is available to the `run` command, then the execution engine will use the image map contained in the lock file to replace tagged image values with checksummed image values at runtime. The goal of this functionality is to help users comply with the guidance in the [specification](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#container) to use the most specific URI to refer to containers, typically the digest, while allowing users and developers to continue using the human-readable tag values in their source. This functionality can also help ensure a level of consistency across runs for a cohort by "locking" the container values to immutable checksums, rather than mutable tags that are commonly utilized in WDL source.
```
but otherwise this paragraph seems spot on 💪

docs/cli/lock.md
```diff
@@ -0,0 +1,8 @@
+# Lock Files
+
+> [!CAUTION]
+> This functionality is considered experimental. There is also currently no consumer for the information that this command writes out. In the future, the `run` command will be modified to utilize this information at execution time.
+
+The Sprocket lock command searches the given input (or the current working directory if no input is provided) for WDL documents. Each WDL document is then loaded along with any of its imported dependencies. Within the WDL documents, each WDL `task` is checked for a `container` key in the `runtime` section. If a single literal value is found, the lock command attempts to get the detailed manifest from the container registry. For each `container` value that is found, the command gets the checksum for that image tag from the manifest. It then writes a the `sprocket.lock` file with a record of each image encountered and the corresponding checksum at the time the lock command was run. It also records a timestamp of the command invocation.
```
acfoltzer:
```suggestion
The Sprocket lock command searches the given input (or the current working directory if no input is provided) for WDL documents. Each WDL document is then loaded along with any of its imported dependencies. Within the WDL documents, each WDL `task` is checked for a `container` key in the `runtime` section. If a single literal value is found, the lock command attempts to get the detailed manifest from the container registry. For each `container` value that is found, the command gets the checksum for that image tag from the manifest. It then writes a `sprocket.lock` file with a record of each image encountered and the corresponding checksum at the time the lock command was run. It also records a timestamp of the command invocation.
```

docs/cli/lock.md
```diff
@@ -0,0 +1,8 @@
...
+
+The `sprocket.lock` file that is generated can then be consumed by the `run` subcommand. If the `sprocket.lock` file is available to the `run` command, then the execution engine will use the image map contained in the lock file to replace tagged image values with checksummed image values at runtime. The goal of this functionality is to help users comply with the guidance in the [specification](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#container) to use the most specific URI to refer to containers, typically the digest, while allowing users and developers to continue using the human-readable tag values in their source. This functionality can also help ensure a level of consistency across runs for a cohort by "locking" the container values to immutable checksums, rather than mutable tags that are commonly utilized in WDL source.
```
acfoltzer, line 11: Perhaps move the caution note at the top closer to this paragraph, or make it clear that this is future tense and not behavior that is currently implemented?

src/commands.rs
```diff
@@ -41,6 +42,9 @@ pub enum Commands {
     /// Lints a document or a directory containing documents.
     Lint(check::LintArgs),
 
+    /// Locks Docker images to a sha256 digest.
```
acfoltzer: Since there's not yet any enforcement mechanism for the lock files, we should try out best to make sure no users see the docs for a new command and get confused why it's not having an effect.

```suggestion
    /// (Experimental) Locks Docker images to a sha256 digest.
```

src/commands/lock.rs
```diff
@@ -0,0 +1,125 @@
...
+                                text.unescape_to(&mut buffer);
+                                images.insert(buffer.clone());
+                            }
+                        }
+                    }
+                }
+            }
+            buffer.clear();
+        }
+    }
```
acfoltzer:
```suggestion
    for result in results {
        let doc = result.document().root();

        for task in doc.ast().as_v1().expect("should be a v1 document").tasks() {
            if let Some(runtime) = task.runtime() {
                if let Some(container) = runtime.container() {
                    if let Ok(image) = container.value() {
                        if let Expr::Literal(LiteralExpr::String(s)) = image.expr() {
                            if let Some(text) = s.text() {
                                let mut buffer = String::new();
                                text.unescape_to(&mut buffer);
                                images.insert(buffer);
                            }
                        }
                    }
                }
            }
        }
    }
```
This shouldn't result in any extra copying compared to cloning and clearing the same buffer, but it's easier to convince myself that this version is doing the right thing.

acfoltzer (changes requested):
This seems like a good start for experimentation. I've left specific comments on things that should be addressed before merging this first step, and I'm including some thoughts in this message for what we should look out for in future development.

- It would be very useful to have a set of concrete user stories to consider when future development turns from recording data to providing some kind of enforcement or warning mechanisms. In particular, examples that speak to:
  - Negative outcomes possible today but mitigated by the new functionality
  - Limitations of the approach so they can be documented appropriately
- How should non-literal `container:` values be handled at lockfile generation time? At the very least it seems like a warning would be appropriate, but depending on how people want to use this feature a more strict rejection might be useful.
- If many `container:` values in practice contain interpolation that might include inputs or other values which change from run to run, we might want to consider an alternate approach that produces the lockfile as a trace from an invocation of `run` and warns if subsequent runs deviate from that trace.

src/commands.rs
a-frantz: #107 is an awkward spot, but I wonder if we should do some cherrypicking here and introduce the `sprocket dev` namespace in this PR instead of #107?

src/commands/lock.rs
a-frantz: lol 😭 I'll take responsibility for this bad suggestion - https://github.com/stjude-rust-labs/sprocket/pull/120#discussion_r2127245286

mea culpa, Andrew wrote it right the first time

src/commands.rs
adthrasher: @a-frantz - Your usernames have too many letters in common, at least for my eyes on a Monday morning. I read this multiple times and thought it was Adam's comment as well. I was just about to tag you to ask your thoughts.

I think putting this behind the `dev` subcommand is a great idea. That should help make it clear that it isn't really production-ready due to the lack of consumer, but also hopefully will denote that it may well change in the future.

a-frantz: lol have never touched the auto-generated profile picture github assigned my account at creation. Maybe it's time to personalize the account some 🤔

src/commands/lock.rs
acfoltzer: Hah! I skipped reading resolved comments, so I missed that one. While you're right that there is some theoretical benefit in allocation avoidance, the real cost is on the copying which should be the same for both. Personally, I would recommend doing things the straightforward way until profiling or user complaints motivate a closer look.

docs/cli/lock.md
adthrasher: I wanted to ensure that readers had the information at the top, before they started reading the document. I added another block to specifically callout that the second paragraph is all unimplemented functionality at this point.

src/commands/lock.rs
```diff
@@ -0,0 +1,156 @@
...
+                        result.document().path()
+                    );
+                }
+            } else {
+                tracing::warn!(
+                    "Skipping task {} in document {} with no runtime section",
+                    task.name().inner().text(),
+                    result.document().path()
+                );
+            }
```
acfoltzer: Some style goodies here: it's a perfect opportunity to employ the new-ish [`let...else` syntax](https://doc.rust-lang.org/book/ch06-03-if-let.html#staying-on-the-happy-path-with-letelse) to avoid rightward drift! My suggestion also uses the named format parameter syntax with some normal `let` bindings to reduce repetition, and chains one of the unwrapping pairs that share a warning message by using `and_then()`.

```suggestion
            let task_name_token = task.name();
            let task_name = task_name_token.inner().text();
            let doc_path = result.document().path();
            let Some(runtime) = task.runtime() else {
                tracing::warn!(
                    "Skipping task {task_name} in document {doc_path} with no runtime section",
                );
                continue;
            };
            let Some(image) = runtime.container().and_then(|c| c.value().ok()) else {
                tracing::warn!(
                    "Skipping task {task_name} in document {doc_path} with no container image",
                );
                continue;
            };
            let Expr::Literal(LiteralExpr::String(s)) = image.expr() else {
                tracing::warn!(
                    "Skipping image with non-literal value in task {task_name} in document {doc_path}",
                );
                continue;
            };
            let Some(text) = s.text() else {
                tracing::warn!(
                    "Skipping image with placeholder value in task {task_name} in document {doc_path}",
                );
                continue;
            };
            let mut buffer = String::new();
            text.unescape_to(&mut buffer);
            images.insert(buffer);
```

adthrasher: That is so much nicer. Thanks!

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   no longer require the `--name` option if passed a WDL document containing a
   single task and no workflow ([#121](github.com/stjude-rust-labs/sprocket/pull/121)).
 
+### Added
+
+* Added `lock` subcommand to store container manifest checksums ([#120](https://github.com/stjude-rust-labs/sprocket/pull/120))
```
```suggestion
* Added `dev lock` subcommand to store container manifest checksums ([#120](https://github.com/stjude-rust-labs/sprocket/pull/120))
```
and then also an entry for just the `dev` namespace would be good

README.md
```diff
@@ -38,6 +38,7 @@
 - **`sprocket format`** formats a document or directory of documents.
 - **`sprocket inputs`** writes template input file (JSON or YAML) for a task or workflow.
 - **`sprocket lint`** performs static analysis on a document or directory of documents with additional linting rules enabled (effectively a shortcut for `check --lint`).
+- **`sprocket lock`** write container manifest checksums to a lock file.
```
a-frantz: I would replace this with a "`sprocket dev` development and experimental command" or something, and not mention `lock` in the main README at this point

src/commands.rs
```diff
@@ -53,4 +54,15 @@ pub enum Commands {
     /// It will not catch potential runtime errors that may occur when running
     /// the task or workflow.
     Validate(validate::Args),
+
+    /// Development commands.
```
a-frantz:
```suggestion
    /// Developmental and experimental commands.
```

src/commands/lock.rs
```diff
@@ -0,0 +1,144 @@
...
+use wdl::cli::Analysis;
+use wdl::cli::analysis::Source;
+
+/// Name for the lock file.
+const LOCK_FILE: &str = "sprocket.lock";
+
+/// Arguments for the `lock` subcommand.
+#[derive(Parser, Debug)]
+pub struct Args {
+    /// A source WDL file or URL.
```
a-frantz:
```suggestion
    /// A source WDL file, directory, or URL.
```
?

src/commands/lock.rs
```diff
@@ -0,0 +1,144 @@
...
+        let data = toml::to_string_pretty(&lock)?;
+        std::fs::write(output_path, data)?;
+    } else {
+        let lock = Lock {
+            timestamp: time.to_string(),
+            images: HashMap::new(),
+        };
+
+        let data = toml::to_string_pretty(&lock)?;
+        std::fs::write(output_path, data)?;
```
a-frantz: this writes to the file twice. The if-else statement can be removed. We can just always write out the map regardless of emptiness

docs/cli/lock.md
```diff
@@ -0,0 +1,11 @@
+# Lock Files
+
+> [!CAUTION]
+> This functionality is considered experimental. There is also currently no consumer for the information that this command writes out. In the future, the `run` command will be modified to utilize this information at execution time.
+
+The Sprocket lock command searches the given input (or the current working directory if no input is provided) for WDL documents. Each WDL document is then loaded along with any of its imported dependencies. Within the WDL documents, each WDL `task` is checked for a `container` key in the `runtime` section. If a single literal value is found, the lock command attempts to get the detailed manifest from the container registry. For each `container` value that is found, the command gets the checksum for that image tag from the manifest. It then writes a `sprocket.lock` file with a record of each image encountered and the corresponding checksum at the time the lock command was run. It also records a timestamp of the command invocation.
```
a-frantz:
```suggestion
The Sprocket lock command searches the given input (or the current working directory if no input is provided) for WDL documents. Each WDL document is then loaded along with any of its imported dependencies. Within the WDL documents, each WDL `task` is checked for a `container` key in the `runtime` section. If a single literal value is found, the lock command attempts to get the detailed manifest from the container registry. For each `container` value that is found, the command gets the checksum for that manifest. It then writes a `sprocket.lock` file with a record of each manifest encountered and the corresponding checksum at the time the lock command was run. It also records a timestamp of the command invocation.
```
we don't record specific images if I'm not mistaken

a-frantz (approved):

PR #137, opened 2025-07-09
release: bump version

a-frantz:

release PR 🚀 🎉 

For external contributors:

For all contributors:

### DIFF
--- a/src/commands/check.rs
+++ b/src/commands/check.rs
@@ -26,7 +26,7 @@ use crate::get_display_config;
 pub struct Common {
     /// A set of source documents as files, directories, or URLs.
     #[clap(value_name = "PATH or URL")]
-    pub sources: Option<Vec<Source>>,
+    pub sources: Vec<Source>,
 
     /// Excepts (ignores) an analysis or lint rule.
     ///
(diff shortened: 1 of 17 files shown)


adthrasher (approved):

src/commands/check.rs
```diff
@@ -26,7 +26,7 @@ use crate::get_display_config;
 pub struct Common {
     /// A set of source documents as files, directories, or URLs.
     #[clap(value_name = "PATH or URL")]
-    pub sources: Vec<Source>,
+    pub sources: Option<Vec<Source>>,
```
peterhuene: Hmm, I feel like this doesn't need to be `Option` and instead below we can do:


```
// If no sources were specified, use the default source.
if args.common.sources.is_empty() {
    args.common.sources.push(Source::default());
}
```

src/commands/doc.rs
```diff
@@ -56,6 +57,18 @@ const DEFAULT_OUTPUT_DIR: &str = "docs";
...
+    } else if args.workspace.is_none() {
+        if let Source::Directory(d) = Source::default() {
+            d
+        } else {
+            unreachable!()
+        }
+    } else {
+        bail!("`workspace` must be a local directory for the `doc` command")
+    };
+
```
peterhuene:
```suggestion
    let workspace = if let Some(Source::Directory(workspace)) = args.workspace.unwrap_or_default() {
        workspace
    } else {
        bail!("`workspace` must be a local directory for the `doc` command")
    };
```

src/commands/check.rs
a-frantz: Going to check how both of these render in the `clap` `--help` and decide which looks prettier 👍

Issue #143, opened 2025-07-11
bug: `run` uses `output/` directory as the default


a-frantz:
this contradicts the documentation. Either the docs should be updated (or preferably) `run` should be updated to have better out-dir logic

PR #147, opened 2025-07-17
feat: specify input prefix once

### LINKED ISSUE #141: feature request: infer input prefix
this should work
```bash
$ sprocket run --name validate_string_is_12bit_int data_structures/flag_filter.wdl number="4095"
error: invalid input key `number`: expected the value to be prefixed with the workflow or task name
```
instead I have to do
```bash
$ sprocket run --name validate_string_is_12bit_int data_structures/flag_filter.wdl validate_string_is_12bit_int.number="4095"
```

a-frantz:

Closes #141 

For external contributors:

For all contributors:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     * `--entrypoint` is now required if no inputs are provided.
     * `--entrypoint` will be prefixed to the key of any key-value pairs supplied on the command line.
 
-### Fixed
-
-* Output directory logic for `run` now properly reflects the documentation ([#147](https://github.com/stjude-rust-labs/sprocket/pull/147)).
-
 ## 0.14.1 - 07-10-2025
 
 ### Fixed


Cargo.toml
```diff
@@ -44,6 +44,10 @@ walkdir = "2.5.0"
 wdl = { version = "0.15.1", features = ["cli", "codespan", "engine", "lsp", "doc"] }
 
 
+[patch.crates-io]
+wdl = { git = "https://github.com/stjude-rust-labs/wdl.git", branch = "main" }
+
+
```
acfoltzer: I just took the current head of the `wdl` repo for the revision here, so it might be worth double checking this is what is expected. `patch` sections are best used for temporary overrides in local development, and a symbolic branch name that changes over time is not good for CI stability.

```suggestion
wdl = { version = "0.15.1", features = ["cli", "codespan", "engine", "lsp", "doc"], git = "https://github.com/stjude-rust-labs/wdl.git", rev = "f7ba4c0d16a6b16c7b3bf7d302658799a0eed77f" }
```

The better answer for this is bringing in `wdl` as a submodule or combining the repos, which I am going to write up into a full proposal soon.

src/commands/run.rs
```diff
@@ -56,14 +56,18 @@ pub struct Args {
...
-    /// If no inputs are provided and this argument is not provided, it will be
-    /// assumed you're trying to run the workflow present in the specified WDL
-    /// document.
+    /// If `entrypoint` is not specified, all inputs (from both files and
+    /// key-value pairs) are expected to be prefixed with the name of the
+    /// workflow or task being run.
+    ///
+    /// If `entrypoint` is specified, it will be prefixed (with a `.` delimiter)
+    /// to all key-value pair inputs on the command line. Keys specified within
+    /// files are unchanged by this argument.
```
acfoltzer: The entrypoint "being prefixed" means to me that we're putting something on the front of the entrypoint; hopefully this is more clear.

```suggestion
    /// If `entrypoint` is specified, it will be appended (with a `.` delimiter)
    /// to the front of all key-value pair inputs on the command line. Keys specified within
    /// files are unchanged by this argument.
```

src/commands/run.rs
```diff
@@ -337,36 +312,36 @@ pub async fn run(args: Args) -> Result<()> {
...
+        }
+
+        std::fs::remove_dir_all(&output_dir).with_context(|| {
+            format!(
+                "failed to remove output directory `{dir}`",
+                dir = output_dir.display()
+            )
+        })?;
+    }
+
```
acfoltzer: I think it'll be confusing to have this fix for #143 land in the same commit as the behavior change for #141. Would you mind splitting out the #143 fix into a distinct PR so they won't get squashed together?

src/commands/validate.rs
```diff
@@ -23,14 +23,18 @@ pub struct Args {
...
-    /// If no inputs are provided and this argument is not provided, it will be
-    /// assumed you're trying to validate the workflow present in the specified
-    /// document.
+    /// If `entrypoint` is not specified, all inputs (from both files and
+    /// key-value pairs) are expected to be prefixed with the name of the
+    /// workflow or task being run.
+    ///
+    /// If `entrypoint` is specified, it will be prefixed (with a `.` delimiter)
+    /// to all key-value pair inputs on the command line. Keys specified within
+    /// files are unchanged by this argument.
```
acfoltzer: Whatever wording is desirable for `run`'s `entrypoint` documentation should be mirrored here.

acfoltzer (changes requested):

src/commands/run.rs
adthrasher:
```suggestion
    /// If `entrypoint` is specified, it will be appended with a `.` delimiter and 
    /// then prepended to all key-value pair inputs on the command line. Keys 
    /// specified within files are unchanged by this argument.
```
Dueling suggestions. I think Adam's suggestion reads oddly with the use of `appended` and then `to the front of`. Though the dictionary (at least the one I consulted) doesn't specifically state that `append` is "to the end", that is the common usage.

src/commands/run.rs
```diff
@@ -337,36 +312,36 @@ pub async fn run(args: Args) -> Result<()> {
...
-                    EngineInputs::Task(Default::default()),
-                    origins,
-                )
-            } else {
-                bail!(
-                    "document `{path}` contains no workflow or task",
-                    path = document.path()
-                );
-            }
+            bail!("the `--entrypoint` option is required if no inputs are provided")
```
adthrasher, line 346: In practice, I'm not sure this is bad, but it does deviate from other engines. I would prefer `--entrypoint` to be optional and only used for task execution.

Quoting the spec below. It isn't explicit, but my reading of this is that the workflow should be the default execution.
> If a workflow appears in the primary WDL file, it is called the "top-level" workflow, and any workflows it calls via imports are "subworkflows". Typically, it is an error for the primary WDL file to not contain a workflow; however, an execution engine may choose to support executing individual tasks.

src/commands/run.rs
```diff
@@ -274,37 +278,7 @@ pub async fn run(args: Args) -> Result<()> {
...
-        std::fs::remove_dir_all(&output_dir).with_context(|| {
-            format!(
-                "failed to remove output directory `{dir}`",
-                dir = output_dir.display()
-            )
-        })?;
-    }
-
-    let inferred = Inputs::coalesce(&args.inputs)
+    let inputs = Inputs::coalesce(&args.inputs, args.entrypoint.clone())
```
adthrasher, line 311: Is there a companion PR in the wdl crates to alter the implementation of `coalesce`?

adthrasher: Nevermind, I missed #535 was already merged.

src/commands/run.rs
acfoltzer: I agree; in Slack discussions, I recommended that a bare `sprocket run` should run the top-level workflow, and that `sprocket run --task foo` should be required to invoke a task. The workflow case benefits from both the explicit call-out in the spec, and the lack of ambiguity since there cannot be more than one top-level workflow. @a-frantz mentioned though that she's hoping to remove some of the distinctions between invoking workflows and tasks, so this choice might be in line with that goal.

a-frantz: I do not feel strongly, but as implemented in this PR, the only time this case we're discussing comes up would be for running a workflow without any inputs, which I don't think is happening in practice very often (if it _ever_ happens).

I would prefer not to not special case the top-level workflow, as it just adds extra logic to think through in the implementation and documentation.

src/commands/run.rs
a-frantz: This has been reverted. The change is trivial, but would have merge conflicts if put up right now due to the change from `name` to `entrypoint`, so I'll hold off till this merges and then put up a patch PR

src/commands/run.rs
a-frantz: >  but it does deviate from other engines.

to this point specifically, unlike other engines, we are requiring the entrypoint be provided in our inputs explicitly without any kind of inference applied in any invocation. I think special casing the empty input call to a workflow is odd and inconsistent with that decision.

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Changed
+
+* `--name` option renamed to `--entrypoint` for `validate` and `run` ([#147](https://github.com/stjude-rust-labs/sprocket/pull/147)).
+    * `--entrypoint` is now required if no inputs are provided.
+    * `--entrypoint` will be prefixed to the key of any key-value pairs supplied on the command line.
+
+### Fixed
+
+* Output directory logic for `run` now properly reflects the documentation ([#147](https://github.com/stjude-rust-labs/sprocket/pull/147)).
+
```
acfoltzer: Due to revert
```suggestion
```

a-frantz: Good catch!

src/commands/run.rs
acfoltzer: It's on my todo list to write up our long thread on this more comprehensively in an issue or design doc. For now, without greater changes to how we handle entrypoint inference, I think this is fine. But the reason empty input invocations feel like such a weird special case is that we lean too heavily on the names of inputs to determine what we run.

a-frantz: > the reason empty input invocations feel like such a weird special case is that we lean too heavily on the names of inputs to determine what we run.

yes I would agree with this.

@adthrasher do you have an object to hitting merge on this as is?

adthrasher (approved):

src/commands/run.rs
adthrasher: No issue here. Just added my approval.

PR #149, opened 2025-07-29
feat: new and improved output dir logic

a-frantz:

In short, the default run directories are now located at `./runs/<entrypoint>/<timestamp>/`.
Also we have a `latest` symlink now 😎 

For external contributors:

For all contributors:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* Added new default output directory logic ([#149](https://github.com/stjude-rust-labs/sprocket/pull/149)).
+
 ### Changed
 
 * `--name` option renamed to `--entrypoint` for `validate` and `run` ([#147](https://github.com/stjude-rust-labs/sprocket/pull/147)).
@@ -16,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Output directory logic for the `run` command now always uses the workflow or task name by default ([#148](https://github.com/stjude-rust-labs/sprocket/pull/148)).
+    * this logic replaced in [#149](https://github.com/stjude-rust-labs/sprocket/pull/149).
 
 ## 0.14.1 - 07-10-2025
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     * `--entrypoint` is now required if no inputs are provided.
     * `--entrypoint` will be prefixed to the key of any key-value pairs supplied on the command line.
 
-### Fixed
-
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -16,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Output directory logic for the `run` command now always uses the workflow or task name by default ([#148](https://github.com/stjude-rust-labs/sprocket/pull/148)).
+    * this logic was replaced in [#149](https://github.com/stjude-rust-labs/sprocket/pull/149).
```
acfoltzer: Should delete the previous entry; we aren't obligated to keep anything in this changelog until it's part of a release.

src/commands/run.rs
```diff
@@ -187,6 +206,46 @@ fn progress(kind: ProgressKind<'_>, pb: &tracing::Span, state: &Mutex<State>) {
     pb.pb_set_message(&message);
 }
 
+/// Determines the timestamped execution directory and performs any necessary
+/// staging prior to execution.
+///
+/// Notably, this function does not actually create the exection directory at
```
acfoltzer:
```suggestion
/// Notably, this function does not actually create the execution directory at
```

src/commands/run.rs
```diff
@@ -187,6 +206,46 @@ fn progress(kind: ProgressKind<'_>, pb: &tracing::Span, state: &Mutex<State>) {
...
+    let mut output = root.join(timestamp.format("%F_%H%M%S%f").to_string());
+
+    while output.exists() {
+        tracing::warn!(
+            "`{dir}` was selected for execution but it already exists",
+            dir = output.display()
+        );
+        let timestamp = chrono::Utc::now();
+        output = root.join(timestamp.format("%F_%H%M%S%f").to_string());
+    }
```
acfoltzer: The cases where I can imagine a timestamp with nanosecond precision colliding are so unusual that I'm inclined to just error out instead if this happens. Either way, this shouldn't be tight loop with no sleep, backoff, or retry limit.

src/config.rs
```diff
@@ -90,12 +91,25 @@ pub struct CheckConfig {
...
 /// Represents the configuration for the Sprocket `run` command.
-#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case", deny_unknown_fields)]
 pub struct RunConfig {
     /// The engine configuration.
     #[serde(flatten)]
     pub engine: engine::config::Config,
+
+    /// The "runs" directory under which new `run` invocation's execution
```
acfoltzer:
```suggestion
    /// The "runs" directory under which new `run` invocations' execution
```

src/commands/run.rs
a-frantz: Yeeeaaa, I wasn't quite sure how to handle this case 😅 meant to leave an explicit comment asking for thoughts from the reviewer but it slipped my mind.

Just pushed a commit that bails in this case instead.

dead8309:
> In short, the default run directories are now located at `./runs/<entrypoint>/<timestamp>/`.

thanks 🙏🏻. no more overriding

PR #157, opened 2025-08-06
docs(guided-tour): tweaks to reflect new API

a-frantz:

I thought I had edited the docs as part of the PRs implementing these changes, but it seems I missed the guided tour page.

For external contributors:

For all contributors:

### DIFF
--- a/docs/guided-tour.md
+++ b/docs/guided-tour.md
@@ -161,9 +161,9 @@ Inputs to a Sprocket run are provided as arguments passed after the WDL document
 name is provided. Each input can be specified as either
 
 * a key value pair (e.g., `main.is_pirate=true`)
-* a JSON file containing inputs (e.g., a `defaults.json` file where the
+* a JSON file containing inputs (e.g., a `hello_defaults.json` file where the
   contents are `{ "main.is_pirate": true }`)
-* a YAML file containing inputs (e.g. a `defaults.yaml` file where the
+* a YAML file containing inputs (e.g. a `hello_defaults.yaml` file where the
   contents are `main.is_pirate: true`)
 
 Inputs are _incrementally_ applied, meaning that inputs specified later override
@@ -172,7 +172,7 @@ use a set of default parameters and iterate through sample names in Bash rather
 than create many individual input files.
 
 ```bash
-sprocket run example.wdl defaults.json main.name="Ari"
+sprocket run example.wdl hello_defaults.json main.name="Ari"
 ```
 
 Note that the above command does not specify an entrypoint with the `--entrypoint`
@@ -205,7 +205,7 @@ After a few seconds, this job runs successfully with the following outputs.
 Congrats on your first successful `sprocket run` 🎉!
 
 If you wanted to override some of the defaults for the workflow, you could do
-so by defining the input in a `overrides.json` file:
+so by defining the input in a `hello_overrides.json` file:
 
 ```json
 {
@@ -220,7 +220,7 @@ so by defining the input in a `overrides.json` file:
 and the providing that in the set of inputs to the workflow.
 
 ```shell
-sprocket run example.wdl overrides.json main.name="Sprocket"
+sprocket run example.wdl hello_overrides.json main.name="Sprocket"
 ```
 
 This produces the following output.


docs/guided-tour.md
```diff
@@ -115,14 +115,26 @@ issues](./guided-tour/problems.png){style="margin-top: 30px;"}
...
 
+This will error right away, as we haven't told Sprocket which task or workflow
+to run.
+
+```txt
+error: the `--entrypoint` option is required if no inputs are provided
+```
+
+We want to run the "main" workflow defined in `example.wdl`, so we can try again
+but specify the entrypoint to use this time using the `--entrypoint` flag.
```
adthrasher, line 140: It's a bit late now, but the 0.15.0 changelog should probably have had a stronger emphasis like `BREAKING` to note this change in behavior. Admittedly it's a small subset of workflows that do not provide inputs, but it's non-zero.

adthrasher (dismissed):

docs/guided-tour.md
```diff
@@ -150,22 +162,29 @@ use a set of default parameters and iterate through sample names in Bash rather
 than create many individual JSON input files.
 
 ```bash
-sprocket run workflow.wdl defaults.json main.sample_name="Foo"
+sprocket run workflow.wdl defaults.json workflow.sample_name="Foo"
```
claymcleod: Why is this now `workflow`? Isn't the target date still `main`?

a-frantz: If you scroll up a few lines, the example JSON uses `workflow` instead of `main`. `sample_name` is also not a parameter for the `example.wdl` workflow.

But this is a bit confusing, I think it could be reworked a bit

a-frantz: @claymcleod I've reworked `example.wdl` a little bit and fleshed out the examples some to be a touch more in depth

docs/public/guided-tour/example.wdl
a-frantz, line 1: this change was made so that the guided tour can override both `greetings` and `is_pirate` and see the impact of both in the same run (prior version effectively ignored `is_pirate` if `greetings` wasn't using the default value)

docs/guided-tour.md
```diff
@@ -75,16 +83,17 @@ note[MetaSections]: workflow `main` is missing both `meta` and `parameter_meta`
...
-35 │         String color = "green"
+30 │         String color = "green"
    │                ^^^^^
 ```
 
 Specific lint rules can be ignored with multiple invocations of the `-e` flag.
+Rule matching is case insensitive from command line arguments.
 
 ```shell
-sprocket lint example.wdl -e ContainerUri -e MetaSections
```
claymcleod: This doesn't seem like an improvement to me. Though the CLI can accept case insensitive rules, the rules themselves are still capitalized in the output messages. So I feel it makes sense to keep them as they appear in the messages.

docs/guided-tour.md
```diff
@@ -140,32 +161,41 @@ must be provided before the workflow can run.
...
+
+Note that the above command does not specify an entrypoint with the `--entrypoint`
+flag. This is because every input is using fully qualified dot notation; each
+input is prefixed with the name of the entrypoint and a period, `main.`.
+This fully qualified dot notation is required for inputs provided within a file.
+The dot notation can get repetitive if supplying many key value pairs on the command line,
+so specifying `--entrypoint` allows you to omit the repeated part of the keys.
 :::
 
 Here, we can specify the `name` parameter as a key-value pair on the command
```
claymcleod, line 186: Maybe we should explain that, because the entrypoint is provided, the `main.` can be omitted.

claymcleod (approved):
After these are addressed, looks good.

docs/guided-tour.md
a-frantz: That is explained in the preceding "Understand Inputs" text box. I don't think it needs to be repeated here

docs/guided-tour.md
```diff
@@ -194,22 +204,23 @@ After a few seconds, this job runs successfully with the following outputs.
...
-  ]
+  ],
+  "main.is_pirate": true
 }
 ```
 and the providing that in the set of inputs to the workflow.
 
 ```shell
-sprocket run example.wdl greetings.json main.name="Sprocket"
+sprocket run example.wdl overrides.json main.name="Sprocket"
```
adthrasher: I see why you used `overrides.json` and `defaults.json`, but I wonder if there is potential to cause confusion if readers think those are some sort of magic names for the files.

a-frantz: I hadn't thought of that, but I do see how a particularly literal reading of this example could lead to that confusion.

Maybe rename these to `hello_overrides.json` and `hello_defaults.json`?

PR #164, opened 2025-08-25
feat: update to use Crankshaft events channel.

peterhuene:
This PR updates the progress bar for `sprocket run` to be based off of Crankshaft events, rather than the previous `ProgressKind` from `wdl-engine` that no longer exists.

For external contributors:

For all contributors:


src/commands/run.rs
```diff
@@ -426,7 +469,11 @@ pub async fn run(args: Args) -> Result<()> {
         .unwrap(),
     );
 
-    let state = Mutex::<State>::default();
+    let state = Arc::new(Mutex::<State>::default());
+    let token = CancellationToken::new();
+    let (events_tx, events_rx) = broadcast::channel(100);
```
acfoltzer: Rationale for this value?

Cargo.toml
```diff
@@ -19,12 +18,13 @@ clap-verbosity-flag = { version = "3.0.3", features = ["tracing"] }
 clap_complete = "4.5.55"
 codespan-reporting = "0.12.0"
 colored = "3.0.0"
-crankshaft-docker = "0.2.0"
+crankshaft = { version = "0.4.0", git = "https://github.com/stjude-rust-labs/crankshaft", rev = "b99c2fd" }
+crankshaft-docker = { version = "0.2.0", git = "https://github.com/stjude-rust-labs/crankshaft", rev = "b99c2fd" }
```
claymcleod: I'm not sure why this isn't the case, but I would expect this to be gated behind the `docker` feature in `crankshaft`. Is there a reason I'm forgetting why we didn't do it that way?

peterhuene: I don't know if we anticipated the crate being directly useful for Crankshaft users or not; I believe we use it to ask Docker to resolve tags for the `lock` command.

We might as well put it behind a feature in the main crate and remove this reference.

src/commands/run.rs
peterhuene: None particularly; we need enough of a buffer so that the events don't get dropped (unlike the other channel implementations, the sender never blocks, it'll just drop messages and notify the receivers they're lagging). At the very least, we can comment to that effect here.

Cargo.toml
acfoltzer: If functionality can be selected based on importing one crate vs another, that is a better state of affairs than depending on Cargo features. Typically one only resorts to the latter when the optional parts cannot be shipped in a separate crate due to trait impls, access to private fields, etc.

src/commands/run.rs
peterhuene: I added a constant with some comments around it.

Cargo.toml
peterhuene: I've put up a PR to re-export the docker crate, so we'll be able to remove the extra dependency shortly.

src/commands/run.rs
```diff
@@ -40,6 +46,14 @@ use crate::emit_diagnostics;
 /// very short analyses.
 const PROGRESS_BAR_DELAY_BEFORE_RENDER: Duration = Duration::from_secs(2);
 
+/// The capacity for the Crankshaft events channel.
+///
+/// This is the number of events to buffer in the channel before receivers
+/// become lagged.
```
a-frantz: What does "become lagged" mean?

peterhuene: The events channels are implemented using `tokio::sync::broadcast` (which is required for Crankshaft since we might send events to multiple receivers: a crankshaft monitor server or sprocket, for example).

The implementation of `broadcast` channels doesn't block on sending events into the channel; it will buffer up to some limit waiting for all receivers to have read the message.

When that buffer gets full (for example, the writer is writing the events faster than all receivers can receive them), it'll start dropping older messages in the buffer and receivers that haven't read a dropped message are notified via an error that they're "lagging" behind.

In the unlikely event this happens given our channel capacity size, we'll warn the user that the progress indicator displayed by `sprocket` might not be correct.

peterhuene: I'll update the comment with some context (plus it's not Crankshaft-specific since `cloud-copy` also uses a broadcast channel for its events).

a-frantz: great, that makes sense! Thanks for the explainer!

src/commands/run.rs
```diff
@@ -452,18 +599,25 @@ pub async fn run(args: Args) -> Result<()> {
             error!("execution was interrupted: waiting for evaluation to abort");
             token.cancel();
             evaluate.await.ok();
+            transfer_progress.await.ok();
+            crankshaft_progress.await.ok();
```
acfoltzer, line 603: This is the typical convention for indicating that we don't care about a `Result` or `Option`
```suggestion
            let _ = evaluate.await;
            let _ = transfer_progress.await;
            let _ = crankshaft_progress.await;
```

src/commands/run.rs
```diff
@@ -452,18 +599,25 @@ pub async fn run(args: Args) -> Result<()> {
...
-            Ok(outputs) => {
-                println!("{outputs}", outputs = serde_json::to_string_pretty(&outputs.with_name(&entrypoint))?);
-                Ok(())
-            }
-            Err(EvaluationError::Source(e)) => {
-                emit_diagnostics(&e.document.path(), e.document.root().text().to_string(), &[e.diagnostic], &e.backtrace, args.report_mode.unwrap_or_default(), args.no_color)?;
-                bail!("aborting due to evaluation error");
+        res = &mut evaluate => {
+            transfer_progress.await.ok();
+            crankshaft_progress.await.ok();
```
acfoltzer, line 608:
```suggestion
            let _ = transfer_progress.await;
            let _ = crankshaft_progress.await;
```

src/commands/run.rs
```diff
@@ -452,18 +599,25 @@ pub async fn run(args: Args) -> Result<()> {
...
+
+            match res {
+                Ok(outputs) => {
+                    println!("{outputs}", outputs = serde_json::to_string_pretty(&outputs.with_name(&entrypoint))?);
```
acfoltzer, line 612: When there's only a single format specifier, I don't think the named argument form is very helpful. Especially when, as in this case, the name is already in scope for something else.
```suggestion
                    println!("{}", serde_json::to_string_pretty(&outputs.with_name(&entrypoint))?);
```

src/commands/run.rs
```diff
@@ -452,18 +599,25 @@ pub async fn run(args: Args) -> Result<()> {
...
+                    Ok(())
+                }
+                Err(EvaluationError::Source(e)) => {
+                    emit_diagnostics(&e.document.path(), e.document.root().text().to_string(), &[e.diagnostic], &e.backtrace, args.report_mode.unwrap_or_default(), args.no_color)?;
```
acfoltzer, line 616: `rustfmt` gave up on this because it's in the `select!` macro body, but this is what it would've done:
```suggestion
                    emit_diagnostics(
                        &e.document.path(),
                        e.document.root().text().to_string(),
                        &[e.diagnostic],
                        &e.backtrace,
                        args.report_mode.unwrap_or_default(),
                        args.no_color
                    )?;
```

src/commands/run.rs
```diff
@@ -161,57 +253,102 @@ impl std::fmt::Display for Ids<'_> {
...
-    pb.pb_set_message(&message);
+                pb.pb_set_message(&message);
+            }
+            Err(RecvError::Closed) => break,
+            Err(RecvError::Lagged(_)) => {
+                if !warned {
+                    warned = true;
+                    warn!("event stream is lagging: task progress reporting may be incorrect");
+                }
+            }
```
acfoltzer, line 349: Since there's no recovery from this condition, maybe this should cause the progress bar message to change for the remainder of the execution rather than emitting a single warning? I don't think it'll matter for a typical user who should have a relatively quiet console, but someone mucking around with higher log verbosity settings could pretty easily miss the warning

src/commands/run.rs
```diff
@@ -161,57 +253,102 @@ impl std::fmt::Display for Ids<'_> {
...
-            ids = Ids(&state.ids)
+            s3 = if executing == 1 { "" } else { "s" },
+            sep = if executing == 0 { "" } else { ": " },
+            tasks = Tasks(&state.executing)
         )
-    };
+    }
+
+    let mut warned = false;
+    message(&state.lock().expect("failed to lock state"));
```
acfoltzer, line 294: I do appreciate explicit messages with `expect()` vs `unwrap()` in most cases, but when it comes to `Mutex` poisoning I don't think it adds much: this is only going to be triggered when another thread is already bringing down the program via its own panic that is more specific to the failure.

src/commands/run.rs
peterhuene: That's a great idea, I'll implement that.

src/commands/run.rs
peterhuene: In fact, this doesn't need to be protected with a `Mutex` at all; `state` isn't being accessed from outside of the tokio task any longer.

PR #166, opened 2025-08-28
fix: deny-notes implies deny-warnings

a-frantz:

For external contributors:

For all contributors:

### DIFF
--- a/src/commands/check.rs
+++ b/src/commands/check.rs
@@ -45,6 +45,8 @@ pub struct Common {
     pub deny_warnings: bool,
 
     /// Causes the command to fail if notes were reported.
+    ///
+    /// Implies `--deny-warnings`.
     #[clap(long)]
     pub deny_notes: bool,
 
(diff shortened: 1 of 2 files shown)


adthrasher (approved):

src/commands/check.rs
```diff
@@ -100,8 +102,9 @@ impl CheckArgs {
             .into_iter()
             .chain(config.check.except.clone())
             .collect();
-        self.common.deny_warnings = self.common.deny_warnings || config.check.deny_warnings;
         self.common.deny_notes = self.common.deny_notes || config.check.deny_notes;
+        self.common.deny_warnings =
```
claymcleod, line 106: Should this not also include `self.check.deny_notes`?

tests/cli.rs
```diff
@@ -133,7 +133,8 @@ async fn run_sprocket(test_path: &Path, working_test_directory: &Path) -> Result
     let args_string = fs::read_to_string(&args_path)
         .await
         .with_context(|| format!("failed to read command at path {:?}", &args_path))?;
-    let args = shlex::split(&args_string).ok_or_else(|| anyhow!("failed to split command args"))?;
+    let args = shlex::split(&format!("--skip-config-search {args_string}"))
```
claymcleod, line 136: Why is this needed again?

a-frantz: I have a `sprocket.toml` in my home dir (https://sprocket.bio/cli/config.html#xdg-config-home) that was causing the tests to fail, as it has non-default config. Adding `--skip-config-search` prevents that from happening in the tests.

src/commands/check.rs
a-frantz: `self.common.deny_notes` is set on line 105 (including `config.check.deny_notes`) and then that value is used in line 107, so I don't think that check is needed

acfoltzer (approved):
Looks good to me!

> deduplicates some logic, although I believe at a miniscule perf hit

Maybe I'm missing something, but I don't see anything that's changed in a hot path? Perhaps that's what you mean by miniscule, but it's not worth your time to worry about an extra function call that happens once per execution, especially not when it buys you consolidation of logic.

PR #169, opened 2025-09-04
[WIP]feat: enable/disable lint rules via tag

a-frantz:

depends on https://github.com/stjude-rust-labs/wdl/pull/592

For external contributors:

For all contributors:


src/commands/check.rs
```diff
@@ -1,26 +1,41 @@
...
 
+/// The [`Tag`]s which will run with the default `lint` configuration.
+const DEFAULT_TAG_SET: TagSet = TagSet::new(&[
+    Tag::Completeness,
+    Tag::Naming,
+    Tag::Clarity,
+    Tag::Portability,
+    Tag::Correctness,
+    Tag::Deprecated,
+]);
```
a-frantz: This selection omits the tags: `Spacing`, `Style`, `Sorting`, and `Documentation`.

src/commands/check.rs
```diff
@@ -37,9 +52,47 @@ pub struct Common {
...
 
+    /// Enable all lint rules. This includes additional rules outside the
+    /// default set.
+    ///
+    /// `--except <RULE>` can be used in conjuction with this argument. Using
+    /// this argument will ignore certain configuration file options.
+    #[clap(short, long, conflicts_with_all = ["include_lint_tag", "exclude_lint_tag"])]
+    pub all_lint_rules: bool,
+
+    /// Excludes a lint tag from running. This implies all other lint tags
```
adthrasher: This says `all`. I'm not sure how to make it clearer, but I think this means that it will run all other rules that are otherwise enabled (which may be the default set or any other set).

src/commands/check.rs
```diff
@@ -37,9 +52,47 @@ pub struct Common {
...
+    /// should be run.
+    ///
+    /// Repeat the flag mutliple times to exclude multiple tags. `--except
```
adthrasher:
```suggestion
    /// Repeat the flag multiple times to exclude multiple tags. `--except
```

src/commands/check.rs
```diff
@@ -37,9 +52,47 @@ pub struct Common {
...
+    #[clap(long, value_name = "TAG",
+        value_parser = PossibleValuesParser::new(ALL_TAG_NAMES.iter()),
+        ignore_case = true,
+        action = clap::ArgAction::Append,
+        num_args = 1,
+        conflicts_with_all = ["include_lint_tag"],
+    )]
+    pub exclude_lint_tag: Vec<String>,
+
+    /// Includes a lint tag for running. This implies no other lint tags should
```
adthrasher: So this isn't an incremental addition? If this option is specified only the rule/rules specified will run. That is, it overrides the default set.

src/commands/check.rs
```diff
@@ -37,9 +52,47 @@ pub struct Common {
...
+    /// be run.
+    ///
+    /// Repeat the flag mutliple times to include multiple tags. `--except
```
adthrasher:
```suggestion
    /// Repeat the flag multiple times to include multiple tags. `--except
```

src/commands/check.rs
```diff
@@ -37,9 +52,47 @@ pub struct Common {
...
+    /// <RULE>` can be used in conjuction with this argument. Using this
```
adthrasher:
```suggestion
    /// <RULE>` can be used in conjunction with this argument. Using this
```

src/commands/check.rs
```diff
@@ -37,9 +52,47 @@ pub struct Common {
...
+    /// `--except <RULE>` can be used in conjuction with this argument. Using
+    /// this argument will ignore certain configuration file options.
+    #[clap(short, long, conflicts_with_all = ["include_lint_tag", "exclude_lint_tag"])]
+    pub all_lint_rules: bool,
+
+    /// Excludes a lint tag from running. This implies all other lint tags
+    /// should be run.
+    ///
+    /// Repeat the flag mutliple times to exclude multiple tags. `--except
+    /// <RULE>` can be used in conjuction with this argument. Using this
```
adthrasher:
```suggestion
    /// <RULE>` can be used in conjunction with this argument. Using this
```

src/commands/check.rs
adthrasher: Personally, I think `Documentation` should be in the default set. I think good documentation should be expected (similar to Rust) and I want users to explicitly opt-out if they don't want documentation.

src/config.rs
```diff
@@ -90,6 +90,14 @@ pub struct CheckConfig {
     pub deny_notes: bool,
     /// Hide diagnostics with `note` severity.
     pub hide_notes: bool,
+    /// Enable all lint rules, even those outside the default set.
+    pub all_lint_rules: bool,
+    /// Set of lint tags to opt into. Leave this empty to use the default set of
```
adthrasher, line 97: Based on the above code, I think this comment needs to be updated to reflect that if this is specified, only the specified lints are run.

src/commands/check.rs
a-frantz: That makes sense to me, but does conflict with my plans [here](https://github.com/stjude-rust-labs/wdl/issues/510#issuecomment-3247281613). Specifically, I intended to add:
> * `ExpectedMetaString`
>   * this will warn if any of reserved meta keys (`description`, `help`, `external_help`, `warning`, `group` (for inputs), and `category` (for workflows)) have a non `String` value
> * `DescriptionLength`
>   * any `description` meta entry should be at most 140 characters

which would have the `Documentation` tag, and should _not_ run by default. I think we can add back `Documentation` to the default set, and come up with a new tag for these two rules?

adthrasher: I think `DescriptionLength` could be `Style`.

adthrasher: `ExpectedMetaString` is less clear to me. I think `Style` would also be fine as it is a rule to ensure compliance with our "style", but that's a bit tenuous.

src/commands/check.rs
a-frantz: Maybe it should be renamed to `--only-lint-tag`? Because there needs to be a way for users to run a single tag, and making this incremental makes that impossible (without _another_ flag being added)

src/commands/check.rs
a-frantz: > but I think this means that it will run all other rules that are otherwise enabled (which may be the default set or any other set).

The logic is admittedly not very clear, but this understanding is incorrect (but maybe that is how it should work). You can see further down in the file, that the branch for this case looks like:
```rust
        } else if !args.common.exclude_lint_tag.is_empty() {
            let mut tags = vec![];
            for s in args.common.exclude_lint_tag.iter() {
                let t = Tag::from_str(s).ok().unwrap();
                tags.push(t);
            }
            (TagSet::new(Tag::VARIANTS), TagSet::new(&tags))
```
so setting this enables _everything_. But maybe that's not a useful API?

src/commands/check.rs
a-frantz: Maybe this is bad because it blurs the lines between the crates, but I think it might be most clear to have `Tag::DocCompatibility`?

src/commands/check.rs
a-frantz: Updated `include` to `only` and I think that is clearer

src/commands/check.rs
a-frantz: I'm working on changing this to act as a filter on the other selections (via `only_lint_tag` and `all_lint_rules`). Should have a change pushed shortly

a-frantz:
@adthrasher thanks for the bikeshedding! I think this new approach is clearer. LMK what you think

peterhuene (dismissed):

peterhuene (approved):

PR #179, opened 2025-09-17
release: bump version to `0.17.1`.

peterhuene:

For external contributors:

For all contributors:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sprocket"
-version = "0.17.0"
+version = "0.17.1"
 description = "A command line tool for working with Workflow Description Language (WDL) documents"
 authors = ["The Sprocket project contributors"]
 license = "MIT OR Apache-2.0"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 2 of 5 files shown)


Cargo.toml
a-frantz, line 1: this doesn't update the `wdl` crate 😅

peterhuene: It doesn't need to, it's in the lock file.

peterhuene: Technically we didn't need to bump `wdl` or `wdl-cli` for this either, as the lock file could have upgraded just the `wdl-engine` dependency given it was a patch release.

a-frantz: o! I didn't realize cargo/rust was that smart! I assumed it was a bit dumber than it is apparently 😂

a-frantz (approved):

Cargo.toml
peterhuene: Yeah, `cargo` will automatically put the latest patch release in `Cargo.lock` (if there isn't one) and that's what you build with until updated via `cargo update`. That is unless you pin to specific versions of the dependency.

`cargo install` will always update to the latest patch release at install time too, unless you specify `--locked` to use exactly what's in the lock file.

Given that Rust crate authors aren't always respectful of breaking semver changes in even patch releases, I've generally stuck to using `--locked` in install instructions so that users don't run into build issues simply because someone broke an interface in a patch release.

peterhuene:
Seems related: https://github.com/abiosoft/colima/issues/83

PR #180, opened 2025-09-18
chore: absorb the `wdl` repo into this one

acfoltzer:
WIP: see https://github.com/stjude-rust-labs/wdl/issues/557

For all contributors:


.github/workflows/CI.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         if: runner.os == 'Linux'
 
       - name: Run tests
-        run: cargo --locked test --workspace --all-features
+        run: cargo --locked test --workspace --all-features --no-fail-fast
```
a-frantz, line 121: This makes it so that the failure isn't always the last thing in the GitHub UI (which auto-scrolls to the tail), which makes finding the failing test harder.

acfoltzer: This is a necessary tradeoff, unfortunately. A full report of the test failures is needed, particularly when we have so many different test targets that build upon one another, to avoid going down false paths when debugging failures. The more specific a test is, the more likely that its failure will provide information that is directly actionable. So if we just stop the tests once the end-to-end `sprocket` integration suite fails, we'll miss out on failures that will frequently be closer to the underlying cause in the `wdl` crates, and will be more likely to go down false paths in debugging.

Also, in situations where there are multiple unrelated failures awaiting in a branch, it is a frustrating cycle to only learn about the next failure after you've put together a patch for the first test that happened to fail.

There is a mitigation for the downside you point out: since these are all combined in one `cargo test` command, the bottom of the output will always list all the test targets that contained a failure. Those can be cross-referenced to know whether and where to scroll up. 

This is another area where switching to `cargo nextext` could help, though we need to work on how some of our custom test harnesses work in order to make that competitive in runtime with the basic runners.

a-frantz (dismissed):

acfoltzer:
I tried to change the base branch to something else and then back, per the suggestion on a Github support thread, but now it's back to needing an approving review and therefore won't show whether that fixed the rebasability of the branch 😞

a-frantz (approved):

PR #182, opened 2025-09-19
wip: the start of a specialized LSF+Apptainer backend

acfoltzer:
Closing https://github.com/stjude-rust-labs/wdl/pull/600 in favor of this collaboration draft after rebasing the branch contents onto the new combined repo.

For all contributors:

### DIFF
--- a/crates/wdl-engine/src/backend/lsf_apptainer.rs
+++ b/crates/wdl-engine/src/backend/lsf_apptainer.rs
@@ -55,7 +55,8 @@ const LSF_JOB_NAME_MAX_LENGTH: usize = 4094;
 
 #[derive(Debug)]
 struct LsfApptainerTaskRequest {
-    config: Arc<Config>,
+    engine_config: Arc<Config>,
+    backend_config: Arc<LsfApptainerBackendConfig>,
     name: String,
     spawn_request: TaskSpawnRequest,
     /// The requested container for the task.
@@ -238,9 +239,16 @@ impl TaskManagerRequest for LsfApptainerTaskRequest {
         let lsf_stdout_path = attempt_dir.join("lsf.stdout");
         let lsf_stderr_path = attempt_dir.join("lsf.stderr");
 
-        // TODO ACF 2025-09-11: configurable LSF queue, including handling for the
-        // `short_task` hint
-        let mut bsub_child = Command::new("bsub")
+        let mut bsub_command = Command::new("bsub");
+
+        // If an LSF queue has been configured, specify it. Otherwise, the job will end
+        // up on the cluster's default queue.
+        if let Some(queue) = &self.backend_config.queue {
+            bsub_command.arg("-q");
+            bsub_command.arg(queue);
+        }
+
+        bsub_command
             // Pipe stdout and stderr so we can trace them. This should just be the LSF output like
             // `<<Waiting for dispatch ...>>`.
             //
@@ -279,8 +287,9 @@ impl TaskManagerRequest for LsfApptainerTaskRequest {
                 "rusage[mem={memory_kb}KB/job]",
                 memory_kb = self.memory / 1024
             ))
-            .arg(apptainer_command_path)
-            .spawn()?;
+            .arg(apptainer_command_path);
+
+        let mut bsub_child = bsub_command.spawn()?;
 
         // Take the stdio pipes from the child process and consume them for tracing
         // purposes.
@@ -345,14 +354,16 @@ impl TaskManagerRequest for LsfApptainerTaskRequest {
 
 #[derive(Debug)]
 pub struct LsfApptainerBackend {
-    config: Arc<Config>,
+    engine_config: Arc<Config>,
+    backend_config: Arc<LsfApptainerBackendConfig>,
     manager: TaskManager<LsfApptainerTaskRequest>,
 }
 
 impl LsfApptainerBackend {
-    pub fn new(config: Arc<Config>) -> Self {
+    pub fn new(engine_config: Arc<Config>, backend_config: Arc<LsfApptainerBackendConfig>) -> Self {
         Self {
-            config,
+            engine_config,
+            backend_config,
             // TODO ACF 2025-09-11: the `MAX` values here mean that in addition to not limiting the
             // overall number of CPU and memory used, we don't limit per-task consumption. There is
             // potentially a path to pulling queue limits from LSF for these, but for now we just
@@ -375,7 +386,8 @@ impl TaskExecutionBackend for LsfApptainerBackend {
     ) -> anyhow::Result<super::TaskExecutionConstraints> {
         Ok(super::TaskExecutionConstraints {
             container: Some(
-                v1::container(requirements, self.config.task.container.as_deref()).into_owned(),
+                v1::container(requirements, self.engine_config.task.container.as_deref())
+                    .into_owned(),
             ),
             // TODO ACF 2025-09-11: populate more meaningful values for these based on the given LSF
             // queue. Unfortunately, it's not straightforward to ask "what's the most CPUs I can ask
@@ -409,7 +421,7 @@ impl TaskExecutionBackend for LsfApptainerBackend {
         let hints = request.hints();
 
         let container =
-            v1::container(requirements, self.config.task.container.as_deref()).into_owned();
+            v1::container(requirements, self.engine_config.task.container.as_deref()).into_owned();
         let cpu = v1::cpu(requirements);
         let memory = v1::memory(requirements)? as u64;
         // TODO ACF 2025-09-11: I don't _think_ LSF offers a hard/soft CPU limit
@@ -423,9 +435,9 @@ impl TaskExecutionBackend for LsfApptainerBackend {
         let name = request.id()[0..LSF_JOB_NAME_MAX_LENGTH].to_string();
         self.manager.send(
             LsfApptainerTaskRequest {
-                config: self.config.clone(),
+                engine_config: self.engine_config.clone(),
+                backend_config: self.backend_config.clone(),
                 spawn_request: request,
-                // backend: self.inner.clone(),
                 name,
                 container,
                 cpu,
@@ -453,5 +465,20 @@ impl TaskExecutionBackend for LsfApptainerBackend {
         // uid/gids on files shouldn't be as much of an issue, and using only
         // `apptainer exec` means no longer-running containers to tear down
         None
+    }
+}
+
+#[derive(Debug, Clone, serde::Deserialize, serde::Serialize)]
+pub struct LsfApptainerBackendConfig {
+    // TODO ACF 2025-09-12: add queue option for short tasks
+    queue: Option<String>,
+}
+
+impl LsfApptainerBackendConfig {
+    pub fn validate(&self) -> Result<(), anyhow::Error> {
+        // TODO ACF 2025-09-12: what meaningful work to be done here? Maybe ensure the
+        // queue exists, interrogate the queue for limits and match them up
+        // against prospective future config options here?
+        Ok(())
     }
 }
--- a/crates/wdl-engine/src/backend/lsf_apptainer.rs
+++ b/crates/wdl-engine/src/backend/lsf_apptainer.rs
@@ -1,0 +1,332 @@
+use std::fmt::Write as _;
+use std::fs::Permissions;
+use std::os::unix::fs::PermissionsExt as _;
+use std::process::Stdio;
+use std::sync::Arc;
+
+use anyhow::Context as _;
+use anyhow::anyhow;
+use tempfile::TempDir;
+use tokio::fs::File;
+use tokio::fs::{self};
+use tokio::io::AsyncBufReadExt;
+use tokio::io::BufReader;
+use tokio::process::Command;
+use tokio_util::sync::CancellationToken;
+use tracing::trace;
+
+use super::COMMAND_FILE_NAME;
+use super::TaskManagerRequest;
+use super::TaskSpawnRequest;
+use super::WORK_DIR_NAME;
+use crate::PrimitiveValue;
+use crate::STDERR_FILE_NAME;
+use crate::STDOUT_FILE_NAME;
+use crate::TaskExecutionResult;
+use crate::config::Config;
+use crate::path::EvaluationPath;
+
+/// The name of the file where the Apptainer command invocation will be written.
+const APPTAINER_COMMAND_FILE_NAME: &str = "apptainer_command";
+
+/// The root guest path for inputs.
+const GUEST_INPUTS_DIR: &str = "/mnt/task/inputs/";
+
+/// The guest working directory.
+const GUEST_WORK_DIR: &str = "/mnt/task/work";
+
+/// The guest path for the command file.
+const GUEST_COMMAND_PATH: &str = "/mnt/task/command";
+
+/// The path to the container's stdout.
+const GUEST_STDOUT_PATH: &str = "/mnt/task/stdout";
+
+/// The path to the container's stderr.
+const GUEST_STDERR_PATH: &str = "/mnt/task/stderr";
+
+#[derive(Debug)]
+struct LsfApptainerTaskRequest {
+    config: Arc<Config>,
+    name: String,
+    spawn_request: TaskSpawnRequest,
+    /// The requested container for the task.
+    container: String,
+    /// The requested CPU reservation for the task.
+    cpu: f64,
+    /// The requested memory reservation for the task, in bytes.
+    memory: u64,
+    cancellation_token: CancellationToken,
+}
+
+impl TaskManagerRequest for LsfApptainerTaskRequest {
+    fn cpu(&self) -> f64 {
+        self.cpu
+    }
+
+    fn memory(&self) -> u64 {
+        self.memory
+    }
+
+    async fn run(self) -> anyhow::Result<super::TaskExecutionResult> {
+        let attempt_dir = self.spawn_request.attempt_dir();
+
+        // Create the host directory that will be mapped to the WDL working directory.
+        let wdl_work_dir = attempt_dir.join(WORK_DIR_NAME);
+        fs::create_dir_all(&wdl_work_dir).await.with_context(|| {
+            format!(
+                "failed to create WDL working directory `{path}`",
+                path = wdl_work_dir.display()
+            )
+        })?;
+
+        // Write the evaluated WDL command section to a host file.
+        let wdl_command_path = self.spawn_request.attempt_dir().join(COMMAND_FILE_NAME);
+        fs::write(&wdl_command_path, self.spawn_request.command())
+            .await
+            .with_context(|| {
+                format!(
+                    "failed to write WDL command contents to `{path}`",
+                    path = wdl_command_path.display()
+                )
+            })?;
+        fs::set_permissions(&wdl_command_path, Permissions::from_mode(0o777)).await?;
+
+        // Create an empty file for the WDL command's stdout.
+        let wdl_stdout_path = self.spawn_request.attempt_dir().join(STDOUT_FILE_NAME);
+        let _ = File::create(&wdl_stdout_path).await.with_context(|| {
+            format!(
+                "failed to create WDL stdout file `{path}`",
+                path = wdl_stdout_path.display()
+            )
+        })?;
+
+        // Create an empty file for the WDL command's stderr.
+        let wdl_stderr_path = self.spawn_request.attempt_dir().join(STDERR_FILE_NAME);
+        let _ = File::create(&wdl_stderr_path).await.with_context(|| {
+            format!(
+                "failed to create WDL stderr file `{path}`",
+                path = wdl_stderr_path.display()
+            )
+        })?;
+
+        // Create a temp dir for the container's execution within the attempt dir
+        // hierarchy. On many HPC systems, `/tmp` is mapped to a relatively
+        // small, local scratch disk that can fill up easily. Mapping the
+        // container's `/tmp` and `/var/tmp` paths to the filesystem we're using
+        // for other inputs and outputs prevents this from being a capacity problem,
+        // though potentially at the expense of execution speed if the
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
a-frantz, line 1: `wdl-engine` needs a changelog entry as well

crates/wdl-engine/src/backend/lsf_apptainer.rs
adthrasher, line 1: I think we've discussed this elsewhere, and I don't think it should hold up this PR, but ultimately, I'd like to see the backend executor split from the containerization system. That is likely a much larger effort in sprocket/crankshaft.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,648 @@
...
+                token: self.cancellation_token.clone(),
+            },
+        );
+
+        // Take the stdio pipes from the child process and consume them for event
+        // reporting and tracing purposes.
+        //
+        // TODO ACF 2025-09-23: drop the `-K` from `bsub` and poll status instead? Could
+        // be less intensive from a resource perspective vs having a process and
+        // two loops on the head node per task, but we should wait to observe
```
a-frantz, line 376: maybe a tangent or aside; but I don't think the `sprocket` binary should really be run on head nodes. I guess for typical WDL it would be fine, but `sprocket` is capable of doing heavy work on weird WDL, e.g. a user doing `read_lines(<some 300GB file with a billion lines>)` will end up being performed by whatever node `sprocket` was invoked from.

Generally speaking, I don't do much of _anything_ on the head nodes aside from launching an interactive session for myself.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,648 @@
...
+        crankshaft_events: Option<broadcast::Sender<Event>>,
+    ) -> Self {
+        Self {
+            engine_config,
+            backend_config,
+            // TODO ACF 2025-09-11: the `MAX` values here mean that in addition to not limiting the
+            // overall number of CPU and memory used, we don't limit per-task consumption. There is
+            // potentially a path to pulling queue limits from LSF for these, but for now we just
+            // throw jobs at the cluster.
+            manager: TaskManager::new_unlimited(u64::MAX, u64::MAX),
```
a-frantz, line 486: doesn't need to be this PR, but I think it's probably fine to just expose these as config values that must be set by the user instead of trying to do something like parsing `bqueues`

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
```diff
@@ -0,0 +1,234 @@
...
+//!
+//! We don't currently have a notion of a top-level directory for an entire
+//! workflow execution; the `root` path for each workflow and task evaluator is
+//! specific to _that_ workflow or task, but the point of keeping our own cache
+//! of Apptainer images is to avoid pushing our luck with spotty
+//! container registries by inducing repeated requests for the same image.
+//!
+//! For expedience, this implementation makes the simplifying assumption that we
+//! have one top-level workflow execution per process, and keeps the images
+//! directory in a global variable. This should be replaced with something more
```
a-frantz, line 39: how does this work with sub-workflows?

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
```diff
@@ -0,0 +1,234 @@
...
+async fn try_pull(sif_path: &Path, container: &str) -> Result<(), RetryError<anyhow::Error>> {
+    info!(container, "pulling image");
+    let mut apptainer_pull_child = Command::new("apptainer")
+        // Pipe the stdio handles, both for tracing and to inspect for telltale signs of permanent
+        // errors
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .arg("pull")
+        .arg(sif_path)
+        .arg(format!("docker://{container}"))
```
a-frantz, line 169: is there a way for the container to already have the `docker://` prefix?

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,648 @@
...
+
+/// The maximum length of an LSF job name.
+///
+/// See <https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=o-j>.
+const LSF_JOB_NAME_MAX_LENGTH: usize = 4094;
+
+#[derive(Debug)]
+struct LsfApptainerTaskRequest {
+    backend_config: Arc<LsfApptainerBackendConfig>,
+    name: String,
```
adthrasher, line 77: Does this get any type of prefix? Personally, I'd want something like `sprocket_*` so that I can easily filter these out if I'm doing other work.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,648 @@
...
+            },
+        );
+
+        // Take the stdio pipes from the child process and consume them for event
+        // reporting and tracing purposes.
+        //
+        // TODO ACF 2025-09-23: drop the `-K` from `bsub` and poll status instead? Could
+        // be less intensive from a resource perspective vs having a process and
+        // two loops on the head node per task, but we should wait to observe
+        // real-world performance before complicating things.
```
adthrasher, line 377: I was trying to decide if doing something like `BSUB_QUIET=JOBID` and storing the ID for the task and polling would be better. I have a slight preference for that over parsing LSF stdout.

adthrasher: Another option would be using the LSF C API (https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=SSWRJV_10.1.0/api_reference/index.htm).

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,648 @@
...
+                v1::container(requirements, self.engine_config.task.container.as_deref())
+                    .into_owned(),
+            ),
+            // TODO ACF 2025-09-11: populate more meaningful values for these based on the given LSF
+            // queue. Unfortunately, it's not straightforward to ask "what's the most CPUs I can ask
+            // for and still hope to be scheduled?". A reasonable stopgap would be to make this a
+            // config parameter, but the experience would be unfortunate when having to manually
+            // update that if changing queues, or if handling multiple queues for short jobs.
+            cpu: f64::MAX,
+            memory: i64::MAX,
```
adthrasher, line 513: Clay has previously described a vision where configuration can be done at the cluster level, by a group like HPCF. In our particular case, we have a [page](https://wiki.stjude.org/display/HPCF/HPCF+Overview) which describes our available hardware. So it would be a relatively easy set and forget for a user in our environment. I also don't know that users are going to change queues often (though I may be ignorant of how other institutions have setup their cluster queues).

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,648 @@
...
+    pub apptainer_images_dir: PathBuf,
+}
+
+fn default_max_scatter_concurrency() -> u64 {
+    200
+}
+
+fn default_apptainer_images_dir() -> PathBuf {
+    PathBuf::from("~/.cache/sprocket-apptainer-images")
+}
```
adthrasher, line 631: Should this do something similar to [config](https://github.com/stjude-rust-labs/sprocket/blob/ddd42c3fff176a95f08f45c89889a35443ce8ce1/src/config.rs#L145-L150) and use `XDG_CACHE_HOME` rather than always the user's home directory?

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
```diff
@@ -0,0 +1,234 @@
...
+//! tasks are invoked with the same WDL `container` requirement.
+//!
+//! By explicitly converting to `.sif` and using those files in `apptainer exec`
+//! invocations instead of `docker://` specifiers, we avoid this additional
+//! container registry traffic. To sidestep the issue of staleness between the
+//! locally-converted `.sif` images and the contents of the container
+//! registry, the `.sif` files are only used for a single workflow execution. We
+//! still benefit from the Apptainer cache when building new `.sif` files, so
+//! this is not much of a slowdown, but it does increase disk space consumption
+//! depending on where the images directory is created.
```
adthrasher, line 26: I actually like this behavior. I wouldn't want a tagged image to get updated within a workflow, which otherwise could happen if a push happened between task executions.

crates/wdl-engine/src/backend/lsf_apptainer.rs
adthrasher: That's a good point. I _think_ LSF sets environment variables when a task is running as an LSF job. Maybe `sprocket` should check those if run in LSF+apptainer mode and warn the user if running outside of an LSF job.

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
adthrasher: Good catch. If a URI is provided with the `protocol`, I think this will duplicate.

Relevant spec:

> The format of a container URI is protocol://location, where protocol is one of the protocols supported by the execution engine. Execution engines must, at a minimum, support the docker protocol. If only location is specified, the protocol is assumed to be docker. An execution engine should ignore any URI with a protocol it does not support.

adthrasher: We should probably also support the `oras` protocol for pulling SIFs directly. In addition, if `container` is an array, this could prefer anything with the `oras` protocol.

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
adthrasher: Adam can clarify, but my understanding is that the top-level workflow gets a cache location. Any subworkflows would also get stored there, rather than per-workflow.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,648 @@
...
+    pub fn validate(&self, engine_config: &Config) -> Result<(), anyhow::Error> {
+        if cfg!(not(unix)) {
+            bail!("LSF + Apptainer backend is not supported on non-unix platforms");
+        }
+        if !engine_config.experimental_features_enabled {
+            bail!("LSF + Apptainer backend requires enabling experimental features");
+        }
+        // TODO ACF 2025-09-12: what meaningful work to be done here? Maybe ensure the
+        // queue exists, interrogate the queue for limits and match them up
+        // against prospective future config options here?
```
adthrasher, line 654: Another option is to look for `Job not submitted` in the bsub stdout during the above submission logic. The queue name, specifically, could be validated here, but it might make sense to have a more general check that the job submitted successfully. 

```
$ bsub -K -q my_queue bash
HPCF: WARNING! No Memory was requested!
      A default memory request of 2.50 GB has been placed for this job
      The job will be killed if   2.50 GB of memory is used
my_queue: No such queue. Job not submitted.
```

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: Sorry, I mean head node in the generic sense of "the place where I have an interactive shell", not on our specific cluster's extremely underpowered login nodes.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: This is what I'm leaning toward as well.

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
acfoltzer: In this initial version, there is just a single one per invocation of the executable. It's lazily populated and accessed via a global variable. In practice, that means it's 1-1 with the top-level workflow or task executed by `sprocket run`.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: Looks like `bqueues -l <queue_name>` will do the trick too without actually submitting a job. Whether or not that's desirable is another question...

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: Ah, I pushed a fix that manually tacks `.cache` on, but it's nice to respect the proper env var. I'll push another shortly.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: Oh, it's good to know about that `BSUB_QUIET` option. What a weird place to offer that configuration; I looked all over the actual CLI flags for something that would allow customizing the output and had resigned myself to using regexes.

As much as I love writing C bindings, it doesn't seem like a public SDK exists, so using the API directly would substantially complicate our development, build, and release processes. I'd like to avoid that if at all possible.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: This sounds good, do you mind if we track it in a follow-on rather than the initial release? There's enough subtlety there with the max job name length that I would rather not iterate on it here

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
acfoltzer: Do you have any example `oras://` containers I could try running with? `docker://` already works fine, e.g.,
```
  requirements {
    container: "docker://ubuntu:latest"
  }
```

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: Yep, this is definitely a follow-up I want to pursue. I think there's a fairly clean interface seam here; there's essentially nothing Apptainer-specific about the `bsub` invocation. Doing this well will require some refactoring of the various backend interfaces which is probably overdue anyway, given that many of them were designed for a pre-Crankshaft world.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: I could imagine extending the config with a map from queues to resource constraints. This part could be provided by the cluster admins, and then it'd be up to the user to fill in the specific queue to use for a job.

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
acfoltzer: Oh yes, that part is absolutely intended and desirable behavior. What's not ideal is that we don't have a way for the caches to be reused between top-level invocations, and so the current behavior ends up leaving behind a bunch of detritus on disk.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: I am going to punt this to later, since doing it properly implies doing blocking operations in `validate()`. We'll either have to make it `async` to bubble out the blocking explicitly, or denote the behavior and carefully wrap async calling contexts to avoid scheduler starvation.

crates/wdl-engine/src/backend/lsf_apptainer.rs
a-frantz: > That's a good point. I _think_ LSF sets environment variables when a task is running as an LSF job. Maybe `sprocket` should check those if run in LSF+apptainer mode and warn the user if running outside of an LSF job.

Seems like overkill to me, as it would be perfectly reasonable to run `sprocket` interactively so long as we aren't on a head node [in our cluster]. I think for any "serious" work, users will want to `bsub sprocket run`, but for tasks or test workflows running it interactively would be a common use case.

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
a-frantz: I've mentioned this before in passing I believe, but maybe not recently. I can imagine us having a `sprocket clean` command that purges any locations we might cache things. Not needed now of course, but something we may want in the future

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
adthrasher: I don't. I think GHCR supports it, but I'm not aware of anyone using it. There are a number of URI [formats](https://apptainer.org/docs/user/latest/cli/apptainer_pull.html) that apptainer supports. I doubt `shub://` needs to be supported as Singularity Hub is [deprecated](https://singularityhub.github.io/singularityhub-docs/). This should also be fine to make an issue and revisit later.

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
```diff
@@ -0,0 +1,234 @@
...
+                .max_delay_millis(60_000)
+                .take(10),
+            || try_pull(&sif_path, &container),
+            |e, _| {
+                warn!(e = %e, "`apptainer pull` failed");
+            },
+        );
+
+        tokio::select! {
+            _ = cancellation_token.cancelled() => return Err(anyhow!("task execution cancelled")),
```
claymcleod, line 138: Should this be "image pull cancelled"?

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
```diff
@@ -0,0 +1,234 @@
...
+        .map_err(|e| RetryError::permanent(e.into()))?;
+
+    let is_permanent = Arc::new(Mutex::new(false));
+
+    let child_stdout = apptainer_pull_child
+        .stdout
+        .take()
+        .ok_or_else(|| RetryError::permanent(anyhow!("apptainer pull child stdout missing")))?;
+    let stdout_container = container.to_owned();
+    let _stdout_is_permanent = is_permanent.clone();
```
claymcleod, line 181: Is this worth deleting?

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
acfoltzer: No, the pull operation is failing because the overall task execution has been cancelled.

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
acfoltzer: Even though this isn't currently used, having it in place makes the logic around stdout and stderr more symmetric, and when we inevitably end up having to look at stdout to make this determination, this will be a guide to the author of that change.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,656 @@
...
+        fs::write(&wdl_command_path, self.spawn_request.command())
+            .await
+            .with_context(|| {
+                format!(
+                    "failed to write WDL command contents to `{path}`",
+                    path = wdl_command_path.display()
+                )
+            })?;
+        #[cfg(unix)]
+        fs::set_permissions(&wdl_command_path, Permissions::from_mode(0o777)).await?;
```
claymcleod: I wonder if we'd want to do like `0o700` or, alternatively, make this configurable from the start?

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -0,0 +1,656 @@
...
+        fs::write(&apptainer_command_path, apptainer_command)
+            .await
+            .with_context(|| {
+                format!(
+                    "failed to write Apptainer command file `{}`",
+                    apptainer_command_path.display()
+                )
+            })?;
+        #[cfg(unix)]
+        fs::set_permissions(&apptainer_command_path, Permissions::from_mode(0o777)).await?;
```
claymcleod: Same

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: It looks like the Docker backend is using 0o770, so I switched to that and also used qualified names to clean up a bit of the `cfg(unix)` business.

claymcleod (approved):

crates/wdl-engine/src/backend/lsf_apptainer/images.rs
adthrasher: Filed an issue: https://github.com/stjude-rust-labs/sprocket/issues/358

adthrasher: Does this still need to be updated for the case where the provided URI contains the protocol already? I don't know if we're stripping it off upstream.

acfoltzer: Nope, the example above where `docker://` is present in the WDL works fine. There is some processing of the value performed by the evaluator before it gets sent to the backend. We will probably need to change how this works in order to preserve protocol information to the backend when we undertake implementing more possibilities here.

crates/wdl-engine/src/backend/lsf_apptainer.rs
adthrasher: https://github.com/stjude-rust-labs/sprocket/issues/359

crates/wdl-engine/src/backend/lsf_apptainer.rs
adthrasher: I wouldn't error, but a reminder warning (or info?) message wouldn't be overly noisy.

crates/wdl-engine/src/backend/lsf_apptainer.rs
adthrasher: Fine with me to track for later

PR #349, opened 2025-09-22
revise: provides a better error for array and map type resolving

claymcleod:
This provides a slightly better error for users who get a type mismatch in array and map types.

For external contributors:

For all contributors:


crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -405,7 +405,10 @@ pub fn no_common_type(
         "type mismatch: a type common to both type `{expected}` and type `{actual}` does not exist"
     ))
     .with_label(format!("this is type `{actual}`"), actual_span)
-    .with_label(format!("this is type `{expected}`"), expected_span)
+    .with_label(
+        format!("all elements up until and including this one had a common type `{expected}`"),
```
a-frantz: this seems very wordy to me. A more compact option:
```suggestion
        format!("detected common type `{expected}` from this and prior elements"),
```
?

claymcleod: I was playing around with the wording and didn't find anything I _loved_, so I am open to a better wording. That said, I don't like exactly how you've put it here either. How about this.
```suggestion
        format!("this and all prior elements had a common type `{expected}`"),
```

claymcleod: I went ahead and updated it because I liked it better than any of the other options. I think this is probably the best of all worlds—decently short and still clear.

a-frantz: I'm ok with that! But I wonder if the span should be changed to highlight all the prior elements? That seems clearer to me than just pointing to the last one

crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -405,7 +405,10 @@ pub fn no_common_type(
         "type mismatch: a type common to both type `{expected}` and type `{actual}` does not exist"
     ))
     .with_label(format!("this is type `{actual}`"), actual_span)
-    .with_label(format!("this is type `{expected}`"), expected_span)
+    .with_label(
+        format!("this and all prior elements had a common type `{expected}`"),
```
peterhuene, line 412: The `no_common_type` function is called from `if_expr` as well, which makes this confusing since it has nothing to do with elements.

We can potentially split it into two different diagnostic functions: one where there is always a left and right side (`if` expr) and one that explicitly calls out that the common type was derived from all previous elements.

claymcleod: Resolved in the latest commit.

crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -394,12 +394,34 @@ pub fn call_input_type_mismatch<T: TreeToken>(
...
+        expected_span,
+    )
+}
+
+/// Creates a "no common type" diagnostic for an if statement.
+///
+/// This is called if the return elements of an if statement do not match.
+pub fn no_common_type_if(
```
peterhuene: I think we can probably remove this new function and just use the `type_mismatch` function from `wdl-analysis` directly, which is what we're using now in `wdl-analysis`'s type calculation for `if expr`.

peterhuene (approved):

a-frantz (approved):

Issue #224, opened 2025-07-04
feat(wdl-lint): doc compatibility mode


a-frantz:
**Is your feature request related to a problem? Please describe.**
stjude-rust-labs/wdl#339 introduces a variety of documentation conventions, some of which are already enforced via existing lint rules (e.g. the `description` meta key requirement), but some (e.g. the `description` keys max length) are hard to discover without reading documentation.

**Describe the solution you'd like**
We should introduce more lint rules that enforce conventions that are used by `wdl-doc`. These rules would likely be onerous and unnecessary to follow if the user is not using the Sprocket documentation functions, so they should be opt-in only instead of on by default. They could be enabled with a `--doc-compatibility` flag to `sprocket lint`.

PR #351, opened 2025-09-22
feat: DescriptionLength rule

a-frantz:

Partially addresses #224 by adding a new lint rule.

For external contributors:

For all contributors:


crates/wdl-lint/RULES.md
a-frantz, line 1: I noticed some of these tags were incorrect, so I opted to remove that whole column. IMO it's not needed and just add maintenance burden

crates/wdl-lint/src/rules/description_length.rs
```diff
@@ -0,0 +1,203 @@
...
+    fn description(&self) -> &'static str {
+        "Ensures that `description` meta entries are not too long for display in documentation."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "Descriptions should be kept short so that they can always render in full. If a \
+         `description` is too long, it may be clipped in some contexts during documentation. \
+         `help` meta entries are permitted to be of any length, and may be a better place for long \
+         form text."
+    }
```
a-frantz, line 55: Should these explicitly mention `sprocket [dev] doc`?

crates/wdl-lint/src/tags.rs
```diff
@@ -33,8 +33,11 @@ pub enum Tag {
     /// Rules associated with the use of deprecated language constructs.
     Deprecated,
 
-    /// Rules associeted with documentation.
+    /// Rules associated with documentation.
     Documentation,
+
+    /// Rules associated with compatibility with documentation rendering.
+    DocRendering,
```
a-frantz: I do not love this new tag, but I do not think that this rule should be part of the default set of rules, which includes the `Documentation` tag. Open to hearing alternatives here

crates/wdl-lint/RULES.md
a-frantz: I just deleted the whole file now, as it's redundant with this view on docs.rs: https://docs.rs/wdl/latest/wdl/lint/rules/index.html

And the docs.rs view won't be out of date like RULES.md might be

crates/wdl-lint/src/rules/description_length.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    fn explanation(&self) -> &'static str {
+        "Descriptions should be kept short so that they can always render in full. If a \
+         `description` is too long, it will be clipped in some documentation contexts. `help` meta \
+         entries are never clipped and may be a better place for long form text."
+    }
+
+    fn tags(&self) -> TagSet {
+        TagSet::new(&[Tag::SprocketDoc])
+    }
```
adthrasher, line 59: This is very specific. I wonder if a broader TagSet, like `Tag::Sprocket`, would be more future-proof?

adthrasher (dismissed):

crates/wdl-lint/src/rules/description_length.rs
a-frantz: Yea that might be a better option 🤔 Although whatever is decided here will also be used in #353 and #354 , so there would be at least 3 rules with this tag

a-frantz: I went with `SprocketCompatibility`. What do you think?

crates/wdl-ast/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+* Documentation change to specify what happens when `text()` is called on an empty `LiteralString` ([#351](https://github.com/stjude-rust-labs/sprocket/pull/351)).
```
claymcleod, line 12: I don't really think this requires a changelog entry to be honest. If you still feel it does, at least adding it as like a `#### Chore` or something.

crates/wdl-ast/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
```
a-frantz:
```suggestion
#### Chore
```

crates/wdl-lint/src/rules/description_length.rs
```diff
@@ -0,0 +1,208 @@
...
+        if let Some(description_item) = section
+            .items()
+            .find(|entry| entry.name().inner().to_string() == "description")
+            && let MetadataValue::String(description) = description_item.value()
+            && !description.is_empty()
+        {
+            let mut text = String::new();
+            description
+                .text()
+                .expect("meta strings cannot be interpolated")
```
claymcleod:
```suggestion
                /// SAFETY: `meta` entries aren't interpolated, and we
                /// just checked to ensure it's not empty above.
                .unwrap()
```

Probably need to mention both.

crates/wdl-lint/src/rules/meta_description.rs
```diff
@@ -58,14 +58,17 @@ impl Rule for MetaDescriptionRule {
     }
 
     fn explanation(&self) -> &'static str {
-        "Each task or workflow should have a description in the meta section. This description \
-         should be an explanation of the task or workflow. The description should be written in \
-         active voice and complete sentences. More detailed information can be included in the \
-         `help` key."
+        "Each task, workflow, and struct should have a description in the meta section. The \
+         description should be short, written in active voice, and be in complete sentences. More \
```
claymcleod, line 62:
```suggestion
         description should be short, written in active voice and be in complete sentences. More \
```

crates/wdl-lint/RULES.md
```diff
@@ -1,53 +0,0 @@
-# Rules
```
claymcleod, line 1: I don't view this as a good change, as it's the only place where there is a plaintext description of each of the rules that are available, what their tags are, and what a brief description is. I understand that it sometimes gets out of sync, but another solution is just to ensure we have a check to make sure this table is updated in the PR checklist.

For me, I don't think we should delete this until we have a better strategy around publishing the available rules and associated information in the docs.

crates/wdl-lint/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 //! Lint rules for Workflow Description Language (WDL) documents.
-#![doc = include_str!("../RULES.md")]
```
claymcleod, line 2: This also needs to be added back in.

crates/wdl-lint/RULES.md
claymcleod: And I didn't say this explicitly, but I don't think asking non-Rust-familiar users to dig through our developer documentation to find out this information is a good enough solution in the meantime.

crates/wdl-lint/src/rules/description_length.rs
claymcleod: Seems fine to me.

adthrasher (dismissed):

a-frantz:
@claymcleod I added back `RULES.md`, but have removed the table formatting. IMO it's just an extra maintenance burden that only functions to prettify the source, which is probably not how most people are viewing it. I also added the PR template two lint rule specific checks to make sure that these updates aren't forgotten in the future

claymcleod (approved):

Issue #181, opened 2025-09-19
`check` errors while attempting to recurse my `runs/` directory


a-frantz:
Very unhelpful error messaging:
```console
$ RUST_BACKTRACE=1 cargo run -- check ../workflows/
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.47s
     Running `target/debug/sprocket check ../workflows/`
error: error parsing ignorefile with basename `.sprocketignore`

Stack backtrace:
   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2: std::backtrace::Backtrace::create
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/backtrace.rs:331:13
   3: anyhow::kind::Adhoc::new
             at /Users/afrantz/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/anyhow-1.0.99/src/backtrace.rs:27:14
   4: sprocket::commands::check::check::{{closure}}
             at ./src/commands/check.rs:313:13
   5: sprocket::inner::{{closure}}
             at ./src/main.rs:89:77
   6: sprocket::main::{{closure}}
             at ./src/main.rs:108:29
   7: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /Users/afrantz/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/future/future.rs:124:9
   8: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
             at /Users/afrantz/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.47.1/src/runtime/park.rs:285:71
   9: tokio::task::coop::with_budget
             at /Users/afrantz/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.47.1/src/task/coop/mod.rs:167:5
  10: tokio::task::coop::budget
[...]

PR #365, opened 2025-09-26
feat: re-implement the `format` subcommand with new API

### LINKED ISSUE #357: feat: send formatted document to STDOUT
requested on the openwdl slack. The `format` command has a kind of awkward interface at the moment, with requiring one of the `--overwrite` or `--check` flags. We can change that while adding this feature, although I haven't thought through what's preferable

### LINKED ISSUE #375: Sprocket format does not properly escape things
When formatting this line in a WDL task

```wdl
        ~{"--haplotypecaller-options " + '"' + haplotypecallerPassthroughOptions + '"'} \
```

Sprocket replaces the single quotes with double quotes. However, this makes the string evaluation incorrect now. We need to either detect when a string is intentionally using single quotes for this purpose or escape the double quotes when we convert it.

a-frantz:

closes #357 
closes #375 

new API uses clap subcommands instead of mutually exclusive flags. The subcommands are `check`, `view`, and `overwrite`.

```console
Formats a document or a directory containing documents

Usage: sprocket format [OPTIONS] <COMMAND>

Commands:
  check      Check if files are formatted correctly and print diff if not
  view       Format a document and send the result to STDOUT
  overwrite  Reformat all WDL documents via overwriting
  help       Print this message or the help of the given subcommand(s)

Options:
      --no-color                  Disables color output
  -m, --report-mode <MODE>        The report mode for any emitted diagnostics [possible values: full, one-line]
      --with-tabs                 Use tabs for indentation (default is spaces)
  -i, --indentation-size <SIZE>   The number of spaces to use for indentation levels (default is 4)
      --max-line-length <LENGTH>  The maximum line length (default is 90)
  -v, --verbose...                Increase logging verbosity
  -q, --quiet...                  Decrease logging verbosity
  -c, --config <CONFIG>           Path to the configuration file
  -s, --skip-config-search        Skip searching for and loading configuration files
  -h, --help                      Print help (see more with '--help')
  -V, --version                   Print version
```

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/src/commands/format.rs
+++ b/src/commands/format.rs
@@ -39,7 +39,7 @@ pub struct Args {
     pub report_mode: Option<Mode>,
 
     /// Use tabs for indentation (default is spaces).
-    #[arg(long, global = true)]
+    #[arg(short = 't', long, global = true)]
     pub with_tabs: bool,
 
     /// The number of spaces to use for indentation levels (default is 4).
@@ -136,16 +136,12 @@ fn format_document(
 
 /// Runs the `format` command.
 pub async fn format(args: Args) -> Result<()> {
-    let indent = match Indent::try_new(args.with_tabs, args.indentation_size) {
-        Ok(indent) => indent,
-        Err(e) => bail!("failed to create indentation configuration: {}", e),
-    };
+    let indent = Indent::try_new(args.with_tabs, args.indentation_size)
+        .with_context(|| "failed to create indentation configuration")?;
 
     let max_line_length = match args.max_line_length {
-        Some(length) => match MaxLineLength::try_new(length) {
-            Ok(max_line_length) => max_line_length,
-            Err(e) => bail!("failed to create max line length configuration: {}", e),
-        },
+        Some(length) => MaxLineLength::try_new(length)
+            .with_context(|| "failed to create max line length configuration")?,
         None => MaxLineLength::default(),
     };
 
@@ -216,7 +212,7 @@ pub async fn format(args: Args) -> Result<()> {
                     }
                     errors += 1;
                 } else {
-                    info!("`{}` is formatted correctly", result.document().path())
+                    println!("`{}` is formatted correctly", result.document().path())
                 }
             }
         }
@@ -225,7 +221,10 @@ pub async fn format(args: Args) -> Result<()> {
             match &source {
                 Source::File(_) | Source::Remote(_) => {}
                 Source::Directory(p) => {
-                    bail!("`view` does not support directories: `{}`", p.display())
+                    bail!(
+                        "the `format view` command does not support formatting directory `{path}`",
+                        path = p.display()
+                    );
                 }
             };
 
@@ -240,7 +239,10 @@ pub async fn format(args: Args) -> Result<()> {
             let result = results.filter(&[&source]).next().unwrap();
 
             if let Some(err) = result.error() {
-                bail!("error analyzing `{}`: {}", result.document().path(), err);
+                bail!(
+                    "error analyzing `{path}`: {err:#}",
+                    path = result.document().path()
+                );
             }
 
             let (_source, formatted) = match format_document(
@@ -295,7 +297,10 @@ pub async fn format(args: Args) -> Result<()> {
                     Ok(r) => r,
                     Err(e) => {
                         errors += 1;
-                        warn!("not overwriting `{}`: {e}", result.document().path());
+                        warn!(
+                            "not overwriting document `{path}` due to error: {e:#}",
+                            path = result.document().path()
+                        );
                         continue;
                     }
                 };
--- a/src/commands/format.rs
+++ b/src/commands/format.rs
@@ -43,7 +43,13 @@ pub struct Args {
     pub with_tabs: bool,
 
     /// The number of spaces to use for indentation levels (default is 4).
-    #[arg(short, long, value_name = "SIZE", conflicts_with = "with_tabs", global = true)]
+    #[arg(
+        short,
+        long,
+        value_name = "SIZE",
+        conflicts_with = "with_tabs",
+        global = true
+    )]
     pub indentation_size: Option<usize>,
 
     /// The maximum line length (default is 90).
--- a/src/commands/format.rs
+++ b/src/commands/format.rs
@@ -1,10 +1,7 @@
 //! Implementation of the `format` subcommand.
 
-use std::ffi::OsStr;
 use std::fs;
 use std::io::IsTerminal;
-use std::io::Read;
-use std::path::Path;
 
 use anyhow::Context;
 use anyhow::Result;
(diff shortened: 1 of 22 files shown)


src/commands/format.rs
```diff
@@ -209,42 +153,165 @@ pub fn format(args: Args) -> Result<()> {
...
+                            "{}",
+                            pretty_assertions::StrComparison::new(&source, &formatted)
+                        );
+                    } else {
+                        let diff = similar::TextDiff::from_lines(&source, &formatted);
+                        eprint!("{}", diff.unified_diff().header("input", "formatted"));
+                    }
+                    errors += 1;
+                } else {
+                    info!("`{}` is formatted correctly", result.document().path())
```
peterhuene: Previously we had a `println!` for this, which I think makes more sense than logging it, as it's effectively the "output result" of the command.

src/commands/format.rs
```diff
@@ -209,42 +153,165 @@ pub fn format(args: Args) -> Result<()> {
...
+                }
+            }
+        }
+        FormatSubcommand::View(s) => {
+            let source = s.source;
+            match &source {
+                Source::File(_) | Source::Remote(_) => {}
+                Source::Directory(p) => {
+                    bail!("`view` does not support directories: `{}`", p.display())
```
peterhuene:
```suggestion
                    bail!("the `format view` command does not support formatting directory `{path}`", path = p.display());
```

src/commands/format.rs
```diff
@@ -209,42 +153,165 @@ pub fn format(args: Args) -> Result<()> {
...
+                Err(errors) => {
+                    // SAFETY: this is a non-empty, so it must always have a first
+                    // element.
+                    bail!(errors.into_iter().next().unwrap())
+                }
+            };
+            let result = results.filter(&[&source]).next().unwrap();
+
+            if let Some(err) = result.error() {
+                bail!("error analyzing `{}`: {}", result.document().path(), err);
```
peterhuene: If we're going to bail with the error, we should be using the `#` specifier for context:

```suggestion
                bail!("error analyzing `{path}`: {err:#}", path = result.document().path());
```

src/commands/format.rs
```diff
@@ -209,42 +153,165 @@ pub fn format(args: Args) -> Result<()> {
...
+                &formatter,
+                result.document(),
                 args.report_mode.unwrap_or_default(),
                 args.no_color,
-                args.mode.check,
-            )?;
+            ) {
+                Ok(r) => r,
+                Err(e) => {
+                    bail!("could not view `{}`: {e}", result.document().path());
```
peterhuene: This `bail` should be removed in favor of doing a `.with_context(|| format!("could not view document `{path}`, path = result.document().path()))` on the above call to `format_document`.

src/commands/format.rs
```diff
@@ -87,111 +79,63 @@ impl Args {
...
-pub fn format(args: Args) -> Result<()> {
-    if let Some(Source::Remote(_)) = args.source {
-        bail!("remote sources are not supported for the `format` command");
-    }
-
-    let source = args.source.unwrap_or_default();
+pub async fn format(args: Args) -> Result<()> {
     let indent = match Indent::try_new(args.with_tabs, args.indentation_size) {
         Ok(indent) => indent,
         Err(e) => bail!("failed to create indentation configuration: {}", e),
```
peterhuene: Like a previous comment, this `bail!` should probably be a `with_context` on the call to `try_new`. There may be other places in this file where we don't want to `bail!` with a display of the error too.

src/commands/format.rs
```diff
@@ -209,42 +153,165 @@ pub fn format(args: Args) -> Result<()> {
...
+                .extend_sources(sources.clone())
+                .ignore_filename(Some(IGNORE_FILENAME.to_string()))
+                .run()
+                .await
+            {
+                Ok(results) => results,
+                Err(errors) => {
+                    // SAFETY: this is a non-empty, so it must always have a first
+                    // element.
+                    bail!(errors.into_iter().next().unwrap())
```
peterhuene, line 172: We don't want to use `bail!` like this as it contributes to #181 by erasing the context of the error. I have a fix for this I'll put up in a separate PR.

src/commands/format.rs
```diff
@@ -209,42 +153,165 @@ pub fn format(args: Args) -> Result<()> {
...
+                let (_source, formatted) = match format_document(
+                    &formatter,
+                    result.document(),
+                    args.report_mode.unwrap_or_default(),
+                    args.no_color,
+                ) {
+                    Ok(r) => r,
+                    Err(e) => {
+                        errors += 1;
+                        warn!("not overwriting `{}`: {e}", result.document().path());
```
peterhuene:
```suggestion
                        warn!("not overwriting document `{path}` due to error: {e:#}", path = result.document().path());
```

peterhuene (commented):
Just a few comments related to errors, but otherwise looks good!

src/commands/format.rs
```diff
@@ -87,164 +79,244 @@ impl Args {
...
-    }
-
-    let source = args.source.unwrap_or_default();
-    let indent = match Indent::try_new(args.with_tabs, args.indentation_size) {
-        Ok(indent) => indent,
-        Err(e) => bail!("failed to create indentation configuration: {}", e),
-    };
+pub async fn format(args: Args) -> Result<()> {
+    let indent = Indent::try_new(args.with_tabs, args.indentation_size)
+        .with_context(|| "failed to create indentation configuration")?;
```
peterhuene: Without string formatting, this should just be `context`:

```suggestion
        .context("failed to create indentation configuration")?;
```

src/commands/format.rs
```diff
@@ -87,164 +79,244 @@ impl Args {
...
 
     let max_line_length = match args.max_line_length {
-        Some(length) => match MaxLineLength::try_new(length) {
-            Ok(max_line_length) => max_line_length,
-            Err(e) => bail!("failed to create max line length configuration: {}", e),
-        },
+        Some(length) => MaxLineLength::try_new(length)
+            .with_context(|| "failed to create max line length configuration")?,
```
peterhuene: Without string formatting, this should just be context:

```suggestion
            .context("failed to create max line length configuration")?,
```

src/commands/format.rs
peterhuene: I don't think this comment was resolved.

src/commands/format.rs
```diff
@@ -87,164 +79,244 @@ impl Args {
...
+                }
+            };
+            let sources = sources.iter().collect::<Vec<_>>();
+            let results = results.filter(sources.as_slice()).collect::<Vec<_>>();
+            for result in results {
+                info!("formatting `{}`", result.document().path());
+
+                if let Some(err) = result.error() {
+                    errors += 1;
+                    warn!("error analyzing `{}`: {}", result.document().path(), err);
```
peterhuene:
```suggestion
                    warn!("error analyzing `{path}`: {err:#}", path = result.document().path());
```

peterhuene (commented):
Just a few follow-up comments.

src/commands/format.rs
```diff
@@ -87,164 +79,248 @@ impl Args {
...
-    // Write file because check is not true
-    fs::write(path, formatted)
-        .with_context(|| format!("failed to write `{path}`", path = path.display()))?;
-
-    Ok(0)
+    let ast = document
+        .root()
+        .ast()
+        .into_v1()
+        .expect("only WDL v1.x documents are supported");
```
acfoltzer, line 132: I have a slight preference for `anyhow`ing rather than panicking here, but we have many other parts of the code that panic for unexpected versions already.

src/commands/format.rs
```diff
@@ -87,164 +79,248 @@ impl Args {
...
+                    path = result.document().path()
+                )
+            })?;
+            print!("{}", formatted);
+        }
+        FormatSubcommand::Overwrite(s) => {
+            let mut sources = s.sources;
+            if sources.is_empty() {
+                sources.push(Source::default());
+            }
```
acfoltzer, line 266: A comment here about what `Source::default()` does would be useful. Reading through without cross-referencing, it's not clear why an overwrite without sources shouldn't be an error.

Looking up what `Source::default()` does... I think I would advocate removing this trait impl and replacing it with something more-specifically named method like `Source::cwd()`. But that's kinda beyond the scope of this PR, so OK to save for later.

acfoltzer (dismissed):

acfoltzer (approved):

PR #388, opened 2025-09-30
fix: handle empty metadata string

### LINKED ISSUE #387: Panic on empty parameter meta field
For a task with this parameter meta

```wdl
    parameter_meta {
        file_bam: "Path to the BAM file"
        file_bai: "Path to the BAM index file"
        original_ref_tarball: "Original reference tarball. Required for CRAM input."
        pb_license_bin: "Path to the Parabricks license binary"
        ref: "Name of FASTA reference file, required for CRAM input"
        pb_path: "Parabricks command pbrun"
        pb_docker: "The docker image to use for the task"
        hpc_queue: ""
        n_threads: "Number of threads to use; default 16"
        gb_ram: "GB of RAM to use; default 120"
        disk_gb: "GB of disk space to use; default 0"
        runtime_minutes: "Runtime minutes, set to 600"
    }
```

I get this output

```
thread 'tokio-runtime-worker' panicked at crates/wdl-engine/src/value.rs:114:22:
metadata strings shouldn't have placeholders
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' panicked at crates/wdl-engine/src/eval/v1/workflow.rs:837:22:
failed to join future: JoinError::Panic(Id(6), "metadata strings shouldn't have placeholders", ...)
```

It goes away when `hpc_queue` is not empty.

a-frantz:

closes #387 

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/src/value.rs
+++ b/crates/wdl-engine/src/value.rs
@@ -110,9 +110,9 @@ impl Value {
             v1::MetadataValue::Integer(v) => v.value().expect("number should be in range").into(),
             v1::MetadataValue::Float(v) => v.value().expect("number should be in range").into(),
             v1::MetadataValue::String(v) => PrimitiveValue::new_string(
-                v.text()
-                    .expect("metadata strings shouldn't have placeholders")
-                    .text(),
+                // `v.text()` will return `None` if the `LiteralString` is empty, so we call
+                // `unwrap_or_default()` to handle that case
+                v.text().map(|s| s.text().to_string()).unwrap_or_default(),
             )
             .into(),
             v1::MetadataValue::Null(_) => Self::new_none(Type::None),


peterhuene:
I think we have a few places where this bug is surfaced:

* `crates/wdl-doc/src/meta.rs:53`
* `wdl-doc/src/meta.rs:204`
* `crates/wdl-format/src/v1.rs:59`
* `crates/wdl-format/src/v1.rs:60`
* `src/commands/inputs.rs:179`

peterhuene:
Although now I remember why it's returning `StringText` as it's a red node and thus a temporary we can't return a reference to its text. Hmm, more thought required.

src/commands/inputs.rs
```diff
@@ -175,12 +182,19 @@ impl InputProcessor {
                     for item in m.items() {
                         let (k, v) = item.key_value();
                         let key_name: String = match k {
-                            Expr::Literal(LiteralExpr::String(k)) => {
-                                let text = k.text().expect("should have string text");
-                                let mut buffer = String::new();
-                                text.unescape_to(&mut buffer);
-                                buffer
-                            }
+                            Expr::Literal(LiteralExpr::String(k)) => k
```
a-frantz: I made this change without actually taking a close look at the context (been a long time since I reviewed this bit of code). I'm not sure this is right, but it's late for me. I'll take another look in the morning, but this might need to be changed 🤔

peterhuene:
Naming things is hard :frowning_face: Could probably come up with a better name than `LiteralStringText` to mean "the text of a string without any placeholders".

Additionally, we can probably abstract it such that we store `T` instead of `StringToken<T>` so it can be used from both `LiteralString` and `CommandSection`.

src/commands/inputs.rs
```diff
@@ -174,16 +181,7 @@ impl InputProcessor {
...
-                                text.unescape_to(&mut buffer);
-                                buffer
-                            }
-                            _ => k.text().to_string(),
-                        };
-
+                        let key_name = k.text().to_string();
                         if let Some((_key, value)) =
                             self.expression(ty.clone(), Key::empty(), name, &v)
                         {
```
a-frantz, line 187: @adthrasher this was originally written by you and I'm having trouble piecing together the logic here.

I believe this branch only gets hit iff `--show-expressions` is supplied; this is trying to display a literal map default value.

I'm not sure when a WDL author would want users constructing a `Map` of any type, but I am tempted to "give up" in this case and just tell users "consult the source" instead of this convoluted logic. If we go that direction, I think a few of these other nearby branches could similarly be simplified. What do you think?

peterhuene:
> Is it worth it to abstract this over command text as well?

It looks like there's actually no callers to `CommandSection::text` currently, so perhaps we just remove it for now until such time as someone needs it?

src/commands/inputs.rs
adthrasher: I don't particularly recall. That said, I think the `inputs` command giving up and saying consult the source for certain data types is a poor user experience.

a-frantz: My thinking is that this case is sufficiently complex that I don't think there's anything _useful_ the `inputs` command can return. Instead of trying to embed an entire `Map[*, *]` definition inside of a JSON string, I'd prefer the out JSON to look something like:
```json
{
    "foo.weirdly_complicated_input": "Map[Foo, Bar] (default = ...)"
}
```
not super confident that the `...` literal is sufficiently clear, but I do think that some literal text snippet indicating "we can't embed something this complicated" is better than the current logic

adthrasher: I looked around the repos in Gauntlet. This was the only place where a `Map` was used as an input.

```
% sprocket inputs --name  consolidate_stats https://raw.githubusercontent.com/PacificBiosciences/wdl-common/ae2cd95da6db6eb13eeaa6d5559976af2eb19aad/wdl/tasks/utilities.wdl
{
  "consolidate_stats.id": "String",
  "consolidate_stats.stats": "Map[String, Array[String]]",
  "consolidate_stats.runtime_attributes": "RuntimeAttributes"
}
```

In the vast majority of cases, `Map` was used as a private declaration (often read from a file). So I doubt there's much downside to printing the types of the input and some indicator that it has a default value, rather than printing the full default value.

a-frantz: Yea what's currently happening is confusing and probably not what we want.
```console
$ git diff
diff --git a/tests/cli/inputs/default/complex-types/args b/tests/cli/inputs/default/complex-types/args
index 571e03d..3fe35a8 100644
--- a/tests/cli/inputs/default/complex-types/args
+++ b/tests/cli/inputs/default/complex-types/args
@@ -1 +1 @@
-inputs complex-types.wdl
\ No newline at end of file
+inputs --show-expressions --nested-inputs complex-types.wdl
\ No newline at end of file
diff --git a/tests/cli/inputs/default/complex-types/inputs/complex-types.wdl b/tests/cli/inputs/default/complex-types/inputs/complex-types.wdl
index 38c92ee..386a282 100644
--- a/tests/cli/inputs/default/complex-types/inputs/complex-types.wdl
+++ b/tests/cli/inputs/default/complex-types/inputs/complex-types.wdl
@@ -41,6 +41,12 @@ workflow test {
         Array[Float] y
         String empty = ""
         String interpolated = "weirdly nested string with interpolation: ~{empty}"
+        Map[Int, Baz] oi_vey = {
+            42: Baz {
+                foo: true,
+                bar: 4.2,
+            }
+        }
     }

     call foo as my_call
diff --git a/tests/cli/inputs/default/complex-types/stdout b/tests/cli/inputs/default/complex-types/stdout
index 23f8e75..5294fce 100644
--- a/tests/cli/inputs/default/complex-types/stdout
+++ b/tests/cli/inputs/default/complex-types/stdout
@@ -5,5 +5,12 @@
   "test.x": null,
   "test.y": "Array[Float]",
   "test.empty": "",
-  "test.interpolated": "weirdly nested string with interpolation: ~{empty}"
+  "test.interpolated": "weirdly nested string with interpolation: ~{empty}",
+  "test.oi_vey": {
+    "42": {
+      "foo": true,
+      "bar": 4.2
+    }
+  },
+  "test.my_call.foo": "Foo"
 }
```
The STDOUT value for `test.oi_vey` is neither valid to be used as an input nor clear on what exactly its meant to represent. That said, I think reworking the `inputs` command to be clearer is probably out of scope for this PR.

I can put up another PR which changes `inputs`, and leave this PR focused on the `StringText` changes

adthrasher: The spec has some missing serialization when it comes to `struct` types. I would consider that `stdout` to be correct as Sprocket has coerced the `struct` to an `object`.

a-frantz: But the key `"42"` is wrong. It should be unquoted as it's an `Int`

adthrasher: The spec actually disallows this serialization:

> There is no natural or unambiguous serialization of a Map with a non-String key type. Attempting to serialize a Map with a non-String key type results in an error. A Map with non-String keys must first be converted to a serializable type, e.g., using one of the following suggested methods.

PR #402, opened 2025-10-06
revise: differentiate parse and analysis diagnostics

a-frantz:

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/src/document.rs
+++ b/crates/wdl-analysis/src/document.rs
@@ -477,8 +477,10 @@ struct DocumentData {
     workflow: Option<Workflow>,
     /// The structs in the document.
     structs: IndexMap<String, Struct>,
-    /// The diagnostics for the document.
-    diagnostics: Vec<Diagnostic>,
+    /// The diagnostics from parsing.
+    parse_diagnostics: Vec<Diagnostic>,
+    /// The diagnostics from analysis.
+    analysis_diagnostics: Vec<Diagnostic>,
 }
 
 impl DocumentData {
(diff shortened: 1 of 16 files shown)


crates/wdl-analysis/src/document.rs
```diff
@@ -804,8 +821,9 @@ impl Document {
     pub fn has_errors(&self) -> bool {
```
a-frantz, line 821: I just made this change to get it compiling, but I'm not sure we actually want this as part of the `Document` API as it is now. I'd prefer this to instead `chain` and `filter` _and return the errors_ rather than a boolean. We should be reporting the problems encountered rather than detecting and failing without information.

peterhuene: This solely exists so that evaluation fails up-front when given a document that transitively contains any errors to prevent a `wdl-engine` API misuse rather than surface the errors to the user (which normally would have been done _before_ trying to evaluate a task or workflow).

I'm fine with changing it to an iterator over the errors, but the place where we're calling `has_errors` still won't be displaying them (it's the responsibility of the caller to have already displayed the errors and _not_ call into `wdl-engine` for evaluation when errors are present).

a-frantz: that makes sense to me, thanks for the extra context! I'll add that another thought I had while writing this is that we have this logic (filtering all the diagnostics to just the errors) in a bunch of places. It's trivial logic, but its led to a fair bit of repetition we could nip with a tweak in this API.

crates/wdl-analysis/src/document.rs
```diff
@@ -804,8 +821,9 @@ impl Document {
...
         // Check this document for errors
         if self
-            .diagnostics()
```
peterhuene, line 807: Does this need to change now that we have a `diagnostics` method that does this chain?

a-frantz: oop. Yea you're right

PR #407, opened 2025-10-08
feat(lint): Implemented a lint rule which ensures reserved meta and parameter_meta keys have string values

### LINKED ISSUE #353: request lint rule: ExpectedMetaString
This should warn if any of reserved meta keys (`description`, `help`, `external_help`, `warning`, `group` (for inputs), and `category` (for workflows)) have a non `String` value. All these keys are reserved by `wdl-doc` and using a non-string value will skip their rendering.

Krriishhhhhh:

This PR implements  the `ExpectedMetaString` lint rule, which ensures that all reserved metadata keys used by **wdl-doc** have string values for proper documentation rendering.

Fixes #353 

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,254 @@
...
+         `category`, and `group`) must have string values. Using non-string values will cause \
+         wdl-doc to skip rendering that documentation. This rule ensures all reserved keys have \
+         string values for proper documentation generation."
+    }
+
+    fn tags(&self) -> TagSet {
+        TagSet::new(&[
+            Tag::Correctness,
+            Tag::Documentation,
+            Tag::SprocketCompatibility,
```
a-frantz:
```suggestion
            Tag::SprocketCompatibility,
```
the other tags could apply, but this rule really only applies if someone is using the Sprocket `doc` command. If they _aren't_ using `doc`, there is a good chance they don't care about following this rule, so we should be reserved in when it's enabled.

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,254 @@
...
+    fn description(&self) -> &'static str {
+        "Ensures that reserved meta keys used by wdl-doc have string values."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "The wdl-doc tool reserves certain keys in `meta` and `parameter_meta` sections for \
+         documentation generation. These keys (`description`, `help`, `external_help`, `warning`, \
+         `category`, and `group`) must have string values. Using non-string values will cause \
+         wdl-doc to skip rendering that documentation. This rule ensures all reserved keys have \
+         string values for proper documentation generation."
```
a-frantz: I'd change all these instances of "wdl-doc" to "Sprocket's `doc` command" (or something similar). These messages will be surfaced to users of `sprocket lint`, who may not be aware that `doc` is implemented by the `wdl-doc` crate under the hood, so we should obscure that detail as unimportant.

crates/wdl-lint/src/rules/doc_meta_strings.rs
```diff
@@ -0,0 +1,254 @@
...
+    .with_rule(ID)
+    .with_label(
+        "parameter description must be a string for proper documentation rendering",
+        span,
+    )
+    .with_fix("change the parameter description to a string literal")
+}
+
+/// Gets a human-readable type name for a metadata value.
+fn get_value_type_name(value: &MetadataValue) -> &'static str {
```
a-frantz, line 51: Can you double check the `wdl-ast`/`wdl-grammar` API to see if something to this effect already exists? My instinct here is that this functionality should already exist upstream, and this code would be redundant (but I'm not 100% sure).

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,254 @@
...
+    .with_label(
+        format!("`{key}` must be a string for proper documentation rendering"),
+        span,
+    )
+    .with_fix(format!("change the value of `{key}` to a string literal"))
+}
+
+/// Creates a "non-string parameter description" diagnostic for simple
+/// parameter_meta entries.
+fn non_string_parameter_description(param_name: &str, value_type: &str, span: Span) -> Diagnostic {
```
a-frantz: is there a reason I'm not seeing for having 3 different functions? This should be doable with one function, unless I'm missing something

crates/wdl-lint/src/rules/doc_meta_strings.rs
```diff
@@ -0,0 +1,254 @@
...
+        ]
+    }
+}
+
+impl Visitor for ExpectedMetaStringRule {
+    fn reset(&mut self) {
+        *self = Default::default();
+    }
+
+    fn metadata_section(
```
a-frantz, line 148: this also needs to do a similar search through the `meta { outputs: { ... } }` object if it exists. The contents of `meta.outputs` is expected to be similar to the `parameter_meta` section

crates/wdl-lint/tests/lints/trailing-comma/source.wdl
```diff
@@ -28,13 +28,6 @@ workflow bar {
 
 task foo {
     meta {
```
a-frantz, line 31:
```suggestion
    # except: ExpectedMetaString
    meta {
```
and then add these deleted test cases back

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,254 @@
...
+use wdl_ast::v1::MetadataSection;
+use wdl_ast::v1::MetadataValue;
+use wdl_ast::v1::ParameterMetadataSection;
+
+use crate::Rule;
+use crate::Tag;
+use crate::TagSet;
+
+/// The identifier for the expected meta string rule.
+const ID: &str = "ExpectedMetaString";
```
a-frantz: The issue I created used this name, but on a second look it doesn't fit with our other rule names. It should change, but I haven't decided what exactly works best. Maybe `ReservedMetaKeyString`? That's not quite right either... 🤔 

You can leave it like this for now, but I'll come back on a later review with a better name

crates/wdl-lint/src/rules/doc_meta_strings.rs
Krriishhhhhh: Hey , i checked out both the crates but couldn't find the functionality that this function provides , but i am also not sure either since i am not familier with both of these crates . How should we proceed ?

crates/wdl-lint/src/rules/expected_meta_string.rs
Krriishhhhhh: I can make it one function, but since the diagnostic warnings are different for three cases (meta, parameter_meta, and parameter_meta with an object), the function would end up having three different structures, with conditional logic to handle each part. That’s why I thought dividing them into three separate parts would make it look more structured. Implementation would roughly look like this .


```
enum DiagnosticContext<'a> {
    
    Meta { key: &'a str },
    
    ParameterMeta { param_name: &'a str, key: &'a str },
    
    ParameterDescription { param_name: &'a str },
}


fn non_string_metadata_diagnostic(
    context: DiagnosticContext,
    value_type: &str,
    span: Span,
) -> Diagnostic {
    let (message, label, fix) = match context {
        DiagnosticContext::Meta { key } => (
            ["reserved meta key `", key, "` should have a string value, found ", value_type].concat(),
            ["`", key, "` must be a string for proper documentation rendering"].concat(),
            ["change the value of `", key, "` to a string literal"].concat(),
        ),
        DiagnosticContext::ParameterMeta { param_name, key } => (
            [
                "reserved parameter_meta key `",
                key,
                "` for parameter `",
                param_name,
                "` should have a string value, found ",
                value_type,
            ].concat(),
            ["`", key, "` must be a string for proper documentation rendering"].concat(),
            ["change the value of `", key, "` to a string literal"].concat(),
        ),
        DiagnosticContext::ParameterDescription { param_name } => (
            [
                "parameter `",
                param_name,
                "` description should be a string, found ",
                value_type,
            ].concat(),
            "parameter description must be a string for proper documentation rendering".to_string(),
            "change the parameter description to a string literal".to_string(),
        ),
    };

    Diagnostic::warning(message)
        .with_rule(ID)
        .with_label(label, span)
        .with_fix(fix)
}
```

crates/wdl-lint/tests/lints/trailing-comma/source.wdl
Krriishhhhhh: Previously existing WDL file was not following the new lint rule , so the only options were to remove those test cases , or add the rule in Except , can i put the rule in exception then , cause i coudn't find out another way around it .

crates/wdl-lint/src/rules/expected_meta_string.rs
Krriishhhhhh: How about `MetaKeyStringValue`  or `MetaStringRequired`?

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,254 @@
...
+];
+
+/// Reserved parameter_meta keys that must have string values for wdl-doc
+/// compatibility.
+const RESERVED_PARAMETER_META_KEYS: &[&str] = &[
+    "description",
+    "help",
+    "external_help",
+    "group", // for grouping inputs
+];
```
a-frantz: I don't love splitting these up like this, as it doesn't _quite_ capture the cases that these are checked in `wdl-doc`.

I'm torn between making this check a bit dumber for simplicity and trying to match the `wdl-doc` logic closer 🤔

The major downside to matching the `wdl-doc` logic exactly is that it increases the maintenance burden of the crates, as these two bits of codes aren't directly linked. I'm hesitating to duplicate that bit of complicated logic without a good way of making sure that if the behavior changes upstream that it gets reflected here.

Let's collapse these lists and just treat every entry in either section the same. So there will just be a list of keys that should _always_ have `String` values regardless of where they appear. The downside is that this diagnostic _could_ be emitted as a false positive, but that would be
1. unlikely in practice
2. easy to fix with an `except` statement
3. worth it in exchange for simpler logic here

Please leave this thread unresolved so this is more visible and the next reviewer can weigh in if they disagree.

crates/wdl-lint/src/rules/expected_meta_string.rs
a-frantz: ah right, I forgot to leave a comment for a related thought I had (made the comment here - https://github.com/stjude-rust-labs/sprocket/pull/407/files#r2416292798)

I don't think we need to change the diagnostic language for each of these cases, and we could probably simplify this into one case. So try to rework this into one function, but if you still think the language needs to be more context aware than that feel free to ping on this thread with your reasoning!

crates/wdl-lint/src/rules/doc_meta_strings.rs
a-frantz: Leave this function as it and this thread unresolved for now, and I or the next reviewer can check later 👍

crates/wdl-lint/tests/lints/trailing-comma/source.wdl
a-frantz: the above suggested edit (adding a lint directive above this `meta` section) should work for silencing the diagnostics without removing the test cases

crates/wdl-lint/src/rules/expected_meta_string.rs
a-frantz: The other bit to keep in mind while reworking this text is that `doc` is _currently_ gated behind the `dev` command (invoked with `sprocket dev doc` instead of `sprocket doc`), but the language here shouldn't reflect that (we don't want to need to update this language when we make `doc` a first-class command)

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,257 @@
...
+/// Creates a diagnostic for non-string metadata values.
+fn non_string_value_diagnostic(key: &str, value_type: &str, span: Span) -> Diagnostic {
+    Diagnostic::warning(
+        [
+            "metadata key `",
+            key,
+            "` should have a string value, found ",
+            value_type,
+        ]
+        .concat(),
```
a-frantz: use `format!()` instead of `concat` everywhere this appears please!

Krriishhhhhh:
@a-frantz Done ! There are still some unresolved conversations around the code structure and name of the rule , can you take a look at them.

crates/gauntlet/Arena.toml
```diff
@@ -1705,6 +1705,11 @@ document = "stjudecloud/workflows:/tools/samtools.wdl"
 message = "samtools.wdl:1062:20: note[ContainerUri]: container URI uses a mutable tag"
 permalink = "https://github.com/stjudecloud/workflows/blob/2213187864006b6492ccb0f9318c149bc1a823cd/tools/samtools.wdl/#L1062"
 
+[[diagnostics]]
+document = "stjudecloud/workflows:/tools/samtools.wdl"
+message = "samtools.wdl:1069:9: warning[ExpectedMetaString]: metadata key `description` should have a string value, found object"
+permalink = "https://github.com/stjudecloud/workflows/blob/2213187864006b6492ccb0f9318c149bc1a823cd/tools/samtools.wdl/#L1069"
```
a-frantz, line 1711: nice, found a real bug here!

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+    "external_help",
+    "warning",
+    "category",
+    "group",
+];
+
+/// Creates a diagnostic for non-string metadata values.
+fn non_string_value_diagnostic(key: &str, value_type: &str, span: Span) -> Diagnostic {
+    Diagnostic::warning(format!(
+        "metadata key `{}` should have a string value, found {}",
```
a-frantz:
```suggestion
        "meta key `{}` should have a `String` value, found {}",
```

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+        key, value_type
+    ))
+    .with_rule(ID)
+    .with_label(
+        format!(
+            "`{}` must be a string for proper documentation rendering",
```
a-frantz:
```suggestion
            "`{}` must be a `String` for proper documentation rendering",
```

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+            key
+        ),
+        span,
+    )
+    .with_fix(format!("change the value of `{}` to a string literal", key))
```
a-frantz:
```suggestion
    .with_fix(format!("change the value of `{}` to a `String`", key))
```

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    fn description(&self) -> &'static str {
+        "Ensures that reserved meta keys have string values."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "Sprocket's documentation command reserves certain keys in `meta` and `parameter_meta` \
+         sections for documentation generation. These keys (`description`, `help`, \
+         `external_help`, `warning`, `category`, and `group`) must have string values. Using \
```
a-frantz:
```suggestion
         `external_help`, `warning`, `category`, and `group`) must have `String` values. Using \
```

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+         non-string values will cause the documentation to be rendered incorrectly or not at all. \
```
a-frantz:
```suggestion
         non-`String` values will cause the documentation to be rendered incorrectly or not at all. \
```

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+         This rule ensures all reserved keys have string values for proper documentation \
```
a-frantz:
```suggestion
         This rule ensures all reserved keys have `String` values for proper documentation \
```

a-frantz (commented):
sorry this got left unreviewed for so long! I started one and then looks like I never hit submit :sob:

This looks really good; I just have some nits, and then there are still two unresolved threads I would like someone else on the team to weigh in when they can, but probably won't be till next week. I'll poke someone so we can get this over the finish line! 🐎

crates/wdl-lint/src/rules/expected_meta_string.rs
acfoltzer: I think the `Meta*` names are getting a bit crowded, and that it would be better to group lints related to `doc`. `DocMetaStrings` would be my choice.

crates/wdl-lint/src/rules/doc_meta_strings.rs
acfoltzer: I think it's fine to leave this here for now. Meta values are kinda weird in that they're distinct from but overlap with normal WDL types, which we _do_ have formatting code for. In the long term, it'd be nice to agree on how these should be formatted and put that logic in `wdl-ast`, but that would pretty dramatically expand the scope of this PR so I'm fine not doing it yet.

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+        let value = item.value();
+
+        if !is_string_value(&value) {
+            let value_type = get_value_type_name(&value);
+            diagnostics.exceptable_add(
+                non_string_value_diagnostic(key, value_type, item.span()),
+                SyntaxElement::from(item.inner().clone()),
+                exceptable_nodes,
+            );
+        }
```
acfoltzer: Avoiding negation and early-`continue` makes it easier to see what's happening here.
```suggestion
        let value = item.value();

        if RESERVED_KEYS.contains(&key) && !is_string_value(&value) {
            let value_type = get_value_type_name(&value);
            diagnostics.exceptable_add(
                non_string_value_diagnostic(key, value_type, item.span()),
                SyntaxElement::from(item.inner().clone()),
                exceptable_nodes,
            );
        }
```

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+        let name = item.name();
+        let key = name.text();
+
+        if !RESERVED_KEYS.contains(&key) {
+            continue;
+        }
+
+        let value = item.value();
+
+        if !is_string_value(&value) {
```
acfoltzer: Do we need to consider nested objects here like we do in the `metadata_section()` visitor method?

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+
+            // Check if the value is a string
+            if !is_string_value(&value) {
+                let value_type = get_value_type_name(&value);
+                diagnostics.exceptable_add(
+                    non_string_value_diagnostic(key, value_type, item.span()),
+                    SyntaxElement::from(item.inner().clone()),
+                    &self.exceptable_nodes(),
+                );
+            }
```
acfoltzer: Similar to the above, this is much easier to follow if we avoid negations and combine the conditions we're checking. Don't directly apply this suggestion without looking at the other comment about using `check_object_items()`, though.

```suggestion
            if key == "outputs"
                && let MetadataValue::Object(ref obj) = value
            {
                for output_item in obj.items() {
                    let output_name = output_item.name();
                    let output_key = output_name.text();
                    let output_value = output_item.value();

                    // Check if the key itself is reserved
                    if RESERVED_KEYS.contains(&output_key) && !is_string_value(&output_value) {
                        let value_type = get_value_type_name(&output_value);
                        diagnostics.exceptable_add(
                            non_string_value_diagnostic(output_key, value_type, output_item.span()),
                            SyntaxElement::from(output_item.inner().clone()),
                            &self.exceptable_nodes(),
                        );
                    }

                    // Check nested objects
                    if let MetadataValue::Object(ref nested_obj) = output_value {
                        check_object_items(nested_obj, diagnostics, &self.exceptable_nodes());
                    }
                }
            } else if RESERVED_KEYS.contains(&key) && !is_string_value(&value) {
                let value_type = get_value_type_name(&value);
                diagnostics.exceptable_add(
                    non_string_value_diagnostic(key, value_type, item.span()),
                    SyntaxElement::from(item.inner().clone()),
                    &self.exceptable_nodes(),
                );
            }
```

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+                                SyntaxElement::from(output_item.inner().clone()),
+                                &self.exceptable_nodes(),
+                            );
+                        }
+
+                        // Check nested objects
+                        if let MetadataValue::Object(ref nested_obj) = output_value {
+                            check_object_items(nested_obj, diagnostics, &self.exceptable_nodes());
+                        }
+                    }
```
acfoltzer: If you add the nested object check to `check_object_items()`, this whole block can be replaced with a call to `check_object_items()`.

crates/wdl-lint/src/rules/expected_meta_string.rs
```diff
@@ -0,0 +1,251 @@
...
+                MetadataValue::Object(obj) => {
+                    check_object_items(&obj, diagnostics, &self.exceptable_nodes());
+                }
+
+                // Any other type - check if it would be a reserved key as a simple description
+                // This handles cases like: parameter_name: 123 (instead of parameter_name:
+                // "description")
+                _ => {
+                    let name = item.name();
+                    let _param_name = name.text();
```
acfoltzer: It looks like an incomplete change was made here. The comment describes checking for a reserved key, and the `_param_name` binding is not used.

crates/wdl-lint/src/rules/expected_meta_string.rs
Krriishhhhhh: This one is resolved , right?

crates/wdl-lint/src/rules/expected_meta_string.rs
Krriishhhhhh: Alright , @a-frantz can you also confirm on this one

crates/wdl-lint/src/rules/expected_meta_string.rs
acfoltzer: Looks like it to me 👍🏻

crates/wdl-lint/src/rules/expected_meta_string.rs
acfoltzer: I believe this has been resolved.

crates/wdl-lint/src/rules/expected_meta_string.rs
acfoltzer: Ari is out of the office for a little bit, so I think we should just go ahead with `DocMetaStrings` if you agree that name works.

a-frantz: I'm happy with `DocMetaStrings`!

Krriishhhhhh: Okyy

Krriishhhhhh:
@acfoltzer changed the name and i guess everything is resolved now ! :)

crates/wdl-lint/RULES.md
```diff
@@ -21,7 +21,7 @@ This table documents all `sprocket` lint rules implemented on the `main` branch
...
 | `EndingNewline`             | Spacing, Portability                                        | Ensures that documents end with a single newline character.                                                                                |
 | `ExpectedRuntimeKeys`       | Completeness, Deprecated                                    | Ensures that `runtime` sections have the appropriate keys.                                                                                 |
-| `ExpectedMetaString`        | SprocketCompatibility                                       | Ensures that reserved meta keys have string values for wdl-doc compatibility. |
+| `DocMetaStrings`        | SprocketCompatibility                                       | Ensures that reserved meta keys have string values for wdl-doc compatibility. |
 | `ExpressionSpacing`         | Spacing, Style                                              | Ensures that WDL expressions are properly spaced.                                                                                          |
```
acfoltzer, line 25:
```suggestion
| `ExpectedRuntimeKeys`       | Completeness, Deprecated                                    | Ensures that `runtime` sections have the appropriate keys.                                                                                 |
| `DocMetaStrings`            | SprocketCompatibility                                       | Ensures that reserved meta keys have string values for wdl-doc compatibility.                                                              |
| `ExpressionSpacing`         | Spacing, Style                                              | Ensures that WDL expressions are properly spaced.                                                                                          |
```

acfoltzer (changes requested):
Everything is looking good, but when I went through the checklist I noticed that we're missing changelog entries. Since this is a branch in your fork, I cannot add them myself, so could you please add appropriate entries to `CHANGELOG.md` and `crates/wdl-lint/CHANGELOG.md`?

acfoltzer (approved):
Thanks very much, @Krriishhhhhh!

PR #410, opened 2025-10-09
fix: handle rename for shadowed variables

dead8309:
closes: stjude-rust-labs/sprocket-vscode/issues/40

Rename depends on `find_all_references` which depends on `goto_definition` to correctly identify the symbol that needs to be renamed. The issue was in how `goto_definition` handled shadowed variables.
The issue(in this case) was, it relied entirely on scope lookup as there were no checks present for `BoundDeclNode`.

now the handler checks if the token under the cursor is part of a declaration node (Bound/Unbound) and if it is, the definition is considered to be the declaration itself, and it's own location is returned

For external contributors:

For all contributors:

### DIFF
--- a/crates/wdl-analysis/src/handlers/goto_definition.rs
+++ b/crates/wdl-analysis/src/handlers/goto_definition.rs
@@ -163,11 +163,11 @@ fn resolve_by_context(
         }
 
         SyntaxKind::UnboundDeclNode => {
-            resolve_unbound_decl_definition(parent_node, token, document_uri, lines)
+            resolve_decl_definition::<v1::UnboundDecl>(parent_node, token, document_uri, lines)
         }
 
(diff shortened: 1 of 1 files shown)


crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -594,8 +598,26 @@ fn resolve_access_expression(
     Ok(None)
 }
 
-/// Resolve struct member definitions to themselves.
-fn resolve_struct_member_definition(
+/// Resolve bound declarations to themselves.
+fn resolve_bound_decl_definition(
```
claymcleod: Instead of splitting this into two functions, I'd just make one function with a generic of the AST node to cast to.

PR #411, opened 2025-10-09
feat: adds `else if`/`else` clauses to conditions in WDL v1.3

claymcleod:
This PR introduces `else if`/`else` clauses to the codebase. It accomplishes this by refactoring much of the logic of how conditionals are handled across the crates, namely:

- Conditionals now may have multiple clauses.
- Conditionals now execute and promote variables according to the logic laid out in https://github.com/openwdl/wdl/pull/699. Note that this is intended to be backwards compatible with the way conditionals used to promote variables to the parent scope.

In pursuit of a testable version of the code, WDL v1.3 was added as a valid supported version, with the caveat that the `experimental_versions` feature flag must be set to `true` to run WDL v1.3.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -469,6 +469,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
 
 [[package]]
+name = "bytesize"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f5c434ae3cf0089ca203e9019ebe529c47ff45cefe8af7c85ecb734ef541822f"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
 name = "cc"
 version = "1.2.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1127,6 +1136,12 @@ checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
 dependencies = [
  "cfg-if",
 ]
+
+[[package]]
+name = "env_home"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c7f84e12ccf0a7ddc17a6c41c93326024c42920d7ee630d04950e6926645c0fe"
 
 [[package]]
 name = "equivalent"
@@ -5076,6 +5091,7 @@ dependencies = [
  "approx",
  "bimap",
  "blake3",
+ "bytesize",
  "chrono",
  "cloud-copy",
  "codespan-reporting",
@@ -5114,6 +5130,7 @@ dependencies = [
  "wdl-analysis",
  "wdl-ast",
  "wdl-grammar",
+ "which",
 ]
 
 [[package]]
@@ -5233,6 +5250,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32b130c0d2d49f8b6889abc456e795e82525204f27c42cf767cf0d7734e089b8"
 dependencies = [
  "rustls-pki-types",
+]
+
+[[package]]
+name = "which"
+version = "8.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3fabb953106c3c8eea8306e4393700d7657561cb43122571b172bbfb7c7ba1d"
+dependencies = [
+ "env_home",
+ "rustix",
+ "winsafe",
 ]
 
 [[package]]
@@ -5609,6 +5637,12 @@ checksum = "21a0236b59786fed61e2a80582dd500fe61f18b5dca67a4a067d0bc9039339cf"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "winsafe"
+version = "0.0.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d135d17ab770252ad95e9a872d365cf3090e3be864a34ab46f48555993efc904"
 
 [[package]]
 name = "wit-bindgen"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ anyhow = "1.0.98"
 approx = "0.5.1"
 bimap = "0.6.3"
 blake3 = { version = "1.8.2", features = ["mmap"] }
+bytesize = { version = "2.1.0", features = ["serde"] }
 chrono = "0.4.41"
 clap = { version = "4.5.45", features = ["derive", "string"] }
 clap-verbosity-flag = { version = "3.0.3", features = ["tracing"] }
@@ -110,6 +111,7 @@ urlencoding = "2.1.3"
 uuid = "1.18.0"
 walkdir = "2.5.0"
 wdl = { version = "0.19.0", path = "crates/wdl", features = ["cli", "codespan", "engine", "lsp", "doc"] }
+which = "8.0.0"
 
 [workspace.lints.rust]
 missing_docs = "warn"
--- a/crates/wdl-engine/CHANGELOG.md
+++ b/crates/wdl-engine/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Added support for `else if` and `else` clauses in conditional statements (in support of WDL v1.3) ([#411](https://github.com/stjude-rust-labs/sprocket/pull/411)).
 
+#### Changed
+
+* Added optional CPU and memory limits to the queue definitions in the LSF + Apptainer backend configuration. This is a breaking change for previous LSF configurations, as the queues are now a struct with a required `name` string field, rather than just a bare string ([#429](https://github.com/stjude-rust-labs/sprocket/pull/429)).
+
 ## 0.9.0 - 10-14-2025
 
 #### Added
--- a/crates/wdl-engine/Cargo.toml
+++ b/crates/wdl-engine/Cargo.toml
@@ -19,6 +19,7 @@ codespan-reporting = ["codespan", "dep:codespan-reporting"]
 anyhow = { workspace = true }
 bimap = { workspace = true }
 blake3 = { workspace = true }
+bytesize = { workspace = true }
 chrono = { workspace = true }
 cloud-copy = { workspace = true }
 codespan-reporting = { workspace = true, optional = true }
@@ -61,6 +62,7 @@ pretty_assertions = { workspace = true }
 toml = { workspace = true }
 tracing-subscriber = { workspace = true }
 walkdir = { workspace = true }
+which = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/wdl-engine/src/backend.rs
+++ b/crates/wdl-engine/src/backend.rs
@@ -28,11 +28,13 @@ use crate::Value;
 use crate::http::Transferer;
 use crate::path::EvaluationPath;
 
+mod apptainer;
 mod docker;
 mod local;
 mod lsf_apptainer;
 mod tes;
 
+pub use apptainer::*;
 pub use docker::*;
 pub use local::*;
 pub use lsf_apptainer::*;
@@ -234,6 +236,29 @@ impl TaskSpawnRequest {
     /// The temp directory for the evaluation.
     pub fn temp_dir(&self) -> &Path {
         &self.temp_dir
+    }
+
+    /// The default host-side location of the script generated from the task
+    /// `command`.
+    pub fn wdl_command_host_path(&self) -> PathBuf {
+        self.attempt_dir.join(COMMAND_FILE_NAME)
+    }
+
+    /// The default host-side location of the task's working directory.
+    pub fn wdl_work_dir_host_path(&self) -> PathBuf {
+        self.attempt_dir.join(WORK_DIR_NAME)
+    }
+
+    /// The default host-side location where the `command`'s stdout will be
+    /// written.
+    pub fn wdl_stdout_host_path(&self) -> PathBuf {
+        self.attempt_dir.join(STDOUT_FILE_NAME)
+    }
+
+    /// The default host-side location where the `command`'s stderr will be
+    /// written.
+    pub fn wdl_stderr_host_path(&self) -> PathBuf {
+        self.attempt_dir.join(STDERR_FILE_NAME)
     }
 }
 
--- a/crates/wdl-engine/src/backend/apptainer.rs
+++ b/crates/wdl-engine/src/backend/apptainer.rs
@@ -1,0 +1,290 @@
+//! Support for using Apptainer (a.k.a. Singularity) as an in-place container
+//! runtime for task execution.
+//!
+//! There are two primary responsibilities of this module: `.sif` image cache
+//! management and command script generation. The entrypoint for both of these
+//! is [`ApptainerConfig::prepare_apptainer_command()`].
+
+use std::fmt::Write as _;
+use std::path::Path;
+use std::path::PathBuf;
+
+use anyhow::Context as _;
+use anyhow::anyhow;
+use images::sif_for_container;
+use tokio_util::sync::CancellationToken;
+
+use super::TaskSpawnRequest;
+use crate::Value;
+
+mod images;
+
+/// The guest working directory.
+const GUEST_WORK_DIR: &str = "/mnt/task/work";
+
+/// The guest path for the command file.
+const GUEST_COMMAND_PATH: &str = "/mnt/task/command";
+
+/// The path to the container's stdout.
+const GUEST_STDOUT_PATH: &str = "/mnt/task/stdout";
+
+/// The path to the container's stderr.
+const GUEST_STDERR_PATH: &str = "/mnt/task/stderr";
+
+/// Configuration for the Apptainer container runtime.
+#[derive(Debug, Clone, serde::Deserialize, serde::Serialize)]
+pub struct ApptainerConfig {
+    /// Additional command-line arguments to pass to `apptainer exec` when
+    /// executing tasks.
+    pub extra_apptainer_exec_args: Option<Vec<String>>,
+    /// The directory in which temporary directories will be created containing
+    /// Apptainer `.sif` files.
+    ///
+    /// This should be a location that is accessible by all jobs on the LSF
+    /// cluster.
+    ///
+    /// By default, this is `$HOME/.cache/sprocket-apptainer-images`, or
+    /// `/tmp/sprocket-apptainer-images` if the home directory cannot be
+    /// determined.
+    #[serde(default = "default_apptainer_images_dir")]
+    pub apptainer_images_dir: PathBuf,
+}
+
+fn default_apptainer_images_dir() -> PathBuf {
+    if let Some(cache) = dirs::cache_dir() {
+        cache.join("sprocket-apptainer-images")
+    } else {
+        std::env::temp_dir().join("sprocket-apptainer-images")
+    }
+}
+
+impl Default for ApptainerConfig {
+    fn default() -> Self {
+        Self {
+            extra_apptainer_exec_args: None,
+            apptainer_images_dir: default_apptainer_images_dir(),
+        }
+    }
+}
+
+impl ApptainerConfig {
+    /// Prepare for an Apptainer execution by ensuring the image cache is
+    /// populated with the necessary container, and return a Bash script
+    /// that invokes the task's `command` in the container context.
+    ///
+    /// # Shared filesystem assumptions
+    ///
+    /// The returned script should be run in an environment that shares a
+    /// filesystem with the environment where this method is invoked, except
+    /// for node-specific mounts like `/tmp` and `/var`. This assumption
+    /// typically holds on HPC systems with shared filesystems like Lustre or
+    /// GPFS.
+    pub async fn prepare_apptainer_command(
+        &self,
+        container: &str,
+        cancellation_token: CancellationToken,
+        spawn_request: &TaskSpawnRequest,
+    ) -> Result<String, anyhow::Error> {
+        let container_sif = sif_for_container(self, container, cancellation_token).await?;
+        self.generate_apptainer_script(&container_sif, spawn_request)
+            .await
+    }
+
+    /// Generate the script, given a container path that's already assumed to be
+    /// populated.
+    ///
+    /// This is a separate method in order to facilitate testing, and should not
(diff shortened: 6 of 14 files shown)


crates/wdl-analysis/src/types.rs
```diff
@@ -308,6 +308,13 @@ impl Type {
             return Some(Self::Compound(ty, self.is_optional()));
         }
 
+        // Check for a call type to have a common type with itself
```
claymcleod, line 302: This seems strange out of context, but it's required for getting the common type between calls. Peter and I discussed this in Slack.

crates/wdl-analysis/src/eval/v1.rs
```diff
@@ -835,7 +916,18 @@ impl<N: TreeNode> WorkflowGraphBuilder<N> {
                     }
                 }
                 WorkflowGraphNode::Conditional(statement, _) => {
-                    self.add_expr_edges(from, statement.expr(), graph, diagnostics);
+                    for clause in statement.clauses() {
```
claymcleod, line 912: Note to self: is this right?

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1330,8 +1421,105 @@ fn resolve_call_type(
     }
 }
 
-/// Promotes the names in the current to the parent scope.
-fn promote_scope(scopes: &mut [Scope], index: ScopeIndex, skip: Option<&str>, kind: PromotionKind) {
+/// A scope union.
+///
+/// A scope union takes the union of names within a number of given scopes and
+/// computes a set of common, output names for a (presumed parent) scope. This
```
peterhuene:
```suggestion
/// computes a set of common output names for a (presumed parent) scope. This
```

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1330,8 +1421,105 @@ fn resolve_call_type(
...
+    /// Adds a scope to the union.
+    pub fn insert(&mut self, scope_index: ScopeIndex, exhaustive: bool) {
+        self.scope_indices.push((scope_index, exhaustive));
+    }
+
+    /// Resolves the scope union to names and types that should be accessible
+    /// from the parent scope.
+    ///
+    /// Returns an error if any issues are encountered during resolving.
+    pub fn resolve(self) -> Result<Vec<(String, Name)>, Vec<Diagnostic>> {
```
peterhuene: As we're basically just iterating on the returned value anyway, could we save collecting into a `Vec` by simply returning the `HashMap` we're building internally instead?

crates/wdl-analysis/src/eval/v1.rs
peterhuene: I think this duplicates the work being done with the`WorkflowGraphNode::ConditionalClause` case.

My recommendation would be to remove the `WorkflowGraphNode::ConditionalClause` case.

crates/wdl-analysis/src/eval/v1.rs
```diff
@@ -835,7 +912,18 @@ impl<N: TreeNode> WorkflowGraphBuilder<N> {
...
+                        let Some(expr) = clause.expr() else { continue };
+                        self.add_expr_edges(from, expr, graph, diagnostics);
+                    }
+                }
+                WorkflowGraphNode::ConditionalClause(statement, idx, _) => {
+                    // Only add edges for the expression of this specific clause
+                    if let Some(clause) = statement.clauses().nth(idx)
```
peterhuene: As calling `nth` on `clauses` is O(n), I think we should probably eliminate the `WorkflowGraphNode::ConditionalClause` case in favor of the `WorkflowGraphNode::Conditional` case.

crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -129,16 +129,32 @@ impl From<NameContext> for Context {
 }
 
 /// Creates a "name conflict" diagnostic.
-pub fn name_conflict(name: &str, conflicting: Context, first: Context) -> Diagnostic {
-    Diagnostic::error(format!("conflicting {conflicting} name `{name}`"))
+pub fn name_conflict(
+    name: &str,
+    conflicting: Context,
+    first: Context,
+    others: Option<Vec<Context>>,
```
peterhuene: I couldn't find where we're passing in `Some` for `others` to this function.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1330,8 +1421,105 @@ fn resolve_call_type(
...
+        if !errors.is_empty() {
+            return Err(errors);
+        }
+
+        Ok(names.into_iter().collect())
+    }
+}
+
+/// Promotes names from a scatter scope to the parent scope.
+fn promote_scope(scopes: &mut [Scope], index: ScopeIndex, skip: Option<&str>) {
```
peterhuene:
```suggestion
fn promote_scatter_scope(scopes: &mut [Scope], index: ScopeIndex, skip: Option<&str>) {
```

Given that this is now called only in one place, I'd be fine with inlining it too (provided the calling function isn't already too verbose).

crates/wdl-analysis/src/diagnostics.rs
claymcleod: Good catch, I ended up not needing this.

crates/wdl-analysis/src/document/v1.rs
claymcleod: I've decided to inline it.

crates/wdl-ast/src/v1/workflow.rs
```diff
@@ -722,19 +723,132 @@ impl<N: TreeNode> WorkflowStatement<N> {
     }
 }
 
+/// A kind of conditional statement clause.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
peterhuene: Might as well throw a `Hash` on here for anyone that might need it (since it's a public type).

```suggestion
#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
```

crates/wdl-ast/src/v1/workflow.rs
```diff
@@ -722,19 +723,132 @@ impl<N: TreeNode> WorkflowStatement<N> {
...
+    /// Gets all of the clauses of a conditional statement.
+    pub fn clauses(&self) -> impl Iterator<Item = ConditionalStatementClause<N>> {
+        ConditionalStatementClause::children(&self.0)
     }
 
-    /// Gets the statements of the conditional body.
-    pub fn statements(&self) -> impl Iterator<Item = WorkflowStatement<N>> + use<'_, N> {
-        WorkflowStatement::children(&self.0)
+    /// Gets the initial `if` clause.
+    pub fn r#if(&self) -> ConditionalStatementClause<N> {
```
peterhuene: Nit: I personally try to avoid using raw keywords in (especially public) interfaces.

Could we use `if_clause` instead?

crates/wdl-ast/src/v1/workflow.rs
```diff
@@ -722,19 +723,132 @@ impl<N: TreeNode> WorkflowStatement<N> {
...
+        self.clauses()
+            .find(|clause| clause.kind() == ConditionalStatementClauseKind::If)
+            .expect("missing required conditional statement `if` clause")
+    }
+
+    /// Gets the `else if` clauses, if any exist in the conditional statement.
+    pub fn else_if(&self) -> impl Iterator<Item = ConditionalStatementClause<N>> {
```
peterhuene: nit: if we go with `if_clause`, we should probably rename this to `else_if_clauses`.

crates/wdl-ast/src/v1/workflow.rs
```diff
@@ -722,19 +723,132 @@ impl<N: TreeNode> WorkflowStatement<N> {
...
+        self.clauses()
+            .filter(|clause| clause.kind() == ConditionalStatementClauseKind::ElseIf)
+    }
+
+    /// Gets the final `else` clause, if it exists.
+    pub fn r#else(&self) -> Option<ConditionalStatementClause<N>> {
```
peterhuene: Nit: same as above, suggested name of `else_clause`.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -861,6 +861,13 @@ impl TaskEvaluator {
             return Err(anyhow!("cannot evaluate a document with errors").into());
         }
 
+        // WDL v1.3 requires experimental features to be enabled
+        if let Some(SupportedVersion::V1(V1::Three)) = document.version()
+            && !self.config.experimental_features_enabled
+        {
+            return Err(anyhow!("WDL v1.3 requires enabling experimental features").into());
+        }
```
peterhuene: This check feels like it should be in `crates/wdl-analysis/src/validation/version.rs`. Checking it at task evaluation time is too late, in my opinion.

crates/wdl-engine/src/eval.rs
```diff
@@ -440,8 +440,9 @@ impl Scope {
 
     /// Inserts a name into the scope.
     pub fn insert(&mut self, name: impl Into<String>, value: impl Into<Value>) {
-        let prev = self.names.insert(name.into(), value.into());
-        assert!(prev.is_none(), "conflicting name in scope");
+        let name = name.into();
```
peterhuene: I'm not sure if we want to clone this for an assertion message that should never occur (although I see the value of it when you were developing the feature).

At the very least, we could restrict the clone to a `cfg(debug_assertions)` perhaps.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -495,8 +495,66 @@ impl Subgraph {
                 }
 
                 match &graph[index] {
-                    WorkflowGraphNode::Conditional(_, exit)
-                    | WorkflowGraphNode::Scatter(_, exit) => {
+                    WorkflowGraphNode::Conditional(_, exit) => {
```
peterhuene: Would it be easier to just split for a `WorkflowGraphNode::ConditionalClause` variant rather than at the conditional node?

The conditional node itself no longer needs to be a subgraph.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -619,6 +677,13 @@ impl WorkflowEvaluator {
             return Err(anyhow!("cannot evaluate a document with errors").into());
         }
 
+        // WDL v1.3 requires experimental features to be enabled
+        if let Some(SupportedVersion::V1(V1::Three)) = document.version()
+            && !self.config.experimental_features_enabled
+        {
+            return Err(anyhow!("WDL v1.3 requires enabling experimental features").into());
+        }
```
peterhuene: Same comment as above, this belongs in static analysis.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -850,14 +915,10 @@ impl WorkflowEvaluator {
                             "evaluation of call statement has completed",
                         )
                     }
-                    WorkflowGraphNode::Conditional(stmt, _) => debug!(
+                    WorkflowGraphNode::Conditional(..) => debug!(
```
peterhuene, line 867: Could this be `ConditionalClause`? It would be good to know which clause completed.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -1230,97 +1293,260 @@ impl WorkflowEvaluator {
...
+                let target = edge.target();
+                match &state.graph[target] {
+                    WorkflowGraphNode::ConditionalClause(_, idx, _) => Some((*idx, target)),
+                    _ => None,
+                }
+            })
+            .collect();
+
+        // Sort by clause index to ensure they're evaluated in order
+        clause_nodes.sort_by_key(|(idx, _)| *idx);
```
peterhuene: I feel like we could be utilizing the AST here rather than building up a list and sorting it.

We first look at `stmt.if_clause()`, then each `stmt.else_clauses()`, then the `stmt.else_clause()` as needed.

For the first if/else if with conditional expression evaluated to true (or none evaluated to true and there's an else), we effectively do what we did before: alloc a scope, evaluate the subgraph associated with the clause node, and then promote the scope to the parent (but now by unioning the names and common types).

Otherwise, if none of the clauses evaluated to true and there's no else clause, we promote the union of names and common types with `None` values to the parent scope.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -1230,97 +1293,260 @@ impl WorkflowEvaluator {
...
+                        let mut all_names: HashMap<String, Type> = HashMap::new();
+                        for clause in stmt.clauses() {
+                            if let Some(braced_scope_span) = clause.braced_scope_span() {
+                                let clause_scope = state
+                                    .document
+                                    .find_scope_by_position(braced_scope_span.start())
+                                    .expect("should have scope");
+
+                                for (name, n) in clause_scope.names() {
+                                    all_names.insert(name.to_string(), n.ty().clone());
```
peterhuene: I think we'll need to determine the common type here as well; we should be able to do so with an expectation rather than emit a diagnostic, since evaluation assumes (and verifies) that there are no error diagnostics from static analysis on what's being evaluated.

Could we perhaps refactor and reuse `ScopeUnion` here such that it works by storing `ScopeRef` rather than the scopes slice and working on indexes?

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -1230,97 +1293,260 @@ impl WorkflowEvaluator {
...
+                                                },
+                                            ))
+                                            .into(),
+                                        ),
+                                    );
+                                } else {
+                                    parent.insert(name, Value::new_none(ty.optional()));
+                                }
+                            }
+                        }
```
peterhuene: Instead of looping twice and constructing a `HashSet`, could we just loop on the names and types from the resolved "scope union" and check to see if the name is present in the evaluation scope? If so, use the value, otherwise use a `None` value?

peterhuene (commented):
Review comments from `wdl-ast` and `wdl-engine`.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1675,3 +1855,130 @@ fn type_check_expr(
...
+    fn smoke() {
+        let scopes = example_scopes();
+
+        // Test with else clause (exhaustive)
+        let mut scope_union = ScopeUnion::new(&scopes);
+        scope_union.insert(ScopeIndex(0), false);
+        scope_union.insert(ScopeIndex(1), false);
+        scope_union.insert(ScopeIndex(2), true);
+
+        let mut results = scope_union.resolve().expect("should resolve").into_iter().collect::<Vec<_>>();
```
peterhuene: Nit: now that we're returning the map we could just `results["a"]` etc instead of collecting it into a `Vec` and doing the `find` calls below.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -1751,8 +1751,10 @@ fn bound_decl(
     Ok(())
 }
 
-/// Parses a conditional statement in a workflow.
-fn conditional_statement(
+/// Parses a conditional statement clause.
```
peterhuene:
```suggestion
/// Parses a conditional statement `if` clause.
```

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -1768,6 +1770,44 @@ fn conditional_statement(
         WORKFLOW_STATEMENT_RECOVERY_SET,
         workflow_statement
     );
+    marker.complete(parser, SyntaxKind::ConditionalStatementClauseNode);
+    Ok(())
+}
+
+/// Parses both `else if` and `else` clause in a conditional statement.
+///
+/// Returns whether we should continue looping over conditional clauses.
```
peterhuene:
```suggestion
```

peterhuene (commented):
Remaining review comments.

crates/wdl-engine/src/eval/v1/workflow.rs
claymcleod: I'll do you one better: we just need to iterate through `stmt.clauses()`, as they're already in order there.

crates/wdl-analysis/src/config.rs
```diff
@@ -151,6 +157,16 @@ impl Config {
...
         }
     }
+
+    /// Return a new configuration with the previous [`FeatureFlags`]
+    /// replaced by the argument.
+    pub fn with_feature_flags(&self, feature_flags: FeatureFlags) -> Self {
+        let mut inner = (*self.inner).clone();
+        inner.feature_flags = feature_flags;
+        Self {
+            inner: Arc::new(inner),
```
peterhuene, line 167: For a future PR, but this `Config` is begging for a proper builder abstraction to prevent these repeated `Arc::new` calls (at the very least, perhaps a `Arc::make_mut`).

crates/wdl-analysis/src/document.rs
```diff
@@ -196,7 +200,7 @@ pub struct ScopeRef<'a> {
 
 impl<'a> ScopeRef<'a> {
     /// Creates a new scope reference given the scope index.
-    fn new(scopes: &'a [Scope], index: ScopeIndex) -> Self {
+    pub fn new(scopes: &'a [Scope], index: ScopeIndex) -> Self {
```
peterhuene: Does this actually need to be public? Other crates shouldn't be creating a `ScopeRef`, but getting one via introspection of a `Document`.

crates/wdl-analysis/src/document.rs
```diff
@@ -154,11 +158,11 @@ impl Name {
 
 /// Represents an index of a scope in a collection of scopes.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-struct ScopeIndex(usize);
+pub struct ScopeIndex(pub usize);
```
peterhuene: Likewise, I don't think this type needs to be public (provided `ScopeRef::new` doesn't need to be public).

crates/wdl-analysis/CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+#### Added
+
+* Added support for `else if` and `else` clauses in conditional statements (in support of WDL v1.3) ([#411](https://github.com/stjude-rust-labs/sprocket/pull/411)).
+* Added feature flags support to enable experimental WDL versions ([#411](https://github.com/stjude-rust-labs/sprocket/pull/411)).
+
+#### Changed
+
+* Refactored `ScopeUnion` to use `ScopeRef` instead of direct slice indexing, allowing it to be reused in the runtime engine for proper type reconciliation across conditional branches ([#411](https://github.com/stjude-rust-labs/sprocket/pull/411)).
+* Made `Scope`, `ScopeIndex`, and `ScopeRef::new()` public to support cross-crate usage ([#411](https://github.com/stjude-rust-labs/sprocket/pull/411)).
```
peterhuene: Reminder to remove this line if we make them private again.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -1230,97 +1260,257 @@ impl WorkflowEvaluator {
         stmt: &ConditionalStatement<SyntaxNode>,
         max_concurrency: u64,
     ) -> EvaluationResult<()> {
-        let expr = stmt.expr();
+        // Build a map from clause inner node to graph node index
+        let clause_map: HashMap<_, _> = state
```
peterhuene: I think we could probably do without this map at all and, upon the evaluation of a clause, do a search for its corresponding graph node.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -1230,97 +1260,257 @@ impl WorkflowEvaluator {
...
+                    _ => None,
+                }
+            })
+            .collect();
+
+        for clause in stmt.clauses() {
+            let clause_node = clause_map[clause.inner()];
+            let kind = clause.kind();
+
+            match kind {
```
peterhuene: I feel like this match could be made more precise by instead checking to see if the clause has an expression and if it evaluates to false, continue. Otherwise, evaluate the clause statements, promote scope values to the parent, and return.

Then if we fall through the loop, we promote `None` values instead.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -1230,97 +1260,257 @@ impl WorkflowEvaluator {
...
+                    .find_scope_by_position(braced_scope_span.start())
+                    .expect("should have scope");
 
-            // Promote all values in the scope to the parent scope as optional
-            let mut scopes = state.scopes.write().await;
-            let (parent, child) = scopes.parent_mut(scope);
-            for (name, value) in child.local() {
-                parent.insert(name.to_string(), value.clone());
+                for (name, n) in scope.names() {
+                    all_names.insert(name.to_string(), n.ty().clone());
```
peterhuene: This is also not merging the common type, so we'll probably want to make use of `ScopeUnion` here too.

peterhuene (commented):
A few more comments to go through.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -495,8 +496,31 @@ impl Subgraph {
...
+                                    &graph[target],
+                                    WorkflowGraphNode::ConditionalClause(..)
+                                ) {
+                                    nodes.remove(&target);
+                                    Some(target)
+                                } else {
+                                    None
+                                }
+                            })
+                            .collect();
```
peterhuene: Question: do we need to build this list? Could we not just process each `WorkflowGraphNode::ConditionalClause` (as it also stores the exit node) instead of `WorkflowGraphNode::Conditional`?

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -1230,98 +1260,172 @@ impl WorkflowEvaluator {
...
+                    match &state.graph[target] {
+                        WorkflowGraphNode::ConditionalClause(c, _)
+                            if c.inner() == clause.inner() =>
+                        {
+                            Some(target)
+                        }
+                        _ => None,
+                    }
+                })
+                .expect("clause node should exist in graph");
```
peterhuene: I think this search can be moved down below right before the comment `If we reach here, this clause should be executed`; we don't need the clause's graph node unless we're about to evaluate the subgraph.

peterhuene (commented):
Last two comments, but otherwise gtg!

peterhuene (approved):
Looks great! Thanks for sticking through all the review comments as it's a pretty complex feature in terms of the changes to evaluation.

Issue #395, opened 2025-10-01
Sprocket: periodic progress logging would be nice to have


peterhuene:
When redirecting the output of `sprocket` to a file, we intentionally hide the progress bars as to not draw them repeatedly into the file.

However, this means that there is no indication of the current progress in the file.

Suggestion: when redirected, respond to a timer event by periodically logging the same information that the progress bar would have displayed.

PR #412, opened 2025-10-09
added simple progress change printing if the output is redirected

dmgorsky:
open draft proposal to #395


src/commands/run.rs
```diff
@@ -332,6 +343,7 @@ async fn progress(mut events: broadcast::Receiver<Event>, pb: tracing::Span) {
                 };
 
                 pb.pb_set_message(&message);
+                print_state_change(&message);
```
peterhuene: Unfortunately the message here is one that only makes sense to print as a progress bar.

I think we'll want to format a more concise message that conveys roughly the same statistics.

Cargo.toml
```diff
@@ -144,6 +144,7 @@ readme = "README.md"
 
 [dependencies]
 anyhow.workspace = true
+atty = "0.2.14"
```
peterhuene: We should pull this dependency from the workspace like the other dependencies of this crate.

peterhuene (changes requested):
Hi @dmgorsky. Thanks for the PR!

There's just a few things I think need addressing before we can move forward with it.

src/commands/run.rs
dmgorsky: You are right, the log message can be whatever interests the user, I just logged the same screen output as an example of diagnostics event - to keep my proposal concise. And also to avoid duplicating of messages construction for the screen and for file output - to not have to maintain both with the similar logic. In case these functions are in different parts of code, it could be hard to update both - if, say, we change the struct. But OK, I'll look into 1) having the function formatting the log message; 2) having it somehow bound or at least closer to original progress.message()

Cargo.toml
dmgorsky: In theory, as I checked, it is needed only in this single crate of workspace, you think this should be expanded?

peterhuene: We keep the versioned dependencies at the workspace level so that they're all in one place. We prefer that each crate sources their dependencies from the workspace, even if the dependency is only used in a single crate.

src/commands/run.rs
dmgorsky: changed to
```
[2025-10-16 15:59:14] running workflow: hello_world
[2025-10-16 15:59:14] tasks: 0 completed, 1 ready, 0 executing
[2025-10-16 15:59:14] tasks: 0 completed, 0 ready, 1 executing: hello-K7ao3jimQ8JR
```

PR #424, opened 2025-10-15
feat: introduces the `split` standard library function

claymcleod:
This PR introduces the `split` standard library function proposed in https://github.com/openwdl/wdl/issues/553.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/tests/tasks.rs
+++ b/crates/wdl-engine/tests/tasks.rs
@@ -273,7 +273,15 @@ fn compare_result(path: &Path, result: &str) -> Result<()> {
 
 /// Runs a single test.
 async fn run_test(test: &Path, config: config::Config) -> Result<()> {
-    let analyzer = Analyzer::default();
+    use wdl_analysis::Config as AnalysisConfig;
+    use wdl_analysis::FeatureFlags;
+
+    let analyzer = Analyzer::new(
+        AnalysisConfig::default()
+            .with_feature_flags(FeatureFlags::default().with_experimental_versions()),
+        |(), _, _, _| async {},
+    );
+
     analyzer
         .add_directory(test)
         .await


claymcleod:
Adding @peterhuene as well since I _think_ this is the first stdlib function someone else has added (want to make sure I did all the pieces needed).

crates/wdl-engine/src/stdlib/split.rs
```diff
@@ -0,0 +1,126 @@
...
+    debug_assert!(context.return_type_eq(ANALYSIS_STDLIB.array_string_type().clone()));
+
+    let input = context
+        .coerce_argument(0, PrimitiveType::String)
+        .unwrap_string();
+    let delimiter = context
+        .coerce_argument(1, PrimitiveType::String)
+        .unwrap_string();
+
+    let regex = Regex::new(delimiter.as_str())
```
acfoltzer, line 36: If we're using the `regex` crate to implement this, we should note that we're not supporting the full POSIX ERE (which I think is way overkill for this type of function and probably fine in practice)

> The regex syntax supported by this crate is similar to other regex engines, but it lacks several features that are not known how to implement efficiently. This includes, but is not limited to, look-around and backreferences.

crates/wdl-analysis/src/stdlib.rs
```diff
@@ -2094,6 +2094,55 @@ workflow test_sub {
...
+```wdl
+version 1.3
+
+workflow test_split {
+  String in = "Here's an example\nthat takes up multiple lines"
+
+  output {
+    String split_by_word = sub(in, " ")
+    String split_by_newline = sub(in, "\\n")
+    String split_by_both = sub(in, "\s")
```
peterhuene:
```suggestion
    Array[String] split_by_word = split(in, " ")
    Array[String] split_by_newline = split(in, "\\n")
    Array[String] split_by_both = split(in, "\s")
```

peterhuene (changes requested):
Just the one comment issue. Otherwise, looks good 👍

crates/wdl-engine/src/stdlib/split.rs
peterhuene: Yeah, it's an unfortunate limitation of the `regex` crate, but I doubt it'll impact many users. We should probably add a notice to `sprocket.bio` or something that calls out that not all of POSIX ERE is implemented (mostly due to there being no efficient way to implement those features) and thus not all regexes are supported in our stdlib implementation.

acfoltzer: Is there still time to change the spec language before 1.3, or is there WDL out there in the wild that really depends on it?

peterhuene: Probably something that could be addressed maybe for a 2.0; feels like potential breakage for 1.x (although I have no idea what regexes are in the wild).

acfoltzer: Ahh, I was just thinking about it in the context of this new `split` function; I see now that we specify ERE compatibility for a number of other functions where lookaround and backreferences might be realistically relevant. Agreed on proposing this as an all-at-once change rather than downgrading capability just for this one new function.

peterhuene (approved):

Issue #183, opened 2025-09-19
Output total execution time for sprocket run?


sckott:
As I've been using `sprocket run` more I've been wanting to have an easy way to know how long a run took. I can do that quick subtraction math if the sprocket logs are still around in the shell session, but maybe it would be nice to have something there telling me e.g, 

```
{ some output ... }
[time 3:45]
```

a-frantz:
This should quick and easy to add! Being added to the TODO list for the next release.

I'll leave it unassigned for the moment in case someone wants to swoop in and knock it out.

dead8309:
is there any reason `time sprocket run` cannot be used here @sckott 🤔

a-frantz:
> is there any reason `time sprocket run` cannot be used here [@sckott](https://github.com/sckott) 🤔

Sprocket often has to do some networking (pulling Docker images for example) that can add a decent amount to the runtime, but aren't actually useful timing information. I think what Scott is asking here for is just the time from execution start to finish, which would be the most useful time metric in my understanding.

Although, getting those two timepoints is actually more involved than what I was thinking when I first marked this "good first issue" 🤔

claymcleod:
It's worth noting that, once we get Sprocket working on a variety of backends, we do want to make some improvements to actually profiling your tasks and workflows. So, even if `time` is the solution in the short term, there will definitely be a long-term solution around more granular profiling.

a-frantz:
> [@a-frantz](https://github.com/a-frantz) I would like to work on this Should I reimplement cleanly or base it on the closed PR?

hi! I just removed the "good first issue" label, as I think the vision needs some more refinement before it's ready to be worked on. We definitely want to add WDL profiling to Sprocket, but at least I have not thought very hard about what exactly that would look like. The issue is more complex than just having a "start" and "end" of execution timestamp (which was what this issue started as asking for, but I think we can be more ambitious).

This issue will remain open for tracking and discussion purposes, but the idea needs more detail before we'll be ready to accept a PR.

PR #426, opened 2025-10-16
feat: print elapsed time at the end of a workflow run

nleroy917:
Prints the total runtime of the workflow upon completion of `sprocket run`. Addressing #183.

Only, thought I had is I wonder if it makes sense to make these opt-in (or opt-out). If someone wants to pipe the output to some other downstream process, the total runtim would pollute that; although I am not sure how common that is.

For external contributors:

- [X] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.

For all contributors:


claymcleod:
Oh! And one more thing, it won't always be workflow that's executing, so you'll want to do something like this to get the kind of thing running and include it in the message:

```rust
let run_kind = match &inputs {
    EngineInputs::Task(_) => "task",
    EngineInputs::Workflow(_) => "workflow",
};
```

a-frantz (approved):
looks great to me, thanks!

peterhuene (changes requested):
Looking at the [implementation of `HumanDuration`](https://docs.rs/indicatif/latest/src/indicatif/format.rs.html#107), I actually don't think it's what we want to use here.

`HumanDuration` gives an approximate duration that's appropriate for displaying an ETA in a progress bar, not a precise duration which is what I would expect for a wall clock time.

`cloud-copy` (which is also a transitive dependency from `wdl-engine`) does have a `TimeDeltaExt::english` extension method for displaying a time delta in English with second precision. That operates on `chrono::TimeDelta` and not `std::time::Duration`, though (there is a conversion between the two).

nleroy917:
Hey @peterhuene thanks for the insight!

I think my original solution utilized something to the effect of:
```rust
let execution_time = start_time.elapsed();
let tot = execution_time.as_secs_f64();
```

Perhaps I could swap back to that. I'll differ to you guys for what makes sense there.

claymcleod:
@nleroy917 are you planning on working on this? Or do you want me to finish it out for you?

PR #429, opened 2025-10-16
feature: add per-queue CPU and memory limits for the LSF backend

acfoltzer:
For now, these are manually configured based on the user's understanding of each queue's behavior. LSF tooling doesn't offer a straightforward way to ask "what's the most CPU I can schedule on this queue?", but we could probably stitch a reasonable heuristic together in the future by joining `bqueues` and `bhosts` output.

Adding these limits also means that the entire `wdl-engine/tasks` suite can run and pass in a suitable HPC environment. To that end, this change also extends that suite to check whether LSF and Apptainer tooling is available, and to run the tests with that backend if so. For the moment, this facilitates manual testing of the HPC backend rather than being hooked into CI, but it is an improvement over the previous state-of-the-art which was me continually rebasing a set of HPC-specific changes to push over to the HPC on a separate in-situ branch.

For all contributors:


crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -392,20 +395,80 @@ impl TaskExecutionBackend for LsfApptainerBackend {
...
-        _hints: &std::collections::HashMap<String, crate::Value>,
+        hints: &std::collections::HashMap<String, crate::Value>,
     ) -> anyhow::Result<super::TaskExecutionConstraints> {
+        let mut required_cpu = v1::cpu(requirements);
+        let mut required_memory = ByteSize::b(v1::memory(requirements)? as u64);
+
+        // Determine whether CPU or memory limits are set for this queue, and clamp or
+        // deny them as appropriate if the limits are exceeded
+        //
+        // TODO ACF 2025-10-16: refactor so that we're not duplicating logic here (for
```
claymcleod, line 415: I'd recommend going ahead and refactoring this to like a `apply_queue_resource_limits()` method and replacing here and below with that.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -392,20 +395,80 @@ impl TaskExecutionBackend for LsfApptainerBackend {
     fn constraints(
         &self,
         requirements: &std::collections::HashMap<String, crate::Value>,
-        _hints: &std::collections::HashMap<String, crate::Value>,
+        hints: &std::collections::HashMap<String, crate::Value>,
     ) -> anyhow::Result<super::TaskExecutionConstraints> {
+        let mut required_cpu = v1::cpu(requirements);
+        let mut required_memory = ByteSize::b(v1::memory(requirements)? as u64);
```
claymcleod, line 410: If `v1::memory(requirements)` were to come back with a negative, wouldn't this wrap silently? Do we want to do a check here to make sure it's > 0? This is just a thought, it's possible that this is checked earlier upstream for us, but perhaps not bad to have a defensive check here. Same for line 500.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -568,17 +764,17 @@ impl LsfApptainerBackendConfig {
         // the external tools changes based on where a job gets dispatched, but
         // querying from the perspective of the current node allows
         // us to get better error messages in circumstances typical to a cluster.
-        if let Some(queue) = &self.default_lsf_queue {
-            validate_lsf_queue("default", queue).await?;
+        if let Some(queue) = self.default_lsf_queue.as_ref() {
```
claymcleod, line 794: Nit: can these be validated in parallel?

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -483,6 +606,79 @@ impl TaskExecutionBackend for LsfApptainerBackend {
...
+    max_memory_per_task: Option<ByteSize>,
+}
+
+impl LsfApptainerQueueConfig {
+    /// Create an [`LsfApptainerQueueConfig`].
+    pub fn new(
+        name: String,
+        max_cpu_per_task: Option<u64>,
+        max_memory_per_task: Option<ByteSize>,
+    ) -> Self {
```
claymcleod, line 643: Any validation that would be appropriate here?

claymcleod: I can think of (a) `name` not empty and (b) `max_cpu_per_task`/`max_memory_per_task` not 0.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: There is an upcoming refactor that is going to probably change this in a way that would revert one side of this. I considered it but don't think it's worth the churn.

claymcleod: Okay, makes sense.

crates/wdl-engine/src/backend/lsf_apptainer.rs
claymcleod: Upon further inspection, if this is just checking `bqueues`, it may not be worth it. Up to you.

crates/wdl-engine/tests/common/mod.rs
claymcleod, line 1: So far, we've avoided using `mod.rs` in the codebase and we're just using the newer named module files (i.e., `common.rs` in the parent folder).

acfoltzer: This is a special case for the root directory of integration tests, benches, and examples in Cargo projects. If we don't use `tests/common/mod.rs`, `tests/common.rs` will get treated as an integration test in its own right, and will be compiled and run unnecessarily.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: It looks like this is checked at some point in evaluation, as trying to run a task with a negative integer value gives an appropriately-informative error message. I am surprised it does _not_ get flagged by `check` or `lint` though, probably because it might need some evaluation in the general case where variables or more complex expressions could be present here. I'll open a new issue to look into this more holistically.

acfoltzer: https://github.com/stjude-rust-labs/sprocket/issues/431

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: Yeah, the extra complexity to launch and gather these for a one-time startup operation with at most 4 queues to check doesn't seem worthwhile. I have considered more intensive checks that might, for example, submit a synthetic job to test for schedulability or properties of the compute node environment. These would be worthwhile candidates to parallelize for sure.

crates/wdl-engine/src/backend/lsf_apptainer.rs
acfoltzer: Addressed in https://github.com/stjude-rust-labs/sprocket/pull/429/commits/9c2dcaae5f3505cec972895cc7f635167252beb8

`name` not empty is a good one. I don't think we'd be vulnerable to any shell injection weirdness since we're using the `Command` API when invoking both `bqueues` and `bsub` with that value, but it's better to keep malformed shell arguments as far from the realm of possibility as we can.

I also added the non-zero-if-present checks for CPU and memory here, but those would've been straightforwardly rejected as soon as any task tried to execute. It does feel a bit weird to allow a one-byte memory limit, but there's also not a clear minimum to apply here. In any case, if someone gets confused about the units of this field, they should still get an informative error message with units that should show them where they've gone astray.

PR #432, opened 2025-10-17
feat: adds pre-evaluation `task` availability alongside `task.previous` in post-evaluation contexts

claymcleod:
This PR adds context-aware retries to WDL v1.3 by implementing the feature proposed in https://github.com/openwdl/wdl/issues/696.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-ast/src/v1/task.rs
```diff
@@ -246,6 +254,8 @@ pub const TASK_FIELD_META: &str = "meta";
 pub const TASK_FIELD_PARAMETER_META: &str = "parameter_meta";
 /// The name of the `ext` task variable field.
 pub const TASK_FIELD_EXT: &str = "ext";
+/// The name of the `max_retries` task variable field.
+pub const TASK_FIELD_MAX_RETRIES: &str = "max_retries";
```
claymcleod, line 258: @peterhuene I guess `max_retries` was not added before? Looking [at the spec](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#runtime-access-to-requirements-hints-and-metadata), it indeed appears that `max_retries` is missing from the description of the `task` type. I'll make an issue.

crates/wdl-analysis/src/types/v1.rs
```diff
@@ -127,12 +129,73 @@ use crate::stdlib::FunctionBindError;
...
+    }
+}
+
+/// Gets the type of a `task` variable member for post-evaluation contexts.
+///
+/// This is used in command and output sections where all task fields are
+/// available.
+///
+/// Returns [`None`] if the given member name is unknown.
+pub fn task_task_post_evaluation_member_type(
```
claymcleod: This name was confusing me because `task` in the other methods means "in the context of a task" whereas here is was meaning "the `task` variable". So, just to make it explicit, I duplicated it—don't love that, but it makes the function name more clear IMO.

crates/wdl-analysis/tests/analysis/task-scope-output/source.wdl
```diff
@@ -0,0 +1,33 @@
...
+    memory: 2000000000
+    cpu: 2
+  }
+
+  command <<<
+    echo "test"
+  >>>
+
+  output {
+    # Post-evaluation fields are available.
```
claymcleod, line 16: Some of these tests may test things twice, I just had made a variety of them so I decided to keep them. Let me know if that is bothersome.

crates/wdl-engine/src/eval/v1/expr.rs
```diff
@@ -1473,10 +1473,16 @@ impl<C: EvaluationContext> ExprEvaluator<C> {
                 Some(value) => Ok(value.clone()),
                 None => Err(not_an_object_member(&name)),
             },
-            Value::Task(task) => match task.field(name.text()) {
+            Value::TaskPreEvaluation(task) => match task.field(name.text()) {
                 Some(value) => Ok(value.clone()),
                 None => Err(not_a_task_member(&name)),
             },
+            Value::TaskPostEvaluation(task) => {
+                match task.field(self.context.version(), name.text()) {
```
claymcleod, line 1483: Version needs to be passed in here because `previous` is not available in WDL v1.2.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1336,6 +1347,14 @@ impl TaskEvaluator {
             .document
             .version()
             .expect("document should have version");
+
+        // In WDL 1.3+, use `TASK_SCOPE_INDEX` to access the `task` variable.
+        let scope_index = if version >= SupportedVersion::V1(V1::Three) {
+            TASK_SCOPE_INDEX
```
claymcleod, line 1365: @peterhuene is this right?

crates/wdl-engine/src/value.rs
```diff
@@ -2717,37 +2757,178 @@ struct TaskData {
...
+    /// Constructs a new post-evaluation task value with the given name,
+    /// identifier, and constraints.
+    pub(crate) fn new<N: TreeNode>(
         name: impl Into<String>,
         id: impl Into<String>,
         definition: &v1::TaskDefinition<N>,
         constraints: TaskExecutionConstraints,
         attempt: i64,
     ) -> Self {
+        let previous_ty = TASK_PREVIOUS_TYPE.clone();
```
claymcleod: TODO: unify with pre-evaluation?

crates/wdl-engine/src/value.rs
```diff
@@ -2889,23 +3074,47 @@ impl TaskValue {
...
     }
 
+    /// Sets the previous requirements for retry attempts.
+    pub(crate) fn set_previous(
+        &mut self,
+        version: SupportedVersion,
+        requirements: &HashMap<String, Value>,
+    ) {
+        // SAFETY: this should always have a type as it is statically defined.
+        let ty = task_task_post_evaluation_member_type(version, TASK_FIELD_PREVIOUS).unwrap();
```
claymcleod: TODO: unify with the creation above.

crates/wdl-engine/tests/tasks/task-scope-command/config.toml
```diff
@@ -0,0 +1 @@
+backends.default = { type = "local" }
```
claymcleod: Sadly, this means that I cannot check the `container` field being set. This is due to constraints in our MacOS CI environment which, apparently, cannot run container-based tests for tasks. I suppose this still gets the job done of checking that it's unset, but still unsatisfying.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -531,12 +531,24 @@ fn add_task(config: &Config, document: &mut DocumentData, definition: &TaskDefin
...
+                    scopes[index.0].insert(TASK_VAR_NAME, task_name.span(), task_type);
+                }
+            }
+            Type::TaskPostEvaluation => {
+                // Post-evaluation task type is available in v1.2+.
+                if version >= Some(SupportedVersion::V1(V1::Two)) {
+                    scopes[index.0].insert(TASK_VAR_NAME, task_name.span(), task_type);
+                }
+            }
+            _ => {}
```
peterhuene: Should this panic, as it seems like it should be one of the two variants above?

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1611,10 +1675,10 @@ impl TaskEvaluator {
             ),
         };
 
-        // Update or insert the `task` variable in the task scope
-        // TODO: if task variables become visible in `requirements` or `hints` section,
-        // this needs to be relocated to before we evaluate those sections
-        if state.document.version() >= Some(SupportedVersion::V1(V1::Two)) {
+        // Now that those are evaluated, insert a [`TaskPostEvaluation`] for
+        // `tasks` which includes those calculates requirements before the
```
adthrasher:
```suggestion
        // `tasks` which includes those calculated requirements before the
```

crates/wdl-analysis/src/types/v1.rs
peterhuene: I personally don't see why it can't be called `task_member_type_pre_evaluation` and `task_member_type_post_evaluation`; they're both getting the type of a `task` variable member, just in different evaluation contexts.

crates/wdl-analysis/src/types/v1.rs
```diff
@@ -127,12 +129,73 @@ use crate::stdlib::FunctionBindError;
 use crate::stdlib::MAX_PARAMETERS;
 use crate::stdlib::STDLIB;
 use crate::types::Coercible;
-/// Gets the type of a `task` variable member type.
+
+/// The type for `task.previous`.
+pub static TASK_PREVIOUS_TYPE: LazyLock<Type> = LazyLock::new(|| {
+    Type::Compound(
+        CompoundType::Struct(Arc::new(StructType::new(
```
peterhuene: Should this just be an `Object` rather than a named struct?

crates/wdl-ast/src/v1/task.rs
peterhuene: That's correct (and I repeated exactly this before I saw this question in the `max_retries` PR).

crates/wdl-engine/src/eval/v1/task.rs
peterhuene: I think so; the task scope can be reused across the different sections provided we update (which this PR does) the task variable value before each section's evaluation.

crates/wdl-analysis/src/types/v1.rs
peterhuene: Using a struct type here feels wrong and I see why we wouldn't want to use an `Object` because then we can't statically determine the fields being accessed.

The spec should probably define a hidden type for the "previous requirements" field of the task hidden type.

As such, we should probably have a new variant in the `Type` enumeration to represent the type of the "previous requirements" field of the task hidden type.

I think the type of the field would be `Type::PreviousRequirements` (a special hidden type; if these hidden types are getting out of hand, perhaps we could extract them out into a `Type::Hidden(HiddenType)` variant and `HiddenType::PreviousRequirements` instead).

From there we basically follow the same pattern we do for the `Type::TaskPreEvaluation` and `Type::TaskPostEvaluation` (i.e. we have a function that can be used to determine the legal fields of a "previous requirements" hidden type).

crates/wdl-engine/src/value.rs
```diff
@@ -2717,37 +2756,178 @@ struct TaskData {
...
+            // SAFETY: an empty `previous` struct should always create, as all
+            // of the member types are optional.
+            previous: Struct::new(None, previous_ty, std::iter::empty::<(String, Value)>())
+                .unwrap(),
+        }
+    }
+
+    /// Sets the previous requirements for retry attempts.
+    pub(crate) fn set_previous(&mut self, requirements: &HashMap<String, Value>) {
+        let previous_ty = TASK_PREVIOUS_TYPE.clone();
```
peterhuene: So if we end up using a new hidden type to represent the "previous requirements" of a task variable, we could have a new `Value` variant to represent a value of the previous requirements.

That value could store an `Arc<HashMap<String, Value>>` and we could cheaply clone the previous requirements without having to do any filtering.

That "previous requirements" value would have a `field` method, just like we do for these task values, that would only access the names that valid for the previous requirements hidden type.

I would greatly prefer that than having to "extract out valid fields" to recreate the previous value each time.

crates/wdl-analysis/src/types/v1.rs
peterhuene: An additional benefit would not having `struct TaskPrevious` showing up in diagnostics (when I see that, I would go searching for a definition of that struct).

We could effectively have the diagnostics we do now for accessing an invalid field for the `task` variable, only it would be an invalid field for the `previous` field.

crates/wdl-engine/src/value.rs
```diff
@@ -2717,37 +2770,172 @@ struct TaskData {
...
+pub struct PreviousRequirementsValue(Arc<HashMap<String, Value>>);
+
+impl PreviousRequirementsValue {
+    /// Creates a new previous requirements value.
+    pub fn new(requirements: HashMap<String, Value>) -> Self {
+        Self(Arc::new(requirements))
+    }
+
+    /// Creates an empty previous requirements value.
+    pub fn empty() -> Self {
```
claymcleod, line 2769: I recognize we could `#[derive(Default)]` here, but I like keeping `empty()` to indicate to the reader in the downstream methods the intention.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1065,6 +1081,16 @@ impl TaskEvaluator {
 
                 attempt += 1;
 
+                let task = state.scopes[TASK_SCOPE_INDEX.0]
```
claymcleod: I'm not sure if there is a better way to do this, come back and consider.

crates/wdl-analysis/tests/analysis/task-scope-requirements/source.wdl
```diff
@@ -9,7 +9,7 @@ task test_requirements_scope {
     cpu: task.attempt + 1
     container: if task.id != "" then "ubuntu:latest" else "debian:latest"
     disks: if select_first([task.previous.memory, 0]) > 0 then ["20 GiB"] else ["10 GiB"]
-    gpu: select_first([task.previous.gpu, false])
+    gpu: select_first([task.previous.memory, 0]) > 1000000000
```
adthrasher: Why is this using the prior memory to set gpu?

claymcleod: It's just a proof of concept to show that these fields compile in a pre-evaluation context. This is not something you would do in a real WDL document.

crates/wdl-engine/src/eval/v1/expr.rs
```diff
@@ -1473,10 +1476,22 @@ impl<C: EvaluationContext> ExprEvaluator<C> {
...
+                    Some(value) => Ok(value.clone()),
+                    None => Err(not_a_task_member(&name)),
+                }
+            }
+            Value::PreviousTaskData(prev) => match previous_task_data_member_type(name.text()) {
+                Some(ty) => Ok(prev
+                    .field(name.text())
+                    .unwrap_or_else(|| Value::new_none(ty))),
+                None => Err(not_a_previous_task_data_member(&name)),
```
peterhuene: Should this not be implemented like accessing the pre/post task variables above? That is to say, we call `prev.field` and if it returns `None`, we error instead of returning a WDL `None` value (and thus don't need to call `previous_task_data_member_type`)?

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1065,6 +1081,16 @@ impl TaskEvaluator {
...
+                    .names
+                    .get(TASK_VAR_NAME)
+                    // SAFETY: task variable should always exist in scope at this point
+                    .unwrap()
```
peterhuene: Is this true, though? I don't think we insert a `task` variable into the scope for < 1.2 documents.

Feels like we should only be doing this if `get` returns `Some`.

crates/wdl-engine/src/value.rs
```diff
@@ -2705,49 +2748,196 @@ struct TaskData {
...
+    /// Contains the evaluated task data from the previous attempt.
+    ///
+    /// On the first attempt, this is empty.
+    previous: PreviousTaskDataValue,
 }
 
-impl TaskValue {
-    /// Constructs a new task value with the given name and identifier.
-    pub(crate) fn new_v1<N: TreeNode>(
+impl TaskPreEvaluationValue {
```
peterhuene, line 2888: Can we move this impl closer to `TaskPreEvaluationValue`'s declaration?

crates/wdl-engine/src/value.rs
```diff
@@ -2775,30 +2965,32 @@ impl TaskValue {
...
             attempt,
+            meta: definition
+                .metadata()
+                .map(|s| Object::from_v1_metadata(s.items()))
+                .unwrap_or_else(Object::empty),
+            parameter_meta: definition
+                .parameter_metadata()
+                .map(|s| Object::from_v1_metadata(s.items()))
+                .unwrap_or_else(Object::empty),
+            ext: Object::empty(),
```
peterhuene: I wonder if we could just reuse the `meta`, `parameter_meta`, and `ext` from the `TaskPreEvaluationValue` that we construct earlier in task evaluation and thus won't need to walk the AST again.

peterhuene (commented):
Looks great (I especially like refactoring for hidden types). I think there's now far fewer allocations made during task evaluation due to these changes.

I just have a few more comments to go through.

crates/wdl-engine/src/eval/v1/expr.rs
claymcleod: Importantly, the behavior is a bit different—when fields are supported but have no value, they are returned as `None`. That being said, I can refactor the code so that the coercion to `None` happens in the `field` method, which I _think_ is what you're alluding to here (the outcome will be that this code you've commented on will be structured similarly to the above).

crates/wdl-engine/src/eval/v1/task.rs
claymcleod: Ah good catch, this is not true for < WDL v1.2.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1586,8 +1637,48 @@ impl TaskEvaluator {
         definition: &TaskDefinition<SyntaxNode>,
         inputs: &TaskInputs,
         attempt: u64,
+        previous_task_data: Option<Arc<TaskPostEvaluationData>>,
     ) -> EvaluationResult<EvaluatedSections> {
-        // Start by evaluating requirements and hints
+        let version = state.document.version();
+
+        // Extract task metadata once to avoid walking the AST multiple times
+        let task_meta = definition
```
claymcleod, line 1645: These are calculated whether or not they are needed (even for < WDL v1.2)—I felt that was fine as they are relatively quick operations. Alternatively, we could lazily evaluate them, but the locking overhead would probably be slow enough that it's not worth the added code complexity.

peterhuene: Agreed, metadata evaluation is relatively inexpensive in that it doesn't invoke expression evaluation. I'm fine with this approach.

peterhuene (approved):
Great work!

PR #433, opened 2025-10-20
feat: adds `max_retries` to the `task` type

claymcleod:
This PR adds the `max_retries` member to the `task` type in support of https://github.com/openwdl/wdl/issues/732 and https://github.com/openwdl/wdl/pull/733. Those links, especially #732, give the best (albeit short) defense of why I think this member needs to be present. It is essentially an argument of completeness combined with the fact that I can think of a few scenarios where one might want to use the member to crank up the resources on the last attempt.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/src/types/v1.rs
+++ b/crates/wdl-analysis/src/types/v1.rs
@@ -200,7 +200,7 @@ pub fn task_task_post_evaluation_member_type(
         n if n == TASK_FIELD_NAME || n == TASK_FIELD_ID => Some(PrimitiveType::String.into()),
         n if n == TASK_FIELD_CONTAINER => Some(Type::from(PrimitiveType::String).optional()),
         n if n == TASK_FIELD_CPU => Some(PrimitiveType::Float.into()),
-        n if n == TASK_FIELD_MEMORY || n == TASK_FIELD_ATTEMPT => {
+        n if n == TASK_FIELD_MEMORY || n == TASK_FIELD_ATTEMPT || n == TASK_FIELD_MAX_RETRIES => {
             Some(PrimitiveType::Integer.into())
         }
         n if n == TASK_FIELD_GPU || n == TASK_FIELD_FPGA => {
--- a/crates/wdl-analysis/tests/analysis/task-scope-full/source.diagnostics
+++ b/crates/wdl-analysis/tests/analysis/task-scope-full/source.diagnostics
@@ -1,18 +1,18 @@
 error: struct `TaskPrevious` does not have a member named `not_a_member`
-   ┌─ tests/analysis/task-scope-full/source.wdl:23:24
+   ┌─ tests/analysis/task-scope-full/source.wdl:24:24
    │
-23 │     gpu: task.previous.not_a_member
+24 │     gpu: task.previous.not_a_member
    │                        ^^^^^^^^^^^^
 
 error: the `task` variable does not have a member named `cpu`
-   ┌─ tests/analysis/task-scope-full/source.wdl:26:16
+   ┌─ tests/analysis/task-scope-full/source.wdl:27:16
    │
-26 │     fpga: task.cpu > 2
+27 │     fpga: task.cpu > 2
    │                ^^^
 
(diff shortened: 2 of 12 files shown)


crates/wdl-analysis/tests/analysis/task-scope-full/source.diagnostics
```diff
@@ -1,18 +1,18 @@
 error: struct `TaskPrevious` does not have a member named `not_a_member`
-   ┌─ tests/analysis/task-scope-full/source.wdl:23:24
+   ┌─ tests/analysis/task-scope-full/source.wdl:24:24
    │
-23 │     gpu: task.previous.not_a_member
+24 │     gpu: task.previous.not_a_member
```
claymcleod: Offsets from adding the `max_retries` line above.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1626,7 +1621,7 @@ impl TaskEvaluator {
         inputs: &TaskInputs,
         attempt: u64,
         previous_requirements: Option<Arc<HashMap<String, Value>>>,
-    ) -> EvaluationResult<EvaluatedSections> {
+    ) -> EvaluationResult<(EvaluatedSections, u64)> {
```
claymcleod: I don't love this, but I didn't love mixing it in with the very clean `EvaluationSections` struct either. The alternative is calculating the field twice, which I think is just a recipe for a bug.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -987,16 +990,6 @@ impl TaskEvaluator {
                 )
                 .await?;
 
-            // Get the maximum number of retries, either from the task's requirements or
-            // from configuration
```
claymcleod, line 1004: This is moved verbatim down to `evaluate_sections()` below.

crates/wdl-engine/src/eval/v1/task.rs
acfoltzer: Good instinct here to not want to duplicate it, but returning a bare integer here is awkward. I think the way to go is to make a standalone function similar to `cpu`, `max_cpu`, and friends toward the top of this file. Then `evaluate_sections()` can use it for the purposes of populating the `PostEvaluationTask`, and `perform_evaluation` can use it to actually limit attempts.

acfoltzer (changes requested):

crates/wdl-engine/src/eval/v1/task.rs
claymcleod: Alright, will do.

acfoltzer (approved):
Nice!

crates/wdl-analysis/src/types/v1.rs
```diff
@@ -200,7 +200,7 @@ pub fn task_task_post_evaluation_member_type(
         n if n == TASK_FIELD_NAME || n == TASK_FIELD_ID => Some(PrimitiveType::String.into()),
         n if n == TASK_FIELD_CONTAINER => Some(Type::from(PrimitiveType::String).optional()),
         n if n == TASK_FIELD_CPU => Some(PrimitiveType::Float.into()),
-        n if n == TASK_FIELD_MEMORY || n == TASK_FIELD_ATTEMPT => {
+        n if n == TASK_FIELD_MEMORY || n == TASK_FIELD_ATTEMPT || n == TASK_FIELD_MAX_RETRIES => {
```
peterhuene, line 164: Question: as `max_retries` is [not defined for the task variable in the 1.2 spec](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#runtime-access-to-requirements-hints-and-metadata), should we be limiting the availability of this field to 1.3+?

claymcleod: Good catch, I'll add it.

claymcleod:
This has now diverged significantly from #432 so, rather than fixing all the merge commits, I'm going to just incorporate these changes directly in #432.

Issue #434, opened 2025-10-21
feat: Add a Slurm + Apptainer HPC backend


acfoltzer:
This will at first look very similar to the LSF + Apptainer backend.

Issue #360, opened 2025-09-25
Improve cache management for Apptainer backend


acfoltzer:
From #182:
> Docker-style OCI images are converted to Apptainer's `.sif` format once upon first use in order to avoid issues where large scatters would encounter network failures and rate limiting with certain container backends. These images are currently left behind in a temp directory created for each run, and should be deleted by users regularly to avoid consuming too much disk space. Future work will relieve this burden by making a smarter, automatically-managed cache.

Some considerations here:
- `apptainer` has built-in caching of image layers, so it's not the end of the world if we `apptainer pull` more than is strictly necessary. The main downsides of overdoing it are related to disk space consumption.
- Concurrent `apptainer pull`s to the same destination are bad news and will result in errors.
- Throughout a single top-level execution, a given WDL `container` specifier should resolve to the same image. That is, even if a new `myimage:latest` tag gets pushed in the middle of an execution, we should not start running subsequent tasks of the same top-level execution using the updated image.
- The implementation should be robust to multiple `sprocket`s running concurrently on the same system. This could mean using lock files or similar mechanisms if we attempt to use a shared cache. Lots of potential pitfalls here, like accidental serialization of executions given that the lifetime of a particular cached image needs to be longer than just the `pull` step.
- The simplest improvement would be to use `tempdir` to clean up the cache after each execution, but it's desirable to be able to edit and run generated scripts in-place in the `runs/` directory following a failure. If paths to the `.sif` files are baked into these scripts, it makes expiration and purging of the cache fraught, as either of those operations could invalidate those scripts. Perhaps with call caching this won't be as big of a deal.

PR #384, opened 2025-09-30
chore: fix defunct references to the old WDL repo.

peterhuene:
Before submitting this PR, please make sure:

For external contributors:

- [x] You have read the `wdl` crates' [CONTRIBUTING](https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md) guide in its entirety.
- [x] You have not used AI on any parts of this pull request.

For all contributors:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added tests (when appropriate).
- [x] You have added an entry in the CHANGELOG (when appropriate).
- [x] You have updated the README or other documentation to account for these changes (when appropriate).
- [x] You have either (a) made a PR to the `next` branch in the sprocket.bio repository [if you are a Sprocket team member] or (b) have suggested any changes you _think_ need to be made to sprocket.bio in the description of your PR [if you are an external contributor].

For PRs containing lint rule changes:

- [x] You have updated any and all effected entries within `RULES.md`.
- [x] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

PR #435, opened 2025-10-21
revise: improve configurability of integration test suites and make WDL 1.3 flag specific to that version

acfoltzer:
These changes are primarily aimed at improving the suitability of the test suite for execution in HPC environments in service of #434, but also give `wdl-engine` test authors more flexibility to vary test configurations.

## Engine config from environment

The `SPROCKET_TEST_ENGINE_CONFIG` environment variable can be set to a file path that will be read in place of the default engine configuration(s) whenever a task or workflow will be executed by a test. Many of our tests will, by default, run using the local backend and the Docker backend. By setting the environment variable to an alternate configuration like `test-configs/lsf-apptainer-engine.toml`, these default backends will be omitted and only the specified backend will be used.

So far, this capability has been added to the `sprocket/cli/run` suite and the `wdl-engine/{tasks,workflows}` suites, but it is portable to any test suite which may require the use of `wdl-engine` evaluation in the future.

This replaces the recently-added mechanism that checks for LSF and Apptainer binaries on the PATH and performs a similar config substitution. This approach quickly fell over when encountering HPC environments with multiple scheduler binaries on the PATH.

## Config overrides instead of replacements

To make this change tractable in combination with the existing ability to override engine configs on a test-by-test basis via `config.toml` or `config.linux.toml` in the test root, the latter capability has been changed to be a means to modify the base config rather than replacing it in full.

Instead of a full config in TOML format, the test runners for the `tasks` and `workflows` suites look for a `config-override.yaml` file and apply it using [Figment's `merge` semantics](https://docs.rs/figment/latest/figment/struct.Figment.html#conflict-resolution). YAML was chosen due to its ability to override `Some(_)` options with `None`; TOML does not have a `null` so this is not easily expressible.

## Config overrides for `Analysis`

#424 added configs for the `Analysis` step of the `wdl-engine` tests beyond the defaults. Taking advantage of the new capability to more-surgically override config fields for individual tests, this change makes the analysis config overridable using the same `config-override.yaml` as the engine config.

## Code consolidation in the `wdl-engine` integration tests

Much of the logic in the `tasks` and `workflows` suites was redundant, and has been combined in a `common` module.

## Increased specificity of the feature flag enabling WDL 1.3 features

While integrating the new override for the analysis config, I noticed that this new feature flag enables all potential future versions, not just WDL 1.3. This scope appeared to be unintentionally broad, so I narrowed it to 1.3-only with the intent that it can be deprecated and removed once 1.3 stabilizes.

## Shell expansion on `apptainer_images_dir` config flag

This is needed in order to specify a safe place for these cache directories on HPCs where `$TMPDIR` cannot be reliably used across hosts. This will likely be obviated by #360 but will be useful in the meantime.

## Fixed a quoting bug when passing environment variables to Apptainer

How about that, improving the test suite found a bug!

### DIFF
--- a/crates/wdl-engine/tests/common/mod.rs
+++ b/crates/wdl-engine/tests/common/mod.rs
@@ -1,3 +1,9 @@
+//! Common logic for the `wdl-engine` integration tests.
+//!
+//! This is located in `common/mod.rs` rather than `common.rs` in order to avoid
+//! `cargo test` treating this as an integration test target and needlessly
+//! compiling a test executable from its source.
+
 use std::collections::HashMap;
 use std::env;
 use std::fs;


crates/wdl-engine/tests/tasks.rs
acfoltzer, line 1: Not much has actually changed in this file, but between the code motion and the extra indentation for an `async {}` block, the diff is horrendous. I recommend looking at `common/mod.rs` and this file in full rather than trying to piece apart chunks of the diff.

crates/wdl-engine/tests/workflows.rs
acfoltzer, line 1: Not much has actually changed in this file, but between the code motion and the extra indentation for an `async {}` block, the diff is horrendous. I recommend looking at `common/mod.rs` and this file in full rather than trying to piece apart chunks of the diff.

crates/wdl-engine/tests/common/mod.rs
```diff
@@ -1,16 +1,211 @@
...
 }
 
-/// Are LSF and Apptainer tools available?
-pub fn lsf_apptainer_available() -> Result<bool, anyhow::Error> {
-    let lsf_available = can_find_binary("bsub")?;
-    let apptainer_available = can_find_binary("apptainer")?;
-    Ok(lsf_available && apptainer_available)
+/// Get the baseline configs for executing the tests.
+///
+/// These configs may be modified by merging with `*-config-override.json` files
```
a-frantz:
```suggestion
/// These configs may be modified by merging with `*-config-override.yaml` files
```

crates/README.md
a-frantz, line 1: OOS for this PR, but I wonder if we really want to keep this file around instead of merging its contents into the root README?

tests/cli.rs
```diff
@@ -160,6 +159,9 @@ async fn run_sprocket(test_path: &Path, working_test_directory: &Path) -> Result
         .wait_with_output()
         .context("failed while waiting for command to finish")?;
 
+    // Make sure the temporary config isn't dropped prematurely
+    drop(env_config);
```
a-frantz, line 163: can you explain what this is doing for me? I haven't seen this pattern in Rust code before

crates/wdl-engine/tests/common/mod.rs
a-frantz, line 1: is there a reason for this being at `common/mod.rs` instead of just `common.rs`?

crates/README.md
acfoltzer: @peterhuene did some of that work in #384, but it's worth considering whether that's complete with two READMEs.

tests/cli.rs
acfoltzer: Good question! The short version is: we're making sure that cleanup logic that could delete the temporary file introduced for `env_config` doesn't run until after we're done using it.

The call to `drop` doesn't inherently _do_ anything, you can define your own `drop` that works perfectly well as:

```rust
fn my_drop<T>(_: T) {}
```

The logic in a type's `Drop` impl implicitly runs when a value of that type is no longer in scope. Using `drop` lets us explicitly control when this happens, which might be sooner or later than the compiler would otherwise be free to do it.

In general, I like using the explicit `drop` for types where the effects of the `Drop` have a really big effect on other code, even if inserting it explicitly doesn't actually change where it happens vs just following the normal rules of scope. 

Here for example, `env_config` isn't used after we set up the arguments to `sprocket`, but we the developers know that if that file gets cleaned up before `sprocket` gets run, we'll hit an error. So it's best to use our extra knowledge of the system to show the compiler how long `env_config` really ought to live.

Another common case where I prefer explicit dropping is when a `MutexGuard` or similar concurrency primitive is bound to a variable and possibly used several times. Here, it might be useful to use `drop` to _shorten_ the lifetime of the guard in order to release the lock as soon as we're done messing with what's inside it.

Some more references [here](https://doc.rust-lang.org/book/ch15-03-drop.html) and [here](https://doc.rust-lang.org/rust-by-example/trait/drop.html)

crates/wdl-engine/tests/common/mod.rs
acfoltzer: Yes, `cargo test` treats all `tests/*.rs` files as individual test entrypoints. If we had `tests/common.rs`, Cargo would needlessly compile a separate test executable for that file.

acfoltzer: Since you're not the first to ask this question, I went ahead and added an explanation inline: 1cc45b5e

tests/cli.rs
a-frantz: Thanks for this thorough writeup! Makes a lot of sense :D 

I've written a custom `Drop` impl before (https://github.com/stjude-rust-labs/sprocket/blob/main/crates/wdl-format/src/element.rs#L35), but have not seen the explicit `drop()` call used before. Is it possible for the compiler to drop `env_config` without this explicit call? I would _expect_ it to persist till the variable left scope, but as it isn't used anywhere would the compiler drop it early (without this explicit call ofc)?

acfoltzer: You've got the right understanding. The compiler will _always_ insert code to run drop logic when a variable leaves scope, whether that's via the end of a curly-brace expression block, an early `return` or `break`, or even via unwinding from a panic. It will never insert the drop logic when the variable is still live, but as you say there are many cases where a variable is no longer live, but we haven't e.g. reached the end of the curly-brace block. 

In theory, the compiler is free to insert the drop logic anywhere in that span. In practice, it has almost always inserted it at the end of the scope, but this is all too subtle and requires too much thinking about the compiler for my tastes. So, I prefer to use `drop` explicitly when it's possible that the timing of the drop will make an observable difference.

a-frantz: great, learned something new! 🎉

PR #427, opened 2025-10-16
refactor: split Apptainer logic into a separate module

acfoltzer:
This is a mostly-code-motion refactor in service of #359.

- The Apptainer script generation was lifted out of `LsfApptainerTaskRequest::run()` into a standalone module.

- The Apptainer configuration is now its own struct, though for the moment we flatten it into the `LsfApptainerBackendConfig` struct to avoid breaking current configs.

- Some basic unit tests were added to ensure that script generation does not trivially error, and passes a `shellcheck` test. Since `shellcheck` is already a dev environment prerequisite, this does not need a README update.

For all contributors:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added tests (when appropriate).
- [ ] You have added an entry in the CHANGELOG (when appropriate).
- [ ] You have updated the README or other documentation to account for these changes (when appropriate).
- [ ] You have either (a) made a PR to the `next` branch in the sprocket.bio repository [if you are a Sprocket team member] or (b) have suggested any changes you _think_ need to be made to sprocket.bio in the description of your PR [if you are an external contributor].

PR #372, opened 2025-09-27
wip: snack-sized feature addons for LSF + Apptainer HPC backend

acfoltzer:
So far addresses, at least partially:
- #368 
- #369 

For all contributors:

- [ ] You have added a few sentences describing the PR here.
- [ ] You have added yourself or the appropriate individual as the assignee.
- [ ] You have added at least one relevant code reviewer to the PR.
- [ ] Your code builds clean without any errors or warnings.
- [ ] You have added tests (when appropriate).
- [ ] You have added an entry in the CHANGELOG (when appropriate).
- [ ] You have updated the README or other documentation to account for these changes (when appropriate).
- [ ] You have either (a) made a PR to the `next` branch in the sprocket.bio repository [if you are a Sprocket team member] or (b) have suggested any changes you _think_ need to be made to sprocket.bio in the description of your PR [if you are an external contributor].

For PRs containing lint rule changes:

- [ ] You have updated any and all effected entries within `RULES.md`.
- [ ] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

PR #436, opened 2025-10-22
feat: experimental Slurm + Apptainer HPC backend

### LINKED ISSUE #434: feat: Add a Slurm + Apptainer HPC backend

acfoltzer:
Closes #434.

Adds an experimental backend using Slurm as a job scheduler and Apptainer (a.k.a. Singularity) as a container runtime, very similarly to the LSF + Apptainer backend initially added in #182. This builds on the same improvements delivered to that backend in #372, #379, #404, #427, #429, and #435, so this is much more complete and better-tested than the LSF backend initially was.

However, this is still experimental because testing still must be performed manually rather than in CI, and because the backend has only been tested on a small number of example workflows. We anticipate immediate feature requests and bug reports once folks begin using it anger.

Here is a sample `sprocket.toml` that uses the Slurm backend on St. Jude's Slurm cluster:

```toml
# The Slurm + Apptainer backend requires explicitly opting into experimental
# features.
run.experimental_features_enabled = true

# Set the default backend to Slurm + Apptainer.
run.backends.default.type = "slurm_apptainer"

# The Slurm partition used by default for task execution.
#
# This parameter is optional. If it's absent and no other applicable queues
# are specified, jobs will be submitted to your Slurm cluster's default
# partition.
run.backends.default.default_slurm_partition.name = "gpu"
# The largest number of CPUs and memory that can be reserved for a single job
# on this partition.
#
# These parameters are optional, and should be set according to site-specific
# information about the hosts available to dispatch work from the partition.
# They can also be set for the other types of partitions, but this example
# leaves them unconstrained by default.
run.backends.default.default_slurm_partition.max_cpu_per_task = 64
run.backends.default.default_slurm_partition.max_memory_per_task = "96 GB"

# The Slurm partition used for short tasks.
#
# This parameter is optional, and overrides `default_slurm_partition`.
run.backends.default.short_task_slurm_partition.name = "short"

# The Slurm partition used for GPU tasks.
#
# This parameter is optional, and overrides `default_slurm_partition` and
# `short_task_slurm_partition`.
run.backends.default.gpu_slurm_partition.name = "gpu"

# The Slurm partition used for FPGA tasks.
#
# This parameter is optional, and overrides `default_slurm_partition` and
# `short_task_slurm_partition`.
run.backends.default.fpga_slurm_partition.name = "fpga"

# Additional command-line arguments to pass to `sbatch` when submitting jobs to
# Slurm.
run.backends.default.extra_sbatch_args = ["--time=60"]

# The maximum number of subtasks each `scatter` will try executing at once.
#
# This is *not* a direct limit on the total number of concurrent tasks, but
# can affect the number of jobs that get queued at one time.
run.backends.default.max_scatter_concurrency = 200

# The directory where Apptainer `.sif` images will be cached.
#
# This must be a location that is accessible from both the node where Sprocket
# is running and any nodes where Slurm may dispatch a job for execution. In many
# cluster configurations, this means that `/tmp` will *not* work.
run.backends.default.apptainer_images_dir = "/scratch_space/$USER/sprocket-apptainer-images"
```

      is implemented.

### DIFF
--- a/crates/wdl-engine/src/backend/slurm_apptainer.rs
+++ b/crates/wdl-engine/src/backend/slurm_apptainer.rs
@@ -28,7 +28,8 @@ use tokio::io::BufReader;
 use tokio::process::Command;
 use tokio::sync::broadcast;
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error};
+use tracing::debug;
+use tracing::error;
 use tracing::trace;
 use tracing::warn;
 


crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,886 @@
...
+        self.required_memory.as_u64()
+    }
+
+    async fn run(self) -> anyhow::Result<super::TaskExecutionResult> {
+        let crankshaft_task_id = crankshaft::events::next_task_id();
+
+        let attempt_dir = self.spawn_request.attempt_dir();
+
+        // Create the host directory that will be mapped to the WDL working directory.
+        let wdl_work_dir = attempt_dir.join(WORK_DIR_NAME);
```
claymcleod: nit: pulling out the process of creating a file, maybe setting it's permissions, and maybe writing contents to it would make the first part of this method more readable and shrink it down significantly.

claymcleod: Additionally, in the LSF/Apptainer backend, you're using `wdl_work_dir_host_path()`—it makes sense to use that here instead of manually joining? Same comment for the rest of these.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,886 @@
...
+                        bail!(
+                            "task requires at least {required_memory} GiB of memory{env_specific}",
+                            required_memory = required_memory.as_u64() as f64 / ONE_GIBIBYTE
+                        );
+                    }
+                }
+            }
+        }
+
+        let _max_cpu = v1::max_cpu(hints);
```
claymcleod, line 598: Are these used anywhere? I don't see them.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,886 @@
...
+    /// [`gpu_slurm_partition`][Self::gpu_slurm_partition] or
+    /// [`fpga_slurm_partition`][Self::fpga_slurm_partition] for tasks which
+    /// require specialized hardware.
+    pub short_task_slurm_partition: Option<SlurmPartitionConfig>,
+    /// Which partition, if any, to specify when submitting [tasks which require
+    /// a GPU](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#hardware-accelerators-gpu-and--fpga)
+    /// to Slurm.
+    pub gpu_slurm_partition: Option<SlurmPartitionConfig>,
+    /// Which partition, if any, to specify when submitting [tasks which require
+    /// a GPU](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#hardware-accelerators-gpu-and--fpga)
```
claymcleod:
```suggestion
    /// an FPGA](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#hardware-accelerators-gpu-and--fpga)
```

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,886 @@
...
+                }
+                trace!(stdout = line, task_name);
+            }
+        });
+        let sbatch_stderr = sbatch_child
+            .stderr
+            .take()
+            .ok_or_else(|| anyhow!("sbatch child stderr missing"))?;
+        let task_name = self.name.clone();
+        let _stderr_crankshaft_events = self.crankshaft_events.clone();
```
claymcleod: Similarly, is this used?

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,886 @@
...
+                        "failed to validate {name}_slurm_partition `{partition}`"
+                    ))
+                } else {
+                    Ok(())
+                }
+            }
+            Err(_) => Err(anyhow!(
+                "timed out trying to validate {name}_slurm_partition `{partition}`"
+            )),
+        }
```
acfoltzer, line 742: Use `scontrol show partition` instead

crates/wdl-engine/src/backend/slurm_apptainer.rs
acfoltzer: Addressed the second point in d809089b. As for the first, there's enough variation between these calls (whether it just creates vs writes, whether it's a directory or a file, whether it sets permissions, what the semantic string is in the error) that I don't think this is worth changing.

crates/wdl-engine/src/backend/slurm_apptainer.rs
acfoltzer: Added a TODO for further investigation of these in d809089b

crates/wdl-engine/src/backend/slurm_apptainer.rs
acfoltzer: Got this in d809089b, along with the same copy-paste error that made it into the LSF backend

crates/wdl-engine/src/backend/slurm_apptainer.rs
acfoltzer: I sure was hoping it would be, but as I describe in the comment above this, `sbatch --wait` doesn't offer the relevant output here. Removed this in d809089b

crates/wdl-engine/src/backend/slurm_apptainer.rs
acfoltzer: d809089b

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,879 @@
...
+            // anything, but can be useful for debugging if the scripts get modified.
+            .arg("-o")
+            .arg(slurm_stdout_path)
+            .arg("-e")
+            .arg(slurm_stderr_path)
+            // An explicit task count is required for some options
+            .arg("--ntasks=1")
+            // CPU request is rounded up to the nearest whole CPU
+            .arg(format!(
+                "--cpus-per-task={}",
```
claymcleod, line 248: Since we're using `--cpus-per-task` here, would it be worth doing `--gpus-per-task` above instead of `--gpus-per-node`?

acfoltzer: Good catch! 63a8304e

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,879 @@
...
+                self.required_cpu.ceil() as u64
+            ))
+            // Memory request is specified per node in megabytes; we round the request up to the
+            // next megabyte
+            .arg(format!(
+                "--mem={}",
+                (self.required_memory.as_u64() as f64 / bytesize::MIB as f64).ceil() as u64
+            ))
+            .arg(apptainer_command_path);
+
```
claymcleod, line 266: Does it make sense to `debug!()`/`trace!()` the full command here since it's not being written out to a file?

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,879 @@
...
+        // json suitable for this.
+        let sbatch_stdout = sbatch_child
+            .stdout
+            .take()
+            .ok_or_else(|| anyhow!("sbatch child stdout missing"))?;
+        let task_name = self.name.clone();
+        let stdout_crankshaft_events = self.crankshaft_events.clone();
+        tokio::spawn(async move {
+            let mut lines = BufReader::new(sbatch_stdout).lines();
+            while let Ok(Some(line)) = lines.next_line().await {
```
claymcleod, line 295: Since we're scanning for a specific line here, does it make sense to do something in the error case—even if that's just ignoring the error?

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,879 @@
...
+            }
+        });
+        let sbatch_stderr = sbatch_child
+            .stderr
+            .take()
+            .ok_or_else(|| anyhow!("sbatch child stderr missing"))?;
+        let task_name = self.name.clone();
+        tokio::spawn(async move {
+            let mut lines = BufReader::new(sbatch_stderr).lines();
+            while let Ok(Some(line)) = lines.next_line().await {
```
claymcleod, line 320: Same question here—if we encounter an error, do we want this to stop printing or should we keep printing and just skip the error line?

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -0,0 +1,879 @@
...
+            // CPU request is rounded up to the nearest whole CPU
+            .arg(format!(
+                "--cpus-per-task={}",
+                self.required_cpu.ceil() as u64
+            ))
+            // Memory request is specified per node in megabytes; we round the request up to the
+            // next megabyte
+            .arg(format!(
+                "--mem={}",
+                (self.required_memory.as_u64() as f64 / bytesize::MIB as f64).ceil() as u64
```
claymcleod, line 263:
```suggestion
                (self.required_memory.as_u64() as f64 / bytesize::MB as f64).ceil() as u64
```

From the docs, the default units are in _megabytes_ (https://slurm.schedmd.com/sbatch.html).

acfoltzer: The docs do say "megabytes", but if you look at the other places where unit suffixes are specified in more detail like `tres` and `gres`, they are interpreted as the base-2 units.

Frustratingly, I cannot find first-party Slurm documentation that explicitly says this is the case for memory, but searching around the web one can find multiple third-party sources that say it's base-2, for example https://info.nrao.edu/computing/guide/cluster-processing/appendix/memory-options and https://wcmscu.atlassian.net/wiki/spaces/WIKI/pages/327731/Using+Slurm

Because this is annoyingly non-obvious, I have extended the comment here and also added the explicit(ly ambiguous) `M` suffix in 7fdd28d6

crates/wdl-engine/src/backend/slurm_apptainer.rs
acfoltzer: Good idea; 7fdd28d6 adds this

crates/wdl-engine/src/backend/slurm_apptainer.rs
acfoltzer: I'm not confident we can do this in a way that doesn't risk an infinite loop unless we do some careful digging through the tokio sources to determine what kinds of error kinds can arise here. I'm also skeptical that we'd be able to reliably still find the line we're looking for if some non-UTF-8 output has shown up before it.

Since the consequences of getting this wrong are a misreporting of the number of active tasks to the CLI, and a premature interruption of `trace`-level output, I'm inclined to not open this can of worms unless we actually see an instance of it in practice.

PR #439, opened 2025-10-22
feat: adds GPU support to the Docker backend

claymcleod:
This PR adds GPU support to the Docker backend and centralizes the calculation of the number of GPU requests across backends. It's related to https://github.com/stjude-rust-labs/crankshaft/pull/64.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -187,29 +187,11 @@ impl TaskManagerRequest for LsfApptainerTaskRequest {
         // If GPUs are required, pass a basic `-gpu` flag to `bsub`. If this is a bare
         // `requirements { gpu: true }`, we request 1 GPU per host. If there is
         // also an integer `hints: { gpu: n }`, we request `n` GPUs per host.
```
acfoltzer: Trimming this comment down, since the logic now lives over in `v1::gpu()`
```suggestion
        // If GPUs are required, pass a basic `-gpu` flag to `bsub`.
```

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -235,6 +240,54 @@ pub(crate) fn max_memory(hints: &HashMap<String, Value>) -> Result<Option<i64>>
...
+            }
+            // This should be validated prior to calling this function
+            Some(_) => unreachable!("`gpu` hint should be an integer or string"),
+            None => DEFAULT_GPU_COUNT,
+        };
+        Some(count)
+    } else {
+        None
+    }
+}
```
acfoltzer, line 296: Seeing this as a standalone function made me rethink a lot of my original code that you based this on, so this is as much a suggestion to my past self as it is a commentary on the new code. There are several style changes and one semantic change; if you like some but not the others I'm happy to rewrite the suggestion to include a subset:

- Semantics: if the GPU hint is not a positive integer, emit a warning and return `None` rather than falling back to the default. I feel fairly strongly that this should be the behavior when the hint is zero. If it's negative, well, that's just weird but I think it's somewhat less surprising to request no GPU in this case than to fall back to a default. I'm also receptive to the idea that this function should return `Result` instead of `Option`, and a negative should be an `Err`.
- Style: nested conditionals can get hard to follow, and now that it's a standalone function we can take advantage if let-else to tell a more top-to-bottom rather than left-to-right story of how we determine the request.
- Style: since this is not a public function, the audience for the docstring is internal development, which is better served by inline comments describing the behavior as it is implemented.

```suggestion
/// Gets the number of required GPUs from requirements and hints.
pub(crate) fn gpu(
    requirements: &HashMap<String, Value>,
    hints: &HashMap<String, Value>,
) -> Option<u64> {
    // If `requirements { gpu: false }` or there is no `gpu` requirement, return
    // `None`.
    let Some(true) = requirements
        .get(TASK_REQUIREMENT_GPU)
        .and_then(|v| v.as_boolean())
    else {
        return None;
    };

    // If there is no `gpu` hint giving us more detail on the request, use the
    // default count.
    let Some(hint) = hints.get(TASK_HINT_GPU) else {
        return Some(DEFAULT_GPU_COUNT);
    };

    // A string `gpu` hint is allowed by the spec, but we do not support them yet.
    // Fall back to the default count.
    //
    // TODO(clay): support string hints for GPU specifications.
    if let Some(hint) = hint.as_string() {
        warn!(
            %hint,
            "string `gpu` hints are not supported; falling back to {DEFAULT_GPU_COUNT} GPU(s)"
        );
        return Some(DEFAULT_GPU_COUNT);
    }

    match hint.as_integer() {
        // If the hint is a positive integer, that's our request.
        Some(count) if count >= 1 => Some(count as u64),
        // If the hint is zero or negative, it's not clear what the user intends. Maybe they have
        // tried to disable GPUs by setting the count to zero, or have made a logic error. Emit a
        // warning, and continue with no GPU request.
        Some(count) => {
            warn!(
                %count,
                "`gpu` hint specified {count} GPU(s); no GPUs will be requested for execution"
            );
            None
        }
        None => {
            // Typechecking should have already validated that the hint is an integer or
            // a string.
            error!(?hint, "`gpu` hint must be an integer or string");
            unreachable!("`gpu` hint must be an integer or string")
        }
    }
}
```

vendor/crankshaft
acfoltzer, line 1: Adding a comment here to make sure this is updated to a `main` commit before merging

crates/wdl-engine/src/backend/docker.rs
```diff
@@ -402,11 +406,15 @@ impl TaskExecutionBackend for DockerBackend {
             }
         }
 
+        let gpu = gpu(requirements, hints)
+            .map(|count| (0..count).map(|i| format!("gpu-{i}")).collect())
+            .unwrap_or_default();
```
acfoltzer, line 416: This could use a comment explaining the significance of these strings. This falls under "execution engine-specific", but since this is the first backend that puts anything other than `vec![]` in this field, we should probably document that we're picking this as our format for other backends to emulate.

acfoltzer (changes requested):

vendor/crankshaft
claymcleod: This has been updated.

acfoltzer (approved):
🚀

PR #444, opened 2025-10-30
feat: implement "slow" vs "fast" failure modes.

peterhuene:
This PR implements both a "slow" and a "fast" failure mode in Sprocket.

The "slow" failure mode waits for currently executing tasks to complete when an error is encountered or the user interrupts evaluation.

The "fast" failure mode immediately cancels the currently executing tasks and waits for cancellation to complete.

The previous implementation of Sprocket behaved like the "fast" failure mode; the new default is "slow" and can be changed via a setting in `sprocket.toml`.

Additionally, Ctrl-C handling has been reimplemented to support multiple Ctrl-C invocations depending on the failure mode.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/src/eval.rs
+++ b/crates/wdl-engine/src/eval.rs
@@ -1062,23 +1062,47 @@ mod test {
     }
 
     #[test]
-    fn cancellation_error() {
+    fn cancellation_error_slow() {
         let context = CancellationContext::new(FailureMode::Slow);
         assert_eq!(context.state(), CancellationContextState::NotCanceled);
 
-        // An error should cancel the context
+        // An error should not cancel the token
         context.error(&EvaluationError::Canceled);
         assert_eq!(context.state(), CancellationContextState::Waiting);
         assert!(!context.user_canceled());
         assert!(!context.token.is_cancelled());
 
-        // A repeated error should not cancel the token
+        // A repeated error should not cancel the token either
         context.error(&EvaluationError::Canceled);
         assert_eq!(context.state(), CancellationContextState::Waiting);
         assert!(!context.user_canceled());
         assert!(!context.token.is_cancelled());
 
         // However, another cancellation will
+        assert_eq!(context.cancel(), CancellationContextState::Canceling);
+        assert_eq!(context.state(), CancellationContextState::Canceling);
+        assert!(!context.user_canceled());
+        assert!(context.token.is_cancelled());
+    }
+
+    #[test]
+    fn cancellation_error_fast() {
+        let context = CancellationContext::new(FailureMode::Fast);
+        assert_eq!(context.state(), CancellationContextState::NotCanceled);
+
+        // An error should cancel the context
+        context.error(&EvaluationError::Canceled);
+        assert_eq!(context.state(), CancellationContextState::Canceling);
+        assert!(!context.user_canceled());
+        assert!(context.token.is_cancelled());
+
+        // A repeated error should not change anything
+        context.error(&EvaluationError::Canceled);
+        assert_eq!(context.state(), CancellationContextState::Canceling);
+        assert!(!context.user_canceled());
+        assert!(context.token.is_cancelled());
+
+        // Neither should another `cancel` call
         assert_eq!(context.cancel(), CancellationContextState::Canceling);
         assert_eq!(context.state(), CancellationContextState::Canceling);
         assert!(!context.user_canceled());


crates/wdl-engine/src/eval.rs
```diff
@@ -49,6 +55,167 @@ const MAX_STDERR_LINES: usize = 10;
...
+
+/// State used in a [`CancellationContext`].
+#[derive(Default)]
+struct CancellationContextState {
+    /// Whether or not the underlying cancellation token is currently
+    /// cancellable.
+    ///
+    /// This is initially `false` for the "slow" failure mode where the first
+    /// cancellation should transition to "fast" failure mode and the second
+    /// call to `cancel` should cancel the underlying cancellation token.
```
adthrasher: I think I understand this, but I've had to read it a couple of times. When in "slow" failure mode, either an error or a ctrl+c invocation transitions the engine to a "failed" state, where it allows running tasks to continue to completion. When in this failure state, a subsequent ctrl+c (which may be the first or second invocation), but not a task failure, will cause it to "fast" fail and terminate any running tasks. Assuming that my understanding is correct, I think `should transition to "fast" failure mode` could be rewritten. That state isn't the same as the "fast" failure mode, in that, a task failure won't trigger an immediate exit. It is similar to the "fast" failure mode only when the user requests an exit with ctrl+c. Though maybe when I get to the implementation below it will be clearer.

peterhuene: Agreed, this could be reworded better. Your synopsis is correct. I'll amend.

src/commands/run.rs
```diff
@@ -580,42 +580,50 @@ pub async fn run(args: Args) -> Result<()> {
...
+    let mut evaluate = evaluator.run(cancellation.clone(), events).boxed();
+
+    loop {
+        select! {
+            // Always prefer the CTRL-C signal to the evaluation returning.
+            biased;
+
+            _ = tokio::signal::ctrl_c() => {
+                // If the cancellation token was canceled, the user doesn't want to wait for evaluation to complete
+                if cancellation.token().is_cancelled() {
```
claymcleod: nit: should this be `cancellation.is_cancelled()` to handle the case where an error has occurred and we want to interrupt it from waiting for the tasks to finish without having to press Ctrl+C twice? This is also a comment that the line just semantically could be clearer.

crates/wdl-engine/src/eval.rs
```diff
@@ -49,6 +55,167 @@ const MAX_STDERR_LINES: usize = 10;
...
+    /// failure mode. A subsequent call to [`cancel`][Self::cancel] will cancel
+    /// the underlying cancellation token.
+    ///
+    /// If the initial failure mode was [`Fast`](FailureMode::Fast), this will
+    /// immediately cancel the underlying cancellation token.
+    ///
+    /// Returns `true` if the underlying cancellation token was canceled
+    /// (i.e. currently `Fast` mode) or `false` if it was not (i.e. previously
+    /// `Slow` mode).
+    pub fn cancel(&self) -> bool {
```
claymcleod: Technically, there is an unlikely race condition here that is due to the state being spread across two different atomics. 

Consider the following scenario:

- `cancel()` is called, setting `reason` to `CANCEL_REASON_USER`.
- before `cancelable` can be checked, `error()` is called and its `compare_exchange` for `reason` fails (the method does nothing because it's already `CANCEL_REASON_USER`).
- the aforementioned `cancel()` continues, seeing that `cancelable` is false and doesn't cancel the token.

In this way, two canceling events can happen without having the right effect. The way to fix this would be to have a single atomic state variable that is the dot product of the two concerns.

src/commands/run.rs
peterhuene: If the failure mode is `Slow` and an error occurred, we'll have transitioned to waiting for the executing tasks to complete and `CancellationContext::is_canceled` will return true. A Ctrl-C in that case shouldn't immediately exit the process, but instead issue another cancellation to cancel the executing tasks instead via the cancellation token.

It's only after the cancellation token has been canceled would we want to exit the process on another Ctrl-C.

Upon error or initial interruption, we should give the users using the `Slow` failure mode a chance to cleanup (cancel) the executing tasks, and that can only happen with multiple Ctrl-C invocations before we simply exit from Sprocket.

crates/wdl-engine/src/eval.rs
peterhuene: Good catch. They can easily be merged together into a single state. Let me fix that.

crates/wdl-engine/src/eval.rs
```diff
@@ -49,6 +55,167 @@ const MAX_STDERR_LINES: usize = 10;
...
+        }
+
+        // Perform a cancellation and log the error
+        let message: Cow<'_, str> = match error {
+            EvaluationError::Canceled => "evaluation was canceled".into(),
+            EvaluationError::Source(e) => e.diagnostic.message().into(),
+            EvaluationError::Other(e) => format!("{e:#}").into(),
+        };
+
+        if self.cancel() {
```
claymcleod: nit: I find it a bit confusing that `cancel()` is called with the expectation that the atomic update to the `reason` field will fail within. Given that the rest of the function is so small, it might just make sense to inline the pieces of `cancel()` that you want here to avoid confusion.

src/commands/run.rs
peterhuene: With unifying the state, I'll encapsulate this with a method to better reflect what's being checked here.

crates/wdl-engine/src/eval.rs
claymcleod: nit: it might be worth a `#[must_use]` on this method since it doesn't always perform a true cancellation.

crates/wdl-engine/src/eval.rs
```diff
@@ -49,6 +55,167 @@ const MAX_STDERR_LINES: usize = 10;
...
+                }
+                .into(),
+                reason: CANCEL_REASON_NONE.into(),
+            }),
+            token: CancellationToken::new(),
+        }
+    }
+
+    /// Determines if a cancellation has occurred.
+    pub fn is_canceled(&self) -> bool {
```
claymcleod: nit: I don't love the name `is_canceled` here, because it might indicate to some readers that the cancellation token has been cancelled. Perhaps `is_cancellation_pending()` is a better name?

src/commands/run.rs
claymcleod: Okay I think that was 75% of my original concern anyways—it seems like having this hidden behind a semantically named method would make this code a bit clearer.

crates/wdl-engine/src/eval.rs
peterhuene: This text has been removed after refactoring. Hopefully the new comments make more sense.

src/commands/run.rs
peterhuene: There's now a `state` method on the context that this code is using to determine if we are currently cancelling and thus doesn't use the underlying token to determine it.

crates/wdl-engine/src/eval.rs
peterhuene: This has been refactored such that updating a single state variable is atomically done.

crates/wdl-engine/src/eval.rs
peterhuene: This has been refactored such that `error` doesn't call `cancel` internally.

crates/wdl-engine/src/eval.rs
peterhuene: This has been renamed to `state` and returns an enumeration of effectively "not canceled", "waiting", "canceling".

crates/wdl-engine/src/eval.rs
```diff
@@ -49,6 +54,214 @@ const MAX_STDERR_LINES: usize = 10;
...
+    ///
+    /// This bit is mutually exclusive with the `WAITING` bit.
+    const CANCELING: u8 = 2;
+    /// A state bit to denote that cancellation was the result of an error.
+    ///
+    /// This bit will only be set if either the `CANCELING` bit or the `WAITING`
+    /// bit are set.
+    const ERROR: u8 = 4;
+    /// A constant to denote that no cancellation has occurred yet.
+    const NOT_CANCELED: u8 = 0;
```
peterhuene: Eww, why is `cargo fmt` rearranging these declarations?

peterhuene: So `reorder_impl_items` controls this, but it also orders associated items in trait impls, which I agree should always come before the functions.

Unfortunate sorting of the consts here; I'll move to not be impl-associated consts.

Issue #225, opened 2025-07-04
feat(wdl-doc): >=v1.2 struct documentation


a-frantz:
**Is your feature request related to a problem? Please describe.**
Currently (including stjude-rust-labs/wdl#339), wdl-doc treats all structs as equivalent, with rather minimal documentation (just copying the raw WDL definition in a syntax highlighted code block). But in WDLv1.2, meta sections were introduced to struct definitions, which allows for much richer documentation similar to the current treatment of tasks and workflows.

**Describe the solution you'd like**
WDL structs with meta sections should get richer documentation

**Additional context**
This not been implemented yet due to the fact not many of our users are writing v1.2 WDL yet.

PR #445, opened 2025-10-31
feat: adds enumerations

claymcleod:
This PR adds enumerations. It's being developed in tandem with https://github.com/openwdl/wdl/pull/695 as a proof of concept of the implementation (please browse that PR for a formal specification of what's being implemented here). Sorry in advance for the large changeset—come to find out it takes a _lot_ of work to get enumerations pushed all the way through `wdl-*` and `sprocket`.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/src/value.rs
+++ b/crates/wdl-engine/src/value.rs
@@ -2091,11 +2091,7 @@ impl EnumVariant {
     /// # Panics
     ///
     /// Panics if the given variant name is not present in the enum type.
-    pub fn new(
-        enum_ty: impl Into<EnumType>,
-        name: &str,
-        value: impl Into<Value>,
-    ) -> Self {
+    pub fn new(enum_ty: impl Into<EnumType>, name: &str, value: impl Into<Value>) -> Self {
         let enum_ty = enum_ty.into();
         let value = Arc::new(value.into());
 


crates/wdl-engine/tests/workflows.rs
```diff
@@ -64,7 +64,20 @@ fn run_test(test: &Path, config: TestConfig) -> BoxFuture<'_, Result<()>> {
...
         }
         if result.document().has_errors() {
-            bail!("test WDL contains errors; run a `check` on `source.wdl`");
+            let errors: Vec<_> = result
+                .document()
+                .diagnostics()
+                .filter(|d| d.severity() == Severity::Error)
+                .collect();
+            bail!(
+                "test WDL contains {} error(s):\n{}",
```
claymcleod, line 73: I run into this frequently, so I figure it's just a slight improvement to list the errors that are being encountered.

crates/wdl-analysis/src/document.rs
```diff
@@ -29,7 +29,7 @@ use crate::types::CallType;
 use crate::types::Optional;
 use crate::types::Type;
 
-mod v1;
+pub mod v1;
```
claymcleod, line 32: This seems scary at first glance. But the only thing being exported is `infer_type_from_literal`, which I want to use downstream in `wdl-engine` for DRY.

crates/wdl-format/src/v1/expr.rs
```diff
@@ -452,6 +452,11 @@ pub fn format_literal_object_item(element: &FormatElement, stream: &mut TokenStr
 pub fn format_literal_object(element: &FormatElement, stream: &mut TokenStream<PreToken>) {
     let mut children = element.children().expect("literal object children");
 
+    let object_keyword = children.next().expect("literal object keyword");
```
claymcleod, line 455: This was an unrelated bug that I caught in our formatting. Guess we haven't really used `object` that much.

crates/wdl-lsp/tests/goto_definition.rs
```diff
@@ -87,7 +87,7 @@ async fn should_goto_imported_task_definition() {
     assert_eq!(location.uri, ctx.doc_uri("lib.wdl"));
     assert_eq!(
         location.range,
-        Range::new(Position::new(2, 5), Position::new(2, 8)) // `task add` in lib.wdl
+        Range::new(Position::new(8, 5), Position::new(8, 8)) // `task add` in lib.wdl
```
claymcleod, line 90: These are unrelated offsets due to me adding an enum definition to `lib.wdl`.

crates/wdl-analysis/CHANGELOG.md
```diff
@@ -9,12 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 #### Added
 
+* Added type analysis and validation for WDL enumerations in preparation for WDL v1.3 ([#445](https://github.com/stjude-rust-labs/sprocket/pull/445)).
 * Added support for `else if` and `else` clauses in conditional statements (in support of WDL v1.3) ([#411](https://github.com/stjude-rust-labs/sprocket/pull/411)).
 * Added feature flags support to enable experimental WDL versions ([#411](https://github.com/stjude-rust-labs/sprocket/pull/411)).
 * Introduced pre-evaluation task type for all pre-evaluation contexts (task requirements, task hints, and task runtime sections) and expanded support of `task.previous` for post-evaluation sections in WDL v1.3 ([#432](https://github.com/stjude-rust-labs/sprocket/pull/432)).
 
 #### Changed
 
+* Improved error messages for enum type mismatches to include specific type information ([#445](https://github.com/stjude-rust-labs/sprocket/pull/445)).
```
adthrasher: It seems odd to have a "changed" entry for a PR introducing new features.

crates/wdl-doc/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+* Added documentation generation support for WDL enumerations in preparation for WDL v1.3 ([#445](https://github.com/stjude-rust-labs/sprocket/pull/445)).
```
adthrasher:
```suggestion
* Added documentation generation placeholder for WDL enumerations in preparation for WDL v1.3 ([#445](https://github.com/stjude-rust-labs/sprocket/pull/445)).
```

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -388,9 +465,63 @@ fn are_structs_equal(a: &StructDefinition, b: &StructDefinition) -> bool {
...
+    // Compare variants
+    let mut vars_a: Vec<_> = a.variants().collect();
+    let mut vars_b: Vec<_> = b.variants().collect();
+
+    if vars_a.len() != vars_b.len() {
+        return false;
+    }
+
+    vars_a.sort_by(|a, b| a.name().text().cmp(b.name().text()));
+    vars_b.sort_by(|a, b| a.name().text().cmp(b.name().text()));
```
peterhuene: So while this isn't called out in the enum proposal to the spec, I would like to formalize equivalence of enum _definitions_ to make variant definition order relevant.

We have proposed stdlib functions that were kicked down the road relating to being able to enumerate an enum's variants. To support such a function in the future, the definition order of the variants should matter.

I would propose that we just zip the iterators without attempting to sort them.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -445,25 +576,108 @@ fn add_struct(document: &mut DocumentData, definition: &StructDefinition) {
...
+fn add_enum(document: &mut DocumentData, definition: &EnumDefinition) {
+    let name = definition.name();
+
+    // Check if enums are supported in this version
+    let version = document.version.expect("should have version");
+    if version < SupportedVersion::V1(V1::Three) {
+        document
+            .analysis_diagnostics
+            .push(enum_not_supported(version, definition.span()));
+    }
```
peterhuene, line 592: Should we be returning here and effectively ignoring the definition?

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -921,11 +1135,13 @@ fn populate_workflow(config: &Config, document: &mut DocumentData, workflow: &Wo
     )];
     let mut output_scope = None;
 
+    let compiled_enum_names: HashSet<String> = document.enums.keys().cloned().collect();
```
peterhuene: We already have a map built, so I'd prefer not to create a new collection for an existing `O(1)` lookup data structure.

Perhaps we can give it a callback of type `Fn(&str) -> bool` and we simply close over `dcument.enums`?

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1703,6 +1919,171 @@ fn set_struct_types(document: &mut DocumentData) {
...
+                    // Default to `String` type if no value is specified
+                    PrimitiveType::String.into()
+                };
+                variants.push((variant_name, variant_type));
+            }
+
+            let type_param = definition.type_parameter().map(|t| t.ty());
+            Some((index, name.clone(), variants, type_param, e.name_span))
+        })
+        .collect();
```
peterhuene: I think the `collect` here might not be necessary; seems like we could just use the iterator.

crates/wdl-analysis/src/validation/counts.rs
```diff
@@ -75,7 +75,9 @@ impl fmt::Display for Section {
 
 /// Creates a "at least one definition" diagnostic
 fn at_least_one_definition() -> Diagnostic {
-    Diagnostic::error("there must be at least one task, workflow, or struct definition in the file")
+    Diagnostic::error(
+        "there must be at least one task, workflow, struct, or enum definition in the file",
```
peterhuene: I think we need to make this diagnostic version dependent.

I wouldn't want to see "or an enum definition" from a < 1.3 document.

crates/wdl-analysis/src/validation/counts.rs
```diff
@@ -322,6 +327,19 @@ impl Visitor for CountingVisitor {
...
+        &mut self,
+        _: &mut Diagnostics,
+        reason: VisitReason,
+        _: &wdl_ast::v1::EnumDefinition,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        self.has_enum = true;
```
peterhuene: We should only set this flag for >= 1.3 documents. Generally the experience of validation has been that we produce a singular "this thing isn't supported in your WDL version" diagnostic and then otherwise ignore it.

Thus I would expect an `there must be at least one task, workflow, or struct in the file` diagnostic in a < 1.3 document that just happens to have an enum in it.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1703,6 +1919,171 @@ fn set_struct_types(document: &mut DocumentData) {
...
+            }
+            LiteralExpr::Object(obj) => {
+                for item in obj.items() {
+                    let (_, val_expr) = item.name_value();
+                    infer_type_from_literal(&val_expr)?;
+                }
+                Some(Type::Object)
+            }
+            LiteralExpr::Struct(_)
+            | LiteralExpr::None(_)
```
peterhuene: I feel like `None` should be supported; it's a literal expression with a known constant value and as far as I know we're not limiting the enum representation type to non-optional types.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1703,6 +1919,171 @@ fn set_struct_types(document: &mut DocumentData) {
...
+        }
+    }
+}
+
+/// Infers the type of a literal expression.
+///
+/// Returns `None` if the expression is not a literal or contains interpolation.
+/// For struct literals, returns `None` since struct type information is not
+/// available.
+pub fn infer_type_from_literal(expr: &Expr) -> Option<Type> {
```
peterhuene, line 2006: My take on this is that this is probably sufficient for an initial release of enums in WDL, but it feels quite limiting.

I would like to see the spec define a _constant expression_ in terms of:

1. A _constant primitive literal_ expression: `None`, `true`, `false`, integer literal, float literal, or string literal without interpolation.
2. A _constant compound literal_ expression where any inner expressions are also _constant expressions_: pair, array, map, object, or struct (I can't see a reason not to support struct, as opposed to enums which don't make sense to support).
3. Any unary, binary, or ternary expression where the operands are also _constant expressions_.
4. Specifically excepting call expressions since there currently are no "constant functions".

In the future I think we could extend `ExprTypeEvaluator` such that if given a "constant" evaluation context we error upon encountering a non-const expression, rather than have this code here.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1703,6 +1919,171 @@ fn set_struct_types(document: &mut DocumentData) {
...
+                ))
+            }
+            LiteralExpr::Object(obj) => {
+                for item in obj.items() {
+                    let (_, val_expr) = item.name_value();
+                    infer_type_from_literal(&val_expr)?;
+                }
+                Some(Type::Object)
+            }
+            LiteralExpr::Struct(_)
```
peterhuene, line 2064: Can we think of a reason (complexity?) why structs couldn't be constant literals?

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1756,22 +2137,31 @@ impl crate::types::v1::EvaluationContext for EvaluationContext<'_> {
     }
 
     fn resolve_name(&self, name: &str, _: Span) -> Option<Type> {
-        self.scope.lookup(name).map(|n| n.ty().clone())
+        self.scope
+            .lookup(name)
+            .map(|n| n.ty().clone())
+            // Enum types can be the target of member access directly
+            .or_else(|| self.document.enums.get(name).and_then(|e| e.ty().cloned()))
```
peterhuene: I don't think this should change.

Instead, `ExprTypeEvaluator` should handle the type of an access expression similarly to how we have to handle it in `wdl-engine`: if the LHS expression is _specifically_ a name reference expression to a name that is not in scope (i.e. `resolve_name` returns `None`) and also names an enum type then we _do not_ evaluate the LHS expression.

This would cause _any_ name reference expression that names an enum to evaluate to the type of the enum, which is not desired.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1945,4 +2335,50 @@ mod tests {
         assert_eq!(err.len(), 1);
         assert!(err[0].message().contains("type mismatch"));
     }
+
+    #[test]
+    fn enums_equal_different_variant_order() {
```
peterhuene: Per my comments on the proposal for including variant order as part of definition equality, I think we can remove this test.

crates/wdl-analysis/src/eval/v1.rs
```diff
@@ -599,9 +600,21 @@ pub struct WorkflowGraphBuilder<N: TreeNode = SyntaxNode> {
...
+    /// create runtime dependencies.
+    compiled_enum_names: HashSet<String>,
 }
 
 impl<N: TreeNode> WorkflowGraphBuilder<N> {
+    /// Sets the compiled enum names for the builder.
+    ///
+    /// These are enum type names that are valid references but do not create
+    /// runtime dependencies in the workflow graph.
+    pub fn with_compiled_enum_names(mut self, names: HashSet<String>) -> Self {
```
peterhuene: Per my earlier comment, I think this would be better accomplished via a `Fn(&str) -> bool` parameter to `build` rather than via another collection.

crates/wdl-analysis/src/eval/v1.rs
```diff
@@ -1059,7 +1072,11 @@ impl<N: TreeNode> WorkflowGraphBuilder<N> {
                     }
                 }
                 _ => {
-                    diagnostics.push(unknown_name(name.text(), name.span()));
+                    // Check if this is a compiled enum name
+                    // These are valid references but don't create runtime dependencies
+                    if !self.compiled_enum_names.contains(name.text()) {
```
peterhuene: I'd also expect a similar change to `TaskGraphBuilder::add_expr_edges`.

crates/wdl-analysis/src/handlers/completions.rs
```diff
@@ -830,6 +846,20 @@ fn add_struct_completions(document: &Document, items: &mut Vec<CompletionItem>)
     }
 }
 
+/// Adds completions for enum types.
+fn add_enum_type_completions(document: &Document, items: &mut Vec<CompletionItem>) {
```
peterhuene: As I mentioned on the proposal, WDL currently does not prohibit a local name from shadowing a custom type name.

I think `add_enum_type_completions` (and `add_struct_type_completions`) should take an optional scope and filter the completions to any type name _not_ in scope.

crates/wdl-analysis/src/handlers/common/type_eval.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Evaluates the type of an expression, with special handling for enum type
+/// names.
+///
+/// When the expression is a name reference to an enum type (e.g., `Status` in
+/// `Status.Active`), this function returns the enum type directly instead of
+/// evaluating it as a variable reference. For all other cases, it performs
+/// standard type evaluation.
+///
+/// Returns the evaluated type, or `Type::Union` if evaluation fails.
+pub fn evaluate_expr_type(expr: &v1::Expr, scope: ScopeRef<'_>, document: &Document) -> Type {
```
peterhuene: I don't this we should be doing this special handling given to how we're special casing the access expression specifically to support enums.

That is to say not _all_ name reference expressions should return the type of an enum; only specifically the one that is in an access expression where the LHS is directly a name reference expression with a name not in scope that names an enum type.

crates/wdl-analysis/src/handlers/completions.rs
peterhuene, line 403: In lieu of the `evaluate_expr_type` implementation, here is where I'd expect us to handle enum access (given a 1.3 document).

We would need to check the name isn't in scope and names an enum; if so, add its completions.

crates/wdl-analysis/src/handlers/completions.rs
```diff
@@ -508,6 +513,17 @@ fn add_member_access_completions(
                 ..Default::default()
             });
         }
+        (SyntaxKind::Dot, Type::Compound(CompoundType::Enum(e), _)) => {
```
peterhuene: With my above suggestions, I'd expect this match to no longer be necessary.

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -455,14 +475,31 @@ fn resolve_access_expression(
...
-    let mut ctx = TypeEvalContext {
-        scope,
-        document: analysis_doc,
-    };
-    let mut evaluator = ExprTypeEvaluator::new(&mut ctx);
-    let target_type = evaluator
-        .evaluate_expr(&target_expr)
-        .unwrap_or(crate::types::Type::Union);
+    // Check if target is a namespace reference
+    if let v1::Expr::NameRef(name_ref) = &target_expr {
```
peterhuene, line 498: This is more inline with what I was expecting to see in completions, by the way.

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -455,14 +475,31 @@ fn resolve_access_expression(
...
+        let name = name_ref.name().text().to_string();
```
peterhuene: Is the `to_string` necessary here?

crates/wdl-analysis/src/handlers/hover.rs
```diff
@@ -288,6 +295,18 @@ fn resolve_hover_by_context(
                     "right" => (Some(p.right_type().clone()), None),
                     _ => (None, None),
                 },
+                Type::Compound(CompoundType::Enum(e), _) => {
```
peterhuene: Similar to the comments in `completions.rs`, I'd expect this to be done through special casing of access expression.

crates/wdl-analysis/src/types/v1.rs
```diff
@@ -1738,6 +1746,15 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
                     .add_diagnostic(not_a_struct_member(ty.name(), &name));
                 return None;
             }
+            Type::Compound(CompoundType::Enum(ty), _) => {
```
peterhuene: Similar to previous comments, we should not be evaluating the expression if all of the following are true:

* It is a name reference expression
* The name is not in scope
* The name is to an enum type

crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -531,6 +590,13 @@ pub fn else_not_supported(version: SupportedVersion, span: Span) -> Diagnostic {
     .with_fix("use WDL v1.3 or higher to use `else` conditional clauses")
 }
 
+/// Creates an "enum not supported" diagnostic.
+pub fn enum_not_supported(version: SupportedVersion, span: Span) -> Diagnostic {
+    Diagnostic::error(format!("enumerations are not supported in WDL v{version}"))
```
peterhuene: We should be consistent in the diagnostics; seems like we use `enum` mostly:

```suggestion
    Diagnostic::error(format!("enums are not supported in WDL v{version}"))
```

crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -531,6 +590,13 @@ pub fn else_not_supported(version: SupportedVersion, span: Span) -> Diagnostic {
...
+        .with_label("this enumeration is not supported", span)
```
peterhuene:
```suggestion
        .with_label("this enum is not supported", span)
```

crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -531,6 +590,13 @@ pub fn else_not_supported(version: SupportedVersion, span: Span) -> Diagnostic {
...
+        .with_fix("use WDL v1.3 or higher to use enumerations")
```
peterhuene:
```suggestion
        .with_fix("use WDL v1.3 or higher to use enums")
```

crates/wdl-analysis/src/document.rs
peterhuene: One of the reasons I think we could probably fold its logic into `ExprTypeEvaluator`.

crates/wdl-analysis/src/stdlib.rs
```diff
@@ -535,6 +550,77 @@ impl GenericMapType {
...
+                    .params
+                    .get(self.ty.variant_param)
+                    .expect("variant parameter should be present");
+
+                match variant_ty {
+                    Some(Type::Compound(CompoundType::Enum(enum_ty), _)) => {
+                        enum_ty.value_type().fmt(f)
+                    }
+                    // NOTE: non-enums should gracefully fail.
+                    _ => write!(f, "T"),
```
peterhuene: Should this be:


```suggestion
                    _ => write!(f, "{}", self.ty.variant_param),
```

crates/wdl-analysis/src/types.rs
```diff
@@ -911,6 +949,134 @@ impl Coercible for StructType {
     }
 }
 
+/// Represents the type of an enum.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EnumType {
+    /// The name of the enum.
+    name: Arc<String>,
+    /// The common coerced type computed from all variant values.
+    value_type: Arc<Type>,
```
peterhuene: This can just be `Type`.

A reminder to myself: I need to clean these type definitions up a little.

crates/wdl-analysis/src/stdlib/constraints.rs
```diff
@@ -126,6 +127,7 @@ impl Constraint for JsonSerializableConstraint {
                         && type_is_serializable(ty.value_type())
                 }
                 CompoundType::Struct(s) => s.members().values().all(type_is_serializable),
+                CompoundType::Enum(e) => type_is_serializable(e.value_type()),
```
peterhuene: I think this should be `=> true` as enum values are always serializable (they serialize as the variant name, no?).

crates/wdl-analysis/src/types.rs
```diff
@@ -911,6 +949,134 @@ impl Coercible for StructType {
...
+    /// The variants with their actual types (before coercion).
+    variants: Arc<IndexMap<String, Type>>,
```
peterhuene: Is there value in storing the original types of the variants vs. just an `IndexSet<String>` here?

For static analysis, evaluating a variant definition's expression (which should be a _constant expression_), the resulting type must coerce to either the declared value type of the enum or the common infered type; once that takes place, I don't see us particuarly caring about the "original" type of the variant as for all intense and purposes `value(EnumValue)` always produces a type that is the enum's value type and never what the type of the initialization expression was for a particular variant.

I looked though where we're using the `variants` method and we pretty much never use the value in the map; the one place we do I think isn't correct (I'll have a separate comment for it).

crates/wdl-analysis/tests/analysis/enum-requires-feature-flag/source.diagnostics
```diff
@@ -0,0 +1,12 @@
+error: enumerations are not supported in WDL v1.2
+  ┌─ tests/analysis/enum-requires-feature-flag/source.wdl:5:1
+  │  
+5 │ ╭ enum Status {
+6 │ │     Active,
+7 │ │     Pending,
+8 │ │     Complete
+9 │ │ }
+  │ ╰─^ this enumeration is not supported
```
peterhuene: I think we can probably scope the span of this diagnostic to the enum keyword, perhaps.

The span for the entire enum declaration could be quite long.

crates/wdl-analysis/tests/analysis/structs-dont-conflict-with-workflow/source.wdl
```diff
@@ -0,0 +1,11 @@
+version 1.2
+
+struct MyWorkflow {
```
peterhuene, line 3: Same comment as above re: name conflict diagnostic expected.

crates/wdl-analysis/tests/analysis/enums-dont-conflict-with-task/source.wdl
```diff
@@ -0,0 +1,15 @@
+version 1.3
+
+task Status {
```
peterhuene, line 3: So I think this is an existing spec bug in Sprocket where we weren't properly treating something like this as a name conflict.

To quote the spec regarding the names that must be unique in a "WDL document" namespace:

> * The namespace of an imported document equals that of the basename of the imported file by default, but may be aliased using the `as <identifier>` syntax.
>
> * A WDL document may contain a `workflow` and/or `task`s, which are names within the document's namespace.
>
> * A WDL document may contain `struct`s, which are also names within the document's namespace and usable as types in any declarations. Structs from any imported documents are copied into the document's namespace and may be aliased using the `alias <source name> as <new name>` syntax.

So I think we're incorrect in allowing this as I would amend the above spec language to basically say "any custom type" and not just "struct".

crates/wdl-analysis/tests/analysis/structs-dont-conflict-with-task/source.wdl
```diff
@@ -0,0 +1,15 @@
+version 1.2
+
+struct MyTask {
```
peterhuene, line 3: Same comment as above re: name conflict diagnostic expected.

crates/wdl-analysis/tests/analysis/enums-dont-conflict-with-workflow/source.wdl
```diff
@@ -0,0 +1,9 @@
+version 1.3
+
+# Enum with same name as workflow
+enum test {
```
peterhuene, line 4: Same comment as above re: name conflict diagnostic expected.

crates/wdl-doc/src/lib.rs
```diff
@@ -625,6 +625,7 @@ pub async fn document_workspace(config: Config) -> Result<()> {
                         .push((diff_paths(path, &cur_dir).expect("should diff paths"), page));
                 }
                 DocumentItem::Import(_) => {}
+                DocumentItem::Enum(_) => todo!("enum documentation support"),
```
peterhuene, line 628: If we don't already, we should have a tracking issue for adding enum support to `sprocket doc`.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -533,6 +536,78 @@ fn struct_member_decl(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Ma
...
+            found => {
+                let (found, span) = found
+                    .map(|(t, s)| (Some(t.describe()), s))
+                    .unwrap_or_else(|| (None, parser.span()));
+                return Err((marker, expected_found("variant name or `}`", found, span)));
+            }
+        }
+    }
+
+    marker.complete(parser, SyntaxKind::EnumDefinitionNode);
```
peterhuene: Is it possible to split out most of this into an `enum_variant` function and use `braced_items!(parser, marker, None, ENUM_VARIANT_RECOVERY_SET, enum_variant);` similar to how we parse struct items?

crates/wdl-engine/src/value.rs
```diff
@@ -2057,6 +2093,132 @@ impl fmt::Display for Struct {
...
+#[derive(Debug, Clone)]
+pub struct EnumVariant {
+    /// The type of the enum value.
+    ty: Type,
+    /// The name of the enum.
+    enum_name: Arc<String>,
+    /// The name of the variant.
+    variant_name: Arc<String>,
+    /// The value associated with this variant.
+    value: Arc<Value>,
```
peterhuene: I was expecting to see the following stored:

* The enum type associated with the variant (we have that now).
* The index into the type's variants as we don't need to store the name again.

We shouldn't need to store the associated value here, only in some place that the `value` function can access (i.e. via the evaluation context).

crates/wdl-engine/src/value.rs
```diff
@@ -706,6 +707,41 @@ impl Coercible for Value {
                     bail!("cannot coerce `None` to non-optional type `{target}`");
                 }
             }
+            Self::Primitive(PrimitiveValue::String(s)) if target.as_enum().is_some() => {
+                // String -> Enum coercion for deserializing from JSON
+                let enum_ty = target.as_enum().unwrap();
```
peterhuene: So this feels wrong because the proposal (afaik) did not define this coercion; this will otherwise allow `Color color = "Red"` to work, won't it?

If that's desired, we should codify the coercion.

crates/wdl-engine/src/eval.rs
```diff
@@ -549,6 +549,13 @@ pub trait EvaluationContext: Send + Sync {
     /// Resolves a type name to a type.
     fn resolve_type_name(&self, name: &str, span: Span) -> Result<Type, Diagnostic>;
 
+    /// Gets the enums map for this evaluation context.
+    ///
+    /// Returns `None` if the context has not loaded enums.
+    fn enums(&self) -> Option<&HashMap<String, Enum>> {
```
peterhuene: I think it would probably be best to lazy evaluate enum values instead of returning a map of names to "evaluated enums" (nor the need for the `Enum` type used here).

Perhaps a `fn resolve_variant_value(&self, name: &str, variant: &str) -> Result<Value, Diagnostic>` function instead (and called by the `value` stdlib function)?

Naively, this would look up the enum definition in the document associated with the evaluation context. From that, it would find the variant's value expression and evaluate it to produce a `Value`.

Ideally, we would cache that operation in a way that happens once per enum variant per document (the cache would need to be in the top-level evaluator).

crates/wdl-engine/src/stdlib/size.rs
```diff
@@ -256,6 +256,9 @@ async fn compound_disk_size(
 
             Ok(size)
         }
+        CompoundValue::EnumVariant(e) => {
+            calculate_disk_size(transferer, e.value(), unit, base_dir).await
```
peterhuene: An enum variant value is not the same thing as it's value.

I'd expect `size(Files.Foo)` to be an argument type error (isn't sizable); one would need `size(value(Files.Foo))` instead (at which point the argument wouldn't be an enum variant).

That is to say, I believe this should error.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -126,6 +132,197 @@ fn format_id(namespace: Option<&str>, target: &str, alias: &str, scatter_index:
...
+                $type_name, $type_name
+            );
+        };
+    };
+}
+
+/// Extracts a literal value from an AST expression.
+///
+/// Returns `None` if the value cannot be parsed (e.g., integer overflow).
+fn extract_literal_value(ty: &Type, expr: &Expr) -> Option<Value> {
```
peterhuene: This function feels like it belongs in `eval/v1/expr.rs` in the sense that we should just be evaluating a "constant" expression to produce a value (providing it a type seems redundant).

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -126,6 +132,197 @@ fn format_id(namespace: Option<&str>, target: &str, alias: &str, scatter_index:
...
+                    .variants()
+                    .get(&variant_name)
+                    .expect("variant should exist in type");
+
+                // NOTE: `Type::Union` is used as a sentinel value in
+                // `wdl-analysis` to indicate that the variant type could not be
+                // parsed. As such, skip these variants in building values.
+                if matches!(variant_type, Type::Union) {
+                    return None;
+                }
```
peterhuene: Using the variant type from the static analysis of the variant's expression is unnecessary and this is the only place that we actually get the value in the variant map (hence why I think it can just be an `IndexSet` of names from my much earlier comment).

We should be evaluating the constant expression and then coercing it to the enum variant's value type to produce the enum variant's value.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -126,6 +132,197 @@ fn format_id(namespace: Option<&str>, target: &str, alias: &str, scatter_index:
...
+            Some(Value::Compound(crate::CompoundValue::Object(
+                crate::Object::new(members?),
+            )))
+        }
+        _ => None,
+    }
+}
+
+/// Builds enums from the document's analyzed enums.
+pub(crate) fn build_enums(document: &Document) -> HashMap<String, Enum> {
```
peterhuene: To add on to my previous comment, I'd like us to use an on-demand cache of enum variant values rather than build it for every enum in the document regardless of use.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -168,6 +365,10 @@ impl EvaluationContext for WorkflowEvaluationContext<'_, '_> {
         crate::resolve_type_name(&self.state.document, name, span)
     }
 
+    fn enums(&self) -> Option<&std::collections::HashMap<String, Enum>> {
```
peterhuene: I don't see a similar implementation of this in `TaskEvaluationContext` and I suspect we need some tests of tasks that internally use enums and calling `value` (and not just tasks that take an enum value as input).

crates/wdl-doc/src/lib.rs
a-frantz: No current tracking issue, but a related issue that should probably be addressed at around the same time is #225

crates/wdl-analysis/src/document/v1.rs
claymcleod: Peter and I discussed, and collect is necessary to make the borrow checker happy here.

crates/wdl-analysis/src/document/v1.rs
claymcleod: Totally agree. I'll keep this comment unresolved so it's easily findable in the future.

crates/wdl-analysis/src/types.rs
claymcleod: I've resolved this in this PR. That said, I'll leave this comment thread for you to resolve if you want to make an issue.

crates/wdl-analysis/tests/analysis/enums-dont-conflict-with-task/source.wdl
claymcleod: Because this PR is so long, I'm going to advocate we handle this in a follow-up PR. If that sounds good to you, I'll create an issue (I think it can be addressed pretty much immediately).

crates/wdl-engine/src/value.rs
claymcleod: For the reasons we previously discussed, we _do_ need the value here, so I'll resolve this comment without addressing.

crates/wdl-engine/src/value.rs
claymcleod: I've implemented the coercion from `String` -> `EnumVariant` and `EnumVariant` -> `String` as discussed. I will make a note to go back and add this to the RFC at a later time.

crates/wdl-engine/src/eval/v1/workflow.rs
claymcleod: I've moved this to `eval/v1/expr.rs`. As previously discussed, the (newly introduced) target type argument to this function isn't redundant.

crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -839,3 +905,60 @@ pub fn invalid_regex_pattern(
...
+    )
+}
+
+/// Creates a "no common inferred type for enum" diagnostic.
+///
+/// This diagnostic occurs during enum type calculation when no common type can
+/// be inferred from the variant types.
+pub fn no_common_inferred_type_for_enum(
+    enum_name: &str,
+    last_common_variant_type: &Type,
```
peterhuene: Nit: these are uncommonly long variable names. Perhaps `common_type`, `common_span`, `discordant_type`, and `discordant_span`?

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -388,9 +466,59 @@ fn are_structs_equal(a: &StructDefinition, b: &StructDefinition) -> bool {
...
+            if a_ty.ty() != b_ty.ty() {
+                return false;
+            }
+        }
+        (None, None) => {}
+        _ => return false,
+    }
+
+    let vars_a = a.variants().collect::<Vec<_>>();
+    let vars_b = b.variants().collect::<Vec<_>>();
```
peterhuene: Now that we're no longer sorting them, I don't think we need to collect here simply to get the length.

Instead, we should probably be using `itertools::zip_longest` here and returning false if it doesn't yield `EitherOrBoth::Both`.

We should also be doing the same in `are_structs_equal`, which is an existing bug (we're failing to detect if one struct has more fields than the other since there's not even a collect and length check).

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -603,6 +811,7 @@ fn add_task(config: &Config, document: &mut DocumentData, definition: &TaskDefin
         document.version.unwrap(),
         definition,
         &mut document.analysis_diagnostics,
+        |name| document.structs.get(name).is_some() || document.enums.get(name).is_some(),
```
peterhuene:
```suggestion
        |name| document.structs.contains_key(name) || document.enums.contains_key(name),
```

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1703,6 +1915,210 @@ fn set_struct_types(document: &mut DocumentData) {
...
+
+        match result {
+            Ok(enum_ty) => {
+                document.enums[info.index].ty = Some(enum_ty.into());
+            }
+            Err(diagnostic) => {
+                document.analysis_diagnostics.push(diagnostic);
+            }
+        }
+    }
```
peterhuene, line 1998: I think we could probably simplify this code and remove a collect or two by just enumerating on index and accessing the `IndexMap` as needed, like so:

```suggestion
    if document.enums.is_empty() {
        return;
    }

    // Calculate the underlying type for every enum in the document
    for index in 0..document.enums.len() {
        let e = &document.enums[index];

        // Only look at locally defined enums
        if e.namespace.is_some() {
            continue;
        }

        let definition = e.definition();
        let mut variants = Vec::new();
        let mut variant_spans = Vec::new();

        // Populate the variants and their spans
        for variant in definition.variants() {
            let variant_name = variant.name().text().to_string();
            let variant_type = if let Some(value_expr) = variant.value() {
                // Validate that the value is a literal expression
                match parse_literal_value(&document.structs, &value_expr) {
                    Some(ty) => ty,
                    None => {
                        let span = value_expr.span();
                        let adjusted_span = Span::new(span.start() + e.offset, span.len());
                        document
                            .analysis_diagnostics
                            .push(non_literal_enum_value(adjusted_span));
                        Type::Union
                    }
                }
            } else {
                // Default to `String` type if no value is specified
                PrimitiveType::String.into()
            };

            variants.push((variant_name, variant_type));
            variant_spans.push(Span::new(
                variant.span().start() + e.offset(),
                variant.span().len(),
            ));
        }

        // Check for explicit type parameter or infer if one was not specified
        let result = if let Some(type_param) = definition.type_parameter().map(|t| t.ty()) {
            let type_param = convert_ast_type(document, &type_param);
            let e = &document.enums[index];
            EnumType::new(
                e.name.clone(),
                e.name_span,
                type_param,
                variants,
                &variant_spans,
            )
        } else {
            EnumType::infer(document.enums[index].name.clone(), variants, &variant_spans)
        };

        match result {
            Ok(enum_ty) => {
                document.enums[index].ty = Some(enum_ty.into());
            }
            Err(diagnostic) => {
                document.analysis_diagnostics.push(diagnostic);
            }
        }
    }
```

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1703,6 +1915,210 @@ fn set_struct_types(document: &mut DocumentData) {
...
+fn parse_literal_value(structs: &indexmap::IndexMap<String, Struct>, expr: &Expr) -> Option<Type> {
+    // Handle struct literals specially since they need struct definitions
+    if let Expr::Literal(LiteralExpr::Struct(s)) = expr {
+        for item in s.items() {
+            let (_, val_expr) = item.name_value();
+            parse_literal_value(structs, &val_expr)?;
+        }
+
+        let struct_name = s.name().text().to_string();
+        return structs.get(&struct_name).and_then(|st| st.ty.clone());
```
peterhuene:
```suggestion
        return structs.get(s.name().text()).and_then(|st| st.ty.clone());
```

crates/wdl-analysis/src/handlers/completions.rs
```diff
@@ -317,7 +328,7 @@ fn add_member_access_completions(
         .children_with_tokens()
         .find(|t| matches!(t.kind(), SyntaxKind::Dot | SyntaxKind::OpenBracket))
     else {
-        debug!("could not find accessor token ( or [");
+        debug!("could not find accessor token `.` or `[`");
```
claymcleod, line 331: This is an unrelated error I stumbled across.

crates/wdl-analysis/src/handlers/completions.rs
```diff
@@ -464,7 +475,9 @@ fn add_member_access_completions(
 
     // NOTE: we do type evaluation only for non namespaces or complex types
 
-    let Some(scope) = document.find_scope_by_position(node.span().start()) else {
+    // Use the target node's position to find the scope, not the parent node's
+    // position
+    let Some(scope) = document.find_scope_by_position(target_node.span().start()) else {
```
claymcleod, line 478: Another unrelated issue.

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -445,20 +468,46 @@ fn resolve_access_expression(
     // `SyntaxKind::AccessExprNode` in the `resolve_by_context` before
     // calling this function.
     let access_expr = wdl_ast::v1::AccessExpr::cast(parent_node.clone()).unwrap();
-    let (target_expr, member_ident) = access_expr.operands();
+    let (target_expr, variant_ident) = access_expr.operands();
```
peterhuene: Nit: perhaps call the latter `access_ident` as it's not necessarily a "variant" either (nor "member").

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -563,8 +603,99 @@ fn resolve_access_expression(
         return Ok(Some(location_from_span(uri, span, def_lines)?));
     }
 
+    if let Type::TypeNameRef(CustomType::Enum(enum_ty)) = target_type {
+        let original_enum_name = enum_ty.name().as_str();
+
+        // Check for enum definition in imported namespaces.
+        for (_, ns) in analysis_doc.namespaces() {
+            // SAFETY: `ns.source` comes from a `analysis_doc.namespaces` which only
+            // contains namespaces for documents that guaranteed to be present in
```
peterhuene:
```suggestion
            // contains namespaces for documents that are guaranteed to be present in
```

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -563,8 +603,99 @@ fn resolve_access_expression(
...
+        };
+
+        let variant_span = variant.name().span();
+        let span = Span::new(variant_span.start() + enum_def.offset(), variant_span.len());
+        // Returns found enum variant definition location.
+        return Ok(Some(location_from_span(uri, span, def_lines)?));
+    }
+
+    if let Type::TypeNameRef(CustomType::Struct(_)) = target_type {
+        todo!("handle struct member access via `TypeNameRef`")
```
peterhuene: Should this be a `todo` rather than returning `Ok(None)` for the goto definition request, as `MyStruct.member` isn't a valid expression in WDL?

crates/wdl-analysis/src/handlers/hover.rs
```diff
@@ -248,14 +263,30 @@ fn resolve_hover_by_context(
...
+                            e.name(),
+                            member.text(),
+                            e.inner_value_type()
+                        );
+                        return Ok(Some(content));
+                    }
+                    (None, None)
+                }
+                Type::TypeNameRef(CustomType::Struct(_)) => {
+                    todo!("handle struct member access via `TypeNameRef`")
```
peterhuene: Same comment as above: I don't think the LSP should handle these as they aren't valid expressions anywhere in WDL.

crates/wdl-engine/src/eval/v1/expr.rs
```diff
@@ -1494,11 +1501,204 @@ impl<C: EvaluationContext> ExprEvaluator<C> {
...
+                    let value = self
+                        .context()
+                        .enum_variant_value(ty.name(), name.text())
+                        .map_err(|_| {
+                            crate::diagnostics::unknown_enum_variant_access(ty.name(), &name)
+                        })?;
+                    let variant = EnumVariant::new(ty.clone(), name.text(), value);
+                    Ok(Value::Compound(CompoundValue::EnumVariant(variant)))
+                } else {
+                    todo!("support type name references for `{ty:?}`");
```
peterhuene: This should definitely be a `cannot_access` error.

crates/wdl-analysis/src/handlers/semantic_tokens.rs
```diff
@@ -285,7 +286,9 @@ fn resolve_identifier_ty(
     {
         return match name_info.ty() {
             Type::Call(_) => Some(SemanticTokenType::VARIABLE),
-            Type::Compound(CompoundType::Struct(_), _) => Some(SemanticTokenType::STRUCT),
+            Type::Compound(CompoundType::Custom(CustomType::Struct(_)), _) => {
+                Some(SemanticTokenType::STRUCT)
+            }
```
peterhuene: Should we have a corresponding case for `CustomType::Enum` that returns `Some(SemanticTokenType::ENUM)`?

crates/wdl-engine/src/cache/hash.rs
```diff
@@ -64,6 +65,8 @@ enum ValueKind {
     Input,
     /// The value is an `Output` (hidden type).
     Output,
+    /// The value is an `EnumVariant`.
+    EnumVariant,
```
claymcleod, line 69: It's important that this is last so that its addition doesn't bust previously built caches.

crates/wdl-engine/src/cache/hash.rs
```diff
@@ -330,6 +335,13 @@ impl Hashable for Struct {
     }
 }
 
+impl Hashable for EnumVariant {
+    fn hash(&self, hasher: &mut Hasher) {
+        ValueKind::EnumVariant.hash(hasher);
+        self.value().hash(hasher);
```
claymcleod, line 342: This is, perhaps, controversial. I'm interested in what you think, @peterhuene.

crates/wdl-analysis/src/types.rs
```diff
@@ -911,6 +1070,137 @@ impl Coercible for StructType {
...
 
+/// Represents the type of an enum.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EnumType {
+    /// The name of the enum.
+    name: Arc<String>,
+    /// The common coerced type computed from all variant values.
+    inner_value_type: Type,
+    /// The variants.
+    variants: Arc<Vec<String>>,
```
peterhuene: Nit:

```suggestion
    variants: Arc<[String]>,
```

crates/wdl-analysis/src/types.rs
```diff
@@ -911,6 +1070,137 @@ impl Coercible for StructType {
...
+                    enum_span,
+                    variant_name,
+                    variant_spans[variant_idx],
+                    &explicit_inner_type,
+                    variant_type,
+                ));
+            }
+        }
+
+        let variants = variants.into_iter().map(|(name, _)| name).collect();
```
peterhuene: Nit: could probably just have a `Vec::with_capacity(variants.len())` that we push the name into with a single iteration over `variants.into_iter().enumerate()` above (doesn't really matter though).

crates/wdl-analysis/src/types.rs
```diff
@@ -911,6 +1070,137 @@ impl Coercible for StructType {
...
+        variants: Vec<(String, Type)>,
+        variant_spans: &[Span],
+    ) -> Result<Self, Diagnostic> {
+        assert_eq!(variants.len(), variant_spans.len());
+        let enum_name = enum_name.into();
+
+        let common_inner_type = if variants.is_empty() {
+            Type::Union
+        } else {
+            let mut variants = variants.iter().cloned();
```
peterhuene: Like the previous nit comment, I feel this could probably be better accomplished with a names vec that we push into, plus we shouldn't need to clone all of the elements of `variants` just to calculate the common type:

```rust
pub fn infer(
    enum_name: impl Into<String>,
    variants: Vec<(String, Type)>,
    variant_spans: &[Span],
) -> Result<Self, Diagnostic> {
    assert_eq!(variants.len(), variant_spans.len());
    let enum_name = enum_name.into();

    let mut common_ty: Option<Type> = None;
    let mut names = Vec::with_capacity(variants.len());
    for (i, (name, variant_ty)) in variants.into_iter().enumerate() {
        match common_ty {
            Some(current_common_ty) => match current_common_ty.common_type(&variant_ty) {
                Some(new_common_ty) => {
                    common_ty = Some(new_common_ty);
                }
                None => {
                    return Err(crate::diagnostics::no_common_inferred_type_for_enum(
                        &enum_name,
                        &current_common_ty,
                        variant_spans[i - 1],
                        &variant_ty,
                        variant_spans[i],
                    ));
                }
            },
            None => common_ty = Some(variant_ty),
        }

        names.push(name);
    }

    Ok(Self {
        name: Arc::new(enum_name),
        inner_value_type: common_ty.unwrap_or(Type::Union),
        variants: Arc::new(names),
    })
}
```

crates/wdl-engine/tests/tasks.rs
```diff
@@ -88,7 +88,20 @@ fn run_test(test: &Path, config: TestConfig) -> BoxFuture<'_, Result<()>> {
             bail!("parsing failed: {e:#}");
         }
         if result.document().has_errors() {
-            bail!("test WDL contains errors; run a `check` on `source.wdl`");
+            let errors: Vec<_> = result
```
claymcleod, line 91: I just found this easier to debug issues so that we don't have to do an out-of-band process to get the errors out (particularly difficult with feature flags).

src/commands/analyzer.rs
```diff
@@ -59,6 +60,7 @@ pub async fn analyzer(args: Args) -> CommandResult<()> {
         lint: args.lint,
         exceptions: args.except,
         ignore_filename: Some(IGNORE_FILENAME.to_string()),
+        feature_flags: FeatureFlags::default().with_wdl_1_3(),
```
claymcleod: This needs to be bumped down before merge

crates/wdl-engine/src/cache/hash.rs
peterhuene: So I think we need to also hash the variant's name in addition to it's value.

The name is important in case they coerce the variant value to a string (or interpolate).

The value is important in case the variant's constant expression changes between runs.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -152,16 +154,43 @@ impl EvaluationContext for WorkflowEvaluationContext<'_, '_> {
...
+            return Ok(Value::TypeNameRef(s.ty().unwrap().clone()));
+        }
+
+        // If the name is a reference to an enum, return it as a
+        // [`Value::TypeNameRef`].
+        if let Some(e) = self.state.document.enum_by_name(name) {
+            // SAFETY: enums should always have a type at the point we're
+            // evaluating.
+            return Ok(Value::TypeNameRef(e.ty().unwrap().clone()));
+        }
```
peterhuene: After seeing this in a few implementations of `resolve_name`, I wonder if it makes sense to add a `get_custom_type(&self, name: &str) -> Option<Type>` method to `Document` so we can simply do:

```rust
if let Some(ty) = self.state.document.get_custom_type(name) {
    return Ok(Value::TypeNameRef(ty));
}
```

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -645,16 +647,45 @@ impl EvaluationContext for TaskEvaluationContext<'_, '_> {
...
         crate::resolve_type_name(self.state.document, name, span)
     }
 
+    fn enum_variant_value(&self, enum_name: &str, variant_name: &str) -> Result<Value, Diagnostic> {
+        let r#enum = self
+            .state
+            .document
+            .enum_by_name(enum_name)
+            .ok_or(unknown_enum(enum_name))?;
+        resolve_enum_variant_value(r#enum, variant_name)
```
peterhuene, line 675: In the future (can be a follow up PR), I think we should probably cache these at the `TopLevelEvaluator` level so we evaluate a variant's expression (or default) exactly once.

crates/wdl-engine/src/value.rs
```diff
@@ -1998,6 +2061,94 @@ impl fmt::Display for Struct {
...
+/// A variant enum is the name of the enum variant and the type of the enum from
+/// which that variant can be looked up.
+///
+/// This type is cheaply clonable.
+#[derive(Debug, Clone)]
+pub struct EnumVariant {
+    /// The type of the enum containing this variant.
+    enum_ty: EnumType,
+    /// The name of the variant.
+    name: Arc<String>,
```
peterhuene: I still think we could probably store the index of the variant rather than the name here, since we have the associated `EnumType`, which stores the names.

crates/wdl-analysis/src/document/v1.rs
peterhuene: Yikes, sorry for the ridiculously-sized comment.

crates/wdl-analysis/src/handlers/semantic_tokens.rs
claymcleod: Yup! I had added this this morning.

crates/wdl-analysis/tests/analysis/enums-dont-conflict-with-task/source.wdl
claymcleod: Tracked in #507.

crates/wdl-analysis/src/types.rs
```diff
@@ -1966,4 +2250,150 @@ mod test {
...
+                ("IntValue".into(), PrimitiveType::Integer.into()),
+                ("FloatValue".into(), PrimitiveType::Float.into()),
+            ],
+            &[Span::new(0, 0), Span::new(0, 0)],
+        )
+        .unwrap();
+
+        assert_eq!(mixed.name().as_str(), "Mixed");
+        assert_eq!(mixed.inner_value_type(), &Type::from(PrimitiveType::Float));
+        dbg!(mixed.variants());
```
peterhuene:
```suggestion
```

crates/wdl-engine/src/value.rs
```diff
@@ -1998,6 +2061,100 @@ impl fmt::Display for Struct {
...
+impl PartialEq for EnumVariant {
+    fn eq(&self, other: &Self) -> bool {
+        self.enum_ty == other.enum_ty && self.variant_index == other.variant_index
+    }
+}
+
+impl EnumVariant {
+    /// Attempts to create a new enum variant from a enum type and variant name.
+    ///
+    /// This method returns [`None`] if the variant is not in the enum.
```
peterhuene, line 2089:
```suggestion
    /// This method returns [`None`] if the variant is not in the enum.
    ///
    /// # Panics
    ///
    /// Panics if the given variant name is not present in the enum type.
```

crates/wdl-engine/src/value.rs
```diff
@@ -1998,6 +2061,100 @@ impl fmt::Display for Struct {
...
+    pub fn new(
+        enum_ty: impl Into<EnumType>,
+        name: impl Into<String>,
```
peterhuene:
```suggestion
        name: &str,
```

crates/wdl-engine/src/value.rs
```diff
@@ -1998,6 +2061,100 @@ impl fmt::Display for Struct {
...
+        value: impl Into<Value>,
+    ) -> Self {
+        let enum_ty = enum_ty.into();
+        let name = name.into();
```
peterhuene:
```suggestion
```

crates/wdl-engine/src/value.rs
```diff
@@ -1998,6 +2061,100 @@ impl fmt::Display for Struct {
...
+        let value = Arc::new(value.into());
+
+        let variant_index = enum_ty
+            .variants()
+            .iter()
+            .position(|v| v == &name)
```
peterhuene:
```suggestion
            .position(|v| v == name)
```

PR #460, opened 2025-11-13
refactor: make CLI input processing less dependent on mutation

acfoltzer:
This is an intermediate step in a larger refactoring effort to make the interface to `wdl-engine` smaller and more consistent. The next step of the refactoring is likely to be quite large, so even though this is not a complete set of desired changes, we have decided to land it in advance of that to avoid the size of any one PR getting out of control.

Each commit on this branch ought to work independently, and the `wip` commit entries should give a clue as to what's going on. I'll call out a few larger changes at the top level though:

- `ensure_paths_exist()` has been renamed to `resolve_paths()`, and now works in a functional style of returning a new `Value` rather than mutating one in place. On its face, this is unsatisfying as a number of values will get copied unnecessarily, but since this is logic that runs once on a finite set of initial inputs on startup, the cost should be negligible compared to the benefits in code clarity.
- One layer of `join_paths()` has been removed via inlining, though `Task::join_paths()` and `Workflow::join_paths()` remain as functions that mutate inputs. This will be addressed in subsequent work.
- Helper functions in `wdl_engine::path` have been rewritten to parse strings into URLs early and often, rather than relying on our own partial implementations of URL parsing. I am planning in a subsequent step to reduce the number of redundant parses this introduces by using URLs in the representation of values earlier in the evaluation process.
- Instead of "inputs" or anonymous tuples, I've begun using "invocation" to describe the combinations of entrypoint, input values, and origin paths, though this transformation is incomplete as of this PR.
- ~I converted `impl<'de> serde::Deserialize<'de> for Value` into a much simpler `impl TryFrom<JsonValue> for Value`. This is probably going to change into or be augmented by a conversion function with more arguments in future revisions.~

For all contributors:

### DIFF
--- a/crates/wdl-engine/src/value.rs
+++ b/crates/wdl-engine/src/value.rs
@@ -940,30 +940,34 @@ impl<'de> serde::Deserialize<'de> for Value {
             where
                 A: serde::de::SeqAccess<'de>,
             {
-                use serde::de::Error;
+                use serde::de::Error as _;
 
-                let mut elements = Vec::new();
-                while let Some(v) = seq.next_element::<Value>()? {
-                    elements.push(v);
+                let mut elements = vec![];
+                while let Some(element) = seq.next_element::<Value>()? {
+                    elements.push(element);
                 }
 
-                let element_ty = elements
-                    .iter()
-                    .try_fold(None, |mut ty, element| {
-                        let element_ty = element.ty();
-                        let ty = ty.get_or_insert(element_ty.clone());
-                        ty.common_type(&element_ty).map(Some).ok_or_else(|| {
-                            A::Error::custom(format!(
-                                "a common element type does not exist between `{ty}` and \
-                                 `{element_ty}`"
-                            ))
-                        })
+                // Try to find a mutually-agreeable common type for the elements of the array.
+                let mut candidate_ty = None;
+                for element in elements.iter() {
+                    let new_candidate_ty = element.ty();
+                    let old_candidate_ty =
+                        candidate_ty.get_or_insert_with(|| new_candidate_ty.clone());
+                    let Some(new_common_ty) = old_candidate_ty.common_type(&new_candidate_ty)
+                    else {
+                        return Err(A::Error::custom(format!(
+                            "a common element type does not exist between `{old_candidate_ty}` \
+                             and `{new_candidate_ty}`"
+                        )));
+                    };
+                    candidate_ty = Some(new_common_ty);
+                }
+                // An empty array's elements have the `Union` type.
+                let array_ty: Type = ArrayType::new(candidate_ty.unwrap_or(Type::Union)).into();
+                Ok(Array::new(None, array_ty.clone(), elements)
+                    .map_err(|e| {
+                        A::Error::custom(format!("cannot coerce value to `{array_ty}`: {e:#}"))
                     })?
-                    .unwrap_or(Type::Union);
-
-                let ty: Type = ArrayType::new(element_ty).into();
-                Ok(Array::new(None, ty.clone(), elements)
-                    .map_err(|e| A::Error::custom(format!("cannot coerce value to `{ty}`: {e:#}")))?
                     .into())
             }
 


crates/wdl-engine/src/stdlib/read_json.rs
```diff
@@ -156,18 +181,23 @@ mod test {
...
+    #[tokio::test]
+    async fn read_bad_array_json() {
+        let diagnostic = eval_v1_expr(&make_env(), V1::One, "read_json('bad_array.json')")
             .await
             .unwrap_err();
         assert_eq!(
             diagnostic.message(),
-            "call to function `read_json` failed: failed to deserialize JSON file \
-             `bad_array.json`: a common element type does not exist between `Int` and `String` at \
-             line 1 column 11"
```
acfoltzer, line 167: This is the one place where it's revealed that switching away from doing these conversions via `serde` gives errors slightly less information about where a type error has occurred. Production code paths from `sprocket` already are converting first to `serde_json::Value` and then to `wdl_engine::Value` in discrete steps, however, so this does not constitute a regression to the quality of user-facing errors.

It might be nice to eventually add such location information to user-facing errors, but that should be treated as a future feature addition rather than as a reason to keep unnecessary complications in existing code.

crates/wdl-engine/src/inputs.rs
```diff
@@ -888,46 +885,60 @@ impl Inputs {
...
-        object: JsonMap,
-    ) -> Result<(String, Self)> {
+    fn parse_task_inputs(document: &Document, task: &Task, object: JsonMap) -> Result<Self> {
         let mut inputs = TaskInputs::default();
         for (key, value) in object {
             // Convert from serde_json::Value to crate::Value
-            let value = serde_json::from_value(value)
-                .with_context(|| format!("invalid input key `{key}`"))?;
+            let value =
+                Value::try_from(value).with_context(|| format!("invalid input key `{key}`"))?;
```
claymcleod:
```suggestion
                Value::try_from(value).with_context(|| format!("invalid input value for key `{key}`"))?;
```

crates/wdl-engine/src/inputs.rs
```diff
@@ -888,46 +885,60 @@ impl Inputs {
...
+        Ok(Some((entrypoint_name, inputs)))
     }
 
     /// Parses the inputs for a task.
-    fn parse_task_inputs(
-        document: &Document,
-        task: &Task,
-        object: JsonMap,
-    ) -> Result<(String, Self)> {
+    fn parse_task_inputs(document: &Document, task: &Task, object: JsonMap) -> Result<Self> {
```
claymcleod, line 936: Given that `parse_task_inputs` and `parse_workflow_inputs` are so similar, I wonder if it makes sense to extract a generic helper function or make the function generic itself?

acfoltzer: Yes, I am working out a better abstraction for this in the follow-on branch. Part of what made the current code more difficult to follow was the use of closures for type and path lookups; I'm moving away from that and toward a trait interface that'll allow us to document each implementation and follow it more effectively with `rust-analyzer`.

crates/wdl-engine/src/stdlib/read_json.rs
peterhuene: I'm confused by this comment. How is this not a user-facing error? The user's JSON file content isn't representable as a WDL value and we know exactly where in the JSON file the problem is, so the loss of context is undesirable.

I also fail to see how implementing a `TryFrom<serde_json::Value>` reduces complexity. The previous deserialization code wasn't "an unnecessary complication" to begin with as it was contained within a single implementation of `serde::Deserialize` and has the benefit of _context_.

Is there some other virtue of the `TryFrom` implementation other than it's a little less code?

crates/wdl-engine/src/inputs.rs
```diff
@@ -888,46 +885,60 @@ impl Inputs {
...
+            return Ok(None);
+        }
+
+        // Otherwise, build a set of candidate entrypoints from the prefixes of each
+        // input key.
+        let mut entrypoint_candidates = BTreeSet::new();
+        for key in object.keys() {
+            let Some((prefix, _)) = key.split_once('.') else {
+                bail!(
+                    "invalid input key `{key}`: expected the value to be prefixed with the \
```
peterhuene:
```suggestion
                    "invalid input key `{key}`: expected the key to be prefixed with the \
```

The term _value_ is confusing to me in the context of a key-value map (which is what an inputs JSON file is). Given that we used _key_ earlier in the message, I think using _key_ again is unambiguous whereas _value_ might mean either the value of the key or the value mapped by the key.

src/inputs.rs
```diff
@@ -139,26 +137,30 @@ impl FromStr for Input {
     }
 }
 
-/// The inner type for inputs (for convenience).
-type InputsInner = IndexMap<String, (EvaluationPath, Value)>;
+/// The map structure used for parsed inputs that have not yet had their paths
+/// normalized and converted to engine values.
+// TODO ACF 2025-11-11: why is this an `IndexMap` and not a built-in map type?
```
peterhuene: If we're not requiring a stable iteration on the map anywhere (I can't think of a reason why we would be), then we should replace it with `HashMap` and remove the TODO now.

crates/wdl-engine/src/inputs.rs
```diff
@@ -888,46 +885,60 @@ impl Inputs {
...
+
+        // If every prefix is the same, there will be only one candidate. If not, report
+        // an error.
+        let entrypoint_name = match entrypoint_candidates
+            .iter()
+            .take(2)
+            .collect::<Vec<_>>()
+            .as_slice()
+        {
+            [] => bail!("no entrypoint candidates for inputs; report this as a bug"),
```
peterhuene: I'm not a fan of soft errors that are effectively maintaining an invariant (i.e. "collection can't be empty since we just checked for it above"); a descriptive panic makes it clear to the end user that there's a bug to report, along with context.

crates/wdl-engine/src/value.rs
```diff
@@ -1031,6 +949,20 @@ impl PrimitiveValue {
         Self::Directory(Arc::new(path.into()).into())
     }
 
+    /// Create either a new `File` or `Directory` value, depending on whether
+    /// the `is_file` argument is `true`.
+    ///
+    /// This is a bit awkward, but can save a lot of repetition in code that
+    /// treats files and directories largely the same until having to
+    /// remember which enum variant the path needs to be stuffed back into.
+    pub fn new_file_or_directory(is_file: bool, path: impl Into<HostPath>) -> Self {
```
peterhuene: This seems like a helper function (or macro) for the above and not something we would want exposed on the public interface of `PrimitiveValue`.

peterhuene: Unrelated: we should probably update the signatures of `new_file` and `new_directory` to take `path: impl Into<HostPath>`.

crates/wdl-engine/src/stdlib/read_json.rs
acfoltzer: It's not user-facing because no user-facing code ever invoked the `Deserialize` impl straight from a string (or other type with line/column info) to `wdl_engine::Value`. The `sprocket` frontend makes an intermediate stop at `serde_json::Value`, at which point line/column info is lost. 

If you'd like, we could back this part of the change out and open a feature request to hook it up directly, but I think we'd be better served by using something like [`json_spanned_value`](https://docs.rs/json-spanned-value/latest/json_spanned_value/) rather than inherently tying the parsing and typechecking steps together.

peterhuene: This is an error message displayed in an _evaluation context_ from a call to `read_json` in a user's WDL document.

It has nothing to do with how `sprocket` itself parses input JSON files.

acfoltzer: Ah, you're right, I had looked at the lack of regressions in integration tests involving `read_json` and figured that meant that this error wasn't bubbling up there either. I'll change it back 👍🏻

acfoltzer: > It has nothing to do with how `sprocket` itself parses input JSON files.

This is not the case, though: `wdl_engine::Inputs::parse_{task,workflow}_inputs()` both use this impl to convert from `serde_json::Value` to `wdl_engine::Value`, which is how I got here. In the more-ambitious refactor I was planning, I was going to abandon this new `TryFrom` in favor of a multi-argument conversion from JSON values that'd include type and origin path information.

acfoltzer: Revived in 08100b85

src/inputs.rs
acfoltzer: Replaced with `BTreeMap` so that any `Debug` output that gets spit out should be sorted, which can be nice: 74d31366

crates/wdl-engine/src/value.rs
acfoltzer: Addressed both in 74d31366

crates/wdl-engine/src/inputs.rs
acfoltzer: Agreed; 74d31366

crates/wdl-engine/src/inputs.rs
acfoltzer: Agreed; 74d31366

acfoltzer:
> I think we need change log entries, but otherwise good to merge.

Given that the changes shouldn't be relevant to end users, do you think it needs more detail than what I put in the `wdl-engine` changelog?

Issue #398, opened 2025-10-02
Failed jobs show up as "completed"


claymcleod:
Ideally, failed jobs would not show as "completed" (though I guess that's _technically_ true). Rather, they should be set apart as "failed".

PR #461, opened 2025-11-14
feat: implement call caching.

peterhuene:
This PR implements call caching as specified in [RFC https://github.com/stjude-rust-labs/sprocket/pull/2](https://github.com/stjude-rust-labs/rfcs/pull/2).

Call caching can be enabled by setting `run.task.cache` to `on` in
`sprocket.toml`.

If a cached task result is not reused, the reason is logged at the `INFO` level.

This also updates the progress bar displayed by Sprocket to include the number
of "cached" tasks (i.e. task results reused from a previous run) and "failed" tasks .

The progress bar has also been updated to only show counts above 0 and uses
different colors for the different counts.

Closes https://github.com/stjude-rust-labs/sprocket/issues/398.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/src/eval/v1/workflow.rs
+++ b/crates/wdl-engine/src/eval/v1/workflow.rs
@@ -76,7 +76,6 @@ use crate::diagnostics::decl_evaluation_failed;
 use crate::diagnostics::if_conditional_mismatch;
 use crate::diagnostics::runtime_type_mismatch;
 use crate::http::Transferer;
-use crate::path;
 use crate::path::EvaluationPath;
 use crate::tree::SyntaxNode;
 use crate::tree::SyntaxToken;
@@ -1152,7 +1151,7 @@ impl State {
                 self.base_dir.as_local(),
                 Some(self.transferer()),
                 &|path| {
-                    if !path::is_supported_url(path.as_str()) && path.is_relative() {
+                    if path.is_relative() {
                         bail!("relative path `{path}` cannot be used as a workflow output");
                     }
 


crates/wdl-engine/src/cache/hash.rs
```diff
@@ -0,0 +1,320 @@
...
+
+/// The variant tag for hash-based remote content digests.
+const CONTENT_HASH_VARIANT_TAG: u8 = 0;
+/// The variant tag for ETag based remote content digests.
+const CONTENT_ETAG_VARIANT_TAG: u8 = 1;
+
+/// The variant tag for file digests.
+const DIGEST_FILE_VARIANT_TAG: u8 = 0;
+/// The variant tag for directory digests.
+const DIGEST_DIRECTORY_VARIANT_TAG: u8 = 1;
```
acfoltzer: The [`num_enum` crate](https://docs.rs/num_enum/0.7.5/num_enum/) might be nice to use here.

crates/wdl-engine/src/cache/hash.rs
```diff
@@ -0,0 +1,320 @@
...
+impl Hashable for &[u8] {
+    fn hash(&self, hasher: &mut Hasher) {
+        hasher.update(self);
+    }
+}
+
+impl Hashable for Path {
+    fn hash(&self, hasher: &mut Hasher) {
+        hasher.update(&(self.as_os_str().len() as u32).to_le_bytes());
+        hasher.update(self.as_os_str().as_encoded_bytes());
```
acfoltzer: This does not match the behavior specified in the RFC that `File`/`Directory` values are hashed as UTF-8, as [`as_encoded_bytes()`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.as_encoded_bytes) does not necessarily return UTF-8. The doc even says:
> sending the slice over the network or storing it in a file will likely result in incompatible byte slices

We should probably use [`to_string_lossy()`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.to_string_lossy) or something similar that checks the UTF-8 validity of its result.

crates/wdl-engine/src/cache/hash.rs
```diff
@@ -0,0 +1,320 @@
...
+        }
+    }
+}
+
+impl Hashable for Option<PrimitiveValue> {
+    fn hash(&self, hasher: &mut Hasher) {
+        match self {
+            Some(v) => v.hash(hasher),
+            None => {
+                hasher.update(&[NONE_VARIANT_TAG]);
```
acfoltzer: Is it intended that the hash of `None: Option<PrimitiveValue>` is the same as the hash of `Value::None: Value`?

crates/wdl-engine/src/digest.rs
```diff
@@ -0,0 +1,291 @@
...
+            Self::File(hash) => hash.to_hex(),
+            Self::Directory(hash) => hash.to_hex(),
+        }
+    }
+}
+
+/// Keeps track of previously calculated digests.
+///
+/// As WDL evaluation cannot write to existing files, it is assumed that files
+/// and directories are not modified during evaluation.
```
acfoltzer, line 51: Sadly this is not strictly true given the prevalence of read-write input mounts in other engines, but someone would really have to be trying to self-sabotage in order to witness the potential effects of this. But regardless, we have to trust them not to mess with files through some other external process.

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+
+impl Write for LockedFile {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.0.write(buf)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.0.flush()
+    }
+}
```
acfoltzer: This would go nicely in its own module to reduce the size of this one, perhaps also making the error messages less cache-specific so the type could be used for other purposes.

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+struct State {
+    /// The global cache file lock.
+    ///
+    /// Task and workflow evaluation typically acquires a single shared lock on
+    /// the call cache per run.
+    ///
+    /// Operations to clean the cache will acquire an exclusive lock to ensure
+    /// the cache is cleaned only when no evaluations are taking place.
+    #[allow(dead_code)]
+    lock: Arc<LockedFile>,
```
acfoltzer: I'd rename this and add a note, possibly using `//` comments rather than a docstring, about why it's important that this be here even though it's "unused"
```suggestion
    _lock: Arc<LockedFile>,
```

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+    /// The path to the root call cache directory.
+    cache_dir: Arc<PathBuf>,
+    /// The file transferer that can be used for calculating remote file
+    /// digests.
+    transferer: Arc<dyn Transferer>,
+}
+
+impl State {
+    /// Gets the path to anb entry in the cache given the [`Key`].
```
acfoltzer:
```suggestion
    /// Gets the path to an entry in the cache given the [`Key`].
```

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+        Ok(Self {
+            location: path.try_into()?,
+            digest: digest.to_hex(),
+        })
+    }
+
+    /// Converts the [`Content`] to an evaluation path.
+    ///
+    /// Returns an error if the current digest of the location does not match
+    /// the stored digest.
```
acfoltzer: This is "current" as of the first time this content was hashed during this execution, right? This should probably have a heavier caveat.

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+    stdout: Content,
+    /// The task's last stderr content.
+    stderr: Content,
+    /// The task's last work directory content.
+    work: Content,
+}
+
+/// Represents a key for a [`CallCache`].
+///
+/// Additionally stores digests used to validate cache entries.
```
acfoltzer: There should probably be more of an explanation of what this means somewhere in the comments or at least a link to the "Call Cache Hit" section of the RFC.

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+            Some(cache_dir) => cache_dir.into(),
+            None => crate::config::cache_dir()?.join(CALL_CACHE_SUBDIR),
+        };
+
+        info!(
+            "using call cache directory `{cache_dir}`",
+            cache_dir = cache_dir.display()
+        );
+
+        fs::create_dir_all(&cache_dir).with_context(|| {
```
acfoltzer: Could matter on HPCs with slow filesystems to use the `tokio::fs` version instead

crates/wdl-engine/src/cache/hash.rs
peterhuene: Yes, for the purpose of WDL `Map` keys, a Rust `None` is signifying a WDL `None` value. Technically a map key can store any value of type `PrimitiveType?` but we only implement `PartialEq`, `Eq`, and `Hash` for `PrimitiveValue` and not `Value` (where the WDL `None` representation resides).

Hence we store `Option<PrimitiveValue>` in our internal representation.

crates/wdl-engine/src/digest.rs
peterhuene: This comment was meant to only apply to Sprocket; we aren't really concerned with what other engines do in this context.

If a user happened to be running Sprocket at the same time as some other engine that _does_ allow WDL to write to existing files, firstly Sprocket would likely fail its evaluation (depending on how the user's command is authored) and secondly it would be no different than any other external process modifying the files during our evaluation. 

That said, it'll be true of other engines soon enough as that behavior is fully deprecated in 1.2 and required not to occur in a future 2.0.

crates/wdl-engine/src/cache.rs
peterhuene: I'll switch this over to `tokio::fs`

crates/wdl-engine/src/cache/hash.rs
peterhuene: I'll add a comment explaining this.

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+use crate::Value;
+use crate::backend::TaskExecutionResult;
+use crate::http::Transferer;
+use crate::path::EvaluationPath;
+
+/// The current cache entry version.
+///
+/// This is a monotonic value that should increase whenever the serialization of
+/// call cache entries change.
+const CURRENT_CACHE_VERSION: u32 = 0;
```
claymcleod, line 36: Just wanted to make a comment that we're going to need to parse every version of the cache—does it make sense to go ahead and put this code in a `v0` module?

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+                .await?;
+
+            input_digests.insert(
+                input
+                    .path()
+                    .to_str()
+                    .with_context(|| {
+                        format!(
+                            "input path `{path}` is not UTF-8",
+                            path = input.path().as_local().unwrap().display()
```
claymcleod: Is this safe because we assume that all remote files have been localized at this point?

claymcleod: It feels a bit like this might be more natural to just implement `Display` for `EvaluationPath` and call it here.

crates/wdl-engine/src/cache.rs
peterhuene: The monotonic version serves as a mechanism by which we can reliably bust all existing cache entries in the wild when the implementation changes.

The intention is not to parse every previous version of the cache entry file format, but instead force invalidation.

crates/wdl-engine/src/cache.rs
peterhuene: It's safe due to knowing that `to_str` only returns `None` when the path is local and isn't representable in UTF-8; remote URLs always give back a string representation.

Agreed that this would probably better be served as a display representation that we can convert to string. I'll look into fixing.

crates/wdl-engine/src/cache.rs
claymcleod: Right... I had my backwards compatibility hat on because I've been thinking about that for the provenance PR, but that makes total sense for the cache.

crates/wdl-engine/src/cache.rs
claymcleod: Sounds good.

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,635 @@
...
+        self.key.fmt(f)
+    }
+}
+
+/// Represents a request to calculate a [`Key`].
+pub struct KeyRequest<'a> {
+    /// The document containing the task.
+    pub document: &'a Document,
+    /// The task identifier.
+    pub task_id: &'a str,
```
claymcleod: This is a bit non-obvious from just reading this part of the code. I've traced this back to see that it's actually the task name, which makes sense, but I wonder if it makes sense to (a) normalize on that nomeclature (`id` could mean many things, I presumed it was a transient identifier of a task "instance") or (b) add a clarifying comment here.

crates/wdl-engine/src/value.rs
```diff
@@ -2938,7 +2975,9 @@ impl TaskPreEvaluationValue {
             n if n == TASK_FIELD_META => Some(self.meta.clone().into()),
```
claymcleod: In line with your other improvements, I see these can also be simplified by removing the `n if n ==`.

crates/wdl-engine/src/config.rs
```diff
@@ -671,6 +681,34 @@ impl ScatterConfig {
     }
 }
 
+/// Represents the supported call caching modes.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum CallCachingMode {
```
claymcleod, line 691: This brings up an interesting case, which is that I believe this is the first hint that we've implemented that wasn't outlined in the spec. The question arises as to what to do with this. I think we certainly do **not** want to just implement it and never try to get this into the spec. What do you think our policy should be on this kind of thing?

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -712,13 +732,14 @@ impl<'a> State<'a> {
...
-            format!("document `{document_path}` does not have a parent directory")
+        let document_path = document.uri();
+        let base_dir = EvaluationPath::parent_of(document_path.as_str()).with_context(|| {
+            format!(
+                "document `{path}` does not have a parent directory",
+                path = document.path()
+            )
         })?;
 
-        base_dir.make_absolute();
```
claymcleod, line 717: Why was this removed?

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1088,47 +1132,170 @@ impl TaskEvaluator {
...
+                            );
+                            None
+                        }
+                    }
+                } else if self.config.task.cache == CallCachingMode::On {
+                    debug!(
+                        task_id = id,
+                        task_name = state.task.name(),
+                        document = state.document.uri().as_str(),
+                        "task is not cacheable due to `cacheable` hint set to `true`"
```
claymcleod: Wouldn't this be due to the `cacheable` hint on the task being set to `false`?

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1088,47 +1132,170 @@ impl TaskEvaluator {
...
+                    );
+                    None
+                } else {
```
claymcleod, line 1101: I feel a bit uncomfortable burying this logic here—if we add any more variants to the call cache setting, this will be a recipe for having stale logic. Personally, I feel like we should match on the variant and make the compiler force us to handle new cases.

crates/wdl-engine/src/eval/v1/task.rs
peterhuene: The code above was incorrectly using `document.path()` (which always returns the document's path relative to the CWD if the document is a local path - primarily for display purposes), which caused `base_dir` to be `""` if you ran something like `sprocket run foo.wdl`; the `make_absolute` was correcting that oddity, possibly unintentionally.

By using `document.uri()` instead (which is also cheaper to call), `base_dir` is correctly evaluated to be an absolute local path when the document is local; there's no need for `make_absolute`.

crates/wdl-engine/src/eval/v1/task.rs
peterhuene: Yep, typo.


```suggestion
                        "task is not cacheable due to `cacheable` hint set to `false`"
```

crates/wdl-engine/src/config.rs
peterhuene: I think it's reasonable to say that if we add an unofficial hint, we at least raise it for discussion within the spec for its inclusion, perhaps informally to start (i.e. discussion at a governance meeting) and then followed by a PR to add it to the spec if there's a positive consensus.

claymcleod: Okay, for this one, I'm going to leave that as a follow-up item for you. It may be just a little too late to propose for WDL 1.3 though maybe I'm wrong on that.

crates/wdl-engine/src/cache.rs
peterhuene: This will be removed shortly with changes to how we calculate the cache key.

crates/wdl-engine/src/cache/lock.rs
```diff
@@ -173,3 +175,49 @@ impl Write for LockedFile {
...
+    async fn acquire_exclusive() {
+        let path = NamedTempFile::new().unwrap();
+        let _first = LockedFile::acquire_exclusive(path.path()).await.unwrap();
+
+        // As acquiring a shared lock will dead lock, wait at most 3 seconds for
+        // acquiring the shared lock and panic if we acquire it
+        tokio::select! {
+            _ = tokio::time::sleep(Duration::from_secs(3)) => {}
+            _ = LockedFile::acquire_shared(path.path(), true) => panic!("a shared lock should not be acquired")
+        }
```
peterhuene: This is a particularly silly way of doing this; I have a change ready to go to simply try to acquire a shared lock using `File::try_lock_shared` and ensuring the operation returns a would-block error.

Waiting on GitHub to recover from its outage so I can push to my repository.

crates/wdl-engine/src/config.rs
acfoltzer: I would consider proposing a registry that can be included by reference into the spec, but has a lower barrier to entry than making a full revision of the spec. This is commonly done with IETF RFCs, for example the [Cache-Status parameter definitions](https://www.rfc-editor.org/rfc/rfc9211.html#register). The registry is seeded by the initial values defined in the RFC, and then there's a lighter-weight process for accepting entries to the registry. This only works for things where implementation is optional, but that's perfect for something like hints.

crates/wdl-engine/src/cache/lock.rs
```diff
@@ -0,0 +1,220 @@
...
+            }),
+        }
+    }
+
+    /// Acquires an exclusive file lock for the given path.
+    ///
+    /// If the file does not exist, it is created.
+    ///
+    /// If the file exists, it is truncated once the lock is acquired.
+    pub async fn acquire_exclusive(path: impl AsRef<Path>) -> Result<Self> {
```
acfoltzer: This should probably have a name that suggests truncation will occur, otherwise someone who uses it without checking the docstring might have an unfortunate surprise.

crates/wdl-engine/src/cache.rs
```diff
@@ -472,3 +492,425 @@ impl CallCache {
...
+    use wdl_analysis::Config as AnalysisConfig;
+    use wdl_analysis::DiagnosticsConfig;
+
+    use super::*;
+    use crate::GuestPath;
+    use crate::digest::test::DigestTransferer;
+    use crate::digest::test::clear_digest_cache;
+
+    #[tokio::test]
+    async fn test_call_cache() {
```
acfoltzer: It would be nice to break this test up into individual or smaller sets of assertions. The assertions at the end which depend on messing with the state of the filesystem would particularly benefit from being able to run a shared setup routine before making a single modification and checking an assertion, rather than having to modify and then undo those modifications between assertions.

peterhuene: Fair enough, I'll break this apart into multiple tests.

crates/wdl-engine/src/cache.rs
```diff
@@ -0,0 +1,1317 @@
...
+                command: "modified!",
+                ..request
+            })
+            .await
+            .unwrap();
+        assert_eq!(
+            cache.get(&key).await.unwrap_err().to_string(),
+            "the command of the task was modified"
+        );
+    }
```
acfoltzer: It looks like these all have a common prelude, so we could probably save some further boilerplate with something like this: 

```suggestion
    struct TestContext {
        _root_dir: TempDir,
        task: Task,
        cache: CallCache,
    }

    impl TestContext {
        async fn new() -> TestContext {
            // Prepare an evaluated task for the test
            let root_dir = tempdir().expect("failed to create temporary directory");
            let task = prepare_task(root_dir.path()).await;

            // Create the cache
            let cache_dir = tempdir().unwrap();
            let transfer = Arc::new(DigestTransferer::new([]));
            let cache = CallCache::new(Some(cache_dir.path()), transfer)
                .await
                .unwrap();

            // Populate the cache with the initial entry
            populate_cache(&cache, &task).await;

            Self {
                _root_dir: root_dir,
                task,
                cache,
            }
        }
    }

    #[tokio::test]
    async fn modified_command() {
        let ctx = TestContext::new().await;
        let request = ctx.task.key_request();

        // Check for modified command
        let key = ctx
            .cache
            .key(KeyRequest {
                command: "modified!",
                ..request
            })
            .await
            .unwrap();
        assert_eq!(
            ctx.cache.get(&key).await.unwrap_err().to_string(),
            "the command of the task was modified"
        );
    }
```

acfoltzer: Oh, you of course have to keep the `cache_dir` tempdir around too. Maybe cleaner to make a single `test_root` tempdir with the task and cache dirs both within that, but now it's getting a bit big for a single ```suggestion block

crates/wdl-engine/src/cache/hash.rs
```diff
@@ -0,0 +1,992 @@
...
+        let mut hasher = Hasher::new();
+        hasher.update(&3u32.to_le_bytes()); // String length
+        hasher.update("foo".as_bytes()); // Literal string
+        hasher.update(&3u32.to_le_bytes()); // String length
+        hasher.update("bar".as_bytes()); // Literal string
+        assert_eq!(hash, hasher.finalize());
+    }
+
+    #[test]
+    fn hash_wld_none() {
```
claymcleod, line 593:
```suggestion
    fn hash_wdl_none() {
```

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1323,6 +1531,7 @@ impl TaskEvaluator {
 
         // Insert the name into the scope
         state.scopes[ROOT_SCOPE_INDEX.0].insert(name.text(), value.clone());
+        state.inputs.insert(name.text().to_string(), value.clone());
```
claymcleod: Could we just get a sentence on why this was necessary?

crates/wdl-engine/src/cache/hash.rs
```diff
@@ -0,0 +1,992 @@
...
+
+impl Hashable for String {
+    fn hash(&self, hasher: &mut Hasher) {
+        self.as_str().hash(hasher);
+    }
+}
+
+impl Hashable for Path {
+    fn hash(&self, hasher: &mut Hasher) {
+        self.to_string_lossy().as_bytes().hash(hasher);
```
claymcleod, line 139:
```suggestion
        self.as_os_str().as_encoded_bytes().hash(hasher);
```

Avoids the allocation.

crates/wdl-engine/src/eval/v1/task.rs
peterhuene: The call cache needs to know the input values for deriving the cache key.

We need to keep a separate map of _only_ the evaluated input values, which we don't keep in the scope (the root scope shares both inputs and private declarations).

crates/wdl-engine/src/cache/hash.rs
peterhuene: I originally had it as `as_encoded_bytes()` but Adam correctly pointed out that it is not portable across Rust versions.

`to_string_lossy` will at least normalize the non-UTF-8 sequences, which could lead to cache key collisions, but the worst behavior one would observe is unnecessary thrashing of a cache entry.

For example, you pass in two different `File` values to a task input that are different paths but both happen to have the same replacement sequence; the cache key will be the same, but if the file content is different, we'll invalidate the cache entry on the second lookup.

peterhuene: Also, given how much our code base errors on non-UTF-8 representable paths, I doubt one would get all the way to call cache lookup.

claymcleod: Okay makes sense.

PR #463, opened 2025-11-18
feat: new `wdl-engine` evaluation entrypoints and Apptainer cache management

### LINKED ISSUE #360: Improve cache management for Apptainer backend

acfoltzer:
Closes #360, and provides a small measure of unification for the `wdl-engine` evaluator entrypoints.

For implementors, there is now a `TopLevelEvaluator` struct that's available throughout task and workflow evaluation and is suitable for tracking once-per-run data like the Apptainer image cache. Previously, only the root of the current task or workflow's execution was available, which would have led to no image reuse at all if it had been used for Apptainer.

For all contributors:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   was renamed from `experimental_versions` to `wdl_1_3` ([#435](https://github.com/stjude-rust-labs/sprocket/pull/435)).
 * Removed the `wdl-cli` crate, absorbing its code into the `sprocket` library
   crate in preparation for future refactoring ([#450](https://github.com/stjude-rust-labs/sprocket/pull/450)).
-* Apptainer-based backends now store converted container images within each run directory, rather than in a user-specified directory ([#463](https://github.com/stjude-rust-labs/sprocket/pull/460)).
+* Apptainer-based backends now store converted container images within each run directory, rather than in a user-specified directory ([#463](https://github.com/stjude-rust-labs/sprocket/pull/463)).
 
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   was renamed from `experimental_versions` to `wdl_1_3` ([#435](https://github.com/stjude-rust-labs/sprocket/pull/435)).
 * Removed the `wdl-cli` crate, absorbing its code into the `sprocket` library
   crate in preparation for future refactoring ([#450](https://github.com/stjude-rust-labs/sprocket/pull/450)).
+* Apptainer-based backends now store converted container images within each run directory, rather than in a user-specified directory ([#463](https://github.com/stjude-rust-labs/sprocket/pull/460)).
+
 
 ### Fixed
 
--- a/crates/wdl-engine/src/backend/apptainer/images.rs
+++ b/crates/wdl-engine/src/backend/apptainer/images.rs
@@ -24,39 +24,16 @@
 //! still benefit from the Apptainer cache when building new `.sif` files, so
 //! this is not much of a slowdown, but it does increase disk space consumption
 //! depending on where the images directory is created.
-//!
-//! NOTE ACF 2025-09-22: This is currently a ⚠️ Hack Zone ⚠️ and is not meant to
-//! reflect final behavior.
-//!
-//! We don't currently have a notion of a top-level directory for an entire
-//! workflow execution; the `root` path for each workflow and task evaluator is
-//! specific to _that_ workflow or task, but the point of keeping our own cache
-//! of Apptainer images is to avoid pushing our luck with spotty
-//! container registries by inducing repeated requests for the same image.
-//!
-//! For expedience, this implementation makes the simplifying assumption that we
-//! have one top-level workflow execution per process, and keeps the images
-//! directory in a global variable. This should be replaced with something more
-//! robust, but currently fits the execution model of the `sprocket`
-//! CLI well enough to proceed.
-//!
-//! Since an ordinary `cargo test` runs each test executable once, and each
-//! executable can contain many targets, this hack has a particularly distorting
-//! effect on tests of this backend. Consider using `cargo nextest` to run each
-//! target in a separate process in order to eliminate cross-test target
-//! interference.
 
 use std::collections::HashMap;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Stdio;
 use std::sync::Arc;
-use std::sync::LazyLock;
 use std::sync::Mutex;
 
-use anyhow::Context as _;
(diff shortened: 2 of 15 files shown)


crates/wdl-engine/src/backend/apptainer/images.rs
```diff
@@ -70,89 +47,77 @@ use tracing::info;
...
-    })
-    .await
-    .cloned()
+            tokio::select! {
+                _ = cancellation_token.cancelled() =>
+                    return Err(anyhow!("task execution cancelled")),
+                res = retry =>
+                    res.with_context(
+                        || format!("pulling Apptainer image for container `{container}`")
+                    )?,
```
acfoltzer, line 112: `rustfmt` doesn't attempt to format `select` blocks, so I tried to manually reflow these once the column would've exceeded 100 chars. I'm open to suggestions if you'd prefer a different style.

crates/wdl-engine/src/eval/v1.rs
```diff
@@ -31,3 +40,61 @@ fn write_json_file(path: impl AsRef<Path>, value: &impl Serialize) -> Result<()>
...
+///
+/// "Top-level" here means the outermost invocation of a task or workflow across
+/// an entire execution. This type is suitable for once-per-execution values
+/// like a shared container image cache, and new instances should not be created
+/// for evaluating subgraphs of the initial execution.
+///
+/// This type is meant to be cheaply cloned and sendable between threads. When
+/// adding to it, make sure to use an `Arc` for any non-trivially-sized data.
+#[derive(Clone)]
+pub struct TopLevelEvaluator {
```
acfoltzer, line 53: The thing I like least about all of this is right here: I cannot figure out a good name for this. I think it might have to wait for future refactoring of `sprocket::eval` so we can steal the name `Evaluator`, but everything else I could think of was already taken. Suggestions are very welcome.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1215,23 +1153,24 @@ impl TaskEvaluator {
...
         // Write the outputs to the task's root directory
-        write_json_file(root_dir.join(OUTPUTS_FILE), &outputs)?;
+        write_json_file(task_eval_root.join(OUTPUTS_FILE), &outputs)?;
 
         evaluated.outputs = Ok(outputs);
         Ok(evaluated)
     }
+}
 
+impl<'a> State<'a> {
```
acfoltzer, line 1163: I realize it's funky to break up the `impl State` blocks like this, but in addition to minimizing the diff in this PR I think it reads better from top-to-bottom this way.

crates/wdl-engine/src/backend/apptainer/images.rs
claymcleod: Looks fine to me

crates/wdl-engine/src/eval/v1.rs
claymcleod: Two suggestions are `TargetEvaluator` or `RunEvaluator`.

- `target` is now the combined term we're using that roughly means "a task or workflow being run"—not perfect, but I think closer than `TopLevel` maybe.
- Alternatively, a `run` will mean "an evaluated task or workflow" once #470 lands. So you may preemptively name it something that aligns with that.

claymcleod: Or what about just `Evaluator`?

claymcleod (approved):
Looks good to me!

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   was renamed from `experimental_versions` to `wdl_1_3` ([#435](https://github.com/stjude-rust-labs/sprocket/pull/435)).
 * Removed the `wdl-cli` crate, absorbing its code into the `sprocket` library
   crate in preparation for future refactoring ([#450](https://github.com/stjude-rust-labs/sprocket/pull/450)).
+* Apptainer-based backends now store converted container images within each run directory, rather than in a user-specified directory ([#463](https://github.com/stjude-rust-labs/sprocket/pull/460)).
```
claymcleod: Just noticed, these both link to 460.

crates/wdl-engine/src/eval/v1.rs
acfoltzer: @claymcleod as mentioned above, I like `Evaluator`, but there are already types with that name (though fewer than before this change). Even if it wasn't directly conflicting, the number of other types with `Evaluator` in the name make me feel a need to differentiate how this one is different.

I remain skeptical of `target` here; that term is usually used to refer to the desired output of a command, e.g. a Makefile target is the file that evaluating the rules should cause to exist, or the `target/` directory that's the default output of a `cargo` invocation. [Entry point](https://en.wikipedia.org/wiki/Entry_point) is the canonical term for where a program begins execution.

It just occurred to me that `Engine` might work, and since it's in keeping with the crate name could make it more obvious to someone skimming the docs that this is where to start. It does conflict with `crankshaft::Engine`, though, dang. What do you think?

CHANGELOG.md
acfoltzer: Good catch, this is the sort of thing that makes me want to adopt some tooling for the changelog

crates/wdl-engine/src/eval/v1.rs
claymcleod: Hm. I guess, given all of this, I would prefer to just stay with `TopLevelEvaluator` and reconsider this question more fully when we do the interface refactor of `wdl-engine`. I don't think this question should block this PR and the release we hope to do soon.

PR #464, opened 2025-11-18
chore: make ignorefile part of `Analysis::Default`

a-frantz:

very small PR to reduce redundant code. I was working on the new `test` command and ran into a bug caused by my not properly setting the ignore file, which led me to this tiny refactor.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,12 +215,6 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.21.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
-
-[[package]]
-name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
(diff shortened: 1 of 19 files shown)


src/analysis.rs
```diff
@@ -75,6 +77,7 @@ impl Analysis {
     }
 
     /// Sets the ignorefile basename.
+    #[allow(unused)]
```
a-frantz: should this method just be removed? I'm not sure we'd ever want to support alternate ignore file names, which means this can probably just be axed.

peterhuene: If we don't have a corresponding CLI option to expose it, then I think it should go.

src/analysis.rs
```diff
@@ -180,7 +183,7 @@ impl Default for Analysis {
             exceptions: Default::default(),
             enabled_lint_tags: TagSet::new(&[]),
             disabled_lint_tags: TagSet::new(&[]),
-            ignore_filename: None,
+            ignore_filename: Some(IGNORE_FILENAME.to_string()),
```
peterhuene, line 179: Confused by this change. Would `None` not already use the default ignore file name?

a-frantz: No, this is where that default is defined. Before this change, when it was `None`, we weren't using _any_ ignorefile.

peterhuene: Ah gotcha, I think on first glance I was expecting it to be internal to `wdl-analysis`.

Never mind then!

PR #468, opened 2025-11-19
feat: first step towards the `test` command

a-frantz:

As the `wdl-engine` interface is currently undergoing a refactor, this PR does not interact with `wdl-engine` at all. This PR simply defines the YAML which _will_ be used for defining unit tests and does some basic parsing and logging as proof of concept.

https://github.com/stjude-rust-labs/rfcs/pull/1

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/src/commands/test.rs
+++ b/src/commands/test.rs
@@ -49,20 +49,23 @@ pub struct Args {
 const NESTED_TEST_DIR_NAME: &str = "test";
 
 fn find_yaml(wdl_path: &Path) -> Result<Option<PathBuf>> {
-    let mut result = None;
+    let mut result: Option<PathBuf> = None;
     let mut inner = |path: &Path| {
-        let yaml = path.with_extension("yaml");
-        if yaml.exists() && result.is_none() {
-            result = Some(yaml);
-        } else if yaml.exists() {
-            bail!("more than one test YAML for `{}`", wdl_path.display());
+        for ext in ["yaml", "yml"] {
+            let yaml = path.with_extension(ext);
+            match (yaml.exists(), &result) {
+                (true, Some(previous)) => bail!(
+                    "test file `{path}` conflicts with test file `{previous}`",
+                    path = yaml.display(),
+                    previous = previous.display()
+                ),
+                (true, None) => {
+                    result = Some(yaml);
+                }
+                _ => {}
+            }
         }
-        let yml = path.with_extension("yml");
-        if yml.exists() && result.is_none() {
-            result = Some(yml);
-        } else if yml.exists() {
-            bail!("more than one test YAML for `{}`", wdl_path.display());
-        };
+
         Ok(())
     };
     inner(wdl_path)?;


src/commands/test.rs
```diff
@@ -0,0 +1,214 @@
...
+        for (inner_index, possibility) in enumerate(individual_input_with_possible_values) {
+            if outer_index == 0 {
+                result.push(vec![possibility]);
+            } else {
+                result[inner_index].push(possibility);
+            }
+        }
+    }
+    result
+}
```
a-frantz: this function needs a new name and docstring, but I'm not sure about the technical name for the operation performed. It is similar to a classic zip, but works on any number (>0) of vecs and distributes the elements across the nested vecs differently than a zip. So. Welcoming any insight on how to modify the documentation here 🤷‍♀️

adthrasher: `build_input_combinations` or `build_input_permutations`?

src/commands/test.rs
```diff
@@ -0,0 +1,214 @@
...
+
+/// Compute an iterator of [`Run`]s from a user provided matrix of inputs.
```
adthrasher: This could be more verbose. The transformation isn't immediately clear when looking at the function, so I would describe it a little more fully here.

src/commands/test.rs
a-frantz: It's not really "building" anything 🤔 It's reshaping how the elements are shaped/distributed in a nested structure. Maybe `reshape_input_container`? That name kind of raises more questions than answers 😂

src/commands/test.rs
a-frantz: added step annotations 👍

src/commands/test.rs
a-frantz: The example wasn't quite right. It was actually showing an extra nest level, so the example you reviewed was really what `compute_runs_from_matrix()` was seeing instead of what the `zip_inputs()` function was seeing. Updated the example to be 3 separate invocations of the function.

a-frantz: `invert_nested_collection()`? I think the examples I added is illustrative of the inversion. A shape of 1x2 becomes 2x1, a 2x2 swaps around to 2x2, 2x3 becomes 3x2.

src/commands/test.rs
```diff
@@ -0,0 +1,273 @@
...
+/// ],
+/// ```
+fn zip_inputs(inputs_to_zip: Vec<Vec<Input>>) -> Vec<Vec<Input>> {
+    let mut result = Vec::new();
+    let mut initial_len = None;
+    for (outer_index, individual_input_with_possible_values) in enumerate(inputs_to_zip) {
+        if let Some(prev_len) = initial_len
+            && prev_len != individual_input_with_possible_values.len()
+        {
+            panic!("dimensions of input matrix are inconsistent")
```
a-frantz: @peterhuene the implementation you were working at yielded `None` when groups were of inconsistent length. Tagging you in case you want to argue for that behavior here. I'm of the opinion that we should reject those as poor test definitions rather than try to make sense of them. If a user wants to use a null value, they can specify it explicitly.

peterhuene: I agree we should error upon inconsistent subsequence lengths (assuming this is just maintaining an invariant that we enforce more gracefully elsewhere - I haven't completed my review yet); can we document that this particular function will panic when this is the case?

And I disagree with the doc comment: this is performing a zip across the subsequences, but it's also collecting each sequence into a `Vec` at each step (and collecting each step into a top level `Vec`).

I still believe this would be a great candidate for an implementation that returns iterators instead of collecting into `Vec`, but given this isn't performance critical, collecting is fine.

src/test.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl TestDefinition {
+    /// Parse the user-defined input matrix into an iterator of ordered maps of
+    /// input names to values.
+    ///
+    /// Each map represents a set of input keys whose values should be iterated
+    /// through together. The trivial case is a single input key with a set
+    /// of possible values. Groups of inputs that should be iterated through
+    /// together are designated by a YAML map key starting with `$`.
+    pub fn parse_inputs(&self) -> impl Iterator<Item = InputMapping> {
```
peterhuene: This method should be fallible and we should convert the panics into bail outs. Thus, this should probably collect so we can return something like `Result<Vec<InputMapping>>`.

peterhuene: We should also be validating that a group's subsequences are of equal length in this method so that we fail gracefully.

src/test.rs
```diff
@@ -0,0 +1,89 @@
...
+        })
+    }
+
+    /// Parse the defined assertions into an ordered map.
+    pub fn parse_assertions(&self) -> IndexMap<String, Value> {
+        self.assertions
+            .iter()
+            .filter_map(|(k, v)| {
+                if !k.is_string() {
+                    warn!("skipping non-string key: `{:?}`", k);
```
peterhuene: This feels like a "return an error" condition to me rather than a logged warning.

src/commands/test.rs
```diff
@@ -0,0 +1,325 @@
...
+    let results = Analysis::default()
+        .add_source(source.clone())
+        .run()
+        .await
+        .map_err(CommandError::from)?;
+
+    for result in results.filter(&[&source]) {
+        let document = result.document();
+        let wdl_path = PathBuf::from(document.path().as_ref());
+        let mut yaml_path = wdl_path.with_extension("yaml");
```
peterhuene: A discussion for the future, but as a developer I generally don't like to see test files cluttering up my source directory.

I would suggest that we look at a well-known location relative to the source file, like (spitballing) `<parent-of-source>/test/<source>.[yaml|yml]`.

Related to that, I could see paths in the test file treated as relative to the test YAML file; for example, `fixtures/foo.bam` would be at `<parent-of-source>/test/fixtures/foo.bam`. Then one doesn't need a `$FIXTURES` variable.

src/commands/test.rs
```diff
@@ -0,0 +1,325 @@
...
+        .multi_cartesian_product() // step 3
+        .map(|product| product.into_iter().flatten().collect()) // step 4 and 5
+}
+
+/// Performs the `test` command.
+pub async fn test(args: Args) -> CommandResult<()> {
+    let source = args.source.unwrap_or_default();
+    let source = match &source {
+        Source::Remote(_) => {
+            return Err(anyhow!("the `test` subcommand does not accept remote sources").into());
```
peterhuene, line 115: Is there a reason it couldn't in the future, though?

Other than fetching the `yaml` file for us to parse (which is relatively easy to do), it would just be a matter of joining URLs in inputs like we would for any local file paths (inputs that are `File` / `Directory` would also URLs).

src/test.rs
```diff
@@ -0,0 +1,89 @@
...
+                        panic!("expected sequence, got `{nested_val:?}`");
+                    };
+                    new_map.insert(k.to_string(), vals.to_vec());
+                }
+                new_map
+            } else {
+                let Value::Sequence(vals) = val else {
+                    panic!("expected sequence, got `{val:?}`");
+                };
+                IndexMap::from_iter(vec![(k.to_string(), vals.to_vec())])
```
peterhuene: One doesn't need to allocate a `Vec` here; arrays also implement `IntoIterator`.

```suggestion
                IndexMap::from_iter([(k.to_string(), vals.to_vec())])
```

src/test.rs
peterhuene: Additionally, we need to validate that inputs are uniquely specified, i.e. a group doesn't contain an input specified in any other group or at the top level. Using a single `HashSet<&str>` to detect the duplicate keys (both top-level and every key within a group mapping).

src/test.rs
peterhuene: Additionally, this seems like a we need a better abstraction between a single sequence of inputs and a grouping of sequences of inputs.

I would propose something like [this playground entry](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b3b14d58b57df619548ccdac853f9c33).

src/commands/test.rs
```diff
@@ -0,0 +1,325 @@
...
+/// A map of input keys to values which correspond to a single "run" or
+/// execution for Sprocket to test with. Should be a complete set of required
+/// inputs (potentially with values for optional inputs).
+///
+/// e.g.
+/// {
+///   "bams": ["$FIXTURES/test1.bam", "$FIXTURES/test2.bam"],
+///   "prefix": "test.merged",
+/// }
+type Run = IndexMap<String, Value>;
```
peterhuene: Question for here and other places we use mapping data structures: do we actually needs maps or can we store `Vec` in these places?

That is to say: do we expect to lookup entries by key name in these maps?

I think they're generally just iterated over to perform the cartesian product, which itself can produce a set of (name, value) pairs that will be used to construct an `Inputs` to run the workflow or task.

peterhuene (commented):
This looks like a great first step towards the test command.

I have a few feedback comments and at least one "discussion" comment regarding the UX of the feature that we can easily punt to later if desired.

src/commands/test.rs
a-frantz: You're right that it shouldn't be too hard to implement, but I question the usefulness of testing remote WDL. It seems like a feature we certainly could implement without too much extra work, but it strikes me as expanding our maintenance surface area for a feature no one would use.

src/commands/test.rs
a-frantz: The initial RFC did something similar to what you're suggesting, and (I think) Adam called out that as being potentially annoying having to open multiple directories to write WDL+tests in tandem. Although I think you also raise a good point that this implementation will lead to a lot of clutter in source directories, so maybe we need to rethink this 🤔 

What if we allow two places for tests to be written relative to a WDL file? Either what's currently implemented (`./<basename>.yaml`) or nested into a `test/` dir (`./test/<basename>.yaml`)? This facilitates a setup like the below:
```
.
├── tools
│   ├── bar.wdl
│   ├── baz.wdl
│   ├── foo.wdl
│   └── test
│       ├── bar.yml
│       ├── baz.yaml
│       └── foo.yaml
└── workflows
    ├── dna
    │   ├── dna.wdl
    │   └── dna.yaml
    └── rna
        ├── rna.wdl
        └── rna.yaml

```
When you have many WDL files sibling to each other, you can nest the test definitions under `test/`. If you want to keep the tests "closer" or don't want to introduce more nesting you can keep the files sibling.

I think we want still want to prescribe where tests can be defined, as this frees the test YAML from having to point back to the right WDL file. This also introduces an edge case where a user may create two test files for the same WDL file. Not sure if we should prevent that or maybe merge the tests. I'm leaning towards bailing if we find two test files for the same WDL just for simplicity. I can't think of a reason you'd want to split the tests across two files.

a-frantz:
@peterhuene I believe that I've "rustified" this sufficiently. We are yielding iterators, avoiding redundant `String` allocs, and validating+collecting errors as they are encountered then reporting all of the problems with found tests.

I've also gone ahead and added some simple tagging and filtering of tests (`--include-tag` and `--filter-tag` CL args).

Lastly, I've made another change to where tests are looked for. In addition to test YAML being looked for at "sibling" locations (e.g. `tools/foo.wdl` and `tools/foo.yaml`) tests can also be nested under a sibling `test/` directory (e.g. `tools/foo.wdl` and `tools/test/foo.yml`) The `cli:matrix_combinations` test shows an example:
```text
tests/cli/test/matrix_combinations/inputs/
├── example.wdl
├── example.yaml
├── flag_filter.wdl
└── test
    └── flag_filter.yml
```
Test files still have to have the same basename as the WDL they are testing.

I believe the above addresses all the received feedback 🎉 thanks for the review! LMK if there are any other changes we want to get in this PR

src/commands/test.rs
peterhuene: Sounds good, we can do a follow up PR for this.

src/commands/test.rs
```diff
@@ -0,0 +1,187 @@
...
+    ///
+    /// Can be repeated multiple times.
+    #[clap(short, long, value_name = "TAG",
+        action = clap::ArgAction::Append,
+        num_args = 1,
+    )]
+    pub filter_tag: Vec<String>,
+}
+
+const NESTED_TEST_DIR_NAME: &str = "test";
```
peterhuene, line 49: Nit: I wonder if we should use the plural `tests`? Not sure which is more common out in the wild. Happy with either, personally.

src/commands/test.rs
```diff
@@ -0,0 +1,187 @@
...
+        } else if yaml.exists() {
+            bail!("more than one test YAML for `{}`", wdl_path.display());
+        }
+        let yml = path.with_extension("yml");
+        if yml.exists() && result.is_none() {
+            result = Some(yml);
+        } else if yml.exists() {
+            bail!("more than one test YAML for `{}`", wdl_path.display());
+        };
+        Ok(())
```
peterhuene: It might be helpful to include the conflicting file paths, like so:

```rust
for ext in ["yaml", "yml"] {
    let yaml = path.with_extension(ext);
    match (yaml.exists(), &result) {
        (true, Some(previous)) => bail!("test file `{path}` conflicts with test file `{previous}`", path = yaml.display(), previous = previous.display()),
        (true, None) => {
            result = Some(yaml);
        }
        _ => {}
    }
}

Ok(())
```

src/commands/test.rs
a-frantz: I have s light preference for the singular, although I do not have any justification for that preference. I'm going to leave it as is, but if you [or someone else] has other thoughts on the pluralization I'm game to listen.

Issue #390, opened 2025-10-01
Docker backend: max CPU and max memory hints should be verified on the backend


peterhuene:
If we have this task:

```wdl
version 1.2

task test {
    command <<<>>>

    hints {
        max_cpu: 1000
    }
}
```

and run it with:

```
sprocket run test.wdl --entrypoint test
```

we get the following unhelpful error:

```
error: task execution failed for task `test`: Docker responded with status code 400: range of CPUs is from 0.01 to 8.00, as there are only 8 CPUs available
  ┌─ test.wdl:3:6
  │
3 │ task test {
  │      ^^^^ this task failed to execute
```

As we know the number of CPUs available, we should be emitting an error before we hand the task to Crankshaft and thus to the Docker daemon with its unhelpful 400 response.

We already do this for the `cpu` requirement, for example:

```
error: failed to get constraints for task `test`

Caused by:
    task requires at least 1000 CPUs, but the execution backend has a maximum of 8
```

This error message should also be improved to point at the requirement / hint that caused the problem, ideally.

dead8309:
i can work on this, just have to figure out 2nd part of the issue 
> This error message should also be improved to point at the requirement / hint that caused the problem, ideally.

peterhuene:
This was fixed with recent changes to Sprocket.

PR #484, opened 2025-12-01
fix: validate constraints in backend and point error messages to hints/requirements

dead8309:
closes: #390

For external contributors:

For all contributors:

### DIFF
--- a/crates/wdl-engine/src/backend/docker.rs
+++ b/crates/wdl-engine/src/backend/docker.rs
@@ -52,7 +52,6 @@ use crate::config::Config;
 use crate::config::DEFAULT_TASK_SHELL;
 use crate::config::DockerBackendConfig;
 use crate::config::TaskResourceLimitBehavior;
-use crate::path::EvaluationPath;
 use crate::tree::SyntaxNode;
 use crate::v1::container;
 use crate::v1::cpu;
--- a/crates/wdl-engine/src/backend/local.rs
+++ b/crates/wdl-engine/src/backend/local.rs
@@ -49,7 +49,6 @@ use crate::config::DEFAULT_TASK_SHELL;
 use crate::config::LocalBackendConfig;
 use crate::config::TaskResourceLimitBehavior;
 use crate::convert_unit_string;
-use crate::path::EvaluationPath;
 use crate::tree::SyntaxNode;
 use crate::v1::cpu;
 use crate::v1::cpu_from_values;
--- a/crates/wdl-engine/src/backend/lsf_apptainer.rs
+++ b/crates/wdl-engine/src/backend/lsf_apptainer.rs
@@ -47,7 +47,6 @@ use crate::backend::TaskExecutionResult;
 use crate::config::Config;
 use crate::config::LsfApptainerBackendConfig;
 use crate::config::TaskResourceLimitBehavior;
-use crate::path::EvaluationPath;
 use crate::tree::SyntaxNode;
 use crate::v1;
 use crate::v1::cpu;
--- a/crates/wdl-engine/src/backend/slurm_apptainer.rs
+++ b/crates/wdl-engine/src/backend/slurm_apptainer.rs
@@ -46,7 +46,6 @@ use crate::backend::TaskExecutionResult;
 use crate::config::Config;
 use crate::config::SlurmApptainerBackendConfig;
 use crate::config::TaskResourceLimitBehavior;
-use crate::path::EvaluationPath;
 use crate::tree::SyntaxNode;
 use crate::v1;
 


crates/wdl-engine/src/backend/docker.rs
```diff
@@ -343,13 +351,19 @@ impl TaskExecutionBackend for DockerBackend {
...
+    ) -> Result<TaskExecutionConstraints, Diagnostic> {
         let container = container(requirements, self.config.task.container.as_deref());
 
         let mut cpu = cpu(requirements);
         if (self.max_cpu as f64) < cpu {
+            let span = task
+                .runtime()
+                .and_then(|r| r.items().find(|i| i.name().text() == TASK_REQUIREMENT_CPU))
+                .map(|i| i.span())
+                .unwrap_or_else(|| task.span());
```
acfoltzer: Should this be looking in the `requirements` section rather than the deprecated `runtime` section to find the relevant span?

acfoltzer: I'm seeing that a similar pattern repeats elsewhere in this PR: the value is looked up from the `requirements`, but then blame is ascribed to a `runtime` key. I think we should consider changing the `cpu()`, `memory()`, etc helpers so that they look up the respective keys in both `requirements` and `runtime` for WDL 1.2+, erroring if present in both. The initial lookup should return a type with enough information to both get the `f64`/`u64`/etc value along with a span so that we don't have to look them up multiple times.

dead8309: i thought it was `requirements` that got deprecated and not runtime😭. fixed

acfoltzer: Gotcha. What do you think of moving more logic into the helper functions? There is still a lot of repeated lookup logic here that I think would be best to avoid.

crates/wdl-engine/src/backend.rs
```diff
@@ -288,9 +291,10 @@ pub trait TaskExecutionBackend: Send + Sync {
     /// environment or if the task specifies invalid requirements.
     fn constraints(
         &self,
+        task: &TaskDefinition<SyntaxNode>,
         requirements: &HashMap<String, Value>,
         hints: &HashMap<String, Value>,
-    ) -> Result<TaskExecutionConstraints>;
+    ) -> Result<TaskExecutionConstraints, Diagnostic>;
```
acfoltzer, line 336: Just want to flag this signature change for @peterhuene, given his impending plans to do some cleanup in this crate. We'd also previously discussed some unification of this `constraints` method (which is really more about [giving the task's evaluation context access to the requirements](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#runtime-access-to-requirements-hints-and-metadata)) with the often-repeated logic in `spawn`, but that's beyond the scope of this PR.

crates/wdl-engine/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Added shared validation for task runtime and hint numeric settings (memory, max_memory, max_retries, preemptible), rejecting invalid or negative values earlier ([#485](https://github.com/stjude-rust-labs/sprocket/pull/485)).
 
+#### Fixed
+
+* Validate constraints in docker backend and point error messages to problematic `hints`/`requirements` sections ([#484](https://github.com/stjude-rust-labs/sprocket/pull/484)).
```
acfoltzer: This note probably should not be scoped just to the Docker backend.
```suggestion
* Validate constraints in the Docker backend, and point error messages to problematic `hints`/`requirements` sections for all backends ([#484](https://github.com/stjude-rust-labs/sprocket/pull/484)).
```

crates/wdl-engine/src/eval/v1.rs
```diff
@@ -46,6 +47,15 @@ fn write_json_file(path: impl AsRef<Path>, value: &impl Serialize) -> Result<()>
...
 }
 
+/// A helper struct to associate a value with it's source span.
+#[derive(Debug)]
+pub struct Spanned<T> {
+    /// The value
+    pub value: T,
+    /// The span from the source document.
+    pub span: Span,
+}
```
acfoltzer: This is generic enough that it should probably go next to `Span`'s definition in `wdl_grammar::Diagnostic`

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -211,8 +243,10 @@ pub(crate) fn container<'a>(
         })
 }
 
-/// Gets the `cpu` requirement from a requirements map.
-pub(crate) fn cpu(requirements: &HashMap<String, Value>) -> f64 {
+/// Gets the `cpu` requirement from a map of requirement values.
+///
+/// This is intended for use where the task's AST is not available.
+pub(crate) fn cpu_from_values(requirements: &HashMap<String, Value>) -> f64 {
```
acfoltzer, line 263: I just went down a bit of a rabbit hole trying to figure out what circumstance we would possibly want to pull out a `requirements` value without having the task AST on hand, and I'm realizing that the signature for `TaskExecutionBackend::spawn()` does not give the implementor access to the task itself. It seems like this fix is really exposing the contradictions in the current design that motivated my earlier discussion with @peterhuene about wanting a more comprehensive overhaul here.

Since I'm not going to be around on the project to see it through, I don't want to definitively weigh in on whether we should push for more of a refactor in this PR while we're already changing method signatures on `TaskExecutionBackend`, or if we should proceed with this change for the short term and hold off on larger restructuring. @peterhuene, would you mind making this call?

peterhuene: I think we can proceed with this for the short term and I plan on refactoring this area for the future.

PR #486, opened 2025-12-02
feat: add LSF job name prefix and byte limit

### LINKED ISSUE #362: Add customizable prefix to LSF job names and fix length limiting
> Personally, I'd want something like `sprocket_*` so that I can easily filter these out if I'm doing other work.

_Originally posted by @adthrasher in https://github.com/stjude-rust-labs/sprocket/pull/182#discussion_r2376158411_
            
This should be fairly straightforward, but when I tried to implement it as a quick add-on to the initial PR I noticed that the logic to limit the length of job names is flawed. LSF accepts unicode characters in the job name, but limits the length by bytes. 

Limiting by bytes isn't a problem if job names would otherwise be ASCII-only, and Sprocket's lexer only accepts a subset of ASCII for identifiers. So the easiest thing is to probably confirm this is OK (the spec doesn't seem to give an identifier syntax) and then do the simple thing based on bytes.

raman976:
Add a customizable prefix to LSF job names and fix how their length is limited. This PR introduces an optional job_name_prefix for the LSF + Apptainer backend, uses a new helper to build prefix + identifier while enforcing LSF’s 4094‑byte UTF‑8 limit, and adds unit tests plus a sample config update to cover the new behavior.

Fixes #362 

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/CHANGELOG.md
+++ b/crates/wdl-engine/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+* Added a configurable prefix for LSF job names and ensured the byte-length
+  limit matches LSF's behavior ([#362](https://github.com/stjude-rust-labs/sprocket/issues/362)).
+
 #### Fixed
 
 * Fixed an issue where task requirements and hints sources from an inputs file
--- a/crates/wdl-engine/src/backend/lsf_apptainer.rs
+++ b/crates/wdl-engine/src/backend/lsf_apptainer.rs
@@ -59,6 +59,42 @@ const APPTAINER_COMMAND_FILE_NAME: &str = "apptainer_command";
 /// See <https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=o-j>.
 const LSF_JOB_NAME_MAX_LENGTH: usize = 4094;
 
+/// Construct an LSF job name by combining an optional prefix with the task
+/// identifier while ensuring the total length fits within the byte-oriented
+/// LSF limit.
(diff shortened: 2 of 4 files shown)


raman976:
The CHANGELOG diff also shows some older bullets changing from * to -. I think that came from my editor’s Markdown formatting, the content of those entries is unchanged. If you’d prefer, I’m happy to switch them back to the original style.

a-frantz:
> I'm not sure it matters in practice, but truncating the prefix when the name is too long feels incorrect. Anything downstream that depends on the prefix would break in that case.

Job names in LSF are just for display purposes, right? I think the logic here was to avoid truncating the "identity" part of the name (and avoid duplicitous names?), but I don't think that will have any meaningful ramifications, whereas removing an intentional prefix might.

crates/wdl-engine/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
```
peterhuene:
```suggestion
#### Added
```

crates/wdl-engine/CHANGELOG.md
```diff
@@ -9,10 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 #### Added
 
+
+* Added a configurable prefix for LSF job names and ensured the byte-length limit matches LSF's behavior ([#362](https://github.com/stjude-rust-labs/sprocket/issues/362)).
+
 * Added setting for controlling content digests; supported values are `strong`
   for full cryptographic hashing of file content and `weak` to digest based
   solely off file metadata ([#503](https://github.com/stjude-rust-labs/sprocket/pull/503)).
 
+
```
peterhuene:
```suggestion
```

PR #488, opened 2025-12-03
fix: improve digest caching.

peterhuene:
This PR improves digest caching where the calculation of a local directory digest will now cache any descendant file or directory digest that is calculated.

The fix should help improve performance when a task creates a large file in its working directory and a direct reference to the file is later used by another task.

Without this fix, the above would cause the referenced file to be digested twice: once as a descendant of the directory and again for the direct reference to the file.

Note: due to the change in how local directory digests are calculated, this fix will cause call cache misses for any entry that references a local directory (such as a local working directory).

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/CHANGELOG.md
+++ b/crates/wdl-engine/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
 
-* Cache file digests that are calculated during a directory digest calculation ([#488](https://github.com/stjude-rust-labs/sprocket/pull/488)).
+* Cache file digests that are calculated during local directory digest
+  calculation. Note: this change will bust existing local cache entries ([#488](https://github.com/stjude-rust-labs/sprocket/pull/488)).
 
 ## 0.10.0 - 11-21-2025
 
--- a/crates/wdl-engine/src/digest.rs
+++ b/crates/wdl-engine/src/digest.rs
@@ -171,7 +171,7 @@ fn calculate_directory_digest(path: &Path) -> impl Future<Output = Result<Digest
             // Hash the relative path to the entry
             let entry_rel_path = entry_path
                 .strip_prefix(path)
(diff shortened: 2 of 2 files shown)


crates/wdl-engine/src/digest.rs
```diff
@@ -114,6 +115,86 @@ async fn get_content_digest(transferer: &dyn Transferer, url: &Url) -> Result<Ar
...
+                format!(
+                    "failed to read metadata for path `{path}`",
+                    path = entry_path.display()
+                )
+            })?;
+
+            // Hash the relative path to the entry
+            let entry_rel_path = entry_path
+                .strip_prefix(path)
+                .unwrap_or(&entry_path)
```
a-frantz: what case does the `_or` cover? Shouldn't `path` _always_ be a prefix for `entry_path`?

crates/wdl-engine/CHANGELOG.md
```diff
@@ -9,8 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Changed
 
-* Added shared validation for task runtime and hint numeric settings (memory, max_memory, max_retries, preemptible), rejecting invalid or negative values earlier ([#485](https://github.com/stjude-rust-labs/sprocket/pull/485)).
+* Added shared validation for task runtime and hint numeric settings (memory,
+  max_memory, max_retries, preemptible), rejecting invalid or negative values
+  earlier ([#485](https://github.com/stjude-rust-labs/sprocket/pull/485)).
 
+#### Fixed
```
a-frantz, line 16: I think we've only used "Added", "Changed", and "Fixed", but maybe this should be under a "Performance" header? It's not like the prior behavior was "broken" exactly 🤷‍♀️

peterhuene: I generally consider perf improvements "fixes", but if folks think a separate section calling out performance improvements is best, then happy to do that.

PR #489, opened 2025-12-09
feat: execute tests

a-frantz:

Execute the test definitions defined by #468 🚂 

relevant links:
- https://github.com/stjude-rust-labs/rfcs/pull/1
- https://github.com/stjudecloud/workflows/pull/263
- https://github.com/stjude-rust-labs/sprocket.bio/pull/22

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/src/document.rs
+++ b/crates/wdl-analysis/src/document.rs
@@ -894,6 +894,21 @@ impl Document {
         None
     }
 
+    /// Gets a cache key for an enum variant lookup.
+    pub fn get_variant_cache_key(
+        &self,
+        name: &str,
+        variant: &str,
+    ) -> Option<crate::types::EnumVariantCacheKey> {
+        let (enum_index, _, r#enum) = self.data.enums.get_full(name)?;
+        let enum_ty = r#enum.ty()?.as_enum()?;
+        let variant_index = enum_ty.variants().iter().position(|v| v == variant)?;
+        Some(crate::types::EnumVariantCacheKey::new(
+            enum_index,
+            variant_index,
+        ))
+    }
+
     /// Gets the parse diagnostics for the document.
     pub fn parse_diagnostics(&self) -> &[Diagnostic] {
         &self.data.parse_diagnostics
--- a/crates/wdl-analysis/src/types.rs
+++ b/crates/wdl-analysis/src/types.rs
@@ -1064,6 +1064,25 @@ impl Coercible for StructType {
     }
 }
 
+/// Cache key for enum variant values.
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq)]
+pub struct EnumVariantCacheKey {
+    /// The index of the enum in the document.
+    enum_index: usize,
+    /// The index of the variant within the enum.
+    variant_index: usize,
+}
+
+impl EnumVariantCacheKey {
+    /// Constructs a new enum variant cache key.
+    pub(crate) fn new(enum_index: usize, variant_index: usize) -> Self {
+        Self {
+            enum_index,
+            variant_index,
+        }
+    }
+}
+
 /// Represents the type of an enum.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct EnumType {
--- a/crates/wdl-engine/CHANGELOG.md
+++ b/crates/wdl-engine/CHANGELOG.md
@@ -42,6 +42,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
 
+* Cached enum variant values at the Evaluator level to avoid redundant AST
+  lookups and parsing, improving performance for workflows with heavy enum
+  usage ([#511](https://github.com/stjude-rust-labs/sprocket/pull/511)).
 * Fixed enum variant serialization in command placeholders
   ([#534](https://github.com/stjude-rust-labs/sprocket/pull/534)).
 * Fixed `write_json` not translating host paths to guest paths ([#530](https://github.com/stjude-rust-labs/sprocket/pull/530)).
--- a/crates/wdl-engine/src/eval/v1.rs
+++ b/crates/wdl-engine/src/eval/v1.rs
@@ -5,10 +5,12 @@ mod task;
 mod validators;
 mod workflow;
 
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::BufWriter;
 use std::path::Path;
 use std::sync::Arc;
+use std::sync::Mutex;
 
 use anyhow::Context;
 use anyhow::Result;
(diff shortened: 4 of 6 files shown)


crates/wdl-engine/src/eval.rs
```diff
@@ -800,6 +815,54 @@ impl<'a> ScopeRef<'a> {
...
+    #[error("{}", .0.error())]
+    TaskFailed(FailedTask),
+    /// Something else caused the failure.
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+}
+
+/// Represents the exit conditions of a task which failed evaluation.
+#[derive(Debug)]
+pub struct FailedTask {
```
peterhuene: I think a simpler approach to adding a dependency on `thiserror` and adding this `FailedTask` type would be to stop treating a "failed" task as an evaluation error from `TopLevelEvaluator::evaluate_task` (and the internal `perform_task_evaluation` method).

Instead, if a task is considered a failure, we return `Ok(EvaluatedTask)` and on the evaluated task we have two important methods: `failed` which returns `true` if the evaluated task was considered to be a failure based on its requirements and `into_result` which converts the evaluated task into an evaluation result returning its outputs (this would be the place we return an evaluation error for the failed task, i.e. when you try to get the task's outputs).

Internally we already store an `EvaluationResult<Outputs>` which might contain the evaluation error we might format for a failed task, albeit this field isn't used quite like it used to be (it seems to always contain an `Ok` variant). 

Looking this over, this code needs a little refactoring anyway. Would you mind if I quickly put up a PR to fix this in such a way that should enable `sprocket test` to know if a task or workflow evaluation was due to a failed task while also fixing some of the other issues I'm spotting now?

peterhuene: I implemented #498 that I think can be used in favor of your changes to `wdl-engine`.

src/commands/test.rs
```diff
@@ -44,9 +95,34 @@ pub struct Args {
         num_args = 1,
     )]
     pub filter_tag: Vec<String>,
+    /// Do not clean the file system of successful tests.
+    ///
+    /// The default behavior is to remove directories of successful tests,
+    /// leaving only failed and errored run directories on the file system.
+    #[clap(long)]
+    pub do_not_clean: bool,
```
peterhuene: nit: I think that a name of `keep` or perhaps `no-clean` to keep with "no" option naming conventions.

src/commands/test.rs
```diff
@@ -72,62 +148,196 @@ fn find_yaml(wdl_path: &Path) -> Result<Option<PathBuf>> {
...
-    info!("logging each individual execution defined by test matrix");
-    let mut counter = 0;
-    let matrix = test.parse_inputs()?;
-    for run in matrix.cartesian_product() {
-        info!("execution with inputs: {:#?}", run.collect::<Vec<_>>());
-        counter += 1;
+    false
+}
+
+fn file_contains_regexs<'a>(path: &'a str, regexs: &'a [String]) -> Result<Result<(), &'a str>> {
```
peterhuene: Typically compilation of regular expression automata can be expensive depending on the complexity of the regex. This function is called up to twice per test (which itself is a cartesian product), right?

It might be best to compile these into a `Vec<Regex>` in the assertions once and pass them as a `&[Regex]` to this function (perhaps with a name of `file_matches` instead?).

src/commands/test.rs
```diff
@@ -72,62 +148,196 @@ fn find_yaml(wdl_path: &Path) -> Result<Option<PathBuf>> {
...
+    }
+}
+
+enum IterationResult {
+    Success,
+    Fail(anyhow::Error),
 }
 
 /// Performs the `test` command.
 pub async fn test(args: Args) -> CommandResult<()> {
```
peterhuene: I think we can probably break this function up into multiple; at the very least, I think perhaps extracting the "populate the results" loop and the "process the results" loop into their own functions.

src/commands/test.rs
```diff
@@ -44,9 +98,34 @@ pub struct Args {
...
     )]
     pub filter_tag: Vec<String>,
+    /// Do not clean the file system of successful tests.
+    ///
+    /// The default behavior is to remove directories of successful tests,
+    /// leaving only failed and errored run directories on the file system.
+    #[clap(long)]
+    pub no_clean: bool,
+    /// Clean all exectuion directories, even for tests that failed or errored.
+    #[clap(long)]
```
peterhuene, line 107: Nit: this should probably be marked as conflicting with the `no_clean` option.

src/commands/test.rs
```diff
@@ -72,62 +151,420 @@ fn find_yaml(wdl_path: &Path) -> Result<Option<PathBuf>> {
...
     if test.tags.iter().any(|t| filter_tags.contains(t)) {
-        println!("skipping test because of tag");
-        return Ok(());
+        return true;
+    }
+    false
+}
+
+fn file_matches<'a>(path: &'a str, regexs: &'a [Regex]) -> Result<Result<(), &'a str>> {
+    let contents = read_to_string(File::open(path).with_context(|| "opening file")?)
```
peterhuene: These two calls can be replaced with `std::fs::read_to_string`:

```suggestion
    let contents = read_to_string(path).with_context(|| format!("failed to read file `{path}`))?;
```

src/commands/test.rs
```diff
@@ -72,62 +151,420 @@ fn find_yaml(wdl_path: &Path) -> Result<Option<PathBuf>> {
...
     }
     if test.tags.iter().any(|t| filter_tags.contains(t)) {
-        println!("skipping test because of tag");
-        return Ok(());
+        return true;
+    }
+    false
+}
+
+fn file_matches<'a>(path: &'a str, regexs: &'a [Regex]) -> Result<Result<(), &'a str>> {
```
peterhuene: Why the nested `Result`s?

peterhuene: After looking at the call sites, I think this might make more sense being:

```rust
/// Checks that the contents of the given file path match every given regular expression.
///
/// Returns `Ok(None)` if the file's contents match every regular expression.
/// Returns `Ok(Some(regex))` upon the first unmatched regular expression.
fn file_matches<'a>(path: &str, regexs: &'a [Regex]) -> Result<Option<&'a str>>
```

src/commands/test.rs
```diff
@@ -72,62 +151,420 @@ fn find_yaml(wdl_path: &Path) -> Result<Option<PathBuf>> {
...
+                };
+                let run_dir = run_root.join(test_num.to_string());
+                let events = Events::disabled();
+                let name = test_name.clone();
+                let fixtures = fixtures.clone();
+                let engine = engine.clone();
+                let entrypoint = entrypoint.clone();
+                let assertions = assertions.clone();
+                let document = wdl_document.clone();
+                futures.spawn(async move {
```
peterhuene, line 428: A comment to keep in mind for the future, but is fine for this being an "in-development" command:

Given the current "isolated" design of `Evaluator`, we'll definitely want to place some limits on the amount of parallelism supported for the `dev test` command.

The `Evaluator` will need to be refactored a bit to make this more efficient and prevent there being multiple Docker backends being used, all with their own internal resource management which can easily overload a local Docker daemon.

PR #498, opened 2025-12-16
refactor: change `evaluate_task` to return "failed" evaluated tasks.

peterhuene:
This PR refactors `TopLevelEvaluator::evaluate_task` to return the evaluated task when it has an unacceptable exit code.

Previously we were treating it as an evaluation error, which internally gives no context other than the error message as to what the task execution result was; this is needed for implementing assertions in `sprocket test`).

Calling `into_outputs` on the returned evaluated task will return an error if it had an unacceptable exit code; the error will have an associated task execution result that stores the task's exit code.

This also includes the following fixes:

* Only attempt to download `stderr` of the failed task when the maximum retry is exceeded.
* We now properly invoke the execution backend cleanup when the task had an unacceptable exit code.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/src/eval.rs
+++ b/crates/wdl-engine/src/eval.rs
@@ -345,10 +345,6 @@ pub struct CallLocation {
 pub struct SourceError {
     /// The document originating the diagnostic.
     pub document: Document,
-    /// The task execution result that caused the error.
-    ///
-    /// This is `Some` only when a task fails due to an unacceptable exit code.
-    pub execution_result: Option<TaskExecutionResult>,
     /// The evaluation diagnostic.
     pub diagnostic: Diagnostic,
     /// The call backtrace for the error.
@@ -376,21 +372,9 @@ impl EvaluationError {
     pub fn new(document: Document, diagnostic: Diagnostic) -> Self {
         Self::Source(Box::new(SourceError {
             document,
-            execution_result: None,
             diagnostic,
             backtrace: Default::default(),
         }))
-    }
-
-    /// Sets the associated task execution result for the error.
-    ///
-    /// If the error is not a source error, the result is ignored.
-    pub(crate) fn with_execution_result(mut self, result: TaskExecutionResult) -> Self {
-        if let Self::Source(e) = &mut self {
-            e.execution_result = Some(result);
-        }
-
-        self
     }
 
     /// Helper for tests for converting an evaluation error to a string.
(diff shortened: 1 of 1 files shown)


crates/wdl-engine/src/eval.rs
```diff
@@ -359,6 +345,10 @@ pub struct CallLocation {
 pub struct SourceError {
     /// The document originating the diagnostic.
     pub document: Document,
+    /// The task execution result that caused the error.
+    ///
+    /// This is `Some` only when a task fails due to an unacceptable exit code.
```
a-frantz: This actually needs to be populated on  success as well for `test`. #489 doesn't tackle it yet, but we will want to support assertions on the stdout and stderr files of successful executions. That change could be done in a follow up PR/test PR.

peterhuene: Upon success you should be able to get the exit code, stdout, and stderr from the evaluated task though, I think.

peterhuene: In fact, with these changes, you'll get back an `EvaluatedTask` from `evaluate_task` even if it had an unacceptable exit code; this field is only useful when you:

* call `into_outputs` and it returns an error (which should then have the inner task execution result).
* get an evaluation error from `evaluate_workflow` due to an inner call to a task that failed with an unacceptable exit.

I think we're not going to support asserting on the details of a failed task from within a workflow evaluation, correct? If so, I wonder if we should just remove this field.

a-frantz: yes I think you're right. Although this raises something I haven't thought about very hard, which is that there aren't many meaningful ways to test/evaluate a failed workflow. I wonder if we should be removing the ability to test workflows entirely? Or rather, only allow "Success" assertions on workflow runs?

I believe that a workflow can only fail:
- due to a called task failing, in which case the "correct" way to test that is via the task itself rather than trying to extract task info from a workflow evaluation
- a runtime error during evaluation of a private declaration in the workflow, which is... tough to assert on? I'm not sure what that looks like

these are really questions for `test` and are a bit off topic of this PR, but you've got me thinking :thinking:

a-frantz (approved):

crates/wdl-engine/src/eval.rs
peterhuene: I think that testing a workflow is useful, but probably just asserting it evaluates successfully end-to-end (i.e. kind of an integration test) and more specific assertions limited to testing tasks (i.e. kind of a unit test).

PR #501, opened 2025-12-17
wdl-lint: ignore LineWidth diagnostics for import statements

### LINKED ISSUE #355: bug: LineWidth still fires on import statements
https://github.com/stjude-rust-labs/wdl/pull/590 apparently didn't fix this? Not sure what is going wrong here, and looks like we (I) put 590 through without tests 😬 

I tried removing the exceptions here - https://github.com/stjudecloud/workflows/blob/main/workflows/chipseq/chipseq-standard.wdl#L11-L16
and the notes still emitted with the latest sprocket binary

kris70lesgo:
This fixes a case where the LineWidth lint was still triggering on importstatements, even though imports are supposed to be fully excluded from linewidth checks.
The issue was that LineWidth diagnostics could end up attached to child tokens inside an import (for example the URI string, `as` namespace, or alias clauses).
Because of the visitor order, whitespace after an import line could still cause a LineWidth diagnostic to be emitted.
The fix skips LineWidth checks when the target element is inside an ImportStatement node. Tests were added to cover long import URls, imports with as clauses, and WDL 1.2 alias clauses, all of which should not produce any LineWidth diagnostics.

Resolved : #355 

For external contributors:

- [x] You have read the contributing guide in its entirety.

For all contributors:

- [x] You have either (a) made a PR to the `next` branch in the sprocket.bio repository
      [if you are a Sprocket team member] or (b) have suggested any changes you think
      need to be made to sprocket.bio in the description of your PR
      [if you are an external contributor].

      is implemented.


crates/wdl-lint/tests/lints/import-line-width/source.wdl
```diff
@@ -0,0 +1,23 @@
+version 1.2
+
+# Test case 1: Long URI that exceeds 90 characters
+import "this_is_a_very_very_very_very_very_very_very_very_very_very_long_filename_that_exceeds_ninety_characters.wdl"
```
peterhuene: Instead of a very long local file name, could we perhaps use a URL instead, such as `https://raw.githubusercontent.com/stjudecloud/workflows/refs/heads/main/tools/librarian.wdl` (randomly chosen)?

crates/wdl-lint/src/rules/line_width.rs
```diff
@@ -123,14 +123,23 @@ impl Visitor for LineWidthRule {
     }
 
     fn whitespace(&mut self, diagnostics: &mut Diagnostics, whitespace: &Whitespace) {
+        let element = whitespace
+            .inner()
+            .prev_sibling_or_token()
+            .unwrap_or(SyntaxElement::from(whitespace.inner().clone()));
+
+        if element
```
peterhuene:
```suggestion
        // If the whitespace is within an import statement, ignore the line width rule
        // Import statements may often be too long due to import paths or URLs
        if element
```

peterhuene (commented):
Hi @kris70lesgo. Thanks very much for the fix!

I have just a suggestion and a review comment to go over.

Also, can you include an entry under the `#### Fixed` header for the `CHANGELOG.md` in `crates/wdl-lint` with a description that also links to this PR?

peterhuene:
Also, it looks like we'll need to update the blessed diagnostics for arena, as this fix removes diagnostics that shouldn't have been committed previously.

To bless, run this command and commit the resulting change to `Arena.toml`:

```
cargo run -p gauntlet --release --bin gauntlet -- --arena --bless
```

crates/wdl-lint/tests/lints/import-line-width/source.wdl
kris70lesgo: Thanks for the suggestion, updated the test to use the referenced URL import instead long filename

kris70lesgo:
@peterhuene can u review the pr ?

crates/gauntlet/Arena.toml
```diff
@@ -6617,11 +6352,56 @@ document = "stjudecloud/workflows:/tools/util.wdl"
 message = "util.wdl:83:20: note[ContainerUri]: container URI uses a mutable tag"
 permalink = "https://github.com/stjudecloud/workflows/blob/60c4e506136dbebc687c3db7b1e8d3a745f488c8/tools/util.wdl/#L83"
 
+[[diagnostics]]
+document = "stjudecloud/workflows:/workflows/chipseq/chipseq-standard.wdl"
```
peterhuene: Curious as these should not be added. There's something going wrong here and I'm looking into it.

peterhuene: The problem is that we must call `detect_line_too_long` as it has some internal book keeping relating to the previous newline offset.

I'll push up a fix.

peterhuene (approved):
@kris70lesgo thanks for fixing this!

a-frantz (approved):

peterhuene:
@a-frantz hold off on merging this. Something isn't right with the test but it's still passing CI.

a-frantz (approved):

Issue #247, opened 2025-04-11
feature request: `wdl-format` should be more opinionated about lint directives


a-frantz:
specifically, `wdl-format` should add a space after the comma delimiters of `#@ except` comments and also alphabetize the rules.

Probably belongs in a follow up PR to the above (because it's more complicated), but `wdl-format` should also be able to break long lint directives into multiple lines. e.g.
```
#@ except:CommentWhitespace,DeprecatedObject,MetaDescription,InputSorted,ParameterMetaMatched,MatchingOutputMeta
```
would become
```
#@ except: CommentWhitespace, DeprecatedObject, InputSorted
#@ except: MatchingOutputMeta, MetaDescription, ParameterMetaMatched
```

kris70lesgo:
@a-frantz I would like to work on this issue 
My approach : 
I’ve read the contributing guide 
I plan to work on improving wdl-format so that #@ except: directives:

- insert a space after comma delimiters
- alphabetize rule IDs deterministically

I’ll start by handling single-line directives only and add formatter tests to ensure idempotent output
I’ll leave multiline splitting for a follow-up PR unless you prefer it included here
Please let me know if this approach aligns with expectations

PR #505, opened 2025-12-18
Wdl format except directive normalization

kris70lesgo:
This PR updates wdl-format to normalize single-line `#@ except:` lint directives.
the formatter now:
- Inserts space after comma 
- Sorts rule ids alphabetically in a case-insensitive manner
This change is limited to single-line directives only. Handling of multiline splitting is left for another PR, as discussed in the original issue.
Related issue: #247

### For external contributors

- [x] You have read the contributing guide in its entirety.

### For all contributors

- [x] You have either (a) made a PR to the `next` branch in the sprocket.bio repository or (b) have suggested any changes you think need to be made to sprocket.bio in the description of your PR.

### For PRs containing lint rule changes

- [x] You have updated any and all affected entries within `RULES.md`.
- [x] You have added a test case in `crates/wdl-lint/tests/lints` that covers every possible diagnostic emitted for the rule within the file where the rule is implemented.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4027,7 +4027,6 @@ dependencies = [
  "similar",
  "strum",
  "tempfile",
- "thiserror 2.0.17",
  "tokio",
  "tokio-util",
  "toml 0.9.8",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -178,7 +178,6 @@ serde_json.workspace = true
 serde_yaml_ng.workspace = true
 similar.workspace = true
 strum.workspace = true
-thiserror.workspace = true
 tokio.workspace = true
 tokio-util.workspace = true
 toml.workspace = true
--- a/crates/wdl-analysis/CHANGELOG.md
+++ b/crates/wdl-analysis/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Changed
+
+* Removed `Arc` from type-related enumerations ([#510](https://github.com/stjude-rust-labs/sprocket/pull/510)).
+
+#### Fixed
+
+* Fixed a bug where a dependency cycle was incorrectly detected due to an
+  invalid edge being added between nodes in the dependency graph when a scatter
+  variable shadowed an outer declaration of the same name (#[509](https://github.com/stjude-rust-labs/sprocket/pull/509)).
+
 ## 0.15.0 - 11-21-2025
 
 #### Added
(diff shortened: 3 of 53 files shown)


crates/wdl-format/src/token/pre.rs
```diff
@@ -11,6 +11,44 @@ use crate::TokenStream;
...
+/// Normalize single-line `#@ except:` directives
+fn normalize_except_directive(text: &str) -> String {
+    // Check if comment starts with #@
+    let text = text.trim();
+    if !text.starts_with("#@") {
+        return text.to_owned();
+    }
+
+    // Strip #@ prefix and get remainder
+    let remainder = text[2..].trim_start();
```
peterhuene: Nit: this can be combined with:

```suggestion
    let Some(remainder) = text.trim_start().strip_prefix("#@") else { return text.to_owned(); } 
```

crates/wdl-format/src/token/pre.rs
```diff
@@ -11,6 +11,44 @@ use crate::TokenStream;
...
+        return text.to_owned();
+    }
+
+    // Check if single-line (no newlines in content)
+    if text.contains('\n') {
+        return text.to_owned();
+    }
+
+    // Extract the rule list after "except:"
+    let rules_text = &remainder[7..]; // Skip "except:"
```
peterhuene: Same goes for here (note: needs to be `rustfmt`'d):

```suggestion
    let Some(rules_text) = remainder.trim_start().strip_prefix("except:") else { return text.to_owned(); }
    
    // Check if single-line (no newlines in content)
    if rules.contains('\n') {
        return text.to_owned();
    }
```

a-frantz: Peter's edit fixes a bug which would be present for comments with odd whitespace (the `trim_start()` is key here)

We also don't need the `\n` check, as a `SyntaxKind::Comment` can only be a single line (and it always contains a trailing newline, but `trim_end()` is called before this function so this check will never be true)

crates/wdl-format/src/token/pre.rs
```diff
@@ -193,9 +231,9 @@ impl TokenStream<PreToken> {
     fn push_inline_trivia(&mut self, token: &wdl_ast::Token) {
         assert!(!token.inner().kind().is_trivia());
         if let Some(token) = token.inner().inline_comment() {
-            let inline_comment = PreToken::Trivia(Trivia::Comment(Comment::Inline(Rc::new(
-                token.text().trim_end().to_owned(),
-            ))));
+            let normalized = normalize_except_directive(token.text().trim_end());
+            let inline_comment =
+                PreToken::Trivia(Trivia::Comment(Comment::Inline(Rc::new(normalized))));
```
a-frantz: this can be reverted, as lint directives aren't respected when they are an "inline comment". Lint directives are always expected to be "preceding" comments.

a-frantz (approved):

PR #510, opened 2025-12-24
refactor: cleanup of `wdl-engine`.

peterhuene:
This PR includes the following changes:

* Reduced public surface of the `wdl-engine` crate. The crate now exports a more sensible API for evaluating WDL code.
* Hide `Arc` inside of inner types for `Type`-related enumerations.
* `PairType` and `MapType` now hold a single `Arc` instead of two.
* Made `StructType` cheaply cloneable (it was already expected to be).
* Value constructors now take their specific types instead of `Type` and then panic if it is not the expected type.
* Moved engine types to their appropriate places (e.g. configuration into the `config` module and backend inputs into the `backends` module).
* Made `EvaluationPath` internals private so that an appropriate constructor is always used.
* Moved `HostPath` and `GuestPath` to the `values` module as they are inner types for the `File` and `Directory` values.
* Renamed `TopLevelEvaluator` to simply `Evaluator`.
* Hid value constructors that take the evaluation context from the public API.
* Reduces the number of allocations needed for handling URIs in evaluation paths.
* Removed use of `thiserror` from the `sprocket` crate as it complicated error handling and ended up as an `anyhow::Error` anyway.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/CHANGELOG.md
+++ b/crates/wdl-analysis/CHANGELOG.md
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Removed `Arc` from type-related enumerations ([#510](https://github.com/stjude-rust-labs/sprocket/pull/510)).
 
+#### Fixed
+
+* Fixed a bug where a dependency cycle was incorrectly detected due to an
+  invalid edge being added between nodes in the dependency graph when a scatter
+  variable shadowed an outer declaration of the same name (#[509](https://github.com/stjude-rust-labs/sprocket/pull/509)).
+
 ## 0.15.0 - 11-21-2025
 
 #### Added
--- a/crates/wdl-analysis/src/eval/v1.rs
+++ b/crates/wdl-analysis/src/eval/v1.rs
@@ -1032,7 +1032,7 @@ impl<N: TreeNode> WorkflowGraphBuilder<N> {
                         }
                     }
 
-                    // Add edges to other the requested calls
+                    // Add edges to the "after" calls
                     for after in statement.after() {
                         let name = after.name();
                         if let Some(nodes) =
@@ -1184,13 +1184,10 @@ impl<N: TreeNode> WorkflowGraphBuilder<N> {
         name: &str,
         expr: N,
     ) -> Option<SmallVec<[NodeIndex; SMALLVEC_DECLS_LEN]>> {
-        // If the name came from a declaration or call, return the node
-        if let Some(result) = self.names.get(name) {
-            return Some(result.to_owned());
-        }
-
-        // Otherwise, we need to walk up the parent chain looking for a scatter variable
-        // with the name
+        // We need to walk up the parent chain looking for a scatter variable with a
+        // matching name before looking at names in scope; a scatter variable may shadow
+        // names declared outside of it, but an inner declaration cannot shadow an outer
+        // scatter variable
         let mut current = expr;
         while let Some(parent) = current.parent() {
             if let SyntaxKind::ScatterStatementNode = parent.kind() {
@@ -1203,6 +1200,11 @@ impl<N: TreeNode> WorkflowGraphBuilder<N> {
             }
 
             current = parent;
+        }
+
+        // If the name came from a declaration or call, return the node
+        if let Some(result) = self.names.get(name) {
+            return Some(result.to_owned());
         }
 
         None
--- a/crates/wdl-analysis/tests/analysis/scatter-variable-no-cycle/source.wdl
+++ b/crates/wdl-analysis/tests/analysis/scatter-variable-no-cycle/source.wdl
@@ -1,0 +1,35 @@
+## This is a test to ensure scatter variable dependency edges are added correctly
+## It should not generate any diagnostics
+## See: https://github.com/stjude-rust-labs/sprocket/issues/508
+version 1.2
+
+task bar {
+    input {
+        Int foo
+    }
+
+    command <<<
+        echo ~{foo}
+    >>>
+
+    output {
+        Int out = foo
+    }
+}
+
+workflow scopes {
+    scatter (foo in [1, 2, 3, 4, 5]) {
+        # The reference to `foo` here should be to the scatter variable
+        # and not to the private declaration `foo` below
+        call bar {
+            foo
+        }
+    }
+
+    scatter (not_foo in bar.out) {
+        Int foo = not_foo
+        call bar as bar2 {
+            foo
+        }
+    }
+}


crates/wdl-engine/src/config.rs
```diff
@@ -1210,6 +1211,525 @@ impl TesBackendConfig {
...
+    pub extra_apptainer_exec_args: Option<Vec<String>>,
+    /// Deprecated field.
+    ///
+    /// This was kept for compatibility with previous versions of the Apptainer
+    /// configuration fields, and may be removed in a future version. The
+    /// Apptainer images are now stored at the top level root directory of an
+    /// evaluation.
+    #[serde(default)]
+    #[deprecated]
+    pub apptainer_images_dir: Option<String>,
```
a-frantz: this is completely ignored? I'd rather remove it and break existing configs than keep a config value that doesn't do anything; that seems confusing for users.

crates/wdl-engine/src/config.rs
```diff
@@ -1210,6 +1211,525 @@ impl TesBackendConfig {
...
+            }
+            Err(_) => Err(anyhow!(
+                "timed out trying to validate {name}_lsf_queue `{queue}`"
+            )),
+        }
+    }
+}
+
+/// Configuration for the LSF + Apptainer backend.
+// TODO ACF 2025-09-12: add queue option for short tasks
```
a-frantz: this is done, is it not?
```suggestion
```

crates/wdl-engine/src/config.rs
peterhuene: Agreed. We're pre-1.0, so the config file format isn't stable yet. Plus, this deprecation has already been in a release, so I think it's good to remove.

crates/wdl-analysis/src/types.rs
```diff
@@ -1076,7 +1070,7 @@ pub struct EnumType {
     /// The name of the enum.
     name: Arc<String>,
     /// The common coerced type computed from all variant values.
-    inner_value_type: Type,
+    inner_value_type: Arc<Type>,
```
claymcleod, line 1073: I thought in the last PR we determined that `Type` is intended to be cheaply clonable. Are we moving away from that? I see it does lack the annotation in the docstring, so this is more just for my understanding.

crates/wdl-engine/src/backend.rs
```diff
@@ -148,37 +212,24 @@ impl fmt::Debug for TaskSpawnInfo {
...
-        attempt: u64,
-        attempt_dir: PathBuf,
-        task_eval_root: PathBuf,
-        temp_dir: PathBuf,
-    ) -> Self {
+    pub fn new(id: String, info: TaskSpawnInfo, attempt_dir: PathBuf, temp_dir: PathBuf) -> Self {
         Self {
             id,
             info,
-            attempt,
```
claymcleod, line 179: Should these removals be noted in the `CHANGELOG.md`?

crates/wdl-engine/src/cache/hash.rs
```diff
@@ -188,16 +188,19 @@ impl Hashable for Digest {
...
-        match self {
-            Self::Local(path) => {
-                PathKind::Local.hash(hasher);
-                path.hash(hasher);
-            }
-            Self::Remote(url) => {
-                PathKind::Remote.hash(hasher);
-                url.hash(hasher);
-            }
+        if let Some(path) = self.as_local() {
```
claymcleod: Personally, I like things like this to be `match`es unless there is a good reason to switch. I can't think of one here, but perhaps you have one. If, for no other reason, (a) it's implied that the `match` is exhaustive and [avoiding the `unreachable!`] and (b) it forces you to update the code if you add a new variant. That being said, I don't have a strong preference necessarily.

crates/wdl-engine/src/config.rs
```diff
@@ -1210,6 +1211,507 @@ impl TesBackendConfig {
     }
 }
 
+/// Configuration for the Apptainer container runtime.
```
claymcleod, line 1214: Knowing what I know right now, I would have a slight preference to keep these in their respective modules close to the implementation. That being said, I can see a number of reasons why one might want to centralize them here, so I'll just ask: what was the reasoning for moving these?

crates/wdl-engine/src/eval/trie.rs
```diff
@@ -121,18 +121,17 @@ impl InputTrie {
...
+            // Check to see if the path being inserted is already a guest path
+            if let Some(dir) = self.guest_inputs_dir
+                && p.starts_with(dir)
+            {
+                return Ok(None);
             }
-            EvaluationPath::Remote(url) => Ok(Some(self.insert_url(kind, url))),
+
+            self.insert_path(kind, path.unwrap_local()).map(Some)
+        } else {
```
claymcleod: Again, I like the explicitness of `match` here.

crates/wdl-engine/src/stdlib/join_paths.rs
```diff
@@ -49,56 +48,55 @@ fn join_paths_simple(context: CallContext<'_>) -> Result<Value, Diagnostic> {
...
-
-            Ok(PrimitiveValue::new_file(
-                path.join(second)
-                    .clean()
-                    .into_os_string()
-                    .into_string()
-                    .expect("should be UTF-8"),
-            )
-            .into())
+    if first.is_local() {
```
claymcleod, line 51: Same for match.

crates/wdl-engine/src/stdlib/join_paths.rs
```diff
@@ -156,72 +154,69 @@ fn join_paths(context: CallContext<'_>) -> Result<Value, Diagnostic> {
...
-                if !p.is_relative() {
-                    return Err(function_call_failed(
-                        FUNCTION_NAME,
-                        format!("path `{next}` (array index {i}) is not a relative path"),
-                        array_span,
-                    ));
-                }
-
-                path.push(p);
+    if first.is_local() {
```
claymcleod, line 157: Same for match.

claymcleod (changes requested):
Looking good!

crates/wdl-engine/src/backend.rs
claymcleod: I left this as a comment to myself, and I didn't end up seeing in the notes why these were removed. So I'll leave it here.

crates/wdl-analysis/src/types.rs
peterhuene: With the outer `Arc` removed in the `CompoundType` variant, this is needed to prevent type recursion.

crates/wdl-engine/src/backend.rs
peterhuene: It was effectively dead code that was hidden by the fact that too much of the crate was public.

crates/wdl-engine/src/config.rs
peterhuene: The `config` module is public whereas their previous modules are no longer public; it also consolidates all the types exposed via `Config`.

crates/wdl-engine/src/cache/hash.rs
peterhuene: I'll change it to a match and make the internal enum representation of `EvaluationPath` have crate-public visibility.

claymcleod (approved):

PR #511, opened 2025-12-29
perf: cache enum variant values to avoid redundant AST lookups

### LINKED ISSUE #506: Cache variant value calculation.
Currently, every time an enum variant is accessed (e.g., `Status.Active`), the engine re-evaluates the variant's value expression or recreates the default string value by calling `resolve_enum_variant_value()`. This is wasteful for workflows with heavy enum usage, especially in loops or repeated expressions. We should cache evaluated enum variant values at the `TopLevelEvaluator` level so each variant is evaluated exactly once per workflow/task execution.

See [this comment](https://github.com/stjude-rust-labs/sprocket/pull/445#discussion_r2633530957) for more information.

shashidhar-bm:
Implements caching for enum variant values at the Evaluator level to avoid redundant AST lookups and parsing, 

 - Closes #506

---

**For external contributors:**

**For all contributors:**

- [x] You have either (a) made a PR to the [next](cci:1://file:///home/shashi/projects/sprocket/crates/wdl-engine/src/eval/v1/workflow.rs:1421:8-1452:9) branch in the sprocket.bio repository [if you are a Sprocket team member] or (b) have suggested any changes you _think_ need to be made to sprocket.bio in the description of your PR [if you are an external contributor].

**For PRs containing lint rule changes:**

- [x] You have added a test case in `crates/wdl-lint/tests/lints` that covers every possible diagnostic emitted for the rule within the file where the rule is implemented.

### DIFF
--- a/crates/wdl-engine/CHANGELOG.md
+++ b/crates/wdl-engine/CHANGELOG.md
@@ -35,9 +35,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added shared validation for task runtime and hint numeric settings (memory,
   max_memory, max_retries, preemptible), rejecting invalid or negative values
   earlier ([#485](https://github.com/stjude-rust-labs/sprocket/pull/485)).
-* Cached enum variant values at the Evaluator level to avoid redundant AST
-  lookups and parsing, improving performance for workflows with heavy enum
-  usage ([#511](https://github.com/stjude-rust-labs/sprocket/pull/511)).
 * Changed `join_paths` to take a `Directory` instead of `File` as the first
   argument and return `String` instead of `File` per WDL v1.2.1
   ([#519](https://github.com/stjude-rust-labs/sprocket/issues/519),
@@ -45,6 +42,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
 
+* Cached enum variant values at the Evaluator level to avoid redundant AST
+  lookups and parsing, improving performance for workflows with heavy enum
+  usage ([#511](https://github.com/stjude-rust-labs/sprocket/pull/511)).
 * Fixed enum variant serialization in command placeholders
   ([#534](https://github.com/stjude-rust-labs/sprocket/pull/534)).
 * Fixed `write_json` not translating host paths to guest paths ([#530](https://github.com/stjude-rust-labs/sprocket/pull/530)).


crates/wdl-engine/src/eval/v1.rs
```diff
@@ -65,6 +67,8 @@ pub struct Evaluator {
     cache: Option<CallCache>,
     /// The events for evaluation.
     events: Option<broadcast::Sender<EngineEvent>>,
+    /// Cache for evaluated enum variant values to avoid redundant AST lookups.
+    variant_cache: Arc<RwLock<HashMap<(String, String), crate::Value>>>,
```
peterhuene: I'd prefer not to have a `(String, String)` as a cache key if we can avoid it.

I propose we add an `EnumVariantCacheKey` type to `wdl_analysis` and a new method to `wdl_analysis::Document` with signature `fn get_variant_cache_key(&self, name: &str, variant: &str) -> Option<EnumVariantCacheKey>` (returns `None` if an enum of the given name doesn't exist or if the enum doesn't contain the given variant name).

Internally this method would first use `get_index_of` on the enums map and then with its associated type (which must be an `EnumType`) we find the position of a matching variant name.

We then construct a `EnumVariantCacheKey` from these two `usize` values.

The code in `wdl_engine` then first asks the document for the cache key and checks the cache with it (I think a simple `Mutex` to guard it instead of a `RwLock` is sufficient too; a map lookup is a fast operation). If the value isn't in the cache, we then go on to evaluate the variant's expression and store it in the cache with the same key.

shashidhar-bm: Updated with the `EnumVariantCacheKey` approach as suggested. Thanks for the feedback!

crates/wdl-analysis/src/document.rs
```diff
@@ -894,6 +894,23 @@ impl Document {
...
     }
 
+    /// Gets a cache key for an enum variant lookup.
+    pub fn get_variant_cache_key(
+        &self,
+        name: &str,
+        variant: &str,
+    ) -> Option<crate::types::EnumVariantCacheKey> {
+        let enum_index = self.data.enums.get_index_of(name)?;
+        let r#enum = &self.data.enums[enum_index];
```
peterhuene: These two lines can be replaced with:

```suggestion
        let (enum_index, _, r#enum) = self.data.enums.get_full(name)?;
```

crates/wdl-analysis/src/types.rs
```diff
@@ -1064,6 +1064,23 @@ impl Coercible for StructType {
...
+/// Cache key for enum variant values (enum index, variant index).
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq)]
+pub struct EnumVariantCacheKey {
+    enum_index: usize,
+    variant_index: usize,
+}
+
+impl EnumVariantCacheKey {
+    /// Constructs a new enum variant cache key.
+    pub fn new(enum_index: usize, variant_index: usize) -> Self {
```
peterhuene: Construction of the key should only be performed via `get_variant_cache_key`:

```suggestion
    pub(crate) fn new(enum_index: usize, variant_index: usize) -> Self {
```

crates/wdl-analysis/src/types.rs
```diff
@@ -1064,6 +1064,23 @@ impl Coercible for StructType {
     }
 }
 
+/// Cache key for enum variant values (enum index, variant index).
```
peterhuene: As the type is intended to be opaque, we don't need to describe its internals:

```suggestion
/// Cache key for enum variant values.
```

crates/wdl-engine/Cargo.toml
```diff
@@ -29,6 +29,7 @@ libc = { workspace = true }
 nonempty = { workspace = true }
 num_enum = { workspace = true }
 ordered-float = { workspace = true }
+parking_lot = { workspace = true }
```
peterhuene: The stdlib `Mutex` type is now very performant, so I don't think we need to introduce a dependency on `parking_lot` for it.

The critical section guarded by this mutex is also very small and the `stdlib`'s futex implementation is quite fast now.

crates/wdl-engine/src/eval/v1.rs
```diff
@@ -65,6 +67,8 @@ pub struct Evaluator {
     cache: Option<CallCache>,
     /// The events for evaluation.
     events: Option<broadcast::Sender<EngineEvent>>,
+    /// Cache for evaluated enum variant values to avoid redundant AST lookups.
+    variant_cache: Arc<Mutex<HashMap<wdl_analysis::types::EnumVariantCacheKey, crate::Value>>>,
```
peterhuene: I generally prefer to not qualify a name unless there's a conflict or stylistic reason for doing so.

Could we instead add a `use` for both the `EnumVariantCacheKey` and `Value` types?

crates/wdl-engine/src/eval/v1.rs
```diff
@@ -113,6 +117,7 @@ impl Evaluator {
             transferer,
             cache,
             events: events.engine().clone(),
+            variant_cache: Arc::new(Mutex::new(HashMap::new())),
```
peterhuene: Nit:

```suggestion
            variant_cache: Default::default(),
```

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -670,12 +670,26 @@ impl EvaluationContext for TaskEvaluationContext<'_, '_> {
...
+        let cache_key = self
+            .state
+            .document
+            .get_variant_cache_key(enum_name, variant_name)
+            .ok_or_else(|| unknown_enum(enum_name))?;
+
+        let mut cache = self.state.evaluator.variant_cache.lock();
+        if let Some(cached_value) = cache.get(&cache_key) {
+            return Ok(cached_value.clone());
+        }
```
peterhuene: Could we explicitly drop the mutex as to minimize the length of the critical section?

```suggestion
        let mut cache = self.state.evaluator.variant_cache.lock();
        if let Some(cached_value) = cache.get(&cache_key) {
            return Ok(cached_value.clone());
        }
        
        drop(cache);
```

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -670,12 +670,26 @@ impl EvaluationContext for TaskEvaluationContext<'_, '_> {
...
+
         let r#enum = self
             .state
             .document
             .enum_by_name(enum_name)
             .ok_or(unknown_enum(enum_name))?;
-        resolve_enum_variant_value(r#enum, variant_name)
+        let value = resolve_enum_variant_value(r#enum, variant_name)?;
+
+        cache.insert(cache_key, value.clone());
```
peterhuene, line 790: Likewise, we need to reacquire the lock here to insert the value.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -171,12 +171,26 @@ impl EvaluationContext for WorkflowEvaluationContext<'_, '_> {
     }
 
     fn enum_variant_value(&self, enum_name: &str, variant_name: &str) -> Result<Value, Diagnostic> {
+        let cache_key = self
+            .state
+            .document
+            .get_variant_cache_key(enum_name, variant_name)
+            .ok_or_else(|| unknown_enum(enum_name))?;
+
+        let mut cache = self.state.evaluator.variant_cache.lock();
```
peterhuene: Same comments as above re: scoping the cache lock to just the `get` and `insert` operations.

peterhuene (commented):
This is looking great, just a few comments to go through. Thanks!

crates/wdl-engine/src/eval/v1/task.rs
shashidhar-bm: > Could we explicitly drop the mutex as to minimize the length of the critical section?

Great idea—thanks!

peterhuene (approved):
Looks great, thanks for this contribution!

crates/wdl-engine/CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added shared validation for task runtime and hint numeric settings (memory,
   max_memory, max_retries, preemptible), rejecting invalid or negative values
   earlier ([#485](https://github.com/stjude-rust-labs/sprocket/pull/485)).
+* Cached enum variant values at the Evaluator level to avoid redundant AST
+  lookups and parsing, improving performance for workflows with heavy enum
+  usage ([#511](https://github.com/stjude-rust-labs/sprocket/pull/511)).
```
a-frantz: nit: I don't think this should be under the `Changed` header; I'd put it under `Fixed` instead, but I won't hold up a release over it. Logging my preference, but still approving

shashidhar-bm: i've updated it.

PR #524, opened 2026-01-07
revise: updates `gauntlet`

claymcleod:
This PR updates `gauntlet`, which has been disrupted by some changes over at the WILDS repository.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/gauntlet/Gauntlet.toml
+++ b/crates/gauntlet/Gauntlet.toml
@@ -20,11 +20,11 @@ commit_hash = "9213fa5ff0a47fa29dbad35cb9db11f06aab6a70"
 
 [repositories."broadinstitute/palantir-workflows"]
 identifier = "broadinstitute/palantir-workflows"
-commit_hash = "55d00ddc90d36d888c29abc96beb580677239b07"
+commit_hash = "0d6cb8e62f889d0c2b79fb65c28b1d2727ae7992"
 
 [repositories."broadinstitute/warp"]
 identifier = "broadinstitute/warp"
-commit_hash = "960846735b727047484c8ac33a597f679621172c"
+commit_hash = "54032fd9003183eb90029033fd2f88d98ac9a001"
 
 [repositories."chanzuckerberg/czid-workflows"]
 identifier = "chanzuckerberg/czid-workflows"
@@ -32,7 +32,7 @@ commit_hash = "30f0226d2c404b5c39f9a11704196da955c804d5"
 
 [repositories."getwilds/wilds-wdl-library"]
 identifier = "getwilds/wilds-wdl-library"
-commit_hash = "0aa8316110137c700125f78d0e47aaa1cfd4c611"
+commit_hash = "9252e622a2301312601b23324fb4d4a6c5afc753"
 
 [repositories."stjudecloud/workflows"]
 identifier = "stjudecloud/workflows"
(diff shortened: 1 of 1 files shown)


crates/gauntlet/Gauntlet.toml
```diff
@@ -6260,3499 +6310,3779 @@ permalink = "https://github.com/stjudecloud/workflows/blob/60c4e506136dbebc687c3
...
 [[diagnostics]]
 document = "theiagen/public_health_bioinformatics:/workflows/standalone_modules/wf_snippy_variants.wdl"
 message = "wf_snippy_variants.wdl:39:16: error: type mismatch: expected type `Int`, but found type `Int?`"
-permalink = "https://github.com/theiagen/public_health_bioinformatics/blob/567a9d6b553dbd0c00346a22febc25805d37d392/workflows/standalone_modules/wf_snippy_variants.wdl/#L39"
+permalink = "https://github.com/theiagen/public_health_bioinformatics/blob/879b27e87a794dd9d4af712540fba5d1d1ae4af0/workflows/standalone_modules/wf_snippy_variants.wdl/#L39"
+
+[[diagnostics]]
+document = "theiagen/public_health_bioinformatics:/workflows/standalone_modules/wf_snippy_variants.wdl"
+message = "wf_snippy_variants.wdl:40:18: error: type mismatch: expected type `Int`, but found type `Int?`"
+permalink = "https://github.com/theiagen/public_health_bioinformatics/blob/879b27e87a794dd9d4af712540fba5d1d1ae4af0/workflows/standalone_modules/wf_snippy_variants.wdl/#L40"
```
claymcleod, line 6813: In this case, we have an example where the original task has `map_qual` as an `Int` input with a default and the input being passed in is a `Int?`. Consulting the new table in WDL 1.2.1 ([link](https://github.com/openwdl/wdl/blob/wdl-1.2.1/SPEC.md?plain=1#L4046-L4050)), this should actually be allowed and (a) inherit the default if the outer `map_qual` is `None` and (b) take the outer `map_qual` if it's not `None`.

For now, this is fine, but we do need to track this in an issue to update when WDL 1.2.1 comes out.

crates/gauntlet/Gauntlet.toml
```diff
@@ -6260,3499 +6310,3779 @@ permalink = "https://github.com/stjudecloud/workflows/blob/60c4e506136dbebc687c3
...
 [[diagnostics]]
 document = "theiagen/public_health_bioinformatics:/workflows/utilities/wf_digger_denovo.wdl"
 message = "wf_digger_denovo.wdl:74:18: error: type mismatch: expected type `String`, but found type `String?`"
-permalink = "https://github.com/theiagen/public_health_bioinformatics/blob/567a9d6b553dbd0c00346a22febc25805d37d392/workflows/utilities/wf_digger_denovo.wdl/#L74"
+permalink = "https://github.com/theiagen/public_health_bioinformatics/blob/879b27e87a794dd9d4af712540fba5d1d1ae4af0/workflows/utilities/wf_digger_denovo.wdl/#L74"
+
+[[diagnostics]]
+document = "theiagen/public_health_bioinformatics:/workflows/utilities/wf_digger_denovo.wdl"
+message = "wf_digger_denovo.wdl:83:17: error: type mismatch: expected type `String`, but found type `String?`"
+permalink = "https://github.com/theiagen/public_health_bioinformatics/blob/879b27e87a794dd9d4af712540fba5d1d1ae4af0/workflows/utilities/wf_digger_denovo.wdl/#L83"
```
claymcleod, line 7933: Same here.

crates/gauntlet/Gauntlet.toml
peterhuene: Agreed that this looks like a bug.

peterhuene: Actually, these are 1.0 documents and we limit the behavior of being able to specify a `None` to invoke the default to 1.2 documents as it was the first version of the spec that introduced it (see: https://github.com/openwdl/wdl/pull/634).

Changing it to a 1.2 document gets the desired behavior. If we're back porting this change to 1.0/1.1 patch releases of the spec, then we should probably just drop the version restriction.

claymcleod: That's a good point. I'm not sure what we should do, but we can make an issue and figure it out later.

PR #533, opened 2026-01-10
wdl-format: split long #@ except directives into multiple lines

kris70lesgo:

This PR adds formatter support for splitting long single-line `#@ except:` lint directives into multiple lines when they exceed the configured maximum line length.
The splitting logic preserves the normalized (alphabetically sorted) rule order while redistributing rules across lines based on length constraints. Splits occur only at comma boundaries, ensure at least one rule per line, and leave already-multiline directives unchanged. This follows up on the earlier normalization work and completes the formatter behavior discussed in issue #247.
Formatter tests have been added to cover under-length, single-split, multi-split, edge-length, and single-rule cases.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-format/src/token/post.rs
+++ b/crates/wdl-format/src/token/post.rs
@@ -181,24 +181,26 @@ fn can_be_line_broken(kind: SyntaxKind) -> Option<LineBreak> {
     }
 }
 
-/// Split a normalized `#@ except:` directive into multiple lines when needed.
+/// Attempts to split a `#@ except:` directive into multiple lines when it
+/// exceeds `max_len`.
 ///
-/// Splits at comma boundaries when the line exceeds `max_len`. Each line
-/// gets the `#@ except:` prefix and contains at least one rule.
-fn split_except_directive_lines(normalized: &str, max_len: usize) -> Vec<String> {
+/// Returns `None` if the input is not a `#@ except:` directive or if no
+/// splitting is required. Splitting occurs only at comma boundaries and
+/// ensures at least one rule per line.
+fn split_except_directive_lines(value: &str, max_len: usize) -> Option<Vec<String>> {
     // Check if this is an except directive
-    let Some(remainder) = normalized.trim_start().strip_prefix("#@") else {
-        return vec![normalized.to_owned()];
+    let Some(remainder) = value.strip_prefix("#@") else {
+        return None;
     };
 
     let Some(rules_text) = remainder.trim_start().strip_prefix("except:") else {
-        return vec![normalized.to_owned()];
+        return None;
     };
 
-    // If the whole line fits, return as-is
-    if normalized.len() <= max_len {
-        return vec![normalized.to_owned()];
-    }
+    // If the whole line fits, no splitting needed
+    if value.len() <= max_len {
+        return None;
+    };
 
     // Split into individual rules
     let rules: Vec<&str> = rules_text
@@ -208,7 +210,7 @@ fn split_except_directive_lines(normalized: &str, max_len: usize) -> Vec<String>
         .collect();
 
     if rules.is_empty() {
-        return vec![normalized.to_owned()];
+        return None;
     }
 
     let prefix = "#@ except: ";
@@ -244,7 +246,7 @@ fn split_except_directive_lines(normalized: &str, max_len: usize) -> Vec<String>
         lines.push(format!("{}{}", prefix, current_rules.join(", ")));
     }
 
-    lines
+    Some(lines)
 }
 
 /// Current position in a line.
@@ -458,22 +460,21 @@ impl Postprocessor {
         let mut expanded_stream = TokenStream::<PreToken>::default();
         let in_stream = if let Some(max_len) = config.max_line_length() {
             for token in in_stream.iter() {
-                if let PreToken::Trivia(Trivia::Comment(Comment::Preceding(value))) = token {
-                    if value.trim_start().starts_with("#@") {
-                        let lines = split_except_directive_lines(value, max_len);
-                        if lines.len() > 1 {
-                            // Split occurred, emit multiple comment tokens
-                            for line in lines {
-                                expanded_stream.push(PreToken::Trivia(Trivia::Comment(
-                                    Comment::Preceding(Rc::new(line)),
-                                )));
-                            }
-                        } else {
-                            // No split needed, pass through unchanged
-                            expanded_stream.push(token.clone());
-                        }
-                    } else {
-                        expanded_stream.push(token.clone());
+                let PreToken::Trivia(Trivia::Comment(Comment::Preceding(value))) = token else {
+                    expanded_stream.push(token.clone());
+                    continue;
+                };
+
+                if !value.starts_with("#@") {
+                    expanded_stream.push(token.clone());
+                    continue;
+                }
+
+                if let Some(lines) = split_except_directive_lines(value, max_len) {
+                    for line in lines {
+                        expanded_stream.push(PreToken::Trivia(Trivia::Comment(
+                            Comment::Preceding(Rc::new(line)),
+                        )));
                     }
                 } else {
                     expanded_stream.push(token.clone());
--- a/crates/wdl-format/src/token/post.rs
+++ b/crates/wdl-format/src/token/post.rs
@@ -456,30 +456,34 @@ impl Postprocessor {
         assert!(self.position == LinePosition::StartOfLine);
         // Preprocess the input stream to split long except directives if needed
         let mut expanded_stream = TokenStream::<PreToken>::default();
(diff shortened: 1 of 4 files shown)


crates/wdl-format/tests/format/except-directive-line-splitting/source.formatted.wdl
```diff
@@ -0,0 +1,35 @@
+version 1.2
+
+# Test 1: Directive under length - should not split
+#@ except: ShortRule
+task test1 {
+}
+
+# Test 2: Directive slightly over length - should split into 2 lines
+#@ except: CommentWhitespace, DeprecatedObject, InputSorted, MetaDescription
```
claymcleod, line 9: Just want to point out that your PR states that this "preserves rule order" but, here, the rule order changes. I think that's the right move, but you may want to update the PR description.

crates/wdl-format/src/token/post.rs
```diff
@@ -388,6 +454,32 @@ impl Postprocessor {
...
+                        // No split needed, pass through unchanged
+                        expanded_stream.push(token.clone());
+                    }
+                } else {
+                    expanded_stream.push(token.clone());
+                }
+            }
+        } else {
+            // No max line length configured, use original stream
+            expanded_stream = in_stream.clone();
```
claymcleod: Is this `clone()` necessary?

crates/wdl-format/src/token/post.rs
```diff
@@ -388,6 +454,32 @@ impl Postprocessor {
     ) {
         assert!(!self.interrupted);
         assert!(self.position == LinePosition::StartOfLine);
+        // Preprocess the input stream to split long except directives if needed
+        let mut expanded_stream = TokenStream::<PreToken>::default();
+        if let Some(max_len) = config.max_line_length() {
+            for token in in_stream.iter() {
+                if let PreToken::Trivia(Trivia::Comment(Comment::Preceding(value))) = token {
+                    let lines = split_except_directive_lines(value, max_len);
+                    if lines.len() > 1 {
```
claymcleod: This calls `split_except_directive_lines` for every `Comment::Preceding` token, but most comments are not `#@ except:` directives. For non-except comments, the function allocates a `Vec` and clones the `String` unnecessarily.

Consider checking for `#@` before calling:

```rust
if value.trim_start().starts_with("#@") {
    let lines = split_except_directive_lines(value, max_len);
    // ...
} else {
    expanded_stream.push(token.clone());
}
```

claymcleod (dismissed):
Please also add a CHANGELOG entry to `crates/wdl-format/CHANGELOG.md`.

crates/wdl-format/src/token/post.rs
kris70lesgo: the clone isn’t necessary here  I’ve updated the code to avoid cloning the stream when no max line length is configured and instead reuse the original input directly

crates/wdl-format/tests/format/except-directive-line-splitting/source.formatted.wdl
kris70lesgo: apologies for my mistake , I have updated the pr description

crates/wdl-format/src/token/post.rs
```diff
@@ -181,6 +181,72 @@ fn can_be_line_broken(kind: SyntaxKind) -> Option<LineBreak> {
     }
 }
 
+/// Split a normalized `#@ except:` directive into multiple lines when needed.
+///
+/// Splits at comma boundaries when the line exceeds `max_len`. Each line
+/// gets the `#@ except:` prefix and contains at least one rule.
+fn split_except_directive_lines(normalized: &str, max_len: usize) -> Vec<String> {
+    // Check if this is an except directive
+    let Some(remainder) = normalized.trim_start().strip_prefix("#@") else {
```
a-frantz:
```suggestion
    let Some(remainder) = normalized.strip_prefix("#@") else {
```
a minor detail, but `Comment` tokens never contain leading whitespace so the `trim_start()` will never be needed

crates/wdl-format/src/token/post.rs
```diff
@@ -388,6 +454,36 @@ impl Postprocessor {
     ) {
         assert!(!self.interrupted);
         assert!(self.position == LinePosition::StartOfLine);
+        // Preprocess the input stream to split long except directives if needed
+        let mut expanded_stream = TokenStream::<PreToken>::default();
+        let in_stream = if let Some(max_len) = config.max_line_length() {
+            for token in in_stream.iter() {
+                if let PreToken::Trivia(Trivia::Comment(Comment::Preceding(value))) = token {
+                    if value.trim_start().starts_with("#@") {
```
a-frantz:
```suggestion
                    if value.starts_with("#@") {
```
same here on the trimming

crates/wdl-format/src/token/post.rs
```diff
@@ -388,6 +454,36 @@ impl Postprocessor {
...
+                        } else {
+                            // No split needed, pass through unchanged
+                            expanded_stream.push(token.clone());
+                        }
+                    } else {
+                        expanded_stream.push(token.clone());
+                    }
+                } else {
+                    expanded_stream.push(token.clone());
+                }
```
a-frantz: could you refactor this part so that it's less repetitive and isn't as nested? Looks like the logic is sound, but this has some code smell with the way it's written. Thanks

crates/wdl-format/src/token/post.rs
```diff
@@ -181,6 +181,72 @@ fn can_be_line_broken(kind: SyntaxKind) -> Option<LineBreak> {
...
+    };
+
+    let Some(rules_text) = remainder.trim_start().strip_prefix("except:") else {
+        return vec![normalized.to_owned()];
+    };
+
+    // If the whole line fits, return as-is
+    if normalized.len() <= max_len {
+        return vec![normalized.to_owned()];
+    }
```
a-frantz:
```suggestion
fn split_except_directive_lines(unnormalized: &str, max_len: usize) -> Option<Vec<String>> {
    // Check if this is an except directive
    let Some(remainder) = normalized.strip_prefix("#@") else {
        return None
    };

    let Some(rules_text) = remainder.trim_start().strip_prefix("except:") else {
        return None
    };

    // If the whole line fits, return as-is
    if normalized.len() <= max_len {
        return None
    }
```
first, the input `&str` isn't normalized yet, so the name is getting ahead of itself! 😅 

Second, I think we should change the signature of this function and have it return `None` for all the cases where the input value shouldn't be normalized. This will avoid needless `String` allocations and be more performant. I'm not particularly concerned with a little bit of extra allocation overhead, but just best practice to avoid inefficiencies like this as they can really add up over a large codebase like ours!

a-frantz (changes requested):
Great work, this looks awesome 🚀 I think we can make this more performant with fewer allocations with a little refactoring of the `split_` fn, but otherwise LGTM!

a-frantz (approved):
awesome work! 🚂 we need one other reviewer to give this a thumbs up then we can merge!

peterhuene (approved):

crates/wdl-format/src/token/post.rs
```diff
@@ -181,6 +181,74 @@ fn can_be_line_broken(kind: SyntaxKind) -> Option<LineBreak> {
...
+/// ensures at least one rule per line.
+fn split_except_directive_lines(value: &str, max_len: usize) -> Option<Vec<String>> {
+    // Check if this is an except directive
+    let Some(remainder) = value.strip_prefix("#@") else {
+        return None;
+    };
+
+    let Some(rules_text) = remainder.trim_start().strip_prefix("except:") else {
+        return None;
+    };
```
a-frantz:
```suggestion
    let remainder = value.strip_prefix("#@")?;
    let rules_text = remainder.trim_start().strip_prefix("except:")?
```

crates/wdl-format/src/token/post.rs
```diff
@@ -181,6 +181,68 @@ fn can_be_line_broken(kind: SyntaxKind) -> Option<LineBreak> {
...
 
+/// Attempts to split a `#@ except:` directive into multiple lines when it
+/// exceeds `max_len`.
+///
+/// Returns `None` if the input is not a `#@ except:` directive or if no
+/// splitting is required. Splitting occurs only at comma boundaries and
+/// ensures at least one rule per line.
+fn split_except_directive_lines(value: &str, max_len: usize) -> Option<Vec<String>> {
+    let remainder = value.strip_prefix("#@")?;
+    let rules_text = remainder.trim_start().strip_prefix("except:")?
```
a-frantz:
```suggestion
    let rules_text = remainder.trim_start().strip_prefix("except:")?;
```

a-frantz (approved):

PR #470, opened 2025-11-20
feat: actor-based execution system

claymcleod:
This PR lays the foundation for the actor-based execution system. Notably, it does not (yet) replace the way `run` works, only focusing on getting a working version of the server command. In a future PR, I plan to clean up the way we do configuration and have `run` use the actor.

Before submitting this PR, please make sure:

For external contributors:

- [ ] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [ ] You have not used AI on any parts of this pull request.

For all contributors:

- [ ] You have added a few sentences describing the PR here.
- [ ] You have added yourself or the appropriate individual as the assignee.
- [ ] You have added at least one relevant code reviewer to the PR.
- [ ] Your code builds clean without any errors or warnings.
- [ ] You have added tests (when appropriate).
- [ ] You have added an entry in the CHANGELOG (when appropriate).
- [ ] You have updated the README or other documentation to account for these changes (when appropriate).
- [ ] You have either (a) made a PR to the `next` branch in the sprocket.bio repository [if you are a Sprocket team member] or (b) have suggested any changes you _think_ need to be made to sprocket.bio in the description of your PR [if you are an external contributor].

For PRs containing lint rule changes:

- [ ] You have updated any and all effected entries within `RULES.md`.
- [ ] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

Issue #160, opened 2025-08-13
feature request: output directory with symlinks


a-frantz:
currently, output files of a run are buried deep in nested directories and you need to dig into the work directories to do anything with them.

miniwdl solves this issue with symlinks in an `<run dir>/out/` directory. Sprocket should have something similar.

nikhil-nari:
I'm still getting familiar with the codebase, but I noticed that commit 62592b7 references this issue. Has this change fully resolved the issue with output files, or is there some remaining work needed? Thank you

nikhil-nari:
Hi @a-frantz,

I’m preparing to work on the feature for organizing workflow run outputs by creating an out/ folder with symlinks inside each run directory. Since I am new to the project, I wanted to confirm my understanding before starting,

I found a function in run.rs that creates the working directory for each run.

For each workflow run, there should be an out/ folder created inside that run’s directory, with symlinks pointing to the output files for that run.

Could you please confirm if this is correct, or let me know if I am missing anything important?

Thank you!

a-frantz:
that sounds right to me! Note that it is a bit more complicated than just creating symlinks within an `out/` directory, as WDL does not require that created files have unique names (so it is possible for name collisions to happen if this isn't implemented right). I think that https://github.com/chanzuckerberg/miniwdl handles this pretty well, and I'd recommend mimicking their logic for creating the `out/` directory.

> I found a function in run.rs that creates the working directory for each run.

Correct, the `out/` directory should be placed within that directory. This logic should also remain within the `sprocket` binary (I don't think any changes in `crates/wdl-engine` would be wise). It is also fine for this to be `unix` only feature, as creating symlinks in windows is a more complicated problem. You can find in `run.rs` where we configure it to only create a `_latest/` symlink on unix platforms.

a-frantz:
o, and another clarification just to be explicit: the `out/` directory should only contain WDL outputs of type `File` or `Directory`, and the other outputs of a workflow or task can be ignored (it doesn't make sense to represent them in the file system)

nikhil-nari:
Thank you so much for the clarification - it really helps a lot!
I am glad to have the opportunity to work on this and will get started soon

a-frantz:
@claymcleod are you intending to implement this as part of #540 ? I just skimmed the `system::v1::fs` module in that PR and didn't see any relevant logic.

[...]

PR #540, opened 2026-01-13
feat: provenance tracking and server mode

claymcleod:
This PR combines the work from #452, #453, and #470 to implement the foundation for the [provenance and server mode RFC](https://github.com/stjude-rust-labs/rfcs/pull/3).

Notably, this includes:

- A SQLite-backed database layer for tracking sessions, runs, and tasks
- An index system for symlinking run outputs for easy access
- An actor-based execution system powering the new `server` command

The `run` command remains unchanged for now—a future PR will integrate it with the actor system.

Closes #452, closes #453, closes #470.

For external contributors:

For all contributors:

      is implemented.


migrations/20251108201315_initial.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    created_at timestamp not null default current_timestamp
+);
+
+-- The sessions table.
+--
+-- Sessions are invocations of the Sprocket command line tool by a particular
+-- user. They are tracked for provenance purposes.
+create table if not exists "sessions" (
+    -- Unique identifier for this session
+    id text primary key not null,
```
peterhuene: Does the `id` column have to be text? Could we instead use `integer primary key`? I usually like to avoid using text for foreign keys (plus we can rely on the SQL engine to populate a value for us).

Otherwise, if we'd not joining these tables and they're otherwise denormalized, then there's no reason to not have a text primary key.

migrations/20251108201315_initial.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    -- Timestamp when the session was created
+    created_at timestamp not null default current_timestamp
+);
+
+-- The runs table.
+--
+-- A "run" represented a targeted WDL task or workflow to execute.
+create table if not exists runs (
+    -- Unique identifier for this run
+    id text primary key not null,
```
peterhuene: Same comment as above re: primary key.

migrations/20251108201315_initial.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    -- Foreign key to the session that submitted this run
+    session_id text not null,
+    -- Name of the run
+    "name" text not null,
+    -- Source WDL file path or URL
+    "source" text not null,
```
peterhuene, line 47: Should the entrypoint be tracked along with the source?

migrations/20251108201315_initial.up.sql
```diff
@@ -0,0 +1,152 @@
...
+create index idx_runs_created_at on runs(created_at);
+
+-- The index log table.
+--
+-- The index log track _all_ entries that have ever been linked into the index.
+-- Its primary purpose is providing a mechanism to reconstruct what the index
+-- looked like at any point in time.
+create table if not exists index_log (
+    -- Unique identifier for this index log entry
+    id integer primary key autoincrement not null,
```
peterhuene: Seems like [`autoincrement` is discouraged](https://www.sqlite.org/autoinc.html) in favor of just `integer primary key` as an alias for `rowid`.

migrations/20251108201315_initial.up.sql
```diff
@@ -0,0 +1,152 @@
...
+create index idx_tasks_run_id on tasks(run_id);
+create index idx_tasks_status on tasks("status");
+create index idx_tasks_created_at on tasks(created_at);
+
+-- The task logs table.
+--
+-- The table keeps track of all stdout and stderr logs from tasks.
+create table if not exists task_logs (
+    -- The unique ID for the task log entry
+    id integer primary key autoincrement not null,
```
peterhuene: Same comment as above re: `autoincrement`.

migrations/20251108201315_initial.up.sql
```diff
@@ -0,0 +1,152 @@
...
+) ranked
+where rn = 1
+order by link_path;
+
+-- The tasks table.
+--
+-- The tasks table tracks the status of tasks that execute underneath runs.
+create table if not exists tasks (
+    -- Task name from WDL
+    "name" text primary key not null,
```
peterhuene, line 118: Same comment about re: text primary keys.

migrations/20251108201315_initial.up.sql
peterhuene: If we do use a text primary key, we should probably turn off `rowid` for the table (via `without rowid`) as we won't use it.

src/commands/run.rs
```diff
@@ -30,6 +30,7 @@ use wdl::ast::AstNode as _;
 use wdl::ast::Severity;
 use wdl::engine::CancellationContext;
 use wdl::engine::CancellationContextState;
+use wdl::engine::Config as WdlEngineConfig;
```
peterhuene: Nit: I feel like the `Wdl` prefix here is redundant.

src/commands/run.rs
```diff
@@ -208,7 +218,7 @@ impl Args {
     }
 
     /// Applies the CLI arguments to the given engine configuration.
-    fn apply_engine_config(&self, config: &mut wdl::engine::config::Config) {
+    fn apply_engine_config(&self, config: &mut wdl::engine::Config) {
```
peterhuene: related nit:

```suggestion
    fn apply_engine_config(&self, config: &mut EngineConfig) {
```

src/commands/server.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub allowed_urls: Vec<String>,
+
+    /// Allowed CORS origins.
+    #[arg(long)]
+    pub allowed_origins: Vec<String>,
+}
+
+impl Args {
+    /// Applies the arguments to the configuration.
+    pub fn apply(mut self, mut config: Config) -> Config {
```
peterhuene: Personally I'd prefer to take `config` by mutable reference than just returning it. Should probably also not be `pub`.

I'm trying to move us away from having "apply" the options to the configuration be a public operation that the sprocket CLI has to call to prepare the configuration that's passed into it. This was already done to the `run` command, but wasn't updated for all the commands.

src/server/api/v1/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// Optional target workflow or task name to execute.
+    ///
+    /// If not provided, will attempt to automatically select:
+    ///
+    /// 1. The workflow in the document (if one exists)
+    /// 2. The single task in the document (if no workflow but exactly one task)
+    /// 3. Error if ambiguous (no workflow and multiple tasks)
+    #[serde(default)]
+    #[schema(example = "target")]
+    pub target: Option<String>,
```
peterhuene, line 41: Do we want this nomenclature to match the `sprocket run` CLI (i.e. "entrypoint")?

src/server/api/v1/sessions.rs
```diff
@@ -0,0 +1,143 @@
...
+            session: response.session.into(),
+        }
+    }
+}
+
+/// The response for a "list sessions" query.
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct ListSessionsResponse {
+    /// The sessions.
+    pub sessions: Vec<Session>,
```
peterhuene, line 79: I see that the `runs` and `tasks` "list" operations include a `total` field. Is that something we need to do here?

src/system/v1/db/sqlite.rs
```diff
@@ -0,0 +1,1219 @@
...
+        sqlx::query("insert into sessions (id, subcommand, created_by) values (?, ?, ?)")
+            .bind(id.to_string())
+            .bind(subcommand)
+            .bind(created_by)
+            .execute(&self.pool)
+            .await?;
+
+        let session: Session = sqlx::query_as(
+            "select id, subcommand, created_by, created_at from sessions where id = ?",
+        )
```
peterhuene, line 155: I believe `sqlite` supports a `RETURNING` clause on its inserts, so would it be possible to combine these "insert and then select out the inserted data" operations into a single one?

src/system/v1/db/sqlite.rs
```diff
@@ -0,0 +1,1219 @@
...
+        .execute(&self.pool)
+        .await?;
+
+        let run: Run = sqlx::query_as(
+            "select id, session_id, name, source, status, inputs, outputs, error, directory, \
+             index_directory, started_at, completed_at, created_at from runs where id = ?",
+        )
+        .bind(id.to_string())
+        .fetch_one(&self.pool)
+        .await?;
```
peterhuene: Same comment as above re: `returning` clause.

src/system/v1/db/sqlite.rs
```diff
@@ -0,0 +1,1219 @@
...
+                .await?;
+
+        let id = result.last_insert_rowid();
+
+        let entry: IndexLogEntry = sqlx::query_as(
+            "select id, run_id, link_path, target_path, created_at from index_log where id = ?",
+        )
+        .bind(id.to_string())
+        .fetch_one(&self.pool)
+        .await?;
```
peterhuene, line 415: Same comment as above re: `returning` clause.

src/system/v1/db/sqlite.rs
```diff
@@ -0,0 +1,1219 @@
...
+            "insert into tasks (name, run_id, status)
+             values (?, ?, ?)",
+        )
+        .bind(name)
+        .bind(run_id.to_string())
+        .bind(TaskStatus::Pending)
+        .execute(&self.pool)
+        .await?;
+
+        self.get_task(name).await
```
peterhuene: Same comment as above re: `returning` clause.

src/system/v1/exec/svc/run_manager/commands.rs
```diff
@@ -0,0 +1,218 @@
...
+pub struct SessionResponse {
+    /// The session data.
+    pub session: Session,
+}
+
+/// Response for session list query.
+#[derive(Debug)]
+pub struct ListSessionsResponse {
+    /// The sessions.
+    pub sessions: Vec<Session>,
```
peterhuene, line 72: Same comment as above re: including a `total` for this response.

src/system/v1/exec/svc/run_manager/commands.rs
```diff
@@ -0,0 +1,218 @@
...
+#[derive(Debug)]
+pub struct ListTaskLogsResponse {
+    /// The task logs.
+    pub logs: Vec<TaskLog>,
+    /// Total count before pagination.
+    pub total: i64,
+}
+
+/// Commands sent to the run manager.
+#[derive(Debug)]
```
peterhuene, line 103: So one problem with the model of having the API send requests to the run manager for purely data operations (like listing things in the database, for example) is that we don't really get to take advantage of the database's connection pool.

API requests will block waiting on the run manager queue to service them even though it could just as easily have directly queried the data itself and make use of the connection pool's parallelism.

The approach mainly taken in `planetary` is to have the API service directly read from and write to the database and only dispatch to the orchestrator service the "kick off the orchestration of task `foo`" message, as an example. Though, the only writes planetary's API service does is to insert a new task and update the task's state to "canceling" prior to notifying the orchestrator to cancel the task; the rest of the writing is done by the orchestrator in response to k8s events.

Given that most of the command messages here are data retrieval, I think we should look into not having to dispatch the operations to the run manager's queue in a follow-up PR, perhaps.

src/system/v1/exec/svc/run_manager.rs
```diff
@@ -0,0 +1,654 @@
...
+
+pub(crate) mod commands;
+
+pub use commands::*;
+
+/// Channel capacity for events.
+///
+/// This number represents a reasonable, arbitrary buffer size to handle burst
+/// event production.
+const EVENTS_CHANNEL_CAPACITY: usize = 2048;
```
peterhuene, line 43: We should probably measure how big a buffer this is allocating as there are multiple underlying broadcast channels for each `Events` object and that can add up per-run.

Off the cuff calculation: `(2048 * (24 + 16)) + (2048 * (64 + 16)) + (2048 * (208 + 16))`: `24` for size of the engine's events, `64` for the size of crankshaft events, and `208` for the size of transfer events (it holds `Url` which is a chonky boi due to all the internal offsets they keep into the string).

Probably looking at the better part of megabyte per run.

At the very least we could trim the allocations of the engine events and transfer events with a change to `wdl-engine` given how we're only subscribing to the crankshaft events.

src/system/v1/exec/svc/run_manager.rs
```diff
@@ -0,0 +1,654 @@
...
+    let async_semaphore = semaphore;
+    let async_db = db;
+    let async_ctx = ctx.clone();
+    let async_target = target.clone();
+    let async_document = result.document().clone();
+    let async_cancellation = cancellation.clone();
+    let async_runs = runs.clone();
+    let async_events = events.clone();
+
+    let handle = tokio::spawn(async move {
```
peterhuene: It feels like the spawn should encompass more of the operations above, such as analyzing the document and preparing for evaluation; otherwise, the throughput on requests to submit runs would block on more than it probably should.

I'd probably break this down into: insert the data in the database, update the `runs` data structure, and then spawn the evaluation (including everything above not related to inserting data into the database, such as analysis).

src/system/v1/exec/svc/run_manager.rs
```diff
@@ -0,0 +1,654 @@
...
+                                id
+                            }
+                            Err(e) => {
+                                let _ = rx.send(Err(SubmitRunError::Database(e)));
+                                continue;
+                            }
+                        }
+                    };
+
+                    let result = submit_run(
```
peterhuene: Feels like a good chunk of these arguments could we removed if we simply made this a method taking `&self`).

src/system/v1/db/sqlite.rs
```diff
@@ -0,0 +1,1219 @@
...
+             index_directory, started_at, completed_at, created_at from runs where id = ?",
+        )
+        .bind(id.to_string())
+        .fetch_optional(&self.pool)
+        .await?;
+
+        Ok(run)
+    }
+
+    async fn list_runs(
```
peterhuene, line 334: Are list operations for all sessions desirable?

Would users generally want to see their own runs vs. others?

src/system/v1/db/sqlite.rs
```diff
@@ -0,0 +1,1219 @@
...
+        } else {
+            sqlx::query_as("select count(*) from runs")
+                .fetch_one(&self.pool)
+                .await?
+        };
+
+        Ok(count.0)
+    }
+
+    async fn list_runs_by_session(&self, session_id: Uuid) -> Result<Vec<Run>> {
```
peterhuene, line 386: Saw this after my last question: this only appears to be utilized from tests.

src/server/api/v1/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+pub struct ListRunsQueryParams {
+    /// Filter by status.
+    #[serde(default)]
+    pub status: Option<RunStatus>,
+    /// Number of results to return (default: `100`).
+    #[serde(default)]
+    pub limit: Option<i64>,
+    /// Number of results to skip (default: `0`).
+    #[serde(default)]
+    pub offset: Option<i64>,
```
peterhuene: So this is probably not something we want directly exposed in the API; at the very least we should label this field as "next token" or something similar so that we don't bake in the fact we're currently using `offset` in the underlying queries (something that doesn't scale).

It should probably be an opaque string that the client hands back to us and, for now at least, it can just be a string-ified offset.

src/server/api/v1/sessions.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// Query parameters for listing sessions.
+#[derive(Debug, Clone, Serialize, Deserialize, IntoParams, ToSchema)]
+pub struct ListSessionsQueryParams {
+    /// Number of results to return (default: `100`).
+    #[serde(default)]
+    pub limit: Option<i64>,
+    /// Number of results to skip (default: `0`).
+    #[serde(default)]
+    pub offset: Option<i64>,
```
peterhuene: Same comment as above re: `offset`.

src/server/api/v1/tasks.rs
```diff
@@ -0,0 +1,265 @@
...
+    pub run_id: Option<Uuid>,
+    /// Filter by status.
+    #[serde(default)]
+    pub status: Option<TaskStatus>,
+    /// Number of results to return (default: `100`).
+    #[serde(default)]
+    pub limit: Option<i64>,
+    /// Number of results to skip (default: `0`).
+    #[serde(default)]
+    pub offset: Option<i64>,
```
peterhuene: Same comment as above re: `offset`.

src/system/v1/db/sqlite.rs
```diff
@@ -0,0 +1,1219 @@
...
+        Ok(session)
+    }
+
+    async fn list_sessions(&self, limit: Option<i64>, offset: Option<i64>) -> Result<Vec<Session>> {
+        let limit = limit.unwrap_or(DEFAULT_PAGE_SIZE);
+        let offset = offset.unwrap_or(DEFAULT_OFFSET);
+
+        let sessions: Vec<Session> = sqlx::query_as(
+            "select id, subcommand, created_by, created_at from sessions order by created_at desc \
+             limit ? offset ?",
```
peterhuene: Although I realize we're using `sqlite` here and not a _real™_ database, probably a good idea to file an issue to not use `offset` clauses for pagination in case we have to scale to decent size row counts in the future.

Something I meant to do with `planetary` but also ended up using `offset` for simplicity (it's more important to fix there, though).

src/system/v1/exec/config.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Default output directory function for serde.
+fn default_output_directory() -> PathBuf {
+    PathBuf::from(DEFAULT_OUTPUT_DIRECTORY)
+}
+
+/// Configuration validation errors.
+#[derive(Debug, thiserror::Error)]
+pub enum ConfigError {
+    /// File does not exist.
+    #[error("file does not exist: `{0}`")]
```
peterhuene: nit for consistency with other error messages:

```suggestion
    #[error("file `{0}` does not exist")]
```

src/system/v1/exec/config.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// File path not in allowed file paths.
+    #[error("file path `{0}` is not in an allowed directory")]
+    FilePathForbidden(PathBuf),
+
+    /// URL not in allowed URLs.
+    #[error("url `{0}` does not have an allowed prefix")]
+    UrlForbidden(Url),
+
+    /// Failed to canonicalize file path.
+    #[error("failed to canonicalize path: `{0}`")]
```
peterhuene:
```suggestion
    #[error("failed to canonicalize path `{0}`")]
```

src/system/v1/exec.rs
```diff
@@ -0,0 +1,472 @@
...
+    analyzer
+        .add_document(uri.clone())
+        .await
+        .context("failed to add document to analyzer")?;
+
+    let results = analyzer
+        .analyze(())
+        .await
+        .context("failed to analyze document")?;
+
```
peterhuene, line 394: We probably want to check for analysis errors in _all_ of the results here, not just the entrypoint document which we return here and then check for errors in `validate_analysis_results` below.

Otherwise, if we pass an otherwise correct entrypoint document to evaluation and one of its transitive dependencies has an analysis error, then the engine will immediately fail the evaluation with an opaque "can't evaluate a document that has errors" error.

src/system/v1/exec.rs
```diff
@@ -0,0 +1,472 @@
...
+use wdl::analysis::Analyzer;
+use wdl::analysis::Config as AnalysisConfig;
+use wdl::analysis::Document as AnalysisDocument;
+use wdl::ast::Severity;
+use wdl::engine::CancellationContext;
+use wdl::engine::Config as WdlConfig;
+use wdl::engine::Events;
+use wdl::engine::Inputs;
+use wdl::engine::Outputs;
+use wdl::engine::v1::Evaluator as WdlEvaluator;
```
peterhuene: Is this alias necessary?

src/system/v1/exec.rs
```diff
@@ -0,0 +1,472 @@
...
+    document: &AnalysisDocument,
+    config: Arc<WdlConfig>,
+    cancellation: CancellationContext,
+    events: Events,
+    inputs: &JsonValue,
+    run_dir: &RunDirectory,
+) -> Result<Outputs> {
+    let workflow_inputs = parse_workflow_inputs(db, ctx, inputs, document, run_dir).await?;
+
+    let evaluator = WdlEvaluator::new(run_dir.root(), config, cancellation, events)
```
peterhuene: As I mentioned in Slack, we need to refactor `Evaluator` such that backends (which maintain their own permitting and limits) can be shared between all `Evaluator` instances.

This would necessitate that all evaluators for the process share the same engine configuration, but I think that will usually be the case.

Otherwise, each independent evaluator may overwhelm the backend when otherwise it should be working in concert between the evaluators.

src/config.rs
```diff
@@ -31,6 +41,10 @@ pub struct Config {
     pub analyzer: AnalyzerConfig,
     /// Configuration for the `run` command.
     pub run: RunConfig,
+    /// Configuration for the `server` command.
+    pub server: ServerConfig,
+    /// Shared execution configuration for both `run` and `server` commands.
+    pub execution: ExecutionConfig,
```
peterhuene: I don't think this should exist at the top-level of the configuration, to be honest.

We're not utilizing it from `sprocket run` at all and I would want `run` and `server` engine settings to _not_ be shared anyway.

Additionally, `ExecutionConfig` has server-specific fields in it and they feel like they belong in `ServerConfig` anyway.

Plus, we've generally kept the top-level sections to directly correspond with sprocket commands, which "execution" is not one of.

I'd probably take the approach of `RunConfig` for `ServerConfig` and flatten its own engine `Config` into it or, alternatively, don't flatten it so that the engine settings are explicit with `[server.engine]` (I think I prefer _not_ flattening it).

tests/cli/config-init/run/stdout
```diff
@@ -38,6 +38,37 @@ digests = "weak"
...
 [run.storage.google]
 
+[server]
+host = "127.0.0.1"
+port = 8080
+allowed_origins = []
+
+[server.database]
+
+[execution]
```
peterhuene: With my previous comment regarding configuration, I'd expect the non-engine settings here to be under `server` and the engine settings to perhaps be under `server.engine` with the `execution` section removed entirely.

peterhuene (commented):
Great work on getting this feature off the ground!

I had two general comments on the entire PR:

1. As I mentioned in our offline discussion, I am growing concerned with how much "kitchen sink" functionality is built into the `sprocket` executable, the time it takes to build sprocket (and its hundreds of dependencies currently), and the ultimate size of the `sprocket` executable. I could easy see `sprocket analyzer` and `sprocket server` potentially spun off into their own executables, for example.

2. To that end, I think it would be ideal to move the server implementation into its own crate rather than living in inside of the sprocket crate.

None of the above needs to be addressed in this PR, but I would like to see us move in that direction sooner than later.

migrations/20251108201315_initial.up.sql
claymcleod: We now use a integer primary key. It has the downside of forcing joins on quite a few of the queries since the uuid is not stored directly in multiple tables, but I suppose that's a fair tradeoff.

src/system/v1/db/sqlite.rs
claymcleod: In my mind, if users can list runs, it makes sense they can also list the sessions that created them. I think, for the filtering, we can address that in a follow-up PR if needed (we'll need to do that for more than just sessions).

src/system/v1/exec/svc/run_manager/commands.rs
claymcleod: Filed here: https://github.com/stjude-rust-labs/sprocket/issues/578

src/system/v1/exec/svc/run_manager.rs
claymcleod: Filed here: https://github.com/stjude-rust-labs/sprocket/issues/579.

src/system/v1/exec/svc/run_manager.rs
claymcleod: Okay, I've done the following:

- Moved document analysis, validation, target resolution, and directory creation into the spawned task
- Made target and directory nullable columns (populated after async resolution)
- Introduced `RunnableExecutor` to encapsulate the execution pipeline
- Errors during analysis now mark the run as failed rather than returning an API error

src/system/v1/db/sqlite.rs
claymcleod: Filed: https://github.com/stjude-rust-labs/sprocket/issues/581.

src/system/v1/exec.rs
claymcleod: Filed: https://github.com/stjude-rust-labs/sprocket/issues/582.

claymcleod:
As per Ari's comment on #160, there is a known deficiency in the current symlinking logic: it assumes outputs should be flat and uniquely named files, which won't hold for all WDL outputs (e.g., nested types, arrays, maps, structs). We're waiting to see if @kris70lesgo wants to take on that issue or if I should. I consider that future work and this mergeable as is.

src/system/v1/exec.rs
```diff
@@ -0,0 +1,740 @@
...
+#[derive(Debug, Error)]
+pub enum SelectTargetError {
+    /// Target not found.
+    #[error("target not found: `{0}`")]
+    TargetNotFound(String),
+    /// No tasks or workflows in document.
+    #[error("no tasks or workflows in document; the target is unable to be inferred")]
+    NoExecutableTarget,
+    /// No workflows and multiple tasks in document.
+    #[error("no workflows and multiple tasks in document; the target is unable to be inferred")]
```
peterhuene: nit:

```suggestion
    #[error("a target cannot be inferred because the document contains multiple tasks and no workflow")]
```

src/system/v1/exec.rs
```diff
@@ -0,0 +1,740 @@
...
+}
+
+/// Error type for target selection.
+#[derive(Debug, Error)]
+pub enum SelectTargetError {
+    /// Target not found.
+    #[error("target not found: `{0}`")]
+    TargetNotFound(String),
+    /// No tasks or workflows in document.
+    #[error("no tasks or workflows in document; the target is unable to be inferred")]
```
peterhuene: Nit:


```suggestion
    #[error("a target cannot be inferred because the document contains no tasks and no workflow")]
```

src/system/v1/exec.rs
```diff
@@ -0,0 +1,740 @@
...
+    results
+        .into_iter()
+        .find(|result| **result.document().uri() == uri)
+        .context("analyzer didn't return analysis results for document")
+}
+
+/// Validates the analysis result for parsing and diagnostic errors.
+///
+/// Returns an error if the analysis failed or contains error-level diagnostics.
+pub async fn validate_analysis_results(result: &AnalysisResult) -> Result<()> {
```
peterhuene: Given that we're now checking all of the results in `analyze_wdl_document`, is this function still needed?

peterhuene (approved):

PR #543, opened 2026-01-14
fix: respect requirements and hints specified via inputs.

### LINKED ISSUE #211: runtime and hints can't be overrode if nested inputs not allowed
**Describe the bug**

The WDL spec carves out an exception in the nested inputs behavior for runtime values and hints, which should always be overridable. However if nested inputs aren't allowed for a workflow, `wdl-engine` does not permit that.

**Expected behavior**

The spec should be followed.

**Additional context**

> Unlike inputs, a WDL implementation must support overriding requirements and hints regardless of whether it supports the allow_nested_inputs workflow hint. Requirements and hints specified in the input JSON always supersede values supplied directly in the WDL. Any hints that are not supported by the execution engine are ignored.

### LINKED ISSUE #539: Pass requirements/hints from inputs to backends
**Describe the bug**
When I specify `requirements` parameters in the JSON input, it isn't being respected. Sprocket is not passing requirements/hints that it receives as inputs to the backends properly.

**Expected behavior**
The backend should receive the requirements/hints that were specified in the inputs.

peterhuene:
This PR ensures that task requirements and hints that are specified for a
task via input JSON override what is read from the task's `requirements` or
`hints` sections.

Additionally, this contains a fix allowing setting task requirements and hints in nested calls
even if the workflow does not accept nested inputs.

Code relating to reading requirements and hints has been extracted out of `task.rs` into their own respective modules.

Fixes #539.
Fixes #211.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/src/eval/v1/task/requirements.rs
+++ b/crates/wdl-engine/src/eval/v1/task/requirements.rs
@@ -54,8 +54,7 @@ const SIF_EXTENSION: &str = "sif";
 /// Represents the source of a container image.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ContainerSource {
-    /// A Docker registry image (e.g., `ubuntu:22.04` or
-    /// `docker://ubuntu:22.04`).
+    /// A Docker registry image (e.g. `docker://ubuntu:22.04`).
     Docker(String),
     /// A Sylabs library image (e.g., `library://sylabs/default/alpine`).
     Library(String),


crates/wdl-engine/tests/workflows/override-requirements/inputs.json
```diff
@@ -0,0 +1,5 @@
+{
+    "w.t.requirements.container": "ubuntu:focal",
```
adthrasher, line 2: Can we choose a different input, since `ubuntu:focal` is the default in the task?

peterhuene: It's `ubuntu:latest` in the task. These values were chosen since we already pull the images before the tests run.

adthrasher: Ah, I was referring to the command block. Admittedly, that should never trigger since there is a default, and we're overriding it.
```
        echo "~{if task.container == None then "ubuntu:focal" else task.container}"
```

peterhuene: It needs to be `ubuntu:focal` here to get the same output baseline for both the local backend (where `container` is always `None`) and the Docker backend.

`task.container` is the container that the backend used, which may not be exactly what was specified in the requirement/inputs.

adthrasher: Ah, that makes sense now

crates/wdl-engine/src/eval/v1/task/requirements.rs
```diff
@@ -0,0 +1,671 @@
...
+const LIBRARY_PROTOCOL: &str = "library://";
+
+/// The OCI Registry as Storage protocol prefix.
+const ORAS_PROTOCOL: &str = "oras://";
+
+/// The file protocol prefix for local container files.
+const FILE_PROTOCOL: &str = "file://";
+
+/// The expected extension for local SIF files.
+const SIF_EXTENSION: &str = "sif";
```
a-frantz, line 52:
```suggestion
/// The Docker registry protocol prefix.
const DOCKER_PROTOCOL: &str = "docker://";
/// The Sylabs library protocol prefix.
const LIBRARY_PROTOCOL: &str = "library://";
/// The OCI Registry as Storage protocol prefix.
const ORAS_PROTOCOL: &str = "oras://";
/// The file protocol prefix for local container files.
const FILE_PROTOCOL: &str = "file://";

/// The expected extension for local SIF files.
const SIF_EXTENSION: &str = "sif";
```
very small nit, but I have a preference for compacting this block of protocols

crates/wdl-engine/src/eval/v1/task/requirements.rs
```diff
@@ -0,0 +1,671 @@
...
+
+/// Represents the source of a container image.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ContainerSource {
+    /// A Docker registry image (e.g., `ubuntu:22.04` or
+    /// `docker://ubuntu:22.04`).
+    Docker(String),
```
a-frantz:
```suggestion
    /// A Docker registry image (e.g. `docker://ubuntu:22.04`).
    Docker(String),
```
for what this enum represents, I think the inclusion of the prefix-less form confuses things

a-frantz (approved):

PR #549, opened 2026-01-15
fix: ignore broken symlinks for call caching.

### LINKED ISSUE #544: dangling symlinks should not cause caches to be discarded
A common (and recommended ) practice is to create directories with symlinks during a task execution.  Those links are to the mount points; so they appear dangling in subsequent task executions and you get error messages like:
 
2026-01-14T18:34:23.105391Z ERROR wdl_engine::eval::v1::task: failed to update call cache entry for task `somalier_extract` (task id `somalier_extract-2`): cache entry has been discard: failed to read metadata of `/research_jude/rgs01_jude/clingen/dev/wc/drosenfe/cluster_code/trunk/sample_identity/test/runs/sample_identity/2026-01-14_183352482762088/calls/somalier_extract-2/attempts/0/work/vcf/SJST037995_D1_TRANSCRIPTOME.sitesfuzzion2.vcf.gz`: No such file or directory (os error 

If possible the dangling symlinks should be ignored and the cache entries not discarded in this situation.

peterhuene:
This fixes the digesting of directories entries when the entry is a broken symlink.

Previously, the broken symlink would cause an error and the error would trigger call caching to discard the cache entry, effectively disabling call caching for any task that left behind a broken symlink in its working directory.

The fix is to ignore a broken symlink from the digest calculation of the directory.

Additionally, discarding a cache entry is now logged at a warning level rather than an error level.

This also fixes the `glob` function's handling of symlinks.

Fixes #544.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/CHANGELOG.md
+++ b/crates/wdl-engine/CHANGELOG.md
@@ -9,8 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed an issue with call caching not working when a directory being cached
   contained a broken symlink ([#549](https://github.com/stjude-rust-labs/sprocket/pull/549)).
-* Fixed the `glob` function not returning broken symlinks as required by the
-  WDL spec ([#549](https://github.com/stjude-rust-labs/sprocket/pull/549)).
+* Fixed the `glob` function not properly resolving symlinks to files ([#549](https://github.com/stjude-rust-labs/sprocket/pull/549)).
 
 ## 0.11.1 - 01-12-2026
 


peterhuene:
It looks like there was some discussion here: https://github.com/openwdl/wdl/issues/718.

I agree that this appears to be at odds with the point of `File` being checked for existence in other places of the space.

Given this, I'm going to back that change out and we can raise a spec issue.

peterhuene:
I've pushed up a commit to exclude them for now.

crates/wdl-engine/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+* Fixed an issue with call caching not working when a directory being cached
+  contained a broken symlink ([#549](https://github.com/stjude-rust-labs/sprocket/pull/549)).
+* Fixed the `glob` function not returning broken symlinks as required by the
```
a-frantz: This needs to be updated

PR #553, opened 2026-01-20
feat(wdl-lint)!: lint rule configuration

### LINKED ISSUE #538: Allow specific additional runtime keys.
**Problem**
I'd like to enable as many lints as possible, but I have a problem with the `ExpectedRuntimeKeys` lint. I am running on `DNANexus`, and occasionally it's very useful to specify `dx_instance_type`. 

Some ability to add additional keys seems reasonable, since [older specs explicitly allow backend-specific object blocks](https://github.com/openwdl/wdl/blob/legacy/versions/1.1/SPEC.md#conventions-and-best-practices), and [v1.3 seems to allow arbitrary keys](https://github.com/openwdl/wdl/blob/wdl-1.3/SPEC.md#-runtime-section).

**Requested solution**
Either of:
1. (Preferred): the ability to whitelist specific runtime keys, e.g. `allowed_runtime_keys = ["dx_instance_type"]` in sprocket.toml
2. the ability to `#@except: ExpectedRuntimeKeys` for an individual key

**Alternatives I've considered**
I have tried adding `#@ except: ExpectedRuntimeKeys` before the `runtime` section, but it disables all checks for a task, which seems too lax.

[Edit: fixed typo]

Serial-ATA:
This adds a new lint rule `Config` to `wdl-lint`, which is provided from `sprocket.toml`.

Options are specified like so:

```toml
[check.rules]
allowed_runtime_keys = ["dx_instance_type"]
```

For now the only option is `allowed_runtime_keys` for the `ExpectedRuntimeKeys` rule (#538).

This also adds support for setting config values in the UI tests, see [TESTING.md](https://github.com/Serial-ATA/sprocket/blob/ba01ee043d316239406ad87c4e8fb225082530b5/crates/wdl-lint/TESTING.md).

closes #538

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-lint/src/rules/expected_runtime_keys.rs
+++ b/crates/wdl-lint/src/rules/expected_runtime_keys.rs
@@ -143,7 +143,9 @@ fn deprecated_runtime_key(key: &Ident, replacement: &str) -> Diagnostic {
     ))
 }
 
-/// Creates an "non-reserved runtime key" diagnostic.
+/// Creates a "non-reserved runtime key" diagnostic.
+///
+/// Returns `None` if every given key is allowed.
 fn report_non_reserved_runtime_keys(
     keys: &HashSet<TokenText>,
     allowed_runtime_keys: &HashSet<String>,


a-frantz:
There are quite a few lint rules I'd like to see get some form of configuration; they don't all need to be implemented as part of this PR but this is exciting to see! We have been intending to implement something like this since Sprocket's earliest versions, but lint configuration has never reached the top of any of our priority lists.
Evidence we've been thinking about this for so long can be seen in issue #3 :rofl:

crates/wdl-lint/src/rules/expected_runtime_keys.rs
```diff
@@ -264,6 +279,22 @@ pub struct ExpectedRuntimeKeysRule {
     encountered_keys: Vec<Ident>,
     /// All non-reserved keys encountered in the current runtime section.
     non_reserved_keys: HashSet<TokenText>,
+    /// Whitelisted keys from the config.
```
a-frantz:
```suggestion
    /// Allowed keys from the config.
```

crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors
a-frantz, line 1: shouldn't this file be unchanged?

Serial-ATA: looks like the config is randomly empty on some runs. looking into it

Serial-ATA: fixed in [`370732e` (this PR)](https://github.com/stjude-rust-labs/sprocket/pull/553/commits/370732ec86144cca7bf4416f970bb65f8346b407)

crates/wdl-lsp/src/server.rs
```diff
@@ -237,6 +234,17 @@ pub struct ServerOptions {
     pub feature_flags: FeatureFlags,
 }
 
+/// Options for the external linter.
+#[derive(Debug, Default, Clone)]
+pub struct LintOptions {
+    /// Whether or not linting is enabled.
+    pub enabled: bool,
+    /// Analysis or lint rule IDs to except (ignore).
+    pub exceptions: Vec<String>,
```
a-frantz: as these may include analysis rules while `enabled == false`, I'd prefer `exceptions` to remain outside of `LintOptions`

crates/wdl-lint/src/linter.rs
```diff
@@ -66,7 +67,10 @@ impl Linter {
 impl Default for Linter {
     fn default() -> Self {
         Self {
-            rules: rules().into_iter().map(|r| (r.id(), r)).collect(),
+            rules: rules(&Config::default())
```
a-frantz, line 70: nit:
```suggestion
            rule_config: rules(&Config::default())
```
I think naming it `rules` by itself isn't explicit enough. That makes it sound like it's the setting for which rules are enabled IMO

Serial-ATA: not sure `rule_config` makes sense here. it's holding a map of already configured rules

a-frantz: o, I commented on the wrong line of code 😅 hold on. I was looking for the new `rules` table in `sprocket.toml`

Serial-ATA: could also name the table `wdl_lint`, similar to how the Cargo lint tables are named after tools (`lints.rust`, `lints.clippy`, `lints.rustdoc`). what do you think?

src/config.rs
```diff
@@ -100,6 +100,8 @@ pub struct CheckConfig {
     pub only_lint_tags: Vec<String>,
     /// Set of lint tags to filter out of the enabled lint rules.
     pub filter_lint_tags: Vec<String>,
+    /// Lint rule configuration.
+    pub rules: wdl::lint::Config,
```
a-frantz: this is  the line I thought I was commenting on in the other thread
```suggestion
    pub rule_config: wdl::lint::Config,
```

a-frantz: or maybe
```suggestion
    pub lint: wdl::lint::Config,
```

Serial-ATA: thoughts on https://github.com/stjude-rust-labs/sprocket/pull/553#discussion_r2717365725?

a-frantz: I don't like having the `wdl_` prefix. We generally try to keep those crate names from "leaking" out to users, as it's an internal code structure detail that users shouldn't have to concern themselves with.

I wouldn't call `wdl_lint` a tool in itself. That's more of an artifact from a prior split we were trying to enforce between "Sprocket" and the "`wdl` family of crates". We've been discussing some vague plans for the `sprocket/src` directory to absorb more of the code under `sprocket/crates/*`, and I think the entire `wdl_lint` crate is a candidate for migration.

Serial-ATA: alright, I think `lint` would be best then

a-frantz (approved):
Great, excited to get lint config in! Not a requirement for merging this PR, but we should probably update documentation in a few places. Particularly, I'd like to point out that expanding what's configurable in `wdl_lint` is a good area for new contributors to jump into the project. There's a few use cases that should be easier to enable and implement with this infra initialized now

a-frantz:
o failing CI. Once that gets patched up and is passing, I can merge

a-frantz:
> > we should probably update documentation in a few places
> 
> I know of https://sprocket.bio/configuration/overview.html. Is there anything else I should update?


https://sprocket.bio/subcommands/check-lint.html
Could be expanded by a great deal. It's pretty sparse now (i think the "guided tour" page covers a lot that might be better suited for the check/lint page)

But mostly what I have in mind is opening new issues for features that can be added to existing lints.

src/lib.rs
```diff
@@ -72,7 +73,8 @@ struct Cli {
     skip_config_search: bool,
 }
 
-async fn inner() -> CommandResult<()> {
+/// Logic for [`sprocket_main()`].
+async fn real_main() -> CommandResult<()> {
```
a-frantz, line 77: why was this renamed?

Serial-ATA: `real_main` is the name I typically see for fallible inner main functions

crates/wdl-analysis/src/queue.rs
```diff
@@ -881,54 +880,48 @@ where
...
+                            results.push(AnalysisResult::new(node));
                         }
+                        continue;
+                    }
 
-                        let graph = self.graph.clone();
-                        let config = self.config.clone();
-                        let validator = self.validator.clone();
-                        Some(RayonHandle::spawn(move || {
-                            thread_local! {
```
peterhuene, line 899: Is there a reason why the per-thread caching and reuse of the validator was removed?

Each validator may internally allocate many (and that number will continue to grow over time) rules each with their own internal state that may include other allocations that can be reused between analysis rather than deallocated between analysis of a single graph node.

Keep in mind that the LSP server keeps a single `Analyzer` instance alive to service all LSP requests, so validator reuse is more applicable to that use case than the CLI analysis use cases (e.g. `sprocket check`, `sprocket run`, etc.).

crates/wdl-lint/src/rules/expected_runtime_keys.rs
```diff
@@ -145,12 +146,21 @@ fn deprecated_runtime_key(key: &Ident, replacement: &str) -> Diagnostic {
 /// Creates an "non-reserved runtime key" diagnostic.
```
peterhuene:
```suggestion
/// Creates a "non-reserved runtime key" diagnostic.
///
/// Returns `None` if every given key is allowed.
```

crates/wdl-analysis/src/queue.rs
Serial-ATA: See [`370732e` (this PR)](https://github.com/stjude-rust-labs/sprocket/pull/553/commits/370732ec86144cca7bf4416f970bb65f8346b407). Since the validator holds the lint config, stale configs would occasionally get reused between steps causing tests to fail.

peterhuene: I see. That's not a problem for the LSP server because config changes cause a restart of the server process entirely.

Let me noodle on this for a minute.

peterhuene: The existing design, in general, can lead to unexpected results for anyone that wants to have more than one `Analyzer` with different validator configuration per process, which _definitely_ isn't great.

If the validator reuse optimization is truly needed, then perhaps each `Analyzer` can manage a pool of validators itself for each worker thread to dip into rather than using a thread local.

Let's proceed with the changes in this PR and we can perhaps restore the optimization without effectively locking a worker thread to a particular validator instance indefinitely.

peterhuene: I'll file an issue for this and we're good to merge this PR 👍

Serial-ATA: Yeah, I have no problem revisiting this in a follow up PR (if there are actual performance concerns). Just needed to get this working

peterhuene: Filed #556.

Issue #223, opened 2025-07-07
better error reporting


a-frantz:
**Is your feature request related to a problem? Please describe.**
Discussed briefly in [this thread](https://github.com/stjude-rust-labs/wdl/pull/339/files#r2183028011) but copying relevant bits here:
> I think in other places where we have to deal with something similar (ensuring documents are error-free before further processing) we usually emit only the first error encountered and then bail. This strikes me as potentially annoying for a user if they are under the impression the error emitted was the only error: they fix it, then re-run, and get hit with a different error. But showing every error level diagnostic also feels not great.

**Describe the solution you'd like**
After Clay raised this concern on stjude-rust-labs/wdl#339, I read up on Rust error handling conventions (specifically for libraries) and I think we'd benefit from switching some/most of our current `anyhow` uses to `thiserror`. 

**Describe alternatives you've considered**
Emitting `thiserror` errors instead of `anyhow` errors from the `wdl` crate doesn't solve the problem outlined in my comment, as we still have to figure out the right UX for the `sprocket` bin. But it would improve the state of this library to punt the problem further down the line to the actual consumers of these errors and let them handle it as deemed most appropriate.

Serial-ATA:
#559 solved the original issue in `wdl-doc`, so I think this can be closed. Transitioning from `anyhow` -> `thiserror` could still be done in the future, but that should be broken up into issues for each crate. Though the diagnostics situation has improved a lot since this was made, so I don't think it's necessary to go out of our way to do. It'd really just be for the sake of following best practices.

PR #559, opened 2026-01-23
refactor(wdl-doc)!: fail on analysis errors

### LINKED ISSUE #555: `wdl-doc` should fail on analysis errors
Right now `wdl-doc` can be used on documents that fail analysis. Meaning it allows documents with issues like:

* enums in versions <1.3
* Multiple `meta`/`parameter_meta` sections
* Multiple `workflows`
* etc...

`sprocket dev doc` should emit the diagnostics like `sprocket check` and exit.

Serial-ATA:
Ignoring analysis errors in `sprocket dev doc` could lead to generating invalid docs (see https://github.com/stjude-rust-labs/sprocket/issues/555#issuecomment-3786905556).

This PR:

* Makes `wdl-doc` error when encountering analysis errors
* Adds a `DocError` type for concrete errors
	* Right now, this is only used to get the analysis results back to the caller, but ideally should be expanded to cover all error cases and replace `anyhow` (#223).
* Makes `sprocket` emit the diagnostics

closes #555

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-doc/src/error.rs
+++ b/crates/wdl-doc/src/error.rs
@@ -50,12 +50,12 @@ impl Display for NpmError {
     }
 }
 
-/// The kinds of errors that can occur
+/// The kinds of errors that can occur.
 #[derive(Debug)]
 pub enum DocErrorKind {
     /// The expected workspace was not found.
     WorkspaceNotFound(PathBuf),
-    /// No WDL documents were found in the workspace
+    /// No WDL documents were found in the workspace.
     NoDocuments,
     /// Failed to run analysis on the workspace.
     Analyzer(anyhow::Error),
@@ -87,7 +87,7 @@ impl DocError {
         }
     }
 
-    /// The kind of error that occured.
+    /// The kind of error that occurred.
     pub fn kind(&self) -> &DocErrorKind {
         &self.kind
     }


src/commands/doc.rs
```diff
@@ -186,12 +201,46 @@ pub async fn doc(args: Args) -> CommandResult<()> {
...
         .prefer_full_directory(!args.prioritize_workflows_view);
 
-    document_workspace(config).await.with_context(|| {
-        format!(
-            "failed to generate documentation for workspace at `{}`",
-            workspace.display()
-        )
-    })?;
+    let mut counts = DiagnosticCounts::default();
+    if let Err(e) = document_workspace(config).await {
```
claymcleod, line 205: This can be condensed down to a single match statement

```
match document_workspace(config).await {
  Ok(()) => {},
  Err(DocError::AnalysisFailed(analysis_results)) => {
  }
  // etc
}
```

crates/wdl-doc/src/lib.rs
```diff
@@ -304,32 +306,16 @@ async fn analyze_workspace(
         .with_context(|| "failed to analyze workspace".to_string())?;
 
     if results.is_empty() {
-        return Err(anyhow!("no WDL documents found in analysis",));
+        return Err(anyhow!("no WDL documents found in analysis",).into());
```
claymcleod:
```suggestion
        return Err(anyhow!("no WDL documents found in analysis").into());
```

src/diagnostics.rs
```diff
@@ -35,6 +36,58 @@ static ONE_LINE_CONFIG: LazyLock<TermConfig> = LazyLock::new(|| TermConfig {
...
+    pub errors: usize,
+    /// The number of warnings encountered.
+    pub warnings: usize,
+    /// The number of notes encountered.
+    pub notes: usize,
+}
+
+impl DiagnosticCounts {
+    /// Returns an error if the `errors` count is 1 or more
+    pub fn fail_errors(&self) -> Option<anyhow::Error> {
```
claymcleod:
```suggestion
    pub fn verify_no_errors(&self) -> Option<anyhow::Error> {
```

src/diagnostics.rs
```diff
@@ -35,6 +36,58 @@ static ONE_LINE_CONFIG: LazyLock<TermConfig> = LazyLock::new(|| TermConfig {
...
+
+        Some(anyhow!(
+            "failing due to {errors} error{s}",
+            errors = self.errors,
+            s = if self.errors == 1 { "" } else { "s" }
+        ))
+    }
+
+    /// Returns an error if the `warnings` count is 1 or more
+    pub fn fail_warnings(&self) -> Option<anyhow::Error> {
```
claymcleod:
```suggestion
    pub fn verify_no_warnings(&self) -> Option<anyhow::Error> {
```

src/diagnostics.rs
```diff
@@ -35,6 +36,58 @@ static ONE_LINE_CONFIG: LazyLock<TermConfig> = LazyLock::new(|| TermConfig {
...
+
+        Some(anyhow!(
+            "failing due to {warnings} warning{s} (`--deny-warnings` was specified)",
+            warnings = self.warnings,
+            s = if self.warnings == 1 { "" } else { "s" }
+        ))
+    }
+
+    /// Returns an error if the `notes` count is 1 or more
+    pub fn fail_notes(&self) -> Option<anyhow::Error> {
```
claymcleod:
```suggestion
    pub fn verify_no_notes(&self) -> Option<anyhow::Error> {
```

CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added setting `check.rules` to `sprocket.toml` for controlling `wdl-lint`
   rule configuration ([#553](https://github.com/stjude-rust-labs/sprocket/pull/553)).
 
+### Changed
+
+* `sprocket dev doc` will now **fail** in the presence of analysis errors ([#559](https://github.com/stjude-rust-labs/sprocket/pull/559))
```
claymcleod: I'm not a fan of introducing two tiers of bulleted items in the changelog. I feel these should be condensed down to one bullet point.

crates/wdl-doc/src/error.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::fmt::Display;
+
+/// Errors that can occur while generating documentation.
+#[derive(Debug)]
+pub enum DocError {
+    /// One or more documents failed analysis.
+    ///
+    /// NOTE: This contains *all* analysis results, including those of valid
+    /// documents.
+    AnalysisFailed(Vec<wdl_analysis::AnalysisResult>),
```
claymcleod: nit: I'd import `AnalysisResult` here.

claymcleod: Also, is there a reason this includes all results? A cursory look suggests we just filter out things that aren't errors later on down the road.

crates/wdl-doc/src/error.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// All other errors (e.g. I/O)
+    Other(anyhow::Error),
```
claymcleod: I'm not sure the `Other` variant is required.

There doesn't appear to be any special handling of this variant other than unwrapping the error and returning it.

You could remove this variant and have `document_workspace` return `anyhow::Result<()>`, using `DocError` only for the analysis failure case:

```rust
// lib.rs
// Only wrap in `DocError` for analysis failures.
if has_errors {
    return Err(DocError::AnalysisFailed(results).into());
}
// Other errors just use ? normally
```

Then in the caller, use `downcast_ref` to check for the special case:

```rust
if let Err(e) = document_workspace(config).await {
  // Check if it's the special analysis failure case
  if let Some(DocError::AnalysisFailed(results)) = e.downcast_ref::<DocError>() {
    // emit diagnostics from results
    for result in results {
      // ...
    }
  } else {
    // Regular error, just return with context
    return Err(e.context(format!(
      "failed to generate documentation for workspace at `{}`",
      workspace.display()
    )).into());
  }
}
```

crates/wdl-doc/src/error.rs
Serial-ATA: I figured it'd make sense to return the full workspace since `wdl-doc` is the one responsible for producing it. If we have that information available, might as well return it to the caller

crates/wdl-doc/src/error.rs
Serial-ATA: The `Other` variant exists so that `wdl-doc` can use `DocError` everywhere with the goal of eventually removing `anyhow`. From the PR description:

> Right now, this is only used to get the analysis results back to the caller, but ideally should be expanded to cover all error cases and replace anyhow (https://github.com/stjude-rust-labs/sprocket/issues/223).

crates/wdl-doc/src/lib.rs
```diff
@@ -12,6 +12,7 @@ include!(concat!(env!("OUT_DIR"), "/assets.rs"));
 mod command_section;
 mod docs_tree;
 mod document;
+pub mod error;
```
claymcleod, line 15: I would just export `DocError` at the root of the crate and not make the whole module public.

crates/wdl-doc/src/lib.rs
```diff
@@ -291,7 +293,7 @@ impl VersionBadge {
 async fn analyze_workspace(
     workspace_root: impl AsRef<Path>,
     config: AnalysisConfig,
-) -> Result<Vec<wdl_analysis::AnalysisResult>> {
+) -> Result<Vec<wdl_analysis::AnalysisResult>, DocError> {
```
claymcleod: nit: I'd import `AnalysisResult` here.

crates/wdl-doc/src/lib.rs
```diff
@@ -345,7 +331,12 @@ async fn analyze_workspace(
         return Err(anyhow!(
             "workspace root `{root}` not found in analysis results",
             root = workspace.display(),
-        ));
+        )
+        .into());
+    }
+
+    if has_errors {
+        return Err(DocError::AnalysisFailed(results));
```
claymcleod: Rather than `has_errors` and returning everything, I'd just do one `filter` and `collect` on results with an error and then check if that is empty here.

src/diagnostics.rs
```diff
@@ -35,6 +36,58 @@ static ONE_LINE_CONFIG: LazyLock<TermConfig> = LazyLock::new(|| TermConfig {
...
+    }
+
+    /// Returns an error if the `warnings` count is 1 or more
+    pub fn verify_no_warnings(&self) -> Option<anyhow::Error> {
+        if self.warnings == 0 {
+            return None;
+        }
+
+        Some(anyhow!(
+            "failing due to {warnings} warning{s} (`--deny-warnings` was specified)",
```
claymcleod: It feels a little strange to just assume `--deny-warnings` is added here. It's a bit awkward, but I might add a variable `flag_provided` that optionally adds this parenthetical to the error. Same for notes below.

crates/wdl-doc/src/error.rs
claymcleod: The `downcast_ref` approach actually gets us closer to that goal. With `Other(anyhow::Error)`, you're embedding `anyhow` inside `DocError`; to remove it later, you'd need to figure out what `Other` was catching and replace it with concrete variants. With this approach, you can keep the `anyhow` handling in just the top-level crate.

crates/wdl-doc/src/error.rs
claymcleod: Personally, I think YAGNI—semantically, if we get to a place where we're returning a `DocError::AnalysisFailed`, the only thing to do in the `doc` command context is to report the errors.

crates/wdl-doc/src/error.rs
claymcleod: After discussing with @Serial-ATA, it became clear that the goal is also to remove `anyhow::Result` from being used in any of the signatures within `wdl-doc/src/lib.rs`. In that case, it makes sense to me to just go ahead and change over all the signatures in one PR and flesh out the full error type.

crates/wdl-doc/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## Unreleased
 
+#### Added
+
+* Added `DocError` for concrete error reporting ([#559](https://github.com/stjude-rust-labs/sprocket/pull/559))
+
+#### Changed
+
+* WDL documents with analysis errors will now **fail**
+  * Ignoring analysis errors could lead to generating invalid documentation (See [#555](https://github.com/stjude-rust-labs/sprocket/issues/555))
```
claymcleod: I think we missed a bullet that should be combined into one here.

crates/wdl-doc/src/error.rs
claymcleod: One alternative to consider: rather than storing the full `Vec<AnalysisResult>` in `AnalysisFailed`, we could store just the relevant diagnostics paired with their source URIs. This could be constructed as:

```rust
let failed: Vec<_> = results
  .iter()
  .filter_map(|r| {
    let errors: Vec<_> = r.document()
      .diagnostics()
      .filter(|d| d.severity() == Severity::Error)
      .cloned()
      .collect();
    if errors.is_empty() {
      None
    } else {
      Some((r.document().uri().clone(), errors))
    }
  })
  .collect();

if !failed.is_empty() {
  return Err(DocError::new(DocErrorKind::AnalysisFailed(failed)));
}
```

The consumer gets exactly what they need to report the errors without carrying the entire analysis state.

claymcleod: Actually I see we do make use of the source to display the error in `doc.rs`. So it appears we do need the full `AnalysisResult`.

crates/wdl-doc/src/error.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            NpmError::Build(e) => write!(f, "failed to run `npm run build`: {e}"),
+            NpmError::Install(e) => write!(f, "failed to run `npm install`: {e}"),
+            NpmError::Tailwind(e) => write!(f, "failed to run `npx @tailwindcss/cli`: {e}"),
+        }
+    }
+}
+
+/// The kinds of errors that can occur
```
claymcleod:
```suggestion
/// The kinds of errors that can occur.
```

crates/wdl-doc/src/error.rs
```diff
@@ -0,0 +1,147 @@
...
+#[derive(Debug)]
+pub enum DocErrorKind {
+    /// The expected workspace was not found.
+    WorkspaceNotFound(PathBuf),
+    /// No WDL documents were found in the workspace
```
claymcleod:
```suggestion
    /// No WDL documents were found in the workspace.
```

crates/wdl-doc/src/error.rs
```diff
@@ -0,0 +1,147 @@
...
+impl DocError {
+    /// Create a new `DocError`.
+    pub fn new(kind: DocErrorKind) -> Self {
+        Self {
+            kind,
+            context: None,
+        }
+    }
+
+    /// The kind of error that occured.
```
claymcleod:
```suggestion
    /// The kind of error that occurred.
```

PR #560, opened 2026-01-24
fix(lint): flag meta.outputs entries when task has no output section

### LINKED ISSUE #487: bug: mismatched `meta.outputs` and `outputs` doesn't always result in diagnostics
A clear and concise description of what the problem is.

spotted the issue in this unrelated PR - https://github.com/stjude-rust-labs/sprocket/pull/449#discussion_r2586125008

A `meta.outputs` field populated with non-existent outputs should get flagged. Only "real" outputs defined in the `output` section should appear within `meta.outputs`

~~new lint rule!~~ bug fix please!

VedantMadane:
## Summary

Fixes #487

The MatchingOutputMeta lint rule was not flagging meta.outputs entries when a task had no output section at all. This was due to an early return when task_outputs.is_empty(), which prevented the diagnostic from being reported.

## Changes

- Removed the condition that required an output section to exist before checking meta.outputs
- The rule now correctly reports diagnostics when:
  - Task has meta.outputs entries
  - Task has no output section (all meta.outputs are unmatched)
- Added test case missing_output_section to verify the fix

## Testing

- Existing tests pass
- New test case added for task without output section
- Manually tested with example WDL files

### DIFF
--- a/crates/wdl-lint/tests/lints/nonmatching-output/source.errors
+++ b/crates/wdl-lint/tests/lints/nonmatching-output/source.errors
@@ -1,3 +1,11 @@
+note[ShellCheck]: running `shellcheck` on command section
+   ┌─ tests/lints/nonmatching-output/source.wdl:13:5
+   │
+13 │     command <<< >>>
+   │     ^^^^^^^ could not find `shellcheck` executable.
+   │
+   = fix: install shellcheck (https://www.shellcheck.net) or disable this lint.
+
 warning[MatchingOutputMeta]: `outputs` key missing in `meta` section for the task `bar`
    ┌─ tests/lints/nonmatching-output/source.wdl:22:5
    │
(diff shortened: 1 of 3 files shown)


peterhuene (approved):

crates/wdl-lint/tests/lints/nonmatching-output/source.errors
```diff
@@ -110,3 +110,19 @@ warning[MatchingOutputMeta]: `v` appears in `outputs` section of the task `quuux
...
+    = fix: ensure the output exists or remove the `nonexistent` key from `meta.outputs`
+
+warning[MatchingOutputMeta]: `also_nonexistent` appears in `outputs` section of the task `no_output_section` but is not a declared `output`
+    ┌─ tests/lints/nonmatching-output/source.wdl:223:13
+    │
+223 │             also_nonexistent: "Neither does this one",
+    │             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    │
+    = fix: ensure the output exists or remove the `also_nonexistent` key from `meta.outputs`
+
```
adthrasher, line 128: @a-frantz - I'm curious what you think since you opened the original issue. I think I'd rather have a lint that just flags that `meta.outputs` exists and is non-empty when the `output` block doesn't exist. This seems less intrusive than having a warning for each entry.

a-frantz: Maybe that would be ideal, but this is also a case that is rather niche and I'm not sure we need to be overly concerned with the intrusiveness of it firing. If the `arena` CI is green, that means this case was never found in our corpus of real-world WDL. I'm ok with this as is.

peterhuene:
@VedantMadane looks like your last commit deleted some files it shouldn't an added `.exe` to the baseline when we expect it not to (we don't generally run the update from Windows, so no fault of yours). I may have to push up a fix or two, but you're all good on your end.

a-frantz (approved):
Awesome, sorry for this review being so delayed!

PR #562, opened 2026-01-26
refactor: improve LSF backend.

### LINKED ISSUE #514: The Apptainer backend doesn't support `run.task.shell`
The Apptainer backend [hardcodes `bash` rather than relying on the value of `run.task.shell`](https://github.com/stjude-rust-labs/sprocket/blob/467ee53e04ad05533a00419270b2f9cbe1db3cc3/crates/wdl-engine/src/backend/apptainer.rs#L196). We should change that.

peterhuene:
This PR improves the LSF backend with the following enhancements:

* `bsub` is now used to queue tasks only.
* A new task monitor implementation periodically runs `bjobs` to monitor the
  current state of tasks.
* `bkill` is now used to cancel tasks.
* The apptainer scripts now respect the configured shell.
* Context was added to more errors.

Two new configuration settings for the LSF backend were added:

* `interval` - controls how often the LSF backend will poll for job states with
  `bsub`.
* `max_concurrency` - controls how many `bsub` and `bkill` processes are
  spawned concurrently.

Closes #514.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -193,8 +216,9 @@ impl ApptainerState {
         writeln!(&mut apptainer_command, "\"{}\" \\", container_sif.display())?;
         writeln!(
             &mut apptainer_command,
-            "bash -c \"\\\"{GUEST_COMMAND_PATH}\\\" > \\\"{GUEST_STDOUT_PATH}\\\" 2> \
-             \\\"{GUEST_STDERR_PATH}\\\"\" \\"
+            "{shell} -c \"\\\"{GUEST_COMMAND_PATH}\\\" > \\\"{GUEST_STDOUT_PATH}\\\" 2> \
```
adthrasher, line 216: I'm not necessarily opposed to making the shell configurable, though the WDL spec repeatedly refers to Bash functionality and a task as:

> A task encapsulates a Bash script and a UNIX environment and presents them as a reusable function.

 I think the documentation for this option needs a pretty heavy warning. Also, the `-c` should probably be configurable. I'm assuming not every shell has a `-c` that matches Bash.

adthrasher: Also, our documentation should probably be clear: "if you use a non-Bash shell, then don't file an issue with us if the task fails".

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -54,45 +73,555 @@ use crate::v1::requirements::ContainerSource;
...
+    /// The LSF job id.
+    #[serde(rename = "JOBID")]
+    job_id: String,
+    /// The job state.
+    #[serde(rename = "STAT")]
+    state: String,
+    /// The job exit code (may be empty).
+    #[serde(rename = "EXIT_CODE")]
+    exit_code: String,
+}
```
adthrasher, line 191: Not necessarily relevant for this PR, but since we're polling jobs with `bjobs`, it'd be awesome to have a way to enable tracking of resource utilization. It's the kind of information we always need to collect and always have to do manually.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -123,18 +656,61 @@ impl LsfApptainerBackend {
...
+            .arg(job_id.to_string())
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null());
+
+        let _permit = self
+            .permits
+            .acquire()
+            .await
+            .context("failed to acquire permit for submitting job")?;
```
adthrasher:
```suggestion
            .context("failed to acquire permit for canceling job")?;
```

crates/wdl-engine/src/backend/apptainer.rs
peterhuene: I'll put more "there be dragons here" in the docs. We already do have this warning, at least:

```
Warning: the use of a shell other than `bash` may lead to tasks that may not be portable to other execution engines.
```

crates/wdl-engine/src/backend/lsf_apptainer.rs
peterhuene: That seems easily enough done. Let me add that and we can log the resources at the very least.

crates/wdl-engine/src/backend/apptainer.rs
peterhuene: Updating to:

```
Warning: the use of a shell other than `bash` may lead to tasks that may not be portable to other execution engines.

The shell must support a `-c` option to run a specific script file (i.e. an evaluated task command).

Note that this option affects all task commands, so every container that is used must contain the specified shell.

If using this setting causes your tasks to fail, please do not file an issue. 
```

crates/wdl-engine/src/backend/lsf_apptainer.rs
peterhuene: Also, for what it's worth, we still send the LSF job report to the `job.<id>.stdout` file in the attempt directory.

crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -207,6 +231,180 @@ impl ApptainerState {
...
+            }
+        });
+
+        let child_result = apptainer_pull_child
+            .wait()
+            .await
+            .context("failed to wait for `apptainer` to exit")
+            // Permanently error if something goes wrong trying to wait for the child process
+            .map_err(RetryError::permanent)?;
+
```
claymcleod, line 368: I know you probably didn't introduce this, but it occurs to me we should wait for the spawned stdout and stderr watchers to join here before checking the condition `is_permanent`

claymcleod (approved):

PR #565, opened 2026-01-27
feat(sprocket test): output assertions

a-frantz:

New user guide for these features - https://github.com/stjude-rust-labs/sprocket.bio/pull/29

There is a strange `serde` bug which causes the YAML syntax used in this PR to deviate from the syntax used in the `serde_yaml_ng` docs. See this commit for details https://github.com/stjude-rust-labs/sprocket/pull/565/changes/9fbc9644c7e2e7a759afa77cba7740b1646f1c71 that link to https://github.com/acatton/serde-yaml-ng/issues/14 which links to more information

This is a bit of a "happy accident" bug IMO, as I prefer the alternate syntax used here as opposed to the YAML tag syntax the documentation uses.

I've started adding these to https://github.com/stjudecloud/workflows/pull/263

For external contributors:

For all contributors:

      is implemented.


src/test/assertions.rs
```diff
@@ -0,0 +1,976 @@
...
+                warn!("ignoring `should_fail` assertion for task");
+            }
+
+            let stdout_regexs = self
+                .stdout
+                .iter()
+                .map(|re| Regex::new(re).with_context(|| format!("compiling user regex: `{re}`")))
+                .collect::<Result<Vec<_>>>()?;
+            let stderr_regexs = self
+                .stdout
```
claymcleod: This should iterate over `self.stderr`, not `self.stdout`. Currently stderr assertions will silently use stdout patterns.

```suggestion
                .stderr
```

src/test/assertions.rs
```diff
@@ -0,0 +1,976 @@
...
+            }
+            Self::IntEquals(should_equal) => {
+                let o = output.as_integer().expect("type should be validated");
+                if *should_equal != o {
+                    bail!("output `{o}` does not equal assertion `{should_equal}`")
+                }
+            }
+            Self::FloatEquals(should_equal) => {
+                let o = output.as_float().expect("type should be validated");
+                if *should_equal != o {
```
claymcleod: Direct floating-point equality comparison may cause false failures due to precision issues (e.g., `0.1 + 0.2 != 0.3`). Consider epsilon-based comparison:

```suggestion
                if (*should_equal - o).abs() > f64::EPSILON * 100.0 {
```

src/commands/test.rs
```diff
@@ -186,6 +188,22 @@ fn file_matches<'a>(path: &str, regexs: &'a [Regex]) -> Result<Option<&'a str>>
...
+    assertions: &HashMap<String, Vec<OutputAssertion>>,
+    outputs: &wdl::engine::Outputs,
+) -> Result<()> {
+    for (name, fns) in assertions {
+        let output = outputs
+            .get(name)
+            .expect("output should have been validated");
+        for func in fns {
+            func.evaluate(output)
+                .with_context(|| format!("evalutating WDL output with name `{name}`"))?
```
claymcleod:
```suggestion
                .with_context(|| format!("evaluating WDL output with name `{name}`"))?
```

src/test/assertions.rs
```diff
@@ -0,0 +1,976 @@
...
+    /// Is the WDL `Boolean` equal to this?
+    BoolEquals(bool),
+    /// Is the WDL `String` equal to this?
+    // TODO(Ari): compile this as an RE?
+    StrEquals(String),
+    /// Is the WDL `Int` equal to this?
+    IntEquals(i64),
+    /// Is the WDL `Float` equal to this?
+    FloatEquals(f64),
+    /// Does the WDL `String` contiain this substring?
```
claymcleod:
```suggestion
    /// Does the WDL `String` contain this substring?
```

src/commands/test.rs
```diff
@@ -214,36 +232,46 @@ impl TestIteration {
...
                 }
-                Err(_eval_err) => {
+                Err(eval_err) => {
                     if self.assertions.should_fail {
                         Ok(IterationResult::Success)
                     } else {
                         Ok(IterationResult::Fail(anyhow!(
                             "test iteration #{num} of `{name}` failed but was expected to \
-                             succeed: see `{dir}`",
+                             succeed; see `{dir}`: {err}",
```
a-frantz:
```suggestion
                             succeed: see `{dir}`: {err}",
```

a-frantz:
my latest commit (https://github.com/stjude-rust-labs/sprocket/pull/565/changes/121fa35db7837bb65af7eca7124828a80fb35a21) shows how easy it is to extend this architecture with more assertions. The hardest part of this commit I just pushed was the tests!

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added "output assertions" `sprocket dev test` for asserting on WDL
```
adthrasher:
```suggestion
* Added "output assertions" to `sprocket dev test` for asserting on WDL
```

src/config.rs
```diff
@@ -288,6 +290,23 @@ impl ServerConfig {
     }
 }
 
+/// `test` command configuration.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct TestConfig {
+    /// Number of test executions to run in parallel
```
adthrasher: Will this be needed long-term? I don't recall all of the details of the conversation yesterday, but I thought we were moving to Docker Swarm (for local testing) in order to let Docker handle the parallelism and scheduling.

src/commands/test.rs
```diff
@@ -214,36 +239,46 @@ impl TestIteration {
                             name = self.name,
                             dir = self.run_dir.display(),
                         )))
+                    } else if let Err(e) = evaluate_outputs(&self.assertions.outputs, &outputs)
+                        .with_context(|| {
+                            format!(
+                                "test iteration #{num} of `{name}` failed output assertions: see \
```
adthrasher: I triggered a few of these messages for the workflows repo. I'm not sure that the iteration # is necessarily helpful. I don't have a great suggestion for alternative language. To me, the test name and the directory are the most important pieces of information, since the iterations are being generated automatically by Sprocket based on the YAML definitions.

adthrasher: Maybe something like:
```suggestion
                                "test `{name}` (iteration {num}) failed output assertions: see \
```

src/config.rs
a-frantz:
```
     Running `target/debug/sprocket dev test -w ../workflows/ ../workflows/ -p 999999`
error: removing prior test dir: `/Users/afrantz/workflows/test/runs/bam_to_fastq/simpler`

Caused by:
    Too many open files (os error 24)
```
at least until we refactor to share the `Evaluator` between executions, but it might still need to be throttled after that as well? I have not addressed the TODO comment about messing with this number to find out what the right throttle level is

src/test/assertions.rs
```diff
@@ -0,0 +1,1100 @@
...
+            }
+            Self::IntEquals(should_equal) => {
+                let o = output.as_integer().expect("type should be validated");
+                if *should_equal != o {
+                    bail!("output `{o}` does not equal assertion `{should_equal}`")
+                }
+            }
+            Self::FloatEquals(should_equal) => {
+                let o = output.as_float().expect("type should be validated");
+                if (*should_equal - o).abs() > f64::EPSILON * 100.0 {
```
claymcleod: The tolerance here (`f64::EPSILON * 100.0` ≈ 2.22e-14) is effectively an exact-equality check. Consider relaxing to something like `1e-3`—for a test assertion, users probably care that the value is "close enough" rather than near bit-identical.

src/test/assertions.rs
```diff
@@ -0,0 +1,1100 @@
...
+                        ) {
+                            valid = true;
+                        }
+                    }
+                }
+                if !valid {
+                    bail!("`{self}` assertion cannot be used on `{prim_ty}` WDL type")
+                } else {
+                    Ok(())
+                }
```
claymcleod, line 189: The `let mut valid = false` / set-in-arm / check-after pattern could be replaced by matching on `(self, prim_ty)` directly, which makes the valid combinations self-documenting:

```rust
match (self, prim_ty) {
    (Self::BoolEquals(_), PrimitiveType::Boolean)
    | (Self::Name(_), PrimitiveType::Directory | PrimitiveType::File)
    | (Self::FloatEquals(_), PrimitiveType::Float)
    | (Self::IntEquals(_), PrimitiveType::Integer)
    | (Self::Contains(_) | Self::StrEquals(_) | Self::Length(_) | Self::Empty(_),
       PrimitiveType::String) => Ok(()),
    _ => bail!("`{self}` assertion cannot be used on `{prim_ty}` WDL type"),
}
```

The same pattern applies to the compound type block below.

src/commands/test.rs
```diff
@@ -454,7 +554,8 @@ async fn process_tests(
     root: &Path,
     clean: bool,
     errors: &mut Vec<Arc<anyhow::Error>>,
-) -> Result<()> {
+) -> Result<BTreeMap<String, (usize, usize, usize)>> {
```
claymcleod, line 538: The `IndexMap<String, IndexMap<String, IndexMap<String, JoinSet<TestIteration>>>>` type is hard to parse at a glance. A couple of type aliases would help:

```rust
type TargetResults = IndexMap<String, JoinSet<TestIteration>>;
type DocumentResults = IndexMap<String, TargetResults>;
```

src/commands/test.rs
```diff
@@ -202,53 +229,63 @@ struct TestIteration {
...
                         )))
                     }
                 }
             },
-            RunResult::Task(result) => match &**result {
+            RunResult::Task(result) => match *result {
                 Ok(evaled_task) => {
                     if evaled_task.exit_code() == self.assertions.exit_code {
-                        if !self.assertions.stdout.is_empty() {
+                        if let Some(regexs) = &self.assertions.stdout {
```
claymcleod:
```suggestion
                        if let Some(regexes) = &self.assertions.stdout {
```

src/commands/test.rs
```diff
@@ -257,20 +294,17 @@ impl TestIteration {
                                 Err(e) => return Err(e),
                             }
                         }
-                        if !self.assertions.stderr.is_empty() {
+                        if let Some(regexs) = &self.assertions.stderr {
```
claymcleod:
```suggestion
                        if let Some(regexes) = &self.assertions.stderr {
```

src/commands/test.rs
claymcleod: This should be line 553.

.gitignore
```diff
@@ -6,6 +6,7 @@ target/
 rustc-ice*
 
 out/
+runs/
```
claymcleod: I'm not sure we need this anymore now that `run` has been updated to use the `out` directory.

src/test/assertions.rs
claymcleod: This should be up a few lines.

src/test/assertions.rs
```diff
@@ -0,0 +1,1100 @@
...
+            stderr,
+            outputs: self.outputs,
+            custom: self.custom,
+        })
+    }
+}
+
+/// Possible assertions on a WDL output.
+#[derive(Debug, serde::Deserialize)]
+pub(crate) enum OutputAssertion {
```
claymcleod, line 111: Do we want to consider making these lowercase to match the conventions of the rest of our config-like files? I would advocate for `snake_case` here.

PR #589, opened 2026-02-03
fix(ExpectedRuntimeKeys): Allow excepting specific runtime items

### LINKED ISSUE #538: Allow specific additional runtime keys.

BKDaugherty:
Because we add each unrecognized key to a collection, and then emit one error in the `Exit` processing of the task_definition, we weren't able to take advantage of `exceptable_add` to allow users to "except" specific runtime items. Let's instead process each unrecognized key individually, and add a diagnostic per instance as we process them.

As discussed in the [issue](https://github.com/stjude-rust-labs/sprocket/issues/538), (thanks @a-frantz!), we've opted for emitting each "non_reserved" runtime item as a separate lint for simplicity. 

Note that this is a breaking change in the way in which our lints are presented, as shown in the `source.errors` changes. Notably...
- We now emit a `Diagnostic` for each "non-reserved runtime item key" instead of one for the whole group
- We will emit `Diagnostic`s for a second runtime section in a task_definition, even though the second definition is invalid (where the previous implementation tracked this explicitly).

Closes #538.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-lint/src/rules/expected_runtime_keys.rs
+++ b/crates/wdl-lint/src/rules/expected_runtime_keys.rs
@@ -199,6 +199,9 @@ fn report_missing_recommended_keys(
 pub struct ExpectedRuntimeKeysRule {
     /// The detected version of the current document.
     version: Option<SupportedVersion>,
+    /// Whether or not we've already processed a `runtime` section within the
+    /// current task.
+    runtime_processed_for_task: bool,
     /// All keys encountered in the current runtime section.
     encountered_keys: Vec<Ident>,
     /// Allowed keys from the config.
@@ -210,6 +213,7 @@ impl ExpectedRuntimeKeysRule {
     pub fn new(config: &Config) -> Self {
         Self {
             version: None,
+            runtime_processed_for_task: false,
             encountered_keys: Vec::new(),
             allowed_runtime_keys: config.allowed_runtime_keys.clone(),
         }
@@ -292,12 +296,31 @@ impl Visitor for ExpectedRuntimeKeysRule {
         self.version = Some(version);
     }
 
+    fn task_definition(
+        &mut self,
+        _: &mut Diagnostics,
+        reason: VisitReason,
+        _: &wdl_ast::v1::TaskDefinition,
+    ) {
+        if reason == VisitReason::Exit {
+            self.runtime_processed_for_task = false;
+        }
+    }
+
     fn runtime_section(
         &mut self,
         diagnostics: &mut Diagnostics,
         reason: VisitReason,
         section: &RuntimeSection,
     ) {
+        // NOTE: if we've already processed a `runtime` section for this task
+        // and we hit this again, that means there are multiple `runtime`
+        // sections in the task. In that case, validation should report that
+        // this cannot occur, and the runtime section should be ignored.
+        if self.runtime_processed_for_task {
+            return;
+        }
+
         match reason {
             VisitReason::Enter => {}
             VisitReason::Exit => {
@@ -341,6 +364,7 @@ impl Visitor for ExpectedRuntimeKeysRule {
                     // clear our tracking container of encountered keys to prepare for the next
                     // runtime section.
                     self.encountered_keys.clear();
+                    self.runtime_processed_for_task = true;
                 }
             }
         }
@@ -352,7 +376,11 @@ impl Visitor for ExpectedRuntimeKeysRule {
         reason: VisitReason,
         item: &RuntimeItem,
     ) {
-        if reason == VisitReason::Exit {
+        // NOTE: if we've already processed a `runtime` section for this task
+        // and we hit this again, that means there are multiple `runtime`
+        // sections in the task. In that case, validation should report that
+        // this cannot occur, and the runtime items should be ignored.
+        if self.runtime_processed_for_task || reason == VisitReason::Exit {
             return;
         }
 
--- a/crates/wdl-lint/tests/lints/runtime-keys-multiple-runtime-sections/source.errors
+++ b/crates/wdl-lint/tests/lints/runtime-keys-multiple-runtime-sections/source.errors
@@ -22,14 +22,6 @@ warning[ExpectedRuntimeKeys]: the runtime key `baz` is not reserved in the WDL 1
    │
    = fix: remove the `baz` key
 
-note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1.1 specification: `container`
-   ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:19:5
-   │
-19 │     runtime {
-   │     ^^^^^^^
-   │
-   = fix: include an entry for the `container` key in the `runtime` section
-
 error: task `a_task_with_multiple_runtimes` contains a duplicate runtime section
    ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:19:5
    │
@@ -38,20 +30,4 @@ error: task `a_task_with_multiple_runtimes` contains a duplicate runtime section
    ·
 19 │     runtime {
    │     ^^^^^^^ this runtime section is a duplicate
-
-warning[ExpectedRuntimeKeys]: the runtime key `foo` is not reserved in the WDL 1.1 specification; arbitrary runtime keys are deprecated
-   ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:20:9
-   │
-20 │         foo: "bar"  # these items should not be processed and flagged.
-   │         ^^^
-   │
-   = fix: remove the `foo` key
-
-warning[ExpectedRuntimeKeys]: the runtime key `baz` is not reserved in the WDL 1.1 specification; arbitrary runtime keys are deprecated
-   ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:21:9
-   │
-21 │         baz: "quux"
-   │         ^^^
-   │
-   = fix: remove the `baz` key
 


crates/wdl-lint/src/rules/expected_runtime_keys.rs
```diff
@@ -433,51 +323,30 @@ impl Visitor for ExpectedRuntimeKeysRule {
...
+                                // behavior.
+                                //
+                                // TODO: Use section.span(), assuming that's an acceptable public
+                                // API change.
+                                section
+                                    .inner()
+                                    .first_token()
+                                    .expect("runtime section should have tokens")
+                                    .text_range()
+                                    .into(),
```
BKDaugherty: @a-frantz - not sure what you think we should do here? The previous implementation used this to create the span for the `runtime_section`. In the spirit of removing code, we could use `section.span()`? Although that would change the look of the lint. 

Given we are opting for a breaking change in this PR already, maybe good to do so here too? Happy to do whatever makes most sense :) 

For example, if I make this change...

```diff
brendon@Brendons-MacBook-Pro sprocket % git diff
diff --git a/crates/wdl-lint/src/rules/expected_runtime_keys.rs b/crates/wdl-lint/src/rules/expected_runtime_keys.rs
index 4a441f0f..22e640c1 100644
--- a/crates/wdl-lint/src/rules/expected_runtime_keys.rs
+++ b/crates/wdl-lint/src/rules/expected_runtime_keys.rs
@@ -323,18 +323,7 @@ impl Visitor for ExpectedRuntimeKeysRule {
                         diagnostics.exceptable_add(
                             report_missing_recommended_keys(
                                 missing_keys,
-                                // Note that we don't use `section.span()`, as this would be a
-                                // breaking change in `wdl-lint`'s
-                                // behavior.
-                                //
-                                // TODO: Use section.span(), assuming that's an acceptable public
-                                // API change.
-                                section
-                                    .inner()
-                                    .first_token()
-                                    .expect("runtime section should have tokens")
-                                    .text_range()
-                                    .into(),
+                                section.span(),
                                 &specification,
                             ),
                             SyntaxElement::from(section.inner().clone()),
brendon@Brendons-MacBook-Pro sprocket % BLESS=1 cargo test -p wdl-lint --all-features
   Compiling wdl-lint v0.19.0 (/Users/brendon/Dev/sprocket/crates/wdl-lint)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 2.11s
     Running unittests src/lib.rs (target/debug/deps/wdl_lint-58da928dcc65c055)
```

Then I get this output change in the affected set of `source.errors` files

```diff
diff --git a/crates/wdl-lint/tests/lints/missing-runtime-block/source.errors b/crates/wdl-lint/tests/lints/missing-runtime-block/source.errors
index 5f84eba0..4dfca0e6 100644
--- a/crates/wdl-lint/tests/lints/missing-runtime-block/source.errors
+++ b/crates/wdl-lint/tests/lints/missing-runtime-block/source.errors
@@ -8,9 +8,10 @@ warning[RuntimeSection]: task `bad` is missing a `runtime` section

 note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1.1 specification: `container`
    ┌─ tests/lints/missing-runtime-block/source.wdl:22:5
-   │
-22 │     runtime {
-   │     ^^^^^^^
-   │
+   │
+22 │ ╭     runtime {
+23 │ │     }
+   │ ╰─────^
+   │
    = fix: include an entry for the `container` key in the `runtime` section

diff --git a/crates/wdl-lint/tests/lints/runtime-keys-multiple-runtime-sections/source.errors b/crates/wdl-lint/tests/lints/runtime-keys-multiple-runtime-sections/source.errors
index f316f335..f72acea4 100644
--- a/crates/wdl-lint/tests/lints/runtime-keys-multiple-runtime-sections/source.errors
+++ b/crates/wdl-lint/tests/lints/runtime-keys-multiple-runtime-sections/source.errors
@@ -1,9 +1,12 @@
 note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1.1 specification: `container`
    ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:12:5
-   │
-12 │     runtime {
-   │     ^^^^^^^
-   │
+   │
+12 │ ╭     runtime {
+13 │ │         foo: "bar"  # these items should be processed and flagged.
+14 │ │         baz: "quux"
+15 │ │     }
+   │ ╰─────^
+   │
    = fix: include an entry for the `container` key in the `runtime` section

 warning[ExpectedRuntimeKeys]: the runtime key `foo` is not reserved in the WDL 1.1 specification; therefore, its inclusion in the `runtime` section is deprecated
@@ -22,14 +25,6 @@ warning[ExpectedRuntimeKeys]: the runtime key `baz` is not reserved in the WDL 1
    │
    = fix: if a reserved key name was intended, correct the spelling; otherwise, remove the `baz` key

-note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1.1 specification: `container`
-   ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:19:5
-   │
-19 │     runtime {
-   │     ^^^^^^^
-   │
-   = fix: include an entry for the `container` key in the `runtime` section
-
 error: task `a_task_with_multiple_runtimes` contains a duplicate runtime section
    ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:19:5
    │
@@ -39,6 +34,17 @@ error: task `a_task_with_multiple_runtimes` contains a duplicate runtime section
 19 │     runtime {
    │     ^^^^^^^ this runtime section is a duplicate

+note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1.1 specification: `container`
+   ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:19:5
+   │
+19 │ ╭     runtime {
+20 │ │         foo: "bar"  # these items should not be processed and flagged.
+21 │ │         baz: "quux"
+22 │ │     }
+   │ ╰─────^
+   │
+   = fix: include an entry for the `container` key in the `runtime` section
+
 warning[ExpectedRuntimeKeys]: the runtime key `foo` is not reserved in the WDL 1.1 specification; therefore, its inclusion in the `runtime` section is deprecated
    ┌─ tests/lints/runtime-keys-multiple-runtime-sections/source.wdl:20:9
    │
diff --git a/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.0/source.errors b/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.0/source.errors
index 14cdefcb..9cb0965a 100644
--- a/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.0/source.errors
+++ b/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.0/source.errors
@@ -2,16 +2,18 @@ note[ExpectedRuntimeKeys]: the following runtime keys are recommended by the WDL
    ┌─ tests/lints/runtime-keys-wdl-1.0/source.wdl:12:5
    │
 12 │     runtime {}  # Two missing keys: "docker" and "memory"
-   │     ^^^^^^^
+   │     ^^^^^^^^^^
    │
    = fix: include entries for the `docker` and `memory` keys in the `runtime` section

 note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1.0 specification: `memory`
    ┌─ tests/lints/runtime-keys-wdl-1.0/source.wdl:33:5
-   │
-33 │     runtime {
-   │     ^^^^^^^
-   │
+   │
+33 │ ╭     runtime {
+34 │ │         docker: "foo"
+35 │ │     }
+   │ ╰─────^
+   │
    = fix: include an entry for the `memory` key in the `runtime` section

 warning[ContainerUri]: container URI is missing a tag
@@ -24,10 +26,12 @@ warning[ContainerUri]: container URI is missing a tag

 note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1.0 specification: `docker`
    ┌─ tests/lints/runtime-keys-wdl-1.0/source.wdl:45:5
-   │
-45 │     runtime {
-   │     ^^^^^^^
-   │
+   │
+45 │ ╭     runtime {
+46 │ │         memory: "foo"
+47 │ │     }
+   │ ╰─────^
+   │
    = fix: include an entry for the `docker` key in the `runtime` section

 warning[ContainerUri]: container URI is missing a tag
diff --git a/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors b/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors
index c214cf20..dd2b6113 100644
--- a/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors
+++ b/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors
@@ -2,7 +2,7 @@ note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1
    ┌─ tests/lints/runtime-keys-wdl-1.1/source.wdl:12:5
    │
 12 │     runtime {}  # Missing `container` key
-   │     ^^^^^^^
+   │     ^^^^^^^^^^
    │
    = fix: include an entry for the `container` key in the `runtime` section

diff --git a/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors.default b/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors.default
index 929b98d2..59ca218d 100644
--- a/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors.default
+++ b/crates/wdl-lint/tests/lints/runtime-keys-wdl-1.1/source.errors.default
@@ -2,7 +2,7 @@ note[ExpectedRuntimeKeys]: the following runtime key is recommended by the WDL 1
    ┌─ tests/lints/runtime-keys-wdl-1.1/source.wdl:12:5
    │
 12 │     runtime {}  # Missing `container` key
-   │     ^^^^^^^
+   │     ^^^^^^^^^^
    │
    = fix: include an entry for the `container` key in the `runtime` section
```

a-frantz: We generally don't care about breaking changes (for now and the short term future) in this repo as all our crates are pre-v1. But I wouldn't want to make this change you're showing here because in IDEs it highlights the entire runtime section with "squigglies", which can look overwhelming.

A brief bit of historical context: a very long time ago, early in this project's life, we got feedback from early testers that they would open an existing WDL file in their IDE and the whole file would light up with highlights from various lint rules. This feedback led to us overhauling all our existing rules to be much more conservative in the spans they report on, preferring to just highlight keywords (like `runtime`) instead of reporting the entire span of the problem region.

So with that lesson learned long ago, we should keep the reported span to only being the `runtime` keyword like you already have it in this PR! So you can remove this TODO comment; I think the currently reported span is correct 💯

crates/wdl-lint/tests/lints/runtime-keys-engine-keys/source.errors
```diff
@@ -19,3 +6,19 @@ warning[ContainerUri]: container URI is missing a tag
    │
    = fix: add a tag to the container URI (e.g., `ubuntu@sha256:foobar` instead of `ubuntu`)
 
+warning[ExpectedRuntimeKeys]: the runtime key `cromwell` is not reserved in the WDL 1.1 specification; therefore, its inclusion in the `runtime` section is deprecated
+   ┌─ tests/lints/runtime-keys-engine-keys/source.wdl:20:9
+   │
+20 │         cromwell: {}
+   │         ^^^^^^^^^^^^
```
a-frantz: small nit: can you change the reported span so it only highlights the key (`cromwell` in this case) instead of the whole runtime object? Same reason as I outlined in the other comment!

crates/wdl-lint/src/rules/expected_runtime_keys.rs
```diff
@@ -143,80 +141,18 @@ fn deprecated_runtime_key(key: &Ident, replacement: &str) -> Diagnostic {
...
-            key.span(),
-        );
-    }
-
-    Some(diagnostic)
+/// Creates a "non-reserved runtime key" diagnostic for a specific `key`
+fn report_non_reserved_runtime_key(key: &str, span: Span, specification: &str) -> Diagnostic {
+    Diagnostic::warning(format!(
+        "the runtime key `{key}` is not reserved in {specification}; therefore, its inclusion in \
+         the `runtime` section is deprecated"
```
a-frantz: while we're in this part of the code making changes, I think these diagnostics are too wordy and could be shortened
```suggestion
        "the runtime key `{key}` is not reserved in {specification}; arbitrary runtime keys are deprecated"
```

crates/wdl-lint/src/rules/expected_runtime_keys.rs
```diff
@@ -143,80 +141,18 @@ fn deprecated_runtime_key(key: &Ident, replacement: &str) -> Diagnostic {
...
+    ))
+    .with_rule(ID)
+    .with_highlight(span)
+    .with_fix(format!(
+        "if a reserved key name was intended, correct the spelling; otherwise, remove the `{key}` \
+         key"
```
a-frantz:
```suggestion
        "remove the `{key}` key"
```

crates/wdl-lint/tests/lints/runtime-keys-engine-keys/source.errors
BKDaugherty: Yee absolutely!

crates/wdl-lint/src/rules/expected_runtime_keys.rs
BKDaugherty: Thanks for the context!

crates/wdl-lint/src/rules/expected_runtime_keys.rs
```diff
@@ -433,51 +321,27 @@ impl Visitor for ExpectedRuntimeKeysRule {
...
-        reason: VisitReason,
-        section: &RuntimeSection,
-    ) {
-        // NOTE: if we've already processed a `runtime` section for this task
-        // and we hit this again, that means there are multiple `runtime`
-        // sections in the task. In that case, validation should report that
-        // this cannot occur, and the runtime section should be ignored.
-        if self.runtime_processed_for_task {
-            return;
-        }
```
peterhuene, line 460: Generally we've taken the approach on the lints to ignore invalid duplicate sections as we're already going to be emitting an _error_ diagnostic for the duplicate section in its entirety, so generating additional lints within it isn't ideal.

Could we add back the `task_definition` visitation such that we keep track if we've seen the runtime section before and reset it when we exit the task definition?

crates/wdl-lint/src/rules/expected_runtime_keys.rs
```diff
@@ -488,11 +352,7 @@ impl Visitor for ExpectedRuntimeKeysRule {
         reason: VisitReason,
         item: &RuntimeItem,
     ) {
-        // NOTE: if we've already processed a `runtime` section for this task
-        // and we hit this again, that means there are multiple `runtime`
-        // sections in the task. In that case, validation should report that
-        // this cannot occur, and the runtime items should be ignored.
-        if self.runtime_processed_for_task || reason == VisitReason::Exit {
+        if reason == VisitReason::Exit {
```
peterhuene: Same comment as above regarding this removed check for duplicate runtime sections.

PR #590, opened 2026-02-04
feat(wdl-lint): Add UnusedDocCommentsRule

### LINKED ISSUE #570: Lint for misplaced doc comments
With #551 bringing support for doc comments, we should lint doc comments that are used in invalid contexts.

Right now, doc comments are supported on:

* `struct`s
* `struct` members
* `enum`s
* `enum` variants

It wouldn't make sense to annotate, say, a `meta` section with doc comments:

```wdl
workflow test {
    ## This does nothing!
    meta {}

    output {}
}
```

Doc comments used outside of contexts that `wdl-doc` understands should produce a warning saying that the comment will not generate documentation.

**Additional context**
This is pretty much the `unused_doc_comments` lint in Rust, which looks like:

```rust
fn main() {
    /// foo
    println!("Hello, world!");
}

warning: unused doc comment
 --> src/main.rs:2:5
  |
2 |     /// foo
  |     ^^^^^^^ rustdoc does not generate documentation for macro invocations
  |
  = help: to document an item produced by a macro, the macro must produce the documentation as part of its expansion
  = note: `#[warn(unused_doc_comments)]` (part of `#[warn(unused)]`) on by default
```

BKDaugherty:
As discussed in [this issue](https://github.com/stjude-rust-labs/sprocket/issues/570]), this PR adds a lint rule for doc comments that are placed "on" syntax nodes that aren't supported.

Fixes #570.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-lint/tests/lints/unused-doc-comments/source.errors
+++ b/crates/wdl-lint/tests/lints/unused-doc-comments/source.errors
@@ -8,7 +8,7 @@ note[UnusedDocComments]: unused doc comment
 43 │       command <<<
    │       ------- documentation will not be generated for this item
    │  
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
    ┌─ tests/lints/unused-doc-comments/source.wdl:39:5
@@ -21,7 +21,7 @@ note[UnusedDocComments]: unused doc comment
 43 │       command <<<
    │       ------- documentation will not be generated for this item
    │  
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
    ┌─ tests/lints/unused-doc-comments/source.wdl:51:5
@@ -34,7 +34,7 @@ note[UnusedDocComments]: unused doc comment
 64 │       command <<<>>>
    │       ------- documentation will not be generated for this item
    │  
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
    ┌─ tests/lints/unused-doc-comments/source.wdl:56:5
@@ -50,7 +50,7 @@ note[UnusedDocComments]: unused doc comment
 64 │       command <<<>>>
    │       ------- documentation will not be generated for this item
    │  
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[ElementSpacing]: extra blank line(s) found
    ┌─ tests/lints/unused-doc-comments/source.wdl:62:47
@@ -71,7 +71,7 @@ note[UnusedDocComments]: unused doc comment
 75 │     meta {
    │     ---- documentation will not be generated for this item
    │
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
    ┌─ tests/lints/unused-doc-comments/source.wdl:79:5
@@ -81,7 +81,7 @@ note[UnusedDocComments]: unused doc comment
 80 │     input {
    │     ----- documentation will not be generated for this item
    │
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
    ┌─ tests/lints/unused-doc-comments/source.wdl:83:22
@@ -91,7 +91,7 @@ note[UnusedDocComments]: unused doc comment
    │         │             
    │         documentation will not be generated for this item
    │
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
    ┌─ tests/lints/unused-doc-comments/source.wdl:84:23
@@ -101,7 +101,7 @@ note[UnusedDocComments]: unused doc comment
    │         │              
    │         documentation will not be generated for this item
    │
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
    ┌─ tests/lints/unused-doc-comments/source.wdl:89:5
@@ -111,7 +111,7 @@ note[UnusedDocComments]: unused doc comment
 90 │     call test_task {}
    │     ---- documentation will not be generated for this item
    │
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
    ┌─ tests/lints/unused-doc-comments/source.wdl:92:5
@@ -122,7 +122,7 @@ note[UnusedDocComments]: unused doc comment
 94 │     Person p = Person {
    │     ------ documentation will not be generated for this item
    │
-   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+   = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
     ┌─ tests/lints/unused-doc-comments/source.wdl:99:5
@@ -132,7 +132,7 @@ note[UnusedDocComments]: unused doc comment
 100 │     if (apple) {
     │     -- documentation will not be generated for this item
     │
-    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
     ┌─ tests/lints/unused-doc-comments/source.wdl:104:5
@@ -142,7 +142,7 @@ note[UnusedDocComments]: unused doc comment
 105 │     else if (banana) {
     │     ---- documentation will not be generated for this item
     │
-    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
     ┌─ tests/lints/unused-doc-comments/source.wdl:106:42
@@ -152,7 +152,7 @@ note[UnusedDocComments]: unused doc comment
     │         │                                 
     │         documentation will not be generated for this item
     │
-    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
     ┌─ tests/lints/unused-doc-comments/source.wdl:109:5
@@ -162,7 +162,7 @@ note[UnusedDocComments]: unused doc comment
 110 │     else {
     │     ---- documentation will not be generated for this item
     │
-    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
     ┌─ tests/lints/unused-doc-comments/source.wdl:111:9
@@ -172,7 +172,7 @@ note[UnusedDocComments]: unused doc comment
 112 │         String favorite_fruit = "Chocolate"
     │         ------ documentation will not be generated for this item
     │
-    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
     ┌─ tests/lints/unused-doc-comments/source.wdl:115:5
@@ -182,7 +182,7 @@ note[UnusedDocComments]: unused doc comment
 116 │     output {
     │     ------ documentation will not be generated for this item
     │
-    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 
 note[UnusedDocComments]: unused doc comment
     ┌─ tests/lints/unused-doc-comments/source.wdl:121:1
@@ -190,5 +190,5 @@ note[UnusedDocComments]: unused doc comment
 121 │ ## Trailing Doc Comment should be linted
     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     │
-    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element
+    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it documents the intended item
 


crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,194 @@
...
+    .with_fix("If you're intending to use a regular comment here, replace the `##` with `#`")
+}
+
+/// Detects whether a doc comment has been placed atop a Node that we do not
+/// generate documentation for.
+#[derive(Default, Debug, Clone)]
+pub struct UnusedDocCommentsRule {
+    // TODO: This seems wasteful? Is there a better unique identifier to use
+    // associated with a comment token in the AST so I can ensure that I don't
+    // revisit?
```
BKDaugherty: @Serial-ATA - can you think of a better way to track the `seen_comments` here? Is there some unique identifier I can use for an AST element so I know not to double process a "doc comment block"?

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,194 @@
...
+                continue;
+            }
+
+            // TODO: This feels pretty jank. Is there a better way to target a field of a
+            // struct?
+            let struct_definition_element = sibling.kind() == SyntaxKind::UnboundDeclNode
+                && sibling
+                    .parent()
+                    .map(|t| t.kind() == SyntaxKind::StructDefinitionNode)
+                    .unwrap_or_default();
```
BKDaugherty: @Serial-ATA - It looks like there isn't a SyntaxKind corresponding directly to the field of a Struct, so i came up with this. For `wdl-doc` do you have a good way of identifying this? Is there something less jank I could use?

crates/wdl-lint/tests/lints/preamble-comment-after-version/source.errors
```diff
@@ -28,3 +28,13 @@ note[PreambleCommentPlacement]: preamble comment after the version statement
    = help: freestanding doc comments (`##`) are assumed to be preamble comments
    = fix: move the preamble comment before the `version` statement
 
+note[UnusedDocComments]: wdl-doc does not generate documentation for comment
+   ┌─ tests/lints/preamble-comment-after-version/source.wdl:30:5
+   │  
+30 │ ╭     ## This one is bad!
+31 │ │     #@ except: MetaDescription
```
BKDaugherty: @Serial-ATA - is there a design for how doc comments and lint directives should interact? I can make the doc comment ignore the lint directive fairly easily, but I'm not sure if there's guidance that's been given on the ordering similar to the [guidance provided for the preamble w.r.t lint directives?](https://github.com/stjude-rust-labs/sprocket/blob/5c6801bed7a0974ff63ffe35811269de809fba22/crates/wdl-lint/src/rules/preamble_formatted.rs#L128-L129)?

crates/wdl-lint/src/rules/unused_doc_comments.rs
a-frantz: I'm about to turn off notifications for the night and haven't reviewed your code, but I did see this comment/TODO and have a potentially useful lead. Check out https://github.com/stjude-rust-labs/sprocket/blob/main/crates/wdl-lint/src/rules/preamble_comment_placement.rs

It relies on the fact that "consecutive comments" like this are only ever separated by whitespace tokens. So if any other token or node is visited that isn't a comment or whitespace, we know we've reached the end of this block of consecutive comments.

Hope this helps! I can take a closer look some point tomorrow :)

crates/wdl-lint/tests/lints/preamble-comment-after-version/source.errors
a-frantz: IMO it looks _much_ cleaner to have doc comments come before lint directives. So
```wdl
## description
##
## long help text
#@ except: directives!
workflow foo {
}
```
I think this rule your working on should just ignore directives.

(Open to discussion if anyone disagrees, but I feel that the above example is pretty clearly better than the alternative)

a-frantz: #592

crates/wdl-lint/src/rules/unused_doc_comments.rs
Serial-ATA: In `wdl-doc`, doc comments are read bottom to top starting at known valid tokens: https://github.com/stjude-rust-labs/sprocket/blob/5c6801bed7a0974ff63ffe35811269de809fba22/crates/wdl-doc/src/meta.rs#L416

The logic here *could* be flipped to do something similar. Instead of validating each commented node against a list of allowed kinds, the `Visitor` could check for doc comments attached to known *invalid* nodes. That way you wouldn't need to concern yourself with `struct`s at all.

Something like:

```rust
impl UnusedDocCommentsRule {
    fn lint_doc_comments(
        &mut self,
        diagnostics: &mut Diagnostics,
        kind: SyntaxKind,
        preceding_trivia: &mut impl Iterator<Item = SyntaxToken>,
    ) {
        let mut span: Option<Span> = None;
        let mut last_comment = None;
        for token in preceding_trivia {
            match token.kind() {
                SyntaxKind::Comment => {
                    if !token.text().starts_with(DOC_COMMENT_PREFIX) {
                        break;
                    }

                    span = span.map_or(Some(token.span()), |prev_span| {
                        Some(Span::new(token.span().start(), prev_span.end()))
                    });
                    last_comment = Some(token);
                }
                _ => break,
            }
        }

        if let (Some(span), Some(last_comment)) = (span, last_comment) {
            diagnostics.exceptable_add(
                unused_doc_comment_diagnostic(span, kind),
                SyntaxElement::from(last_comment),
                &self.exceptable_nodes(),
            );
        }
    }
}

impl Visitor for UnusedDocCommentsRule {
    // Top-level doc comments on meta sections are invalid
    fn metadata_section(
        &mut self,
        diagnostics: &mut Diagnostics,
        reason: VisitReason,
        section: &v1::MetadataSection,
    ) {
        if reason == VisitReason::Exit {
            return;
        }

        self.lint_doc_comments(
            diagnostics,
            section.kind(),
            &mut section.keyword().inner().preceding_trivia(),
        )
    }

    fn parameter_metadata_section(
        &mut self,
        diagnostics: &mut Diagnostics,
        reason: VisitReason,
        section: &v1::ParameterMetadataSection,
    ) {
        if reason == VisitReason::Exit {
            return;
        }

        self.lint_doc_comments(
            diagnostics,
            section.kind(),
            &mut section.keyword().inner().preceding_trivia(),
        )
    }

    /// and so on...
}
```

crates/wdl-lint/tests/lints/preamble-comment-after-version/source.errors
Serial-ATA: Yeah, exceptions after docs makes the most sense. That's typically what you see in Rust as well:

```rust
/// This
/// is
/// a
/// Description
#[allow(dead_code)]
pub struct Foo;
```

But this lint can just ignore any comments that start with `wdl_analysis::EXCEPT_COMMENT_PREFIX`

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Ah that makes sense! Thanks for the code pointer too! I'll take a look at how that feels

BKDaugherty: I pushed a commit that flips the logic to start from "known invalid nodes / tokens". 

It does seem to make the processing of each element a little simpler, but I wonder if this will be harder to maintain? It seems more straightforward to me to know that if you add support for documenting a specific syntax item, you also should ensure that item isn't linted. Whereas with this implementation if you add a syntax item, you need to opt in to linting for doc comments. 

Similarly, it seems like there will be value in having a list of `SyntaxKind` or similar identifying element for the set of supported items for doc comments. If this lint could consume this list, then it could be one of many things that don't need to be updated directly when a new node supports doc comments. (IE: [this lint would probably want this too?](https://github.com/stjude-rust-labs/sprocket/issues/592#issuecomment-3849377584). In the current implementation it's a little hard to generate a concrete set of invalid items and have this lint consume this, since it's baked into the implementation of `Visitor`.

For the `struct` specific stuff, I imagine we will need some special handling for `UnboundDecl` to ensure we lint on that if it's not in a struct vs. it is in a struct with this implementation too? I'm not so familiar with the AST though so still need to read more up on what an `UnboundDecl` is and where it's used though so forgive me if I'm misunderstanding :)

Seems like trade offs in both directions, happy to go whichever way y'all think is best! Let me know what you think!

BKDaugherty: It turned out to be the BoundDecl where I needed an extra check on context I think, assuming we want the ability to write doc comments for items of an input or output section.

```rust
fn bound_decl(
        &mut self,
        diagnostics: &mut Diagnostics,
        reason: VisitReason,
        decl: &v1::BoundDecl,
    ) {
        if reason == VisitReason::Exit {
            return;
        }

        if let Some(parent) = decl.inner().parent()
            && (parent.kind() == SyntaxKind::OutputSectionNode
                || parent.kind() == SyntaxKind::InputSectionNode)
        {
            return;
        }

        self.lint_doc_comments(
            diagnostics,
            decl.kind(),
            &mut decl
                .inner()
                .first_token()
                .expect("Must have at least one token")
                .preceding_trivia(),
        );
    }
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,374 @@
...
+        reason: VisitReason,
+        section: &v1::MetadataSection,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        let keyword = section
+            .token::<MetaKeyword>()
+            .expect("MetadataSection must have MetaKeyword");
```
Serial-ATA: You could move the keyword methods onto the AST types:

```rust
impl<N: TreeNode> MetadataSection<N> {
    /// Gets the `meta` keyword of the meta section.
    pub fn keyword(&self) -> MetaKeyword<N::Token> {
        self.token().expect("meta section should have a keyword")
    }
}
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,374 @@
...
+        reason: VisitReason,
+        section: &v1::CommandSection,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        let keyword = section
+            .token::<CommandKeyword>()
+            .expect("CommandSection must have CommandKeyword");
```
Serial-ATA: Same here

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,374 @@
...
+        reason: VisitReason,
+        section: &v1::ParameterMetadataSection,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        let keyword = section
+            .token::<ParameterMetaKeyword>()
+            .expect("ParameterMetadataSection must have ParameterMetaKeyword");
```
Serial-ATA: Same here

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,374 @@
...
+            return;
+        }
+
+        self.lint_doc_comments(
+            diagnostics,
+            stmt.kind(),
+            &mut stmt
+                .inner()
+                .first_token()
+                .expect("Call statement must have at least one token")
```
Serial-ATA: Same here

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,374 @@
...
+        section: &v1::RequirementsSection,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        let token = section
+            .inner()
+            .first_token()
+            .expect("RequirementsSection must have at least one token");
```
Serial-ATA: Same here

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,374 @@
...
+        stmt: &v1::ScatterStatement,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        let token = stmt
+            .inner()
+            .first_token()
+            .expect("Scatter statement must contain at least one token");
```
Serial-ATA: Same here

crates/wdl-lint/src/rules/unused_doc_comments.rs
Serial-ATA: > If this lint could consume this list, then it could be one of many things that don't need to be updated directly when a new node supports doc comments. (IE: https://github.com/stjude-rust-labs/sprocket/issues/592#issuecomment-3849377584.

I wonder if that should be a lint at all? In the case of this lint, documenting a `meta` section for example *is* wrong. There's nothing necessarily wrong about putting a lint exception at the top or bottom of a doc comment. I think that would be better suited for `wdl-format`.

As for other future lints, if it ever gets to a point where the separate implementations are hard to maintain we could just condense them into a single `Visitor`. For example, `Clippy` has ~20 doc comment lints crammed into a single visitor: https://github.com/rust-lang/rust-clippy/blob/1826ec0715d0a394d862d1c78eec6d95277da399/clippy_lints/src/doc/mod.rs#L750

a-frantz: > There's nothing necessarily wrong about putting a lint exception at the top or bottom of a doc comment. I think that would be better suited for wdl-format.

we do use `sprocket lint` to enforce some formatting decisions, which I understand is more opinionated of us than most other lint tooling. I think it makes sense to include as a lint, but we are also a lot stricter with what we enforce than tools in other contexts.

A touch of extra context: for a while now I've been chewing on the idea that `lint` should distinguish between lints that are just about formatting and more "meaningful" lints about clarity/semantics/etc. That is _kind of_ done with Tags and with the distinction between `note` and `warning` diagnostics, but I think a harder division of concerns could be good.

crates/wdl-lint/src/rules/unused_doc_comments.rs
a-frantz: yea that sounds like a good idea 👍

BKDaugherty: Yeah sounds good. I'll write a lil helper function for the "first_token" calls too.

I should put the "keyword" impls in wdl-ast right? And make an update to the changelog?

Serial-ATA: > I should put the "keyword" impls in wdl-ast right? And make an update to the changelog?

Yep, a few types already have `keyword()` methods (e.g. `EnumDefinition`):

https://github.com/stjude-rust-labs/sprocket/blob/91dc81feb128660d7f7e45aa5579c5a119510d67/crates/wdl-ast/src/v1/enum.rs#L26-L29

So they'd be nice to have for consistency.

BKDaugherty: Yeeee on it 🫡

crates/wdl-lint/src/rules/unused_doc_comments.rs
a-frantz: for transparency: we discussed this thread internally (really just the last 2 or 3 comments) which led to the writing of #600 . Looks like we are going the direction Alex suggested, where `lint` will not be concerned with formatting issues (specifically the last paragraph of 600 is relevant). Although timelines and priorities are still undecided, so exact path forward on #592 is unclear

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,365 @@
...
+
+/// The ID for the UnusedDocComments lint.
+const ID: &str = "UnusedDocComments";
+
+/// Creates a diagnostic for a misplaced doc comment.
+fn unused_doc_comment_diagnostic(span: Span, syntax_kind: SyntaxKind) -> Diagnostic {
+    Diagnostic::note(format!(
+        "Doc comments aren't supported on {}",
+        syntax_kind.describe()
+    ))
```
BKDaugherty: @a-frantz - re wordsmithing, do you have a good thought on what I should use here? Is the describe method sufficient you think to represent the element that the user is attempting to put a doc comment on?

Maybe a bit confusing especially for `BoundDeclNode`? Given it is valid to use a doc comment on "declaration with assignment" if it's in an `output` section?

I could also just say something less specific like "Doc comments aren't supported on this WDL item"?

Serial-ATA: Could take inspiration from the `unused_doc_comments` rustc lint:

```rust
fn main() {
    /// foo
    println!("Hello, world!");
}

warning: unused doc comment
 --> src/main.rs:2:5
  |
2 |     /// foo
  |     ^^^^^^^ rustdoc does not generate documentation for macro invocations
  |
  = help: to document an item produced by a macro, the macro must produce the documentation as part of its expansion
  = note: `#[warn(unused_doc_comments)]` (part of `#[warn(unused)]`) on by default
```

So maybe the message could just be "Unused doc comment" with the help message "Documentation will not be generated for this item". Getting into specifics about the type of item is cool, but not *that* important IMO.

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,336 @@
...
+use crate::TagSet;
+
+/// Prefix for defining the start of a doc comment.
+const DOC_COMMENT_PREFIX: &str = "## ";
+
+/// An "empty doc comment" should have this text.
+///
+/// Used to ensure that if we are parsing a doc comment and we don't see
+/// `DOC_COMMENT_PREFIX`, but we see this text we continue parsing.
+const EMPTY_DOC_COMMENT_TEXT: &str = "##";
```
a-frantz: IMO we should properly parse any comment starting with `##` as a doc comment even if it has weird or missing whitespace.

Regardless of that aspect, this can use the `wdl_ast::DOC_COMMENT_PREFIX` coming in with the PR I'm working on.

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,336 @@
...
+    // Doc Comments before the version statement are assumed to be a "preamble" for
+    // now.
+    fn version_statement(
+        &mut self,
+        _diagnostics: &mut Diagnostics,
+        _reason: wdl_analysis::VisitReason,
+        _stmt: &wdl_ast::VersionStatement,
+    ) {
+        self.version_statement_processed = true;
+    }
```
a-frantz: I'm working on removing the special treatment for preambles, so this logic can hopefully go away.

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,336 @@
...
+            &mut section.keyword().inner().preceding_trivia(),
+        );
+    }
+
+    fn comment(&mut self, diagnostics: &mut Diagnostics, comment: &Comment) {
+        if comment.text().starts_with(EXCEPT_COMMENT_PREFIX)
+            || comment.text() == EMPTY_DOC_COMMENT_TEXT
+            || comment.text().starts_with(DOC_COMMENT_PREFIX)
+            || !self.version_statement_processed
+        {
```
a-frantz: this should be replaceable with a new API call that looks like
```suggestion
        if comment.is_doc_comment() {
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: I think we will with the given implementation, we just use this constant for ensuring we don't skip just the text "##", but I could make the comment more clear for sure.

Edit: Oh I see what u mean yeah this would probably screw up if there's just "##\t" on a line!

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
```diff
@@ -0,0 +1,96 @@
...
+    Red,
+    ## And so is this mutltiline
+    ##
+    ## one.
+    Green
+
+}
+
+## Should I get a lint now that there have been changes to the preamble structure?
+## If so, we need to explicitly target whitespace in the `UnusedDocCommentsRule`
```
BKDaugherty: @a-frantz - with the deletion of the [preamble special handling as mentioned here](https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2782033835) , do we want to lint on "floating doc comments now"?

If so, how should I distinguish between this doc comment here on line 16 and 17 which is poorly placed, and the preamble on line 1? (I can bring back the `version_statement_processed` logic if that's the only difference?)

a-frantz: this is a good question and I have to think this one over a little. See my top level comment for how I would revise the overall approach taken by this implementation. Let's leave this thread unresolved for now, and we can come back to the question when the bigger changes have been implemented!

I've got some thoughts on this brewing, and I keep changing my mind whenever I try to write an answer 🤣

a-frantz (changes requested):
I don't think this implementation is taking the right approach here 🤔 my worry is primarily that of maintenance. This approach involves visiting every "not allowed" syntax node and calling `lint_doc_comments()` on it. If `wdl-ast` changes by adding new syntax nodes, there's a good chance no one will think to update the `unused_doc_comment.rs` with this approach.

A different way to approach this:
The set of "allowed" nodes is both finite and relatively small. It is also unlikely to change very often, and we will probably want to reference that set in multiple places across the codebase (`wdl-doc`, `wdl-format` maybe, here in `wdl-lint` ofc).
So I would add a new const in `wdl-ast` that looks like:
```rust
pub const DOC_COMMENT_SUPPORTED: &[SyntaxKind] = &[
    SyntaxKind::WorkflowDefinitionNode,
    <etc>
];
```
Then for this rule, instead of visiting the "not allowed nodes", we instead visit every comment and check if it is "attached" to a `SyntaxKind` not in the `DOC_COMMENT_SUPPORTED` const.

The tests you've written so far have some interesting cases I'm still chewing on the right way to handle 🤔

BKDaugherty:
> I don't think this implementation is taking the right approach here 🤔 my worry is primarily that of maintenance. This approach involves visiting every "not allowed" syntax node and calling `lint_doc_comments()` on it. If `wdl-ast` changes by adding new syntax nodes, there's a good chance no one will think to update the `unused_doc_comment.rs` with this approach.
> 
> A different way to approach this: The set of "allowed" nodes is both finite and relatively small. It is also unlikely to change very often, and we will probably want to reference that set in multiple places across the codebase (`wdl-doc`, `wdl-format` maybe, here in `wdl-lint` ofc). So I would add a new const in `wdl-ast` that looks like:
> 
> ```rust
> pub const DOC_COMMENT_SUPPORTED: &[SyntaxKind] = &[
>     SyntaxKind::WorkflowDefinitionNode,
>     <etc>
> ];
> ```
> 
> Then for this rule, instead of visiting the "not allowed nodes", we instead visit every comment and check if it is "attached" to a `SyntaxKind` not in the `DOC_COMMENT_SUPPORTED` const.
> 
> The tests you've written so far have some interesting cases I'm still chewing on the right way to handle 🤔

Ah yes. @a-frantz - This was actually my original approach (as seen in [this commit](https://github.com/stjude-rust-labs/sprocket/pull/590/changes/893640f02dbec3316d62c8fd4ae7b82bc386e15b), where I switched to this implementation in the [next commit](https://github.com/stjude-rust-labs/sprocket/pull/590/changes/cb31e6d416078b5ef83cdd811c942e41ecc473c3), after @Serial-ATA and I discussed this tradeoff directly in this [comment](https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2766138546) (which I've reproduced here to save some hops). 

> It does seem to make the processing of each element a little simpler, but I wonder if this will be harder to maintain? It seems more straightforward to me to know that if you add support for documenting a specific syntax item, you also should ensure that item isn't linted. Whereas with this implementation if you add a syntax item, you need to opt in to linting for doc comments.

> Similarly, it seems like there will be value in having a list of SyntaxKind or similar identifying element for the set of supported items for doc comments. If this lint could consume this list, then it could be one of many things that don't need to be updated directly when a new node supports doc comments. (IE: https://github.com/stjude-rust-labs/sprocket/issues/592#issuecomment-3849377584. In the current implementation it's a little hard to generate a concrete set of invalid items and have this lint consume this, since it's baked into the implementation of Visitor.

I'm generally happy with either implementation, and have enjoyed looking at both of the two approaches! I'll leave it up to y'all to decide which you would prefer.

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,218 @@
...
+                // of the comment, continue.
+                if let Some(token) = sibling.as_token() {
+                    let comment_is_floating =
+                        token.text().chars().filter(|c| *c == '\n').count() > 1;
+
+                    if comment_is_floating {
+                        // TODO: Decide if we want to lint on floating comments or not, and how that
+                        // relates to preambles.
+                        return None;
+                    }
```
BKDaugherty: @a-frantz - with reference to [this discussion](https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2818757287), I've left this as a TODO for now.

crates/wdl-ast/src/v1/task.rs
```diff
@@ -1230,6 +1234,12 @@ impl<N: TreeNode> CommandSection<N> {
         SectionParent::cast(self.0.parent().expect("should have a parent"))
             .expect("parent should cast")
     }
+
+    /// Gets the `command` keyword.
+    pub fn keyword(&self) -> CommandKeyword<N::Token> {
+        self.token()
+            .expect("CommandSection must have CommandKeyword")
+    }
 }
```
BKDaugherty, line 1258: Now that we are back to the "visiting each comment" approach, we don't really need these for this PR. Happy to keep them here, revert the commit that added them, or PR them separately if people would prefer :)

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
BKDaugherty: (for those following along at home, see also: https://github.com/stjude-rust-labs/sprocket/pull/590/changes#r2824160388)

crates/wdl-ast/src/v1/task.rs
a-frantz: I think these can stay as part of this PR 👍

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/// Detects whether a doc comment has been placed atop a Node that we do not
+/// generate documentation for.
+#[derive(Default, Debug, Clone)]
+pub struct UnusedDocCommentsRule {
+    /// Because each individual doc comment is visited as a single token,
+    /// we need to track each individual [`Comment`] in a multiline doc comment
+    /// block so we don't visit them again.
+    comments_seen: HashSet<Comment>,
```
a-frantz: `Comment`s are cheap to clone, so this is probably ok in terms of performance, but we can completely avoid this alloc by tracking a "skip count" instead. See this rule for an example - https://github.com/stjude-rust-labs/sprocket/blob/wdl-v0.22.1/crates/wdl-lint/src/rules/preamble_comment_placement.rs#L38

crates/wdl-lint/src/rules/unused_doc_comments.rs
a-frantz: how we should handle these is going to be informed by how the rest of the codebase handles them. In #614 we made it so that `format` will consolidate doc comments and prevent them from floating. Similarly, `doc` does not carve out any special handling for whether a doc comment is floating or not, and just concats all the `##` comments belonging to a node regardless of whether they are consecutive or not.

I'm not convinced either of those behaviors are "ideal", but there is a benefit to keeping things simple. So in spirit of simplicity, I would have this lint rule ignore all whitespace, and fire a diagnostic for each _consecutive_ set of doc comments. Example:
```wdl
## doc comment
## more text
## these 3 lines should all be highlighted by one span
#@ directive
# regular comment
## another doc comment block
## that gets a new diagnostic fired with a new span

## but there's a blank line in the middle!
## but we are going to consider these consecutive anyway
## this line and the prior 5 lines should all be in the same highlighted span
# another regular comment for good measure

command <<<>>>
```

a-frantz:
> > @a-frantz - I think because we are using `"##"` as the `DOC_COMMENT_PREFIX` instead of `"## "`, we are failing some arena tests that have some markdown looking content.
> > Do we want to lint on something like [this code here in the wilds-wdl-library](https://github.com/getwilds/wilds-wdl-library/blob/10fb6f78ebf8c8585b1a630e6979734a25e065d4/modules/ww-salmon/testrun.wdl#L14-L15)? I would imagine the answer is yes, but want to confirm before I bless le arena.
> 
> Arena is still sad because of this I think.

per my comment on ignoring whitespace, I actually don't think this linked comment should get linted. In `doc`, it is going to be parsed as an "appropriate" doc comment for the workflow definition. It will look weird and not render right, but that is because the repo being linted is non-conformant to what our tooling expects. So I would consider this a case "user error" (or rather "user isn't trying to follow this rule") so we shouldn't bother trying to special case it, even if it's not ideal. In other words, this would be an acceptable false negative.

There's usually a balance to be reached in regards to false negatives and false positives. Comprehensively eliminating both is sometimes not possible, or introduces enough complexity to tease apart many special cases that are rare enough that the juice stops being worth the squeeze. Of course the exact right balance and the point at which it becomes appropriate to stick with the "good enough" solution over the "perfect" solution is a judgement call that devs may disagree on!

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Oh nice! That seems way better thanks for the pointer!!

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Sweet! This makes a ton of sense! Thank you!

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,212 @@
+//! A lint rule for misplaced doc comments that will not generate documentation.
+
+use rowan::NodeOrToken;
```
a-frantz: small nit: can we use something exported by `wdl-ast` instead of reaching into `rowan`? https://docs.rs/wdl/latest/wdl/ast/type.SyntaxElement.html

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,212 @@
...
+/// The ID for the UnusedDocComments lint.
+const ID: &str = "UnusedDocComments";
+
+/// Creates a diagnostic for a misplaced doc comment.
+fn unused_doc_comment_diagnostic(span: Span) -> Diagnostic {
+    Diagnostic::note("Unused doc comment")
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_help("Documentation will not be generated for this item")
+        .with_fix("If you're intending to use a regular comment here, replace the `##` with `#`")
```
a-frantz:
```suggestion
    Diagnostic::note("unused doc comment")
        .with_rule(ID)
        .with_highlight(span)
        .with_help("documentation will not be generated for this item")
        .with_fix("replace the leading `##` with `#`")
```

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
```diff
@@ -0,0 +1,117 @@
+## This preamble should not be warned as it's not a doc comment!
```
a-frantz:
```suggestion
## The preamble is considered valid doc comments
```

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
```diff
@@ -0,0 +1,117 @@
...
+
+## This doc comment should be allowed
+#@ except: RequirementsSection
+task test_task {
+    meta {
+        description: "Show doc comments on bad placement of elements"
+    }
+
+    ## You can't doc comment a comment - the user should be warned.
+    # Comment about my doc comment
```
a-frantz:
```suggestion
    ## the user should be warned about this
    ## as it's own span
    # comment interrupting doc block resulting in two diagnostics
```

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
```diff
@@ -0,0 +1,117 @@
...
+## This doc comment should be allowed
+workflow test_workflow {
+    ## This doc comment does nothing and the user should be warned!
+    meta {
+        description: "Show doc comments on bad placement of elements"
+    }
+
+    ## This input section cannot have a doc comment.
+    input {
+        ## And so can it's elements
```
a-frantz:
```suggestion
        ## but it's elements can
```

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
```diff
@@ -0,0 +1,117 @@
...
+    ## But doc comments are not!
+    if (apple) {
+        String favorite_fruit = "Apple"
+    }
+        # Comments seem fine here (although maybe a weird choice)
+        ## But doc comments shouldn't be.
+    else if (banana) {
+        String favorite_fruit = "Banana"
+    }
+        # Seemingly, you can also put comments here,
```
a-frantz: comments are allowed pretty much anywhere there is whitespace in WDL. It took a while to figure out the right way to deal with this for `format`, which has a concept of "interrupting comments" to address this. But yea, users are technically allowed to get real freaky with their comment placement 🤣

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
```diff
@@ -0,0 +1,117 @@
...
+        ## but we don't want doc comments here.
+    else {
+            # Weirdly, wdl-lint seems to suggest my comment whitespace is incorrect
```
a-frantz: is this still true after #614 ?

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,212 @@
...
+
+            if sibling.kind() == SyntaxKind::Whitespace {
+                continue;
+            }
+
+            // If we are still looking at a doc comment, then continue, increase the
+            // skip_count and update the span.
+            if let Some(continued_comment) =
+                sibling.as_token().and_then(|t| Comment::cast(t.clone()))
+                && (continued_comment.is_doc_comment() || continued_comment.is_directive())
```
a-frantz:
```suggestion
                && continued_comment.is_doc_comment()
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,212 @@
...
+                self.skip_count += 1;
+
+                // Don't include a trailing directive in the doc comment's span, but ignore a
+                // directive in the middle of a doc comment block for the purposes of this lint.
+                if !continued_comment.is_directive() {
+                    span = Span::new(
+                        comment.span().start(),
+                        continued_comment.span().end() - comment.span().start(),
+                    );
+                }
```
a-frantz:
```suggestion
                    span = Span::new(
                        comment.span().start(),
                        continued_comment.span().end() - comment.span().start(),
                    );
```
Not sure if this is possible, but I wonder if we can avoid this repeated alloc with a `let mut span`? Can `Span`s have their ends mutated? More of a curiosity than a performance concern tbh

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: @a-frantz If we make this change, this will cause directives to break a doc comment block right?

Happy for that, but I thought we had decided that this rule should ignore directives in [this thread](https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2764893209)?

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
BKDaugherty: I haven't checked since but will!

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
BKDaugherty: lol yeah when I was approaching this the other way I was like wow there are a lot of syntax nodes that a user could technically put a bad doc comment

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: 🫡 I shall investigate

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
BKDaugherty: True! Sorry should have updated this

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
BKDaugherty: lol sorry same here, should have updated this

crates/wdl-lint/src/rules/unused_doc_comments.rs
a-frantz: maybe this edit isn't right? TBH I didn't run it locally 😅  But my idea here is that directives should not be included in the highlighted span, and if I read this fn right this would include them, essentially treating them as part of the doc comment block. Instead, I'd like to see _consecutive_ doc comments highlighted, any directive comment skipped over, and then a _new_ diagnostic firing on any doc comment blocks following the directive.
```wdl
## doc comment block
## that is all one diagnostic span
#@ doc comment block is ended by this directive
## a new doc comment block is started
## here that should get a new diagnostic span
meta {}
```

BKDaugherty: Yeah I think that's what this edit will do! I'll go ahead and make that change :)

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Both `start` and `end` are private. I can add a helper function for adjusting a Span's length though if you'd prefer?

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Hm. @a-frantz - Actually upon investigation I don't think this edit is what we want.

E.g: 

```wdl
## doc comment block that should be treated as 
## one emitted diagnostic 
#@ directive should end block in this case
## so a new one is started
## and we don't want a new span for this as it targets a "valid node" after the directive.
#@ but this directive should be ignored as the doc comment is correctly targeting the task.
task my_task {}
```

To enable this behavior, at the time of processing the directive we'd need to look ahead to the next element to know if it should be "ignored", or if it should be something that breaks up the doc comment.

Just to confirm I'm understanding the direction you're looking for, would you expect for us to get an `unused-doc-comment` lint for the top 

```wdl
## doc comment block that should be treated as 
## one emitted diagnostic 
```

but not the bottom?

```wdl
## so a new one is started
## and we don't want a new span for this as it targets a "valid node" after the directive.
```

a-frantz: o, maybe I wasn't clear. There are a lot of different but similar cases to consider!

The example you shared should not emit _any_ diagnostic as task definitions are valid for doc comments. `wdl-doc` is going to concatenate all those doc comments into a single markdown block, even though there are directives in the middle. `wdl-doc` ignores the directives, so this rule should too.

The tricky part is that if that were a meta section (as a random invalid example) instead of a task definition, then _two_ diagnostics should be emitted. One for the top block and one for the bottom block. The code for that is actually rather straight forward, as you don't need to look ahead beyond the end of the invalid doc comment block, but I get that the behavior is not super intuitive 😅 

To walk through what the processing would look like for the task definition example:
1. visit the first comment (`## doc comment block...`)
2. find the sibling this doc comment is "attached" to (task definition)
3. this is valid so end the visit without a diagnostic
4. visit second comment (`## ...`)
5. same as step 2
6. same as step 3
7. visit third comment (`#@ ...`)
8. not a doc comment so end the visit
9. so on

and if it were a meta section instead (so invalid and should be emitting diagnostics)
1. visit the first comment (`## doc comment block...`)
2. find the sibling this doc comment is "attached" to (meta section)
3. this is invalid so we need a diagnostic
4. check the next comment to see if it's part of this doc block (it is)
5. increment skip count
6. check the next comment to see if it's part of this doc block (is not NOT)
7. calculate `Span` and emit a diagnostic
8. end this visit
9. visit second comment (`## ...`)
10. "skip count" is `>0` so decrement and end visit
11. visit third comment (`#@ ...`)
12. end visit without a diagnostic
13. visit fourth comment (`## ...`)
14. see step 3
15. see step 4
16. see step 5
17. see step 6
18. see step 7
19. see step 8
20. so on

crates/wdl-lint/src/rules/unused_doc_comments.rs
a-frantz: > Both `start` and `end` are private. I can add a helper function for adjusting a Span's length though if you'd prefer?

no, that's not needed

> I wonder if we can avoid this repeated alloc 

It should be possible to just cache the start position of the first comment visited and avoid allocating a new span until we finish the "next comment" logic and are ready to emit a diagnostic. Those intermediate steps don't need to alloc a span since nothing is emitted during that step of processing

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
BKDaugherty: It was not true after #614!

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Done!

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: > The code for that is actually rather straight forward, as you don't need to look ahead beyond the end of the invalid doc comment block,

Sorry I probably could have specified that this is what I meant by lookahead in that each visited comment that isn't skipped by skip count needs to run step 2

> find the sibling this doc comment is "attached" to (task definition)

which requires us to look beyond the doc comment block we will visit next. But thanks for the pseudocode! I implemented it this way but lemme know if this isn't what you were thinking!

crates/wdl-ast/src/lib.rs
```diff
@@ -631,6 +631,25 @@ impl Comment {
     pub fn is_doc_comment(&self) -> bool {
         self.text().starts_with(DOC_COMMENT_PREFIX)
     }
+
+    /// Gets whether the comment is an inline comment or not.
+    pub fn is_inline_comment(&self) -> bool {
```
a-frantz: this is a good addition, but it needs some tests

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,200 @@
...
+
+            if let Some(continued_comment) =
+                sibling.as_token().and_then(|t| Comment::cast(t.clone()))
+                && continued_comment.is_doc_comment()
+            {
+                self.skip_count += 1;
+                span_end = continued_comment.span().end();
+                continue;
+            } else {
+                diagnostics.exceptable_add(
```
a-frantz: this add doesn't need to be exceptable if we make the rule only exceptable for the whole document (see next comment)
```suggestion
                diagnostics.add(
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,200 @@
...
+        - Enum Definitions
+        - Enum Variants"
+    }
+
+    fn tags(&self) -> crate::TagSet {
+        TagSet::new(&[Tag::Documentation])
+    }
+
+    fn exceptable_nodes(&self) -> Option<&'static [wdl_ast::SyntaxKind]> {
+        None
```
a-frantz: let's instead have this as just `VersionStatementNode`. I can't think of a case where a user would only want to except a single doc comment, but keep other doc comments. So it should be ok to make this only exceptable in a document wide fashion. That should also prevent any weird comment placements while trying to get annotate a parent node of a comment

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,200 @@
...
+    fn comment(&mut self, diagnostics: &mut Diagnostics, comment: &Comment) {
+        if self.skip_count > 0 {
+            self.skip_count -= 1;
+            return;
+        }
+
+        // If the visited comment isn't a doc comment, or we've already seen it, then
+        // there's no need to process it!
+        //
+        // Also, ignore comment directives.
```
a-frantz:
```suggestion
        // If the visited comment isn't a doc comment, then
        // there's no need to process it!
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,200 @@
...
+        if !comment.is_doc_comment() || comment.is_directive() {
+            return;
+        }
+
+        if comment.is_inline_comment() {
+            diagnostics.exceptable_add(
```
a-frantz:
```suggestion
            diagnostics.add(
```

crates/wdl-lint/tests/lints/unused-doc-comments/source.errors
```diff
@@ -0,0 +1,173 @@
+note[UnusedDocComments]: unused doc comment
+   ┌─ tests/lints/unused-doc-comments/source.wdl:36:5
+   │  
+36 │ ╭     ## The user should be warned about this doc comment as it targets
+37 │ │     ## command, despite the interruption below.
+   │ ╰───────────────────────────────────────────────^
+   │  
+   = help: documentation will not be generated for this item
```
a-frantz: "this item" without pointing to anything is potentially confusing. I checked out your branch and quickly noodled at adding a `with_label()` that highlights the next token and I think it's a big improvement on the reported diagnostics. I'll add some more comments with the changes I prototyped so you don't have to redo any of it!

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,200 @@
...
+const ID: &str = "UnusedDocComments";
+
+/// Creates a diagnostic for a misplaced doc comment.
+fn unused_doc_comment_diagnostic(span: Span) -> Diagnostic {
+    Diagnostic::note("unused doc comment")
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_help("documentation will not be generated for this item")
+        .with_fix("replace the leading `##` with `#`")
+}
```
a-frantz:
```suggestion
fn unused_doc_comment_diagnostic(comment_span: Span, target_span: Span) -> Diagnostic {
    Diagnostic::note("unused doc comment")
        .with_rule(ID)
        .with_highlight(comment_span)
        .with_label("documentation will not be generated for this item", node_span)
        .with_fix("replace the leading `##` with `#`")
}
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,200 @@
...
+                SyntaxElement::from(comment.inner().clone()),
+                &self.exceptable_nodes(),
+            );
+        }
+
+        let target = self.search_siblings_for_doc_comment_target(comment);
+        if let Some(result) = target
+            && !valid_target_for_doc_comment(&result)
+        {
+            self.lint_next_doc_comment_block(diagnostics, comment);
```
a-frantz:
```suggestion
            let target_span = if let Some(token) = result.as_token() {
                token.span()
            } else if let Some(node) = result.as_node() {
                node.first_token().unwrap().span()
            } else {
                unreachable!();
            };
            self.lint_next_doc_comment_block(diagnostics, comment, target_span)
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,200 @@
...
+    ) -> Option<SyntaxElement> {
+        let mut next = comment.inner().next_sibling_or_token();
+        while let Some(sibling) = next {
+            next = sibling.next_sibling_or_token();
+            if !sibling.kind().is_trivia() {
+                return Some(sibling);
+            }
+        }
+        None
+    }
```
a-frantz: this doesn't need to be a method as it doesn't interact with `self`
```suggestion
fn search_siblings_for_doc_comment_target(
    comment: &Comment,
    direction: Direction,
) -> Option<SyntaxElement> {
    let mut siblings = comment.inner().siblings_with_tokens(direction);
    while let Some(sibling) = siblings.next() {
        if !sibling.kind().is_trivia() {
            return Some(sibling);
        }
    }
    None
}
```

crates/wdl-lint/tests/lints/unused-doc-comments/source.errors
BKDaugherty: We don't currently search for the "target" item for "inline" comments since we already know they are invalid. We could...
- make the label optional, or use a different diagnostic wording for inline comments
- perform the forward search for inline comments
- look back to the "previous node", since that's what I would imagine the author probably is intending if they place a doc comment like that.

Do you have a preference here?

a-frantz: > * look back to the "previous node", since that's what I would imagine the author probably is intending if they place a doc comment like that.

I think this is the right move. It can just highlight the closest non-trivia token

BKDaugherty: Sweet! Thanks!

crates/wdl-ast/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 * Support for printing `EnumDefinition`s via `EnumDefinition::display()` and `StructDefinition`s via `Display` ([#551](https://github.com/stjude-rust-labs/sprocket/pull/551)).
+* Support `keyword` method on `CommandSection`, `RequirementsSection`, `MetadataSection`, `ParameterMetadataSection`, `ScatterStatement` and `CallStatement` for accessing the associated keyword for the syntax node ([#590](https://github.com/stjude-rust-labs/sprocket/pull/590))
```
claymcleod:
```suggestion
* Support `keyword` method on `CommandSection`, `RequirementsSection`, `MetadataSection`, `ParameterMetadataSection`, `ScatterStatement` and `CallStatement` for accessing the associated keyword for the syntax node ([#590](https://github.com/stjude-rust-labs/sprocket/pull/590)).
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Creates a diagnostic for a misplaced doc comment.
+fn unused_doc_comment_diagnostic(comment_span: Span, target_span: Span) -> Diagnostic {
+    Diagnostic::note("unused doc comment")
+        .with_rule(ID)
+        .with_highlight(comment_span)
+        .with_label(
+            "documentation will not be generated for this item",
+            target_span,
+        )
+        .with_fix("replace the leading `##` with `#`")
```
claymcleod:
```suggestion
        .with_fix("if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to its associated element")
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+}
+
+impl Rule for UnusedDocCommentsRule {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Reports doc comments that are attached to WDL sections that don't support them."
```
claymcleod:
```suggestion
        "Reports doc comments that are attached to WDL items that don't support them."
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    fn explanation(&self) -> &'static str {
+        "Some Workflow Definition Language items and sections do not support doc comments (`##`). \
```
claymcleod:
```suggestion
        "Some Workflow Definition Language items do not support doc comments (`##`). \
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
claymcleod: I'd just roll with "items" here—I think that's the term we use in the lexer/parser, and the distinction between sections and items is probably not that important.

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,231 @@
...
+         This lint reports if a doc comment is attached to some section or item that isn't \
```
claymcleod:
```suggestion
         This lint reports if a doc comment is attached to an item that isn't \
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,231 @@
...
+            let target_span = if let Some(token) = target.as_token() {
+                token.span()
+            } else if let Some(node) = target.as_node() {
+                node.first_token().unwrap().span()
+            } else {
+                unreachable!();
+            };
+
+            diagnostics.add(unused_doc_comment_diagnostic(comment.span(), target_span));
+        }
```
claymcleod: This needs a `return` after emitting the diagnostic — otherwise execution falls through to the forward-sibling search below and can produce a second diagnostic for the same comment. Currently masked in tests because the forward targets happen to be valid.

```suggestion
            diagnostics.add(unused_doc_comment_diagnostic(comment.span(), target_span));
            return;
        }
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,231 @@
...
+                span_end = continued_comment.span().end();
+                continue;
+            } else {
+                diagnostics.add(unused_doc_comment_diagnostic(
+                    Span::new(comment.span().start(), span_end - comment.span().start()),
+                    target_span,
+                ));
+                break;
+            }
+        }
```
claymcleod, line 163: If the doc comment block extends to the end of the token stream (no non-whitespace, non-doc-comment sibling follows), the loop exits without emitting a diagnostic. Consider adding a fallback `diagnostics.add(...)` after the loop.

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,231 @@
...
+        // If the visited comment isn't a doc comment, then
+        // there's no need to process it!
+        if !comment.is_doc_comment() || comment.is_directive() {
+            return;
+        }
+
+        if comment.is_inline_comment()
+            && let Some(target) = find_inline_doc_comment_target(comment)
+        {
+            let target_span = if let Some(token) = target.as_token() {
```
claymcleod: nit: This target-span extraction pattern (token vs node vs unreachable) appears three times in this file. Would be worth extracting into a small helper.

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,231 @@
...
+        - Enum Definitions
+        - Enum Variants"
+    }
+
+    fn tags(&self) -> crate::TagSet {
+        TagSet::new(&[Tag::Documentation])
+    }
+
+    fn exceptable_nodes(&self) -> Option<&'static [wdl_ast::SyntaxKind]> {
+        Some(&[SyntaxKind::VersionStatementNode])
```
claymcleod, line 204: `exceptable_nodes()` returns `Some(&[VersionStatementNode])` but the diagnostics are added via `diagnostics.add()` rather than `exceptable_add()` — so `#@ except: UnusedDocComments` directives won't actually suppress the diagnostic.

crates/wdl-ast/src/v1/task.rs
```diff
@@ -1230,6 +1234,12 @@ impl<N: TreeNode> CommandSection<N> {
         SectionParent::cast(self.0.parent().expect("should have a parent"))
             .expect("parent should cast")
     }
+
+    /// Gets the `command` keyword.
+    pub fn keyword(&self) -> CommandKeyword<N::Token> {
+        self.token()
+            .expect("CommandSection must have CommandKeyword")
+    }
```
claymcleod, line 1257: These `keyword()` methods don't appear to be used by the lint rule or elsewhere in this PR — just noting in case they were left over from an earlier iteration. I'm fine with keeping them in as someone may need them in the future.

crates/wdl-ast/src/lib.rs
```diff
@@ -631,6 +631,25 @@ impl Comment {
...
+        // If there is a preceeding token that isn't whitespace with a newline, then
```
claymcleod:
```suggestion
        // If there is a preceding token that isn't whitespace with a newline, then
```

crates/wdl-lint/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 * Added `Config` for configuring the behavior of certain lint rules ([#553](https://github.com/stjude-rust-labs/sprocket/pull/553))
+* New lint rule `UnusedDocComments` to ensure users are warned if they document unsupported syntax items ([#590])(https://github.com/stjude-rust-labs/sprocket/pull/590).
```
claymcleod:
```suggestion
* New lint rule `UnusedDocComments` to ensure users are warned if they document unsupported syntax items ([#590](https://github.com/stjude-rust-labs/sprocket/pull/590)).
```

crates/wdl-lint/RULES.md
```diff
@@ -43,4 +43,5 @@ This table documents all `sprocket` lint rules implemented on the `main` branch
...
 | `TodoComment`               | Style                                                       | Flags TODO statements in comments to ensure they are not forgotten.                                                                        |                                                      |
-| `TrailingComma`             | Style                                                       | Ensures that lists and objects have a trailing comma and that there's not extraneous whitespace and/or comments before the trailing comma. |                                                      |
\ No newline at end of file
+| `TrailingComma`             | Style                                                       | Ensures that lists and objects have a trailing comma and that there's not extraneous whitespace and/or comments before the trailing comma. |                                                      |
+| `UnusedDocComments`         | Spacing, Style                                              | Ensures that all doc comments are attached to supported syntax items.                                                                      |                                                      |
```
claymcleod: The tags here are listed as "Spacing, Style" but the implementation returns `TagSet::new(&[Tag::Documentation])`. Should this be "Documentation"?

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
```diff
@@ -0,0 +1,120 @@
+## This preamble is considered a valid doc comment,
+## despite having whitespace between it and the version statement.
+
+version 1.3
+
+## This doc comment is allowed
+enum Color {
+    ## As is this one
+    Red,
+    ## And so is this mutltiline
```
claymcleod:
```suggestion
    ## And so is this multiline
```

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
```diff
@@ -0,0 +1,120 @@
...
+## This doc comment should be allowed.
+workflow test_workflow {
+    ## This doc comment does nothing and the user should be warned!
+    meta {
+        description: "Show doc comments on bad placement of elements"
+    }
+
+    ## This input section cannot have a doc comment.
+    input {
+        ## But it's elements can.
```
claymcleod:
```suggestion
        ## But its elements can.
```

claymcleod (changes requested):
Nice work getting this through 13 commits of iteration — the overall approach is solid. A few things to address, the most important being a missing `return` in the inline comment path and a dropped diagnostic edge case.

crates/wdl-ast/src/v1/task.rs
BKDaugherty: Yeah. Noted here: https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2824164641

Happy to move them to a separate PR or revert if requested!

claymcleod: Ah, missed that comment! They definitely can stay.

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: That makes sense to me. @a-frantz   is there some magic where exceptable_nodes gets used in the global context?

I swapped this based on the following comments.
- https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2839341912
- https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2839352346

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Yikes! Nice catch! I'll add a test case for that too

crates/wdl-lint/tests/lints/unused-doc-comments/source.wdl
BKDaugherty: lol sorry for all the typos! Thanks for the catches!

crates/wdl-lint/src/rules/unused_doc_comments.rs
a-frantz: there is some magic! Directives on the `VersionStatementNode` prevent the rule from executing for the document, making it safe to call `add()` directly instead of `exceptable_add()` - https://github.com/stjude-rust-labs/sprocket/blob/36c67422bc719185d67fc7d85c09e8503cd58267/crates/wdl-lint/src/linter.rs#L98-L104

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: We won't have a `target_span` in this case, so I updated the `unused_doc_comment_diagnostic` to conditionally apply the label, which makes this case look like...

```
+note[UnusedDocComments]: unused doc comment
+    ┌─ tests/lints/unused-doc-comments/source.wdl:121:1
+    │
+121 │ ## Trailing Doc Comment should be linted
+    │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    │
+    = fix: if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this comment so it is bound to it
s associated element
```

Let me know what you think!

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+        if let Some(result) = target
+            && !valid_target_for_doc_comment(&result)
+        {
+            self.lint_next_doc_comment_block(
+                diagnostics,
+                comment,
+                Some(get_span_of_first_token_for_syntax_element(&result)),
+            )
+        }
```
claymcleod, line 251: Consider collapsing the two branches into a single `if`:

```suggestion
        let target = search_siblings_for_doc_comment_target(comment);
        if target.as_ref().map_or(true, |t| !valid_target_for_doc_comment(t)) {
            self.lint_next_doc_comment_block(
                diagnostics,
                comment,
                target.as_ref().map(get_span_of_first_token_for_syntax_element),
            );
        }
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,252 @@
...
+                diagnostics.add(unused_doc_comment_diagnostic(
+                    Span::new(comment.span().start(), span_end - comment.span().start()),
+                    target_span,
+                ));
+                return;
+            }
+        }
+
+        // If the doc comment block extends to the end of the token stream,
+        // we won't add a diagnostic above. Add one here without a `target_span`.
```
claymcleod: This comment says "without a `target_span`" but the code passes `target_span` through — looks like the comment is out of sync with the implementation.

```suggestion
        // If the doc comment block extends to the end of the token stream,
        // we won't add a diagnostic above. Add one here.
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Clippy hath requested

```rust
warning: this `map_or` can be simplified
   --> crates/wdl-lint/src/rules/unused_doc_comments.rs:239:12
    |
239 |           if target
    |  ____________^
240 | |             .as_ref()
241 | |             .map_or(true, |t| !valid_target_for_doc_comment(t))
    | |_______________________________________________________________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#unnecessary_map_or
    = note: `#[warn(clippy::unnecessary_map_or)]` on by default
help: use is_none_or instead
    |
241 -             .map_or(true, |t| !valid_target_for_doc_comment(t))
241 +             .is_none_or(|t| !valid_target_for_doc_comment(t))
    |
```

So I went ahead with that. Just a heads up! :)

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,253 @@
...
+const ID: &str = "UnusedDocComments";
+
+/// Creates a diagnostic for a misplaced doc comment.
+fn unused_doc_comment_diagnostic(comment_span: Span, target_span: Option<Span>) -> Diagnostic {
+    let diagnostic = Diagnostic::note("unused doc comment")
+        .with_rule(ID)
+        .with_highlight(comment_span)
+        .with_fix(
+            "if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this \
+             comment so it is bound to its associated element",
```
a-frantz:
```suggestion
            "if this is a non-doc comment, replace the leading `##` with `#`; otherwise move this \
             comment so it documents the intended item",
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    fn comment(&mut self, diagnostics: &mut Diagnostics, comment: &Comment) {
+        if self.skip_count > 0 {
+            self.skip_count -= 1;
+            return;
+        }
+
+        // If the visited comment isn't a doc comment, then
+        // there's no need to process it!
+        if !comment.is_doc_comment() || comment.is_directive() {
```
a-frantz:
```suggestion
        if !comment.is_doc_comment() {
```

crates/wdl-lint/src/rules/unused_doc_comments.rs
BKDaugherty: Made the change! This wording was a request [from this comment fwiw](https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2855899571)

PR #606, opened 2026-02-05
Share execution utilities between CLI and server

claymcleod:
The `sprocket run` command now uses the same execution infrastructure as the server. Runs are tracked in a SQLite database at `<output_dir>/sprocket.db` with full provenance. The `--runs-dir` and `--output` flags have been replaced with `--output-dir` (defaulting to `./out`), and `--index-on` enables output indexing.

The `AllowedSource` and `Source` types are unified into a single `Source` type. Shared utilities for database operations, session creation, run directory setup, and target execution live in `system::v1::exec`.

The CLI test framework now supports verifying output directory structures with timestamp/UUID normalization and symlink handling. A `verify-output-structure` test exercises `--index-on` and confirms the directory layout.

For all contributors:

### DIFF
--- a/src/system/v1/exec.rs
+++ b/src/system/v1/exec.rs
@@ -753,10 +753,7 @@ async fn execute_task_target(
         Err(e) => return Err(e),
     };
 
-    evaluated_task
-        .into_outputs()
-        .map(Some)
-        .map_err(Into::into)
+    evaluated_task.into_outputs().map(Some)
 }
 
 /// Execute a workflow or task target.
--- a/tests/cli/run/help/stdout
+++ b/tests/cli/run/help/stdout
@@ -75,11 +75,11 @@ Options:
           
           [env: GOOGLE_HMAC_SECRET]
 
-  -v, --verbose...
-          Increase logging verbosity
-
       --no-call-cache
           Disables the use of the call cache for this run
+
+  -v, --verbose...
+          Increase logging verbosity
 
   -q, --quiet...
           Decrease logging verbosity
@@ -91,7 +91,7 @@ Options:
           - auto:   Automatically colorize output depending on output device
           - always: Always colorize output
           - never:  Never colorize output
-
+          
           [default: auto]
 
   -c, --config <CONFIG>
--- a/tests/logging.rs
+++ b/tests/logging.rs
@@ -61,14 +61,8 @@ task hello {
         assert!(str::from_utf8(&result.stderr).unwrap().contains(level));
 
         // Ensure the log file have at least one message at the level
-        let log = fs::read_to_string(
-            dir.path()
-                .join("runs")
-                .join("hello")
-                .join("_latest")
-                .join("output.log"),
-        )
-        .expect("should have output log");
+        let log = fs::read_to_string(dir.path().join("out").join("output.log"))
+            .expect("should have output log");
         assert!(log.contains(level));
     }
 }


src/analysis/source.rs
```diff
@@ -11,8 +11,11 @@ use path_clean::PathClean;
 use url::Url;
 use wdl::analysis::Analyzer;
 
-/// The supported source schemes for both WDL source files and JSON input files.
-const SUPPORTED_SOURCE_SCHEMES: &[&str] = &["https://", "http://", "file://"];
+/// Remote URL schemes that are parsed as `Source::Url`.
+const REMOTE_URL_SCHEMES: &[&str] = &["https://", "http://", "ftp://"];
```
peterhuene: I don't believe we can support `ftp://` with our current source download implementation, though.

src/commands/run.rs
```diff
@@ -114,26 +113,21 @@ pub struct Args {
...
-
-    /// The execution directory.
+    /// The output directory; defaults to `./out`.
     ///
-    /// If this argument is supplied, the default output behavior of nesting
-    /// execution directories using the target and timestamp will be
-    /// disabled.
-    #[clap(long, conflicts_with = "runs_dir", value_name = "OUTPUT_DIR")]
-    pub output: Option<PathBuf>,
+    /// Individual runs are stored at `<output_dir>/runs/<target>/<timestamp>/`.
```
peterhuene, line 132: Question: do we need to nest `out/runs/...` here? The nesting seems redundant to me.

Could the layout perhaps be:

```
runs/
├─ sprocket.db/
├─ <target>/
│  ├─ <timestamp>/
│  ├─ <timestamp>/
│  ├─ ...
├─ .index/
│  ├─ ...
```

The only problem I see is that we'd have to name the index directory with something that can't conflict with a WDL identifier (hence the `.` prefix above, but perhaps something else like `index.data` would do?).

src/commands/run.rs
```diff
@@ -114,26 +113,21 @@ pub struct Args {
...
-    /// Overwrites the execution directory if it exists.
-    #[clap(long, conflicts_with = "runs_dir")]
-    pub overwrite: bool,
+    /// The output name to index on.
+    ///
+    /// If provided, the run outputs will be indexed using the specified output
+    /// name as the key. The index allows efficient lookup of runs by output
+    /// values.
+    #[clap(long, value_name = "OUTPUT_NAME")]
+    pub index_on: Option<String>,
```
peterhuene, line 144: In retrospect, is there a reason why we're indexing using files on the file system rather than an `outputs` table in the SQLite database? 

Ideally, if SQLite supported indexing a JSONB column we could just have the `outputs` column on the `runs` table be structured and then we could very quickly look up runs with a particular output name.

I'm probably misunderstanding the purpose of this option, though.

src/commands/run.rs
```diff
@@ -747,28 +702,27 @@ pub async fn run(mut args: Args, mut config: Config) -> CommandResult<()> {
...
-                            e.document.root().text().to_string(),
-                            &[e.diagnostic],
-                            &e.backtrace,
-                            args.report_mode.unwrap_or_default(),
-                            args.no_color
-                        )?;
-                        Err(anyhow!("aborting due to evaluation error").into())
-                    }
-                    Err(EvaluationError::Other(e)) => Err(e.into())
+                    Err(e) => Err(anyhow!("evaluation failed: {:#}", e).into())
```
peterhuene: We cannot collapse the display of these different types of errors like you've done here.

We must display a source error via `emit_diagnostics`; additionally, other runtime errors should not be wrapped with an `anyhow!` as they'll lose backtrace context.

With what you have now, if we have a source-related evaluation error, we'll actually get the debug formatting of the error.

See for yourself by running this task:

```
version 1.2

task t {
    command <<<
        exit 1
    >>>
}
```

src/commands/run.rs
claymcleod: Because the mental model is that of `runs/` being unorganized with random names with the `index/` logically organized alongside it, I have a slight preference to keep the runs in a `runs/` directory under `out/`. This enforces that mental model while also allowing for us to extend root files/folders under out/ without losing them in the mix of all targets.

src/commands/run.rs
claymcleod: Basically, it's to make it easy to navigate through the files without ever having to use the `sprocket` command line. I anticipate there will be consumers of data in the `index/` directory that don't know about `sprocket`, nor do we want to require them to use it to get the logical organization of these files out.

src/commands/run.rs
peterhuene: I don't quite understand why they need to be side-by-side, to be honest. It's a directory that specifically ties to runs so it makes sense for `runs` to contain it (i.e. it's the "index of runs").

Conceptually, the `runs` directory today _is_ the output of running a workflow or task. But with these changes the output becomes a `runs` directory sitting next to a SQLite database and index directory, both of which are specifically tied to (and aggregate) the runs.

Like if running a workflow or task produced something _in addition to_ runs, then perhaps nesting them in an `out` directory work make more sense to me.

src/commands/run.rs
peterhuene: Perhaps it's just early days of the implementation of the `index` directory and its use will become more apparent to me over time as more is implemented, then.

In its initial implementation, it feels like just giving users a custom name for a "latest" symlink, though instead of pointing at the run directory itself, it points to the outputs file.

src/commands/run.rs
claymcleod: Peter and I discussed this in Slack, and we ultimately agreed to roll with what's implemented for now with an eye towards providing an even better access pattern in the future. The gist was that I expect the index to be consumed by downstream scientists who literally just want a directory (the index) of the latest data that they can build their bespoke infrastructure on top of. This slightly tips the scale towards not intermingling the index within the runs (which could have 50+ targets for production systems) and instead enforce this mental model of a completely separate output directory for downstream consumers using the splitting of the folders into `runs/` and `index/` under output. He advocated that it wouldn't be that hard for such users to wrap `sprocket` in their bash scripts to get more rich information out, to which I agreed and suggested we think harder about the best way to do this once we see the access patterns in practice.

src/commands/run.rs
claymcleod: See the above resolved conversation for where this landed.

PR #610, opened 2026-02-06
feat(sprocket): explain --format=json

Serial-ATA:
This adds an auto-generated lint list site for both `wdl-lint` and `wdl-analysis`, similar to the [Clippy lint list](https://rust-lang.github.io/rust-clippy/master/index.html).

The site can be generated with `cargo generate-lint-site` 

TODO:
- [x] Add config fields to applicable lints
- [x] Reformat all lint explanations (due to indentation they sometimes get rendered as code blocks)
- [x] ~~Add version numbers to all lints~~
- [x] Add examples to all lints
- [x] ~~Version filtering~~

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/src/rules.rs
```diff
@@ -37,12 +37,18 @@ pub trait Rule: Send + Sync {
     /// which a rule is excepted or denied.
     fn id(&self) -> &'static str;
 
+    /// The `wdl-analysis` version in which this rule was added.
+    fn version(&self) -> &'static str;
```
a-frantz: I'm not sure we really want/need to track the version rules were added. ~~At best, this is redundant~~ this information is available by looking through the manually curated changelogs, and it is potentially confusing as some rules have undergone behavior and/or name changes since they were introduced.

crates/wdl-analysis/src/rules.rs
```diff
@@ -37,12 +37,18 @@ pub trait Rule: Send + Sync {
...
+
     /// A short, single sentence description of the rule.
     fn description(&self) -> &'static str;
 
     /// Get the long-form explanation of the rule.
     fn explanation(&self) -> &'static str;
 
+    /// Get a list of examples that would trigger this lint rule.
+    fn examples(&self) -> &'static [&'static str];
```
a-frantz, line 47: this is a great addition to the trait! It's been discussed a few times but no one has followed through on the idea, so really exciting to see this land 🚀

crates/wdl-analysis/src/rules.rs
```diff
@@ -313,6 +426,20 @@ impl Rule for UsingFallbackVersion {
          a different version."
     }
 
+    fn examples(&self) -> &'static [&'static str] {
+        &[r#"```wdl
+# Not a valid version. If a fallback version is configured,
+# the document will be interpreted as that version.
+version 3.0
```
a-frantz:
```suggestion
version development
```

crates/wdl-analysis/src/rules.rs
Serial-ATA: I wouldn't say it's redundant, at least with the current state of the changelogs. I had to dig through the commit history to find a decent number of lint versions

a-frantz: Fair, "redundant" was not the right word to use. It is a much friendlier place to find the information, but what I was trying to express was rather that this information _does_ exist already, and I'm not sure we need to surface it in a friendlier way. I think that if someone is looking for this information (and I'm not sure why they would), it's fair for us to say "search the changelogs, if not there search the git history".

I would feel differently if we were at a more stable stage of development. I'm of the understanding that almost all our users are updating to latest on some regular cadence (maybe they fall behind by a release or two or three, but then they catch up). IMO Sprocket doesn't have enough stability between releases for tracking this to be useful.

I'd also feel differently if we hadn't undergone more than one rule overhaul where many names and/or behaviors changed. I think that renders the "truthiness" of these to be suspect, or at least not very reliable.

Serial-ATA: Yeah I agree, I took the version filtering out. Still need to fix up some design issues from stripping it out

claymcleod: I think it's worth keeping the version a rule was introduced in — it's useful context for users who are upgrading and want to understand which rules are new to them. The maintenance burden is low since it's a one-time annotation per rule. We may need to add a step to the release process to update the version of any newly released rules though.

a-frantz: Which version should be annotated though? The `wdl-analysis` version, or the Sprocket version? The wdl crates have different version numbers and the crates are an abstraction hidden from users. The contents of wdl-analysis should not be tracking Sprocket versions, but wdl-anlaysis version numbers are meaningless or confusing to users.

crates/wdl-analysis/src/rules.rs
```diff
@@ -154,6 +182,22 @@ impl Rule for UnusedInputRule {
...
     }
 
+    fn examples(&self) -> &'static [&'static str] {
+        &[r#"```wdl
+version 1.2
+
+workflow example {
+    meta {}
+
+    inputs {
```
claymcleod:
```suggestion
    input {
```
WDL uses the singular keyword `input`, not `inputs`.

crates/wdl-lint/src/rules/input_sorted.rs
```diff
@@ -198,14 +198,47 @@ impl Rule for InputSortedRule {
...
+    }
+
+    fn examples(&self) -> &'static [&'static str] {
+        &[
+            r#"```wdl
+version 1.2
+
+task describe_person {
+    input {
+        ?File target = stdout()
```
claymcleod: WDL uses postfix `?` for optional types — e.g., `File? target` rather than `?File target`. Both examples in this rule need fixing.

crates/wdl-lint/src/rules/input_sorted.rs
```diff
@@ -198,14 +198,47 @@ impl Rule for InputSortedRule {
...
-         Map[*, *], Pair[*, *]), drop the outermost type (Array, Map, etc.) and recursively apply \
-         above sorting on the first inner type *, with ties broken by the second inner type. \
-         Continue this pattern as far as possible. Once this ordering is satisfied, it is up to \
-         the developer for final order of inputs of the same type."
+         sorting.\n\nIt first sorts by:\n1. required inputs\n2. optional inputs without \
+         defaults\n3. optional inputs with defaults\n4. inputs with a default value\n\nThen by the \
+         type:\n1. File\n2. Array[*]+\n3. Array[*]\n4. struct\n5. Object\n6. Map[*, *]\n7. Pair[*, \
+         *]\n8. String\n9. Boolean\n10. Float\n11. Int\n\nFor ordering of the same compound type \
+         (Array[*], Map[*, *], Pair[*, *]), drop the outermost type(Array, Map, etc.) and \
+         recursively apply above sorting on the first inner type *, with ties brokenby the second \
```
claymcleod: A couple of spaces got lost in the reformat: "brokenby" → "broken by" and "satisfied,it" → "satisfied, it".

crates/wdl-lint/src/rules/known_rules.rs
```diff
@@ -68,6 +68,33 @@ impl Rule for KnownRulesRule {
          mistakes."
     }
 
+    fn examples(&self) -> &'static [&'static str] {
+        &[
+            r#"```wdl
+#@ except: LintThatDoesNotExit
```
claymcleod:
```suggestion
#@ except: LintThatDoesNotExist
```
Typo: "Exit" → "Exist"

crates/wdl-analysis/src/rules.rs
Serial-ATA: The versions can be revisited in the future if necessary. Right now though, it's unclear which version to display and how to handle renamed lints

crates/wdl-lint/src/rules/consistent_newlines.rs
```diff
@@ -57,6 +57,10 @@ impl Rule for ConsistentNewlinesRule {
          your project."
     }
 
+    fn examples(&self) -> &'static [&'static str] {
+        &[]
```
claymcleod, line 61: Does this need an example?

crates/wdl-lint/src/rules/shellcheck.rs
```diff
@@ -231,11 +231,15 @@ impl Rule for ShellCheckRule {
...
 
     fn explanation(&self) -> &'static str {
-        "ShellCheck (https://shellcheck.net) is a static analysis tool and linter for sh / bash. \
+        "[ShellCheck](https://shellcheck.net) is a static analysis tool and linter for sh / bash. \
          The lints provided by ShellCheck help prevent common errors and pitfalls in your scripts. \
          Following its recommendations will increase the robustness of your command sections."
     }
 
+    fn examples(&self) -> &'static [&'static str] {
+        &[]
```
claymcleod: Should we add an example here?

web-common/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "sprocket-web-common",
+  "version": "1.0.0",
+  "description": "Common web components shared between static sites in the sprocket repo.",
+  "main": "dist/bundle.js",
+  "module": "dist/bundle.js",
```
claymcleod, line 7: private = true?

claymcleod (changes requested):

crates/wdl-lint/src/rules/except_directive_valid.rs
```diff
@@ -78,6 +78,37 @@ impl Rule for ExceptDirectiveValidRule {
...
+
+    output {
+        # MatchingOutputMeta exceptions aren't valid
+        # in this context
+        #@ except: MatchingOutputMeta
+        String name = "Jimmy"
+    }
+}
+```"#,
+            r#"Use instead:
```
claymcleod, line 97: The second example is missing its opening ` ```wdl ` code fence. It currently jumps from "Use instead:" straight into `version 1.2` without a fenced block. Every other rule wraps the WDL in a ` ```wdl ` block—would you mind adding it here too?

src/commands/explain.rs
```diff
@@ -86,56 +262,72 @@ pub fn list_all_rules() -> String {
...
     }
 
     if args.list_all_tags {
-        println!("{}", list_all_tags());
+        match args.format {
+            Format::Default => {
+                println!("{}", list_all_tags());
+            }
+            Format::Json => {
+                let mut all_tags = collect_all_tags().into_values().collect::<Vec<_>>();
```
claymcleod, line 325: `applicable_lints` are sorted in the `--tag` JSON path but not in the `--list-all-tags` JSON path. The test fixtures confirm the inconsistency (compare `All-JSON/stdout` vs `Clarity-JSON/stdout` for the Clarity tag). Consider sorting `applicable_lints` inside `collect_all_tags` so both paths produce consistent output.

crates/wdl-lint/src/config.rs
```diff
@@ -5,30 +5,103 @@ use std::collections::HashSet;
...
+    pub struct Config {
+        /// List of keys to ignore in the [`ExpectedRuntimeKeys`] lint.
+        ///
+        /// ##### Example
+        ///
+        /// ```toml
+        /// allowed_runtime_keys = ["foo"]
+        /// ```
+        ///
+        /// [`ExpectedRuntimeKeys`]: crate::rules::ExpectedRuntimeKeysRule.
```
claymcleod: Trailing period is syntactically part of the link path here, which will break intra-doc link resolution. Remove the `.` after `ExpectedRuntimeKeysRule`.

```suggestion
        /// [`ExpectedRuntimeKeys`]: crate::rules::ExpectedRuntimeKeysRule
```

crates/wdl-lint/src/config.rs
```diff
@@ -5,30 +5,103 @@ use std::collections::HashSet;
...
+}
+
+define_lint_rule_config! {
+    /// The configuration for lint rules.
+    #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+    pub struct Config {
+        /// List of keys to ignore in the [`ExpectedRuntimeKeys`] lint.
+        ///
+        /// ##### Example
+        ///
```
claymcleod, line 84: Heading level inconsistency—this field uses `#####` while `allowed_names` below uses `##`. Consider using the same level for both.

src/commands/explain.rs
```diff
@@ -68,6 +94,156 @@ pub struct Args {
...
+pub struct ConfigField {
+    /// The name of the field, as it appears in the config file.
+    pub name: &'static str,
+    /// A Markdown-formatted description of the field.
+    pub description: &'static str,
+    /// The default value of the field as a TOML string.
+    pub default: String,
+}
+
+#[derive(Debug, Serialize)]
```
claymcleod, line 126: Missing doc comment on this struct. Every other public type in this file has one.

src/commands/explain.rs
```diff
@@ -86,56 +262,72 @@ pub fn list_all_rules() -> String {
...
-    };
+    tags
 }
 
-/// Pretty prints an analysis rule to a string.
-pub fn pretty_print_analysis_rule(rule: &dyn analysis::Rule) {
-    println!("{id}", id = rule.id().bold().underline());
-    println!("{desc}", desc = rule.description());
-    println!("\n{explanation}", explanation = rule.explanation());
+pub fn list_all_tags() -> String {
```
claymcleod, line 294: Missing doc comment on this public function.

crates/wdl-lint/src/rules/consistent_newlines.rs
Serial-ATA: I don't think so, an example wouldn't be that useful since the characters are invisible anyway, and the lint's gonna be removed after #668

claymcleod: Okay, that's fine.

crates/wdl-lint/src/config.rs
```diff
@@ -5,30 +5,103 @@ use std::collections::HashSet;
...
+        #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+        #[serde(default)]
+        pub struct $name {
+            $(
+                $(#[doc = $doc])+
+                pub $field: $ty,
+            )+
+        }
+
+        impl Default for Config {
```
peterhuene:
```suggestion
        impl Default for $name {
```

crates/wdl-lint/src/config.rs
```diff
@@ -5,30 +5,103 @@ use std::collections::HashSet;
...
-    /// allowed_names = ["Foo", "counter_int"]
-    /// ```
-    ///
-    /// [`SnakeCase`]: crate::rules::SnakeCaseRule
-    /// [`DeclarationName`]: crate::rules::DeclarationNameRule
-    pub allowed_names: HashSet<String>,
+/// Define the lint rule config and doc generation utilities.
+macro_rules! define_lint_rule_config {
+    (
+        $(#[$meta:meta])*
```
claymcleod, line 11: The `$(#[$meta:meta])*` capture is never applied in the expansion—the macro hardcodes its own derives and doc comment instead. This means the `#[derive(...)]` and `///` at the invocation site are silently discarded. Consider either applying `$(#[$meta])*` in the expansion or removing the capture so the invocation site does not mislead readers into thinking they can customize the attributes.

src/commands/explain.rs
```diff
@@ -68,6 +94,157 @@ pub struct Args {
...
+    /// Crate-specific configuration fields that apply to this rule.
+    pub config: Option<Vec<ConfigField>>,
+}
+
+impl Rule {
+    /// Convert this rule to a string of the given format.
+    fn format(&self, format: Format) -> String {
+        match format {
+            Format::Default => self.to_string(),
+            Format::Json => serde_json::to_string(self).unwrap(),
```
claymcleod: Every other JSON serialization path in this file uses `.map_err(anyhow::Error::from)?`. Would you mind having this method return `Result<String>` or at minimum use `.expect("Rule should be JSON-serializable")` so the failure mode is documented?

src/commands/explain.rs
```diff
@@ -68,6 +94,157 @@ pub struct Args {
...
+        description: rule.description(),
+        explanation: rule.explanation(),
+        examples: rule.examples(),
+        url: None,
+        related: None,
+        config: None,
+    })
+}
+
+/// A line rule group tag.
```
claymcleod:
```suggestion
/// A lint rule group tag.
```

src/commands/explain.rs
```diff
@@ -68,6 +94,157 @@ pub struct Args {
...
+    /// The crate that the rule is defined in.
+    pub source: RuleSource,
+    /// The unique ID for the rule.
+    pub id: &'static str,
+    /// Tags the rule is grouped under, if the crate supports them.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub tags: Option<Vec<String>>,
+    /// A short description of the rule (possibly Markdown formatted).
+    pub description: &'static str,
+    /// An extended descriptions of the rule (possibly Markdown formatted).
```
claymcleod:
```suggestion
    /// An extended description of the rule (possibly Markdown formatted).
```

crates/wdl-lint/src/config.rs
Serial-ATA: Good catch, fixed that and added examples for the new lints. Should be good now

crates/wdl-analysis/src/rules.rs
```diff
@@ -43,6 +43,9 @@ pub trait Rule: Send + Sync {
     /// Get the long-form explanation of the rule.
     fn explanation(&self) -> &'static str;
 
+    /// Get a list of examples that would trigger this lint rule.
```
claymcleod:
```suggestion
    /// Get a list of examples that would trigger this rule.
```

PR #612, opened 2026-02-06
feat: expose `fallback_version` in `sprocket.toml`

claymcleod:
This PR surfaces the existing `fallback_version` option from `wdl-analysis` as a user-facing configuration under `[common.wdl]` in `sprocket.toml`. The option is threaded through all commands and the server so that it applies globally. As part of this work, the `apply` pattern—where CLI args were mutated with config values—has been replaced with resolved local variables at each command's entry point. CLI tests are included demonstrating `version development` failing without config and succeeding with `fallback_version = "1.2"`.

For all contributors:

### DIFF
--- a/crates/wdl-analysis/src/graph.rs
+++ b/crates/wdl-analysis/src/graph.rs
@@ -37,6 +37,7 @@ use wdl_ast::SyntaxNode;
 use crate::Config;
 use crate::IncrementalChange;
 use crate::document::Document;
+use crate::rules::USING_FALLBACK_VERSION;
 
 /// Represents space for a DFS search of a document graph.
 pub type DfsSpace =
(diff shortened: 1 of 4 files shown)


ahvigil:
> I don't think we need to tackle either of the suggested changes in #618 as part of this PR, but linking for posterity

actually looking at things closer, i see there is already a named rule for fallback version- its just not propagated to diagnostic message. I put in a small suggested fix for this branch at https://github.com/stjude-rust-labs/sprocket/pull/620

a-frantz:
> > I don't think we need to tackle either of the suggested changes in #618 as part of this PR, but linking for posterity
> 
> actually looking at things closer, i see there is already a named rule for fallback version- its just not propagated to diagnostic message. I put in a small suggested fix for this branch at #620

that's great! Appreciate it. Tagging @claymcleod . Can you merge #620 into this when you resolve the merge conflicts?

Issue #609, opened 2026-02-06
Remove formatting lints from `wdl-lint`


Serial-ATA:
Currently, `wdl-lint` concerns itself with formatting due to it predating `wdl-format`. Once `wdl-format` is in a better state, the following lints can be removed and have their suggestions be the default behavior:

- [ ] `CallInputSpacing`
- [ ] `CommentWhitespace`
- [ ] `ConsistentNewlines`
- [ ] `ElementSpacing`
- [ ] `EndingNewline`
- [ ] `ExpressionSpacing`
- [ ] `ImportSorted`
- [ ] `ImportWhitespace`
- [ ] `LineWidth`
- [ ] `MetaKeyValueFormatting`
- [ ] `TrailingComma`
- [ ] `VersionStatementFormatted`
- [ ] `Whitespace`

**Additional context**
#600

Issue #600, opened 2026-02-05
get sprocket format into better shape


a-frantz:
`format` has existed for a long time and has not yet reached a state where I'm excited to declare it "production ready". We are _very_ close to that point! But there are some issues that need to be addressed.

This is parent/tracking issue for high level discussion about the future of the command.

The first major issue is that there are niche scenarios where `format` produces awkward output. These are common enough that it has prevented us from adopting a `format check` on the [workflows repo](https://github.com/stjudecloud/workflows).
- the biggest culprit here is the line breaking logic #278 
  - #232 has "oddities". It looks like these are all caused by line breaking
  - If anyone out there reading this runs into other oddities, please share!

another big problem is lack of configuration. This was always a low priority/concern for the tool, but it is time for this to be implemented.
tangential philosophical rant:
> I'm a big believer in the idea that formatting should never get in the way of software development. It should be something that tooling _just handles_ and developers should be able to ignore the details. For that reason, we've never been concerned with adding much config to `format` because fundamentally we don't think devs should be mucking around with formatting options. But we do need to enable more fine tuning for our formatting that could "get in the way" of devs.
- #599
- there are probably other formatting decisions users want to opt out of, but I don't know what they are! `format` has mostly implemented my own preferences, so I'm happy with almost everything it does 😂 
  - other folks, please chime in on what you don't like so we can do better!

this last point is more about `lint` than `format`, but may illuminate some of our needed configuration options. `lint` predates `format` by a large amount of time, so `lint` has a lot of rules that should be handled by `format check` instead of `lint`. Once `format` is in a better place (by addressing the above problems), we should audit our lint rules and remove any that are purely about style and formatting. Rules that get removed from `lint` should probably end up getting configuration opt-outs in `format`.

a-frantz:
all sub issues are closed, so I consider this closed as well. More configuration is coming ( #1200 #1201 #1203 ) but those are improvements beyond what this issue was tracking.

Issue #592, opened 2026-02-04
lint for how doc comments + lint directives should be placed


a-frantz:
**Describe the solution you'd like**

doc comments should come before lint directives. This should be enforced with a new lint.

```wdl
## description
##
## long help text
#@ except: directives!
workflow foo {
}
```
instead of
```wdl
#@ except: directives!
## description
##
## long help text
workflow foo {
}
```

**Additional context**

see https://github.com/stjude-rust-labs/sprocket/pull/590#discussion_r2763780676

a-frantz:
Good questions! In thinking through the problem a bit more, I think this would make the most sense to implement as a `DocCommentFormatted` rule. We are still working through exactly how we want doc comments to look, so my vision is for this PR to initialize a new rule that will be expanded on in follow up PRs. This PR/new rule would highlight the doc comments as being out of place rather than the lint directives. A diagnostic may say "`Doc comments should come before lint directives`" or something like that.

>     * Scope: should this lint apply only to top-level items (workflow, task, struct), or also to sections (input/output/command/metadata)?

it should apply to anywhere doc comments are going to be supported, which is kind of an awkward requirement at the moment because we haven't finished out full support yet. #577 mentions the other contexts we want to support.

#590 is still in progress, but has to deal with some of this logic as well. Ideally, the set of "valid for doc comments" nodes would be defined somewhere centrally (`wdl-analysis` probably) and any lints that need to deal with doc comments could reference that central constant. But that doesn't exist yet! The location of that constant isn't worth blocking this issue over, so you can just take a similar approach as it seems #590 is taking where the set is defined locally with the lint rule. We can come back and refactor it as we add further support.

>     * Directives: should the rule consider only #@ except: directives, or any lint directive using the same prefix?

There aren't concrete plans for this yet, but I imagine in the future we will support directives other than `except` (maybe formatting directives?). So to future proof this rule, I would apply it to any comment starting with `#@`. A brief look at the code and it looks like we have defined the whole `#@ except:` prefix at https://github.com/stjude-rust-labs/sprocket/blob/main/crates/wdl-analysis/src/lib.rs#L66

It's a tangent for this issue, but I think reworking that to be more robust to future changes would be a small task you could fold into this PR (not a requirement for your PR, but would be nice to knock out both at once!)

>     * Severity: would warning be the intended level, or should this be a note?

note level please!

[...]

Issue #591, opened 2026-02-04
fully support doc comments


a-frantz:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.

There is discussion of adopting documentation comments (AKA doc comments) for WDL "officially" and deprecating WDL's meta sections (https://github.com/openwdl/wdl/issues/757). This would be a big change for the language and we should use Sprocket tooling to encourage this switch.

we recently added partial doc comment support ( #551 ) but there are a lot of other features and changes we will need to make to call doc comments fully supported.

**Describe the solution you'd like**

Break down of the features and changes needed per Sprocket command:
- most important is `doc` support!
    - #551 added partial supprt
    - #577 is tracking additional contexts we need to support
    - at the time of writing, we map doc comments into the existing `description`/`help` meta entries. I'd like to see us support other mappings into structured documentation (as opposed to doc comments just being free-flowing text/markdown)
- `lint`
    - existing lints make assumptions that `##` comments should only be used for preambles. This impacts multiple rules and we will need to refactor these to support `##` comments in more places
    - #570 
    - #557 
- `format`
    - would be very useful to keep comments under the max line length by adding line breaks into multi-line doc comments
        - should be able to refactor and re-use the logic implemented in #533 
- `analyzer`/`wdl-lsp`
    - the LSP parses meta sections for additional context on variables and the like. This feature will need to support doc comments as well
    - maybe related to #569 ?

**Additional context**

this is a tracking/parent issue for all the smaller pieces of the puzzle that we need to fit into place to have full support. It's a place for high level discussion of doc comments in Sprocket tooling.

PR #614, opened 2026-02-09
feat: improve format command

a-frantz:

The most impactful changes in the PR are improved formatting of doc comments and directive comments, and removing a big chunk of our "formatting only" lint rules. ~~Bonus last minute addition also disables input sorting by default. This is a pretty disruptive and arbitrary feature at least 2 users have expressed wanting to opt out of.~~

This also improves the `wdl-ast` and `wdl-analysis` APIs around "special" comment use and parsing (doc comments and directive comments). Particularly, it introduces a way for us to add new directives beyond just `except`. This is in preparation for formatting directives, although I haven't solidified that vision yet.

Relevant issues: #600 , #609 , #592 , #591 ,

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/tests/cli/format/check/format_check_does_not_overwrite_file/stderr
+++ b/tests/cli/format/check/format_check_does_not_overwrite_file/stderr
@@ -1,14 +1,11 @@
 --- input
 +++ formatted
 @@ -1,12 +1,10 @@
--## This file is a poorly formatted WDL file
--## The intention of this test is to show how the format --check command will just output 
--## how it would format the WDL file in stdout without overwriting the file
+ ## This file is a poorly formatted WDL file
+ ## The intention of this test is to show how the format --check command will just output 
+ ## how it would format the WDL file in stdout without overwriting the file
 -
 -    version 1.3
-+## This file is a poorly formatted WDL file The intention of this test is to show how the
-+## format --check command will just output how it would format the WDL file in stdout
-+## without overwriting the file
 +version 1.3
  
 -
--- a/tests/cli/format/overwrite/can_format_file_in_place/outputs/file.wdl
+++ b/tests/cli/format/overwrite/can_format_file_in_place/outputs/file.wdl
@@ -1,5 +1,6 @@
-## This file is a poorly formatted WDL file The intention of this test is to show how the
-## format --overwrite command can format the WDL file in place
+## This file is a poorly formatted WDL file
+## The intention of this test is to show how the format --overwrite command can format
+## the WDL file in place
 version 1.3
 
 workflow test {


crates/wdl-format/tests/format/seaseq-case/source.formatted.wdl
a-frantz, line 1: ~~Comments starting with `###` are identified as doc comments starting with a markdown header, and as such get a "blank" doc comment following them.~~

edit: I changed this behavior

crates/wdl-ast/src/lib.rs
```diff
@@ -563,6 +565,39 @@ impl<T: TreeToken> AstToken<T> for Whitespace<T> {
...
+}
+
+impl FromStr for Directive {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let trimmed = s.trim();
+        let (directive, contents) = trimmed
+            .split_once(DIRECTIVE_DELIMITER)
+            .ok_or(s.to_string())?;
```
peterhuene: This unnecessarily always allocates a new string. Prefer `ok_or_else` in this case.

crates/wdl-ast/src/lib.rs
```diff
@@ -563,6 +565,39 @@ impl<T: TreeToken> AstToken<T> for Whitespace<T> {
...
+/// A directive for WDL tools to respect.
+#[non_exhaustive]
+#[derive(Debug, PartialEq, Eq)]
+pub enum Directive {
+    /// Ignore any rules contained in the set.
+    Except(HashSet<String>),
+}
+
+impl FromStr for Directive {
+    type Err = String;
```
peterhuene: What `String` is being returned? It's a little odd for a `FromStr` implementation to return a copy of the original string on error.

If we're not intending to give a proper error message, then I would suggest the error type be `()`.

crates/wdl-ast/src/lib.rs
```diff
@@ -584,6 +619,30 @@ impl<T: TreeToken> AstToken<T> for Comment<T> {
...
+    /// Gets whether the comment starts with DIRECIVE_COMMENT_PREFIX.
+    pub fn is_directive(&self) -> bool {
+        self.text().starts_with(DIRECTIVE_COMMENT_PREFIX)
+    }
+
+    /// Gets the comment's rule exceptions set.
+    pub fn exceptions(&self) -> Option<HashSet<String>> {
+        match self
+            .text()
+            .strip_prefix(DIRECTIVE_COMMENT_PREFIX)?
```
peterhuene: On the surface, this `strip_prefix` feels like it should be part of parsing a comment directive, rather than a directive string being `: ...`.

crates/wdl-ast/src/lib.rs
```diff
@@ -563,6 +565,39 @@ impl<T: TreeToken> AstToken<T> for Whitespace<T> {
     }
 }
 
+/// The prefix for directive comments.
+pub const DIRECTIVE_COMMENT_PREFIX: &str = "#@";
+/// The delimiter between a directive and its contents
+pub const DIRECTIVE_DELIMITER: &str = ":";
+
+/// A directive for WDL tools to respect.
```
peterhuene:
```suggestion
/// A comment directive for WDL tools to respect.
```

crates/wdl-ast/src/lib.rs
```diff
@@ -584,6 +619,30 @@ impl<T: TreeToken> AstToken<T> for Comment<T> {
...
 }
 
+impl Comment {
+    /// Gets whether the comment starts with DIRECIVE_COMMENT_PREFIX.
+    pub fn is_directive(&self) -> bool {
+        self.text().starts_with(DIRECTIVE_COMMENT_PREFIX)
+    }
+
+    /// Gets the comment's rule exceptions set.
+    pub fn exceptions(&self) -> Option<HashSet<String>> {
```
peterhuene: This method feels like it should be `fn directive(&self) -> Option<Directive>` rather than specific to the exceptions directive; let the caller handle the type of directive it is.

crates/wdl-ast/src/lib.rs
```diff
@@ -584,6 +619,30 @@ impl<T: TreeToken> AstToken<T> for Comment<T> {
...
+            .text()
+            .strip_prefix(DIRECTIVE_COMMENT_PREFIX)?
+            .parse::<Directive>()
+        {
+            Ok(Directive::Except(e)) => Some(e),
+            _ => None,
+        }
+    }
+
+    /// Gets whether comment starts with DOC_COMMENT_PREFIX.
```
peterhuene:
```suggestion
    /// Gets whether comment starts with [`DOC_COMMENT_PREFIX`].
```

crates/wdl-analysis/src/lib.rs
```diff
@@ -81,42 +76,23 @@ pub trait SyntaxNodeExt {
 }
 
 impl SyntaxNodeExt for SyntaxNode {
-    fn except_comments(&self) -> impl Iterator<Item = SyntaxToken> + '_ {
+    fn rule_exceptions(&self) -> HashSet<String> {
```
peterhuene, line 84: Similar to my previous comment, I wonder if this should simply be `fn directives(&self) -> impl Iterator<Item = Directive>` and let callers filter out what directives are relevant.

crates/wdl-analysis/src/lib.rs
```diff
@@ -81,42 +76,23 @@ pub trait SyntaxNodeExt {
...
-        for comment in self.except_comments() {
-            if let Some(ids) = comment.text().strip_prefix(EXCEPT_COMMENT_PREFIX)
-                && ids.split(',').any(|i| i.trim() == id)
-            {
-                return true;
-            }
-        }
-
-        false
+        self.rule_exceptions().contains(id)
```
peterhuene, line 103: If we do change the above to `directives` we could simply do a `self.directives().any(|d| match d { Directive::Except(set) => set.contains(id) })`.

crates/wdl-engine/src/tree.rs
```diff
@@ -270,12 +270,8 @@ impl TreeNode for SyntaxNode {
     }
 }
 
+// TODO: find a way to remove this extension trait
```
peterhuene: We have two different concrete `SyntaxNode` types, so I don't think we can get ride of this extension trait.

If anything, perhaps we could just rename it to `Exceptable` with a default implementation of what we have here (no rules are excepted, empty set for exceptions) and implement the method only for `SyntaxNode` from `wdl_ast`.

crates/wdl-format/src/token/post.rs
```diff
@@ -77,6 +110,178 @@ impl Token for PostToken {
                     }
                     PostToken::TempIndent(value) => write!(f, "{value}"),
                     PostToken::Literal(value) => write!(f, "{value}"),
+                    PostToken::Documentation {
+                        num_indents,
+                        contents: markdown,
+                    } => {
+                        let prefix = DOC_COMMENT_PREFIX.to_string();
```
peterhuene: Why the `to_string` call?

crates/wdl-ast/src/lib.rs
```diff
@@ -563,6 +565,39 @@ impl<T: TreeToken> AstToken<T> for Whitespace<T> {
     }
 }
 
+/// The prefix for directive comments.
+pub const DIRECTIVE_COMMENT_PREFIX: &str = "#@";
+/// The delimiter between a directive and its contents
+pub const DIRECTIVE_DELIMITER: &str = ":";
+
+/// A directive for WDL tools to respect.
+#[non_exhaustive]
```
peterhuene: I don't think this needs to be `non_exhaustive`, personally.

crates/wdl-format/tests/format/doc_comments/source.formatted.wdl
```diff
@@ -0,0 +1,59 @@
+## Read groups are defined in the SAM spec:
+##
+## - `ID`: Read group identifier. Each Read Group must have a unique ID.
+## The value of ID is used in the RG tags of alignment records.
+## - `BC`: Barcode sequence identifying the sample or library. This value is the
+## expected barcode bases as read by the sequencing machine in the absence of errors. If
```
peterhuene: Generally, folks like to see list items indented consistently in Markdown or at least preserve the way they indented list items originally.

Personally, I'm wondering if we need to format Markdown comments at all.

I wonder if, instead, we could integrate with a crate like `rumdl` or some other Markdown formatter to format the comments? Perhaps as a future improvement.

crates/wdl-format/tests/format/seaseq-case/source.formatted.wdl
```diff
@@ -158,12 +158,11 @@ workflow seaseq {
 
     String pipeline_ver = "v2.0.0"
 
+    # Process SRRs
     ### ---------------------------------------- ###
     ### ------------ S E C T I O N 1 ----------- ###
     ### ------ Pre-process Analysis Files ------ ###
     ### ---------------------------------------- ###
-
-    # Process SRRs
```
peterhuene, line 166: The formatting differences here make me wonder if non-doc comments should come after doc comments.

In Rust, at least, I generally document an item with `///` and then if there are comments I don't want in the documentation I leave some trailing `//` comments at the end.

crates/wdl-ast/src/lib.rs
a-frantz: Honestly I copied this from somewhere else in the codebase 😅 specifically here - https://github.com/stjude-rust-labs/sprocket/blob/main/crates/wdl-grammar/src/version.rs#L91

peterhuene: Ugh, should probably fix it there too!

crates/wdl-format/tests/format/doc_comments/source.formatted.wdl
a-frantz: I particularly like the formatting of plain paragraphs so they are under the configured line width, but maybe that isn't worth added complexity/mucking up other markdown structures

crates/wdl-format/tests/format/seaseq-case/source.formatted.wdl
a-frantz: Maybe doc comments -> regular comments -> directive comments? I definitely think directives come last, the other two are swappable IMO

peterhuene: I think that would be my preference, at least.

crates/wdl-ast/src/lib.rs
a-frantz: I made the `Err = ()` change here. I tried to do the same in `wdl-grammar` and it seems `Err = String` is required by one of the `serde` derived traits, so I'm leaving unchanged

Issue #580, opened 2026-01-30
Task outputs that are symlinks to inputs do not work


peterhuene:
**Describe the bug**

Create a `repro.wdl` with the following contents:

```wdl
version 1.3

task repro {
    input {
        File file
    }

    command <<<
        ln -s '~{file}' foo
    >>>

    output {
        File out = "foo"
    }
}
```

Create an empty `input.txt` file.

Evaluate the task:

```
$ sprocket run repro.wdl -e test file=input.txt
```

The evaluation fails with:

```
error: failed to evaluate output `out` for task `repro`: file `/home/peterhuene/src/peterhuene/sprocket/runs/repro/2026-01-30_205239130599141/attempts/0/work/foo` does not exist
   ┌─ test.wdl:13:14
   │
13 │         File out = "foo"
   │              ^^^

error: aborting due to evaluation error
```

The symlink exists, but it points at a guest path and not a host path, which is why `sprocket` thinks the file doesn't exist:

```
$ ls -l foo 
lrwxrwxrwx 1 peterhuene peterhuene 28 Jan 30 15:52 foo -> /mnt/task/inputs/6/input.txt
```

**Expected behavior**

For task execution with a local work directory, `sprocket` should "fixup" any symlinks within the task's work directory that point at guest paths (i.e. point at an input file or a path within an input directory) and recreate the links to the corresponding host paths.

An evaluation error should occur if the symlink is _not_ to an input path and does not point to a file within the task's work directory.

**Related Bug**

The planetary transporter should also properly handle symlinks to inputs such that it can skip uploading the output and instead use the original input URL.

PR #621, opened 2026-02-11
fix: remap guest symlinks to localized files correctly.

peterhuene:
This fixes the guest symlink remapping that task evaluation performs such that if the guest path maps to a remote URL, the symlink is remapped to the localized path of the remote URL instead of the URL itself.

This is a continuation of the fix for #580.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -1121,11 +1121,27 @@ impl Evaluator {
...
+                    // Normalize any trailing slash
+                    if url.strip_suffix('/').unwrap_or(url)
+                        == host.strip_suffix('/').unwrap_or(host)
+                    {
+                        i.local_path()
+                    } else {
+                        None
+                    }
+                }) {
+                Cow::Borrowed(local_path)
```
claymcleod: If `find_map` returns `None` here (i.e., the host is a URL but no matching backend input was found), this falls through to the `else if` / `else` branches which call `Path::new(host.0.as_str())` with a URL string—constructing a nonsensical filesystem path. In practice this shouldn't happen, but it might be worth making this an explicit error rather than a silent fallthrough (e.g., bail if the host is a URL and the local path lookup fails).

claymcleod (approved):
Left one minor inline comment.

crates/wdl-engine/src/eval/v1/task.rs
peterhuene: I turned this into an error.

PR #628, opened 2026-02-12
refactor: split `wdl-doc` into a separate binary

Serial-ATA:
This moves the logic from `sprocket dev doc` to a separate `wdl-doc` binary, with `sprocket dev doc` now being a thin wrapper that invokes the binary.

With the split, `wdl-doc` is now responsible for emitting its diagnostics, so the `diagnostics` module of `sprocket` was split into a separate `wdl-diagnostics` crate.

For external contributors:

For all contributors:

      is implemented.


src/commands/doc.rs
```diff
@@ -1,255 +1,94 @@
...
-        return Err(e.into());
+        .with_diagnostics_config(DiagnosticsConfig::except_all())
+        .as_args()
+        .map_err(Into::<anyhow::Error>::into)?;
+    let mut command = Command::new(wdl_doc_bin);
+    command
+        .args(args.doc_args)
+        .arg(format!("--color={color_mode}"))
+        .arg(format!("--output={}", docs_dir.to_string_lossy()))
+        .arg(&docs_dir)
```
claymcleod: This passes `docs_dir` (the output directory) as the positional `workspace` argument to the child binary. The actual `workspace` variable is never forwarded. The child will analyze the output directory instead of the source workspace.

```suggestion
        .arg(&workspace)
```

src/commands/doc.rs
```diff
@@ -1,255 +1,94 @@
...
+        .env("WDL_ANALYSIS_ARGS", analysis_args)
+        .envs(std::env::vars())
```
claymcleod, line 76: `.envs(std::env::vars())` comes after the explicit `.env()` call. If the parent process already has `WDL_ANALYSIS_ARGS` set, the inherited value overwrites the carefully-constructed one. Either swap the order or filter `WDL_ANALYSIS_ARGS` out of the inherited vars.

src/commands/doc.rs
```diff
@@ -1,255 +1,94 @@
...
-    /// Enables support for documentation comments
-    ///
-    /// This option is *experimental* and will be removed in a future major
-    /// version. Follow the pre-RFC discussion here: <https://github.com/openwdl/wdl/issues/757>.
-    #[arg(long)]
-    pub with_doc_comments: bool,
 
     /// The report mode.
     #[arg(short = 'm', long, value_name = "MODE")]
     pub report_mode: Option<Mode>,
```
claymcleod, line 41: `report_mode` is accepted from the user via `--report-mode` / `-m` but is never forwarded to the child `wdl-doc` process. The flag is silently swallowed. Consider adding something like:

```rust
if let Some(mode) = args.report_mode {
    command.arg(format!("--report-mode={mode}"));
}
```

src/commands/doc.rs
```diff
@@ -1,255 +1,94 @@
...
-    }
-
-    if let Some(e) = counts.verify_no_errors() {
-        return Err(e.into());
+        .with_diagnostics_config(DiagnosticsConfig::except_all())
+        .as_args()
+        .map_err(Into::<anyhow::Error>::into)?;
+    let mut command = Command::new(wdl_doc_bin);
+    command
+        .args(args.doc_args)
```
claymcleod, line 71: `.args(args.doc_args)` is the first `.args()` call, placing user pass-through args before `--color`, `--output`, and the positional workspace. If `doc_args` contains positional arguments, clap in the child may bind them incorrectly. Move `.args(args.doc_args)` to after all wrapper-controlled arguments.

crates/wdl-analysis/src/config.rs
```diff
@@ -167,10 +171,98 @@ impl Config {
...
+        let mut args = String::new();
+        diagnostics.write_args(&mut args)?;
+        if let Some(fallback_version) = fallback_version {
+            write!(args, " --fallback-version={fallback_version}")?;
+        }
+        if let Some(ignore_filename) = ignore_filename {
+            write!(args, " --ignore-filename={ignore_filename}")?;
+        }
+        if !all_rules.is_empty() {
+            write!(args, " --rules=")?;
```
claymcleod: `as_args()` writes `--rules=` but `load()` registers `--all-rules`. The name mismatch means rules silently fail to round-trip. Additionally, the trailing comma in `write!(args, "{rule},")` produces a phantom empty-string element on deserialization.

crates/wdl-analysis/src/config.rs
```diff
@@ -224,6 +316,41 @@ impl FeatureFlags {
...
     }
+
+    /// Convert these feature flags into a string of arguments.
+    ///
+    /// See [`Config::as_args()`].
+    pub fn write_args(self, writer: &mut dyn Write) -> fmt::Result {
+        let FeatureFlags { wdl_1_3 } = self;
+
+        if wdl_1_3 {
+            write!(writer, "-Fwdl13")?;
```
claymcleod: This is written without a leading space. If any prior arguments exist, this concatenates onto the previous token (e.g., `--fallback-version=1.1-Fwdl13`). Add a leading space: `" -Fwdl13"`.

crates/wdl-analysis/src/config.rs
```diff
@@ -167,10 +171,98 @@ impl Config {
             inner: Arc::new(inner),
         }
     }
+
+    /// Convert the config into a string of arguments.
+    ///
+    /// These arguments are intended to be used in the `WDL_ANALYSIS_ARGS`
+    /// environment variable, loaded via [`Self::load()`].
+    pub fn as_args(&self) -> Result<String, fmt::Error> {
```
claymcleod: The `as_args()`/`load()` custom arg-string format is fragile (whitespace splitting, name mismatches, values with spaces, lossy round-trips for `wdl_1_3=false`). `Config` already derives `serde::Serialize`/`serde::Deserialize`—consider using `serde_json` for the env var instead. This would eliminate `getopts` as a dependency, remove the second CLI-parsing framework, and make the IPC contract type-safe by construction.

crates/wdl-analysis/src/config.rs
```diff
@@ -167,10 +171,98 @@ impl Config {
...
+                "",
+                "all-rules",
+                "A comma-separated list of rule names",
+                "<RULES>",
+            );
+
+        DiagnosticsConfig::apply_args(&mut options);
+        FeatureFlags::apply_args(&mut options);
+
+        let matches = options.parse(args.split(' '))?;
```
claymcleod: `args.split(' ')` produces empty tokens on multiple/leading/trailing spaces. Use `args.split_whitespace()` instead—it handles arbitrary whitespace and skips empties.

crates/wdl-doc/bin/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+        match e.kind() {
+            DocErrorKind::AnalysisFailed(analysis_results) => {
+                for result in analysis_results {
+                    let path = result.document().path().to_string();
+                    let source = result.document().root().text().to_string();
+
+                    emit_diagnostics(
+                        &path,
+                        source,
+                        result.document().diagnostics().filter(|d| {
```
claymcleod, line 227: The `.filter()` closure only passes through `Severity::Error` diagnostics and increments `counts.errors` as a side effect. Warnings and notes are silently dropped, making `report_mode` ineffective. Consider collecting all diagnostics, counting errors separately, and emitting the full set.

dev-proxy/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+use tempfile::tempdir;
+
+fn main() -> std::io::Result<()> {
+    let tmpdir = setup_temp()?;
+
+    let status = Command::new("cargo")
+        .args(["run", "-p", "sprocket"])
+        .args(std::env::args_os().skip(1))
+        .env("SPROCKET_DATA_DIR", tmpdir.path())
+        .envs(std::env::vars_os())
```
claymcleod, line 26: Same env-var ordering bug as the wrapper: `.env("SPROCKET_DATA_DIR", tmpdir.path())` followed by `.envs(std::env::vars_os())` allows the user's existing `SPROCKET_DATA_DIR` to overwrite the temp path. Swap the order.

dev-proxy/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    let target = std::env::var("CARGO_TARGET_DIR").map_or_else(
+        |_| {
+            Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("..")
+                .join("target")
+        },
+        PathBuf::from,
+    );
+    let profile_dir = if cfg!(debug_assertions) {
```
claymcleod, line 58: `cfg!(debug_assertions)` is evaluated at the proxy's compile time, not the component's. If the proxy is built in debug but `wdl-doc` was built in release (or vice versa), it looks in the wrong `target/` subdirectory. Consider accepting the profile as a CLI arg or detecting which directory contains the artifact.

src/commands/doc.rs
```diff
@@ -1,255 +1,94 @@
...
+        .arg(format!("--color={color_mode}"))
```
claymcleod: `format!("--color={color_mode}")` relies on `Display` or `Debug` for `ColorMode`, but clap's `ValueEnum` typically expects lowercase (`always`/`never`). If the formatting produces `Always`/`Never`, the child's clap parser will reject the argument. Use `clap::ValueEnum::to_possible_value()` to get the canonical string.

src/lib.rs
```diff
@@ -125,6 +156,12 @@ async fn real_main() -> CommandResult<()> {
         (ColorMode::Never, _) => false,
     };
 
+    let color_mode = if colorize {
```
claymcleod: The color_mode derivation only produces `Always` or `Never`, never `Auto`. The child is always forced into an explicit mode. Consider passing `auto` by default and letting the child detect terminal capabilities via `Stdio::inherit()`.

src/commands.rs
```diff
@@ -140,3 +156,51 @@ pub enum DevCommands {
...
+                    key.to_string_lossy(),
+                    value.to_string_lossy()
+                )),
+                None => envs_str.push_str(&format!("{} ", key.to_string_lossy())),
+            }
+        }
+
+        let args_os = self.0.get_args();
+        let mut args_str = String::new();
+        if args_os.len() > 0 {
```
claymcleod, line 187: `DebuggableCommand::fmt` has a double-space bug: `args_str` starts with `push(' ')` and then `args_spacer` adds another space. Also, `args_os.len() > 0` should be `!args_os.is_empty()`. Consider writing directly to the formatter `f` to avoid intermediate allocations.

claymcleod (changes requested):
A few thoughts on this PR — the binary split approach is solid, but there are some wiring issues in the thin wrapper that need attention before this is ready.

src/commands/doc.rs
Serial-ATA: The display impl is already lowercase

crates/wdl-doc/bin/src/main.rs
Serial-ATA: `wdl-doc` only ever produces error diagnostics

dev-proxy/src/main.rs
Serial-ATA: Personally, I don't see that as an issue. It's intended for developers of `sprocket`, so there may be a case for overriding `SPROCKET_DATA_DIR`. I'd rather just add a warning if it's already defined but not outright override it

claymcleod (changes requested):
Thanks for the work here, Alex—the goal of moving `wdl-doc` testing upstream (i.e., decoupling it from the `sprocket` crate so it can be built and tested independently) is a good one, and it has the nice side effect of making the individual binaries smaller. I'd like to see us move forward with that.

Upon closer review, I think `sprocketup` and the component management infrastructure (i.e., `dev-proxy`, `find_component`, `SPROCKET_DATA_DIR`-based discovery, profiles, manifests) adds more surface area than our small team can maintain right now. We'd be taking on a custom distribution system when existing package managers already solve this problem pretty well.

Here's what I think a leaner version of this PR looks like:

1. **Keep the binary split as-is**—`crates/wdl-doc/bin/` as a separate workspace member that depends only on `wdl-doc` (the library) and `wdl-analysis` for config. This achieves the independent testing goal.

2. **Keep `wdl-diagnostics`**—the shared diagnostic reporting crate is useful regardless.

3. **Lean on existing distribution channels instead of sprocketup:**
   - **`cargo install`**: `cargo install sprocket wdl-doc`
   - **Homebrew**: add a `bin.install "target/release/wdl-doc"` line to the existing formula so both binaries are installed together
   - **GitHub releases**: update the release workflow to build `wdl-doc` alongside `sprocket` and include both binaries in the release archive

4. **Discover `wdl-doc` via PATH** instead of the bin-directory scanning approach. `sprocket dev doc` calls `which wdl-doc` (or checks `PATH`), and if it's not found, emits a helpful error like:

   > `wdl-doc` was not found. You can install it with `cargo install wdl-doc` or download it from https://github.com/stjude-rust-labs/sprocket/releases/latest.

5. **Drop** `sprocketup/`, `dev-proxy/`, `find_component`, `sprocket_bin_dir`/`sprocket_data_dir` (for this purpose), the `.cargo/config.toml` alias, and the `--data-dir` global arg.

This keeps the architectural win (independent `wdl-doc` binary with upstream testing and smaller binaries) without the maintenance burden of a custom installer. We can always revisit a component manager later if the project grows to need one.

PR #631, opened 2026-02-13
refactor(wdl-doc): convert struct and enum members to tables

Serial-ATA:
This converts the member lists of structs and enums to tables to match workflows/tasks:

image
image

For external contributors:

For all contributors:

      is implemented.


crates/wdl-doc/src/meta.rs
```diff
@@ -126,15 +127,7 @@ impl MetaMapExt for MetaMap {
...
-            .get(DESCRIPTION_KEY)
-            .map(|v| match v {
-                MetaMapValueSource::MetaValue(MetadataValue::String(s)) => {
-                    let t = s.text().expect("meta string should not be interpolated");
-                    t.text().to_string()
-                }
-                MetaMapValueSource::Comment(s) => s.to_string(),
-                _ => "ERROR: description not of type String".to_string(),
-            })
+            .full_description()
```
claymcleod: Switching to `full_description()` here changes the behavior of `render_description()` — it now includes help text (from `HELP_KEY`) concatenated with the description. For individual struct members and enum choices this is fine (the old code also used `full_description()`), but at the top level (enum/struct page), `render_remaining()` also renders help as a separate section, which could cause duplication.

crates/wdl-doc/src/meta.rs
```diff
@@ -114,6 +114,7 @@ impl MetaMapExt for MetaMap {
         if let Some(mut description) = self.get(DESCRIPTION_KEY).and_then(MetaMapValueSource::text)
         {
             if let Some(help) = help {
+                description.push('\n');
```
claymcleod: Two consecutive `push('\n')` calls — consider using `push_str("\n\n")`.

crates/wdl-doc/src/enum.rs
```diff
@@ -82,20 +76,29 @@ impl Enum {
...
+                        div class="main__grid-header-cell" { "Name" }
+                        div class="main__grid-header-cell" { "Description" }
+                        div class="main__grid-header-separator" {}
+                        @for choice in self.choices.iter() {
+                            @let choice_name = choice.choice.name();
+                            @let choice_id = format!("choice.{}", choice_name.text());
+                            @let choice_anchor = format!("#{choice_id}");
+                            div id=(choice_id) class="main__grid-row" x-data="{ description_expanded: false }" {
+                                div class="main__grid-cell" {
+                                    a href=(choice_anchor) class="hidden" {}
```
claymcleod: The anchor element has `class="hidden"` here but the equivalent in `struct.rs` (line 122) has no class on its anchor. These should be consistent. Since the parent `div` already carries the `id` for fragment linking, consider removing the anchor in both files, or applying `class="hidden"` in both.

crates/wdl-doc/src/enum.rs
```diff
@@ -82,20 +76,29 @@ impl Enum {
...
+                                    (choice_name.text())
```
claymcleod: Enum choice names are rendered as plain text, but struct member names (in `struct.rs`, line 123) use `code { }` wrapping. For visual consistency across all grid layouts, enum choices should also use `code { }`.

crates/wdl-doc/src/meta.rs
```diff
@@ -455,6 +448,7 @@ pub(crate) fn doc_comments<T: TreeToken + SyntaxTokenExt>(token: &T) -> MetaMap
     let help = paragraphs.into_iter().fold(String::new(), |mut acc, p| {
         if !acc.is_empty() {
             acc.push('\n');
+            acc.push('\n');
```
claymcleod: Same as the earlier comment — consider using `push_str("\n\n")`.

crates/wdl-doc/theme/src/main.css
```diff
@@ -301,14 +307,18 @@
   .main__summary-container {
     @apply inline;
   }
-  .main__meta-item-member {
-    @apply mb-[0.75em];
+  .main__grid-full-width-cell p,
+  .main__grid-cell p {
+    margin-top: revert;
```
claymcleod: `margin-top: revert` and `margin-bottom: revert` revert to browser defaults, which vary across browsers. Consider using explicit values (e.g., `1em`) for consistent cross-browser rendering.

claymcleod (dismissed):
Here are some thoughts.

crates/wdl-doc/src/enum.rs
Serial-ATA: Wasn't sure whether or not choices looked better in code blocks. I guess it looks alright either way

a-frantz (approved):
LGTM 👍 

I want to take a pass through `wdl-doc` and refactor the parts related to `*Meta*` so that doc comments can be better integrated. That is largely unrelated to this PR, but pointing it out as on my agenda so we don't end up with multiple cooks in the kitchen 🧑‍🍳

crates/wdl-doc/src/parameter.rs
```diff
@@ -246,10 +248,10 @@ impl Parameter {
                     div class="main__grid-cell" { (self.render_expr(true)) }
                 }
                 div class="main__grid-cell" {
-                    (self.description(true))
+                    (self.meta().render_full_description(true))
                 }
                 div x-show="description_expanded" class="main__grid-full-width-cell" {
-                    (self.description(false))
+                    (self.meta().render_full_description(false))
                 }
```
claymcleod, line 221: Consider keeping `render_description()` here rather than `render_full_description()`. Parameters already render `help` via `render_remaining_meta()` in the additional metadata panel, so using `render_full_description()` in the grid cell would show `help` text twice for parameters with multi-paragraph doc comments.

```suggestion
                div class="main__grid-cell" {
                    (self.render_description(true))
                }
                div x-show="description_expanded" class="main__grid-full-width-cell" {
                    (self.render_description(false))
                }
```

crates/wdl-doc/src/meta.rs
```diff
@@ -96,25 +96,52 @@ pub(crate) trait MetaMapExt {
...
+    if !summarize {
+        return Markdown(text).render();
+    }
+
+    match summarize_if_needed(text, DESCRIPTION_MAX_LENGTH, DESCRIPTION_CLIP_LENGTH) {
+        MaybeSummarized::No(desc) => Markdown(desc).render(),
+        MaybeSummarized::Yes(summary) => {
+            html! {
+                div class="main__summary-container" {
+                    (Markdown(summary))
```
claymcleod, line 128: I noticed the summarized text path changed from raw text rendering `(summary)` to `(Markdown(summary))`. This is more consistent with the non-summarized path, but the Markdown renderer will wrap the truncated text in `<p>` tags inside the `display: inline` `main__summary-container` — worth confirming this doesn't affect the visual layout.

claymcleod (commented):
Nice refactoring — the grid tables are a big improvement in visual consistency. A few observations.

crates/wdl-doc/src/struct.rs
```diff
@@ -111,20 +107,31 @@ impl Struct {
...
+                div class="main__grid-container" {
+                    div class="main__grid-struct-member-container" {
+                        div class="main__grid-header-cell" { "Name" }
+                        div class="main__grid-header-cell" { "Type" }
+                        div class="main__grid-header-cell" { "Description" }
+                        div class="main__grid-header-separator" {}
+                        @for member in self.members.iter() {
+                            @let member_name = member.decl.name();
+                            @let member_id = format!("member.{}", member_name.text());
+                            div id=(member_id) class="main__grid-row" x-data="{ description_expanded: false }" {
```
claymcleod, line 119: The old rendering had `<a href="#{member_id}">` inside each section for self-linking. The new grid rows have `id` attributes (so fragment URLs still work for incoming links), but there's no clickable anchor element anymore. The parameter tables don't have self-links either, so this is consistent — just flagging in case the old behavior was intentional.

crates/wdl-doc/src/parameter.rs
Serial-ATA: yeah, talked to Ari about this before. it makes sense for `description` and `help` to be combined into a full description since they logically go together (at least in the case of doc comments). but that'll need some design changes since that would potentially mean having two expand buttons (one for the full description and one for the additional meta).

crates/wdl-doc/src/meta.rs
Serial-ATA: Seems fine after some quick tests

crates/wdl-doc/src/struct.rs
Serial-ATA: Yeah, the clickable anchors made more visual sense when the member names were headers. I think it'll make more sense to add back when we figure out how we want to render the full descriptions

claymcleod: Ok, makes sense

crates/wdl-doc/src/parameter.rs
```diff
@@ -84,6 +85,12 @@ pub(crate) struct Parameter {
     io: InputOutput,
 }
 
+impl DefinitionMeta for Parameter {
+    fn meta(&self) -> &MetaMap {
+        &self.meta
+    }
```
claymcleod, line 91: Now that `Parameter` implements `DefinitionMeta`, the inherent `meta()` method at line 106 is redundant — could it be removed?

claymcleod (commented):
Looks good, one small note and then we can approve.

Issue #622, opened 2026-02-11
feat: add a `sprocket submit` command for submitting workflows to a running server


claymcleod:
It would be nice to have a `sprocket submit` command that submits workflows to a running `sprocket server` instance. Right now, the only way to do this is through the REST API directly (e.g., with `curl`), which isn't the most ergonomic experience. A dedicated CLI command would make it much easier to interact with a remote server without having to construct HTTP requests manually.

Suggested by @adthrasher in https://github.com/stjude-rust-labs/sprocket.bio/pull/30#discussion_r2759436541.

BKDaugherty:
I think this issue should be closed - I forgot to mention in my [PR](https://github.com/stjude-rust-labs/sprocket/pull/633) that it should close this issue, but @claymcleod ended up carrying over the finish line!

claymcleod:
Thanks for the reminder!

Issue #583, opened 2026-01-30
Consider splitting server into its own crate/executable


claymcleod:
As noted by @peterhuene in #540:

> I am growing concerned with how much "kitchen sink" functionality is built into the `sprocket` executable, the time it takes to build sprocket (and its hundreds of dependencies currently), and the ultimate size of the `sprocket` executable. I could easily see `sprocket analyzer` and `sprocket server` potentially spun off into their own executables, for example.
>
> To that end, I think it would be ideal to move the server implementation into its own crate rather than living inside of the sprocket crate.

This issue tracks the potential refactoring to:

- Move the server implementation into its own crate (e.g., `sprocket-server`)
- Consider whether `sprocket analyzer` should also be split out
- Reduce build times and binary size for users who don't need all features

PR #633, opened 2026-02-14
feat(sprocket): Add submit subcommand

BKDaugherty:
Adds a `submit` subcommand as a thin wrapper around the Sprocket Server's REST API.

https://github.com/stjude-rust-labs/sprocket/issues/622

For external contributors:

For all contributors:

      is implemented.


src/commands.rs
```diff
@@ -137,6 +138,8 @@ pub enum DevCommands {
     Lock(lock::Args),
     /// Runs the HTTP API server for run execution.
     Server(server::Args),
+    /// Submit a workflow to a Sprocket HTTP API server.
+    Submit(submit::Args),
```
BKDaugherty, line 142: I'd imagine this could be `sprocket client` instead of `sprocket submit` to then allow for hitting the other endpoints with a consistent wrapper?

(Similar question to here: https://github.com/stjude-rust-labs/sprocket/issues/622#issuecomment-3901034861)

src/server.rs
```diff
@@ -20,6 +20,9 @@ use crate::system::v1::exec::svc::RunManagerSvc;
 mod api;
 
 pub use api::AppState;
+pub use api::v1::error::ErrorResponse;
+pub use api::v1::runs::SubmitResponse;
+pub use api::v1::runs::SubmitRunRequest;
```
BKDaugherty: @claymcleod - how do you feel about making these public so that the client can re-use the types?

I noted that there's a tentative refactor to split the server into it's own crate / executable in https://github.com/stjude-rust-labs/sprocket/issues/583.

Perhaps it makes sense then to split the server into a `sprocket-api` crate for the types and then the server binary and client binaries can both use that if we decide it's cool to re-use these?

claymcleod (commented):
Sorry for the slow review on this — nice work! A few suggestions.

src/commands/submit.rs
```diff
@@ -0,0 +1,155 @@
...
+        let error_response: ErrorResponse = resp
+            .json()
+            .await
+            .context("Expected a JSON API Error Response from the Sprocket API")?;
+
+        return Err(CommandError::Single(anyhow::anyhow!(
+            "Error from Sprocket Server ({}): {}",
+            error_response.kind,
+            error_response.message
+        )));
```
claymcleod: If the server returns a non-JSON body (e.g., a proxy 502 HTML page), this deserialization fails with a confusing serde error. Consider falling back to the raw response text:

```suggestion
    if !resp.status().is_success() {
        let status = resp.status();
        let body = resp.text().await.unwrap_or_default();
        let msg = serde_json::from_str::<ErrorResponse>(&body)
            .map(|e| format!("{}: {}", e.kind, e.message))
            .unwrap_or_else(|_| format!("HTTP {status}: {body}"));

        return Err(CommandError::Single(anyhow::anyhow!(msg)));
    }
```

src/commands/submit.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/// Handles the `submit` subcommand.
+///
+/// Submits a workflow to a Sprocket server based on the Args / Config.
+pub async fn submit(args: Args, config: Config) -> Result<(), CommandError> {
+    let host = args.client_args.host.unwrap_or(config.server.host);
+    let port = args.client_args.port.unwrap_or(config.server.port);
+
+    let url = format!("http://{host}:{port}/api/v1/runs", host = host, port = port);
```
claymcleod: Consider extracting the base URL logic into `SprocketClientConnectionArgs` so future commands (e.g., status, cancel) can reuse it:

```rust
impl SprocketClientConnectionArgs {
    fn base_url(&self, config: &Config) -> String {
        let host = self.host.as_deref().unwrap_or(&config.server.host);
        let port = self.port.unwrap_or(config.server.port);
        format!("http://{host}:{port}")
    }
}
```

src/commands/submit.rs
```diff
@@ -0,0 +1,155 @@
...
+        } = value;
+
+        let inputs = match (inputs, inputs_path) {
+            (Some(value), None) => {
+                serde_json::from_str(&value).context("Parsing workflow inputs as json")?
+            }
+            (None, Some(path_to_inputs)) => serde_json::from_reader(
+                &File::open(path_to_inputs).context("Opening provided inputs file")?,
+            )?,
+            otherwise => {
```
claymcleod: This arm is unreachable given `required_unless_present` + `conflicts_with` — consider 

```
unreachable!("clap enforces exactly one of `--inputs` or `--inputs-path`")
``` 

to communicate the invariant.

src/server.rs
claymcleod: These can be `pub(crate)` since the only consumer is the `submit` command within this same binary crate.

```suggestion
pub(crate) use api::v1::error::ErrorResponse;
pub(crate) use api::v1::runs::SubmitResponse;
pub(crate) use api::v1::runs::SubmitRunRequest;
```

src/commands.rs
claymcleod: I think `submit` works well as-is — it's clear and direct. If we end up wrapping more endpoints later, we can reconsider the hierarchy then.

src/server.rs
claymcleod: Re-exporting via `pub(crate) use` is the right call for now. It indicates the intent that these shouldn't be used outside of `sprocket` and it can be cleaned up later into like a `sprocket-api-types` crate.

src/commands/submit.rs
```diff
@@ -0,0 +1,155 @@
+//! Implementation of the `submit` subcommand
```
claymcleod:
```suggestion
//! Implementation of the `submit` subcommand.
```

src/commands/submit.rs
```diff
@@ -0,0 +1,155 @@
...
+    client_args: SprocketClientConnectionArgs,
+
+    #[command(flatten)]
+    run_request_args: SubmitRunRequestArgs,
+}
+
+/// Handles the `submit` subcommand.
+///
+/// Submits a workflow to a Sprocket server based on the Args / Config.
+pub async fn submit(args: Args, config: Config) -> Result<(), CommandError> {
```
claymcleod: A few higher-level thoughts on the UX:

- **Positional args for the happy path.** Right now every arg requires a `--flag`, which makes the common case verbose (`sprocket dev submit --source workflow.wdl --inputs-path inputs.json`). Consider making `source` positional and giving `inputs-path` a short alias (`sprocket dev submit workflow.wdl -i inputs.json`).
- **Local validation before submit.** Sprocket already has a full WDL analyzer — this command could validate the WDL and inputs locally before sending to the server. That would catch errors early with better diagnostics than the server's HTTP error responses, and it's where the real value of a dedicated `submit` command over `curl` comes from.
- **Machine-parseable output.** Consider just writing the server's JSON response directly to stdout rather than formatting it. That way the output is naturally composable with other tools (`sprocket dev submit ... | jq .uuid`).

src/commands/submit.rs
```diff
@@ -0,0 +1,201 @@
...
+            // SAFETY: we wouldn't have a workflow inputs if a workflow didn't
+            // exist that matched the user's criteria.
+            let workflow = document.workflow().unwrap();
+            inputs.validate(document, workflow, None)?
+        }
+    }
+
+    let target_json_inputs = serde_json::from_str(&inputs_to_json(&target, &inputs)?)
+        .context("Deserializing previously serialized inputs shouldn't fail")?;
+
```
BKDaugherty, line 142: We incur an extra json serialization step here largely to re-use the `inputs_to_json` function from `run`. This felt worth it to keep the "attaching the target prefix to each key" logic in one place, but I could refactor a bit more if we wanted this to be less wasteful.

src/commands/submit.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    let (target, inputs) = resolve_target_and_inputs(
+        &args.run_request_args.inputs,
+        args.run_request_args.target.clone(),
+        &document,
+    )
+    .await?;
+
+    match &inputs {
```
BKDaugherty: I copied this from `validate`. 

I could pull this into a helper function similar to `resolve_target_and_inputs`, but I was hesitant to do so given the assumptions that this block makes. 

See also the more meta comment here: https://github.com/stjude-rust-labs/sprocket/pull/633/changes#r3004976295

src/commands/submit.rs
```diff
@@ -0,0 +1,201 @@
...
+    client_args: SprocketClientConnectionArgs,
+
+    #[command(flatten)]
+    run_request_args: SubmitRunRequestArgs,
+}
+
+/// Handles the `submit` subcommand.
+///
+/// Submits a workflow to a Sprocket server based on the Args / Config.
+pub async fn submit(args: Args, config: Config, colorize: bool) -> CommandResult<()> {
```
BKDaugherty, line 102: In general in this function, I copied a fair bit from check and validate. We could opt to just call those functions in their entirety and then rebuild the inputs?

I'm not sure if we really care about performance here given what I imagine this will be used for, so it might make sense to build an analysis of the given document multiple times as a trade off for less code / more clarity. Open to whatever y'all think!

src/commands/submit.rs
```diff
@@ -0,0 +1,201 @@
...
+    // Ensure the document is valid before sending to the server.
+    let results = Analysis::default()
+        .add_source(args.run_request_args.source.clone())
+        .fallback_version(config.common.wdl.fallback_version)
+        .run()
+        .await
+        .map_err(CommandError::from)?;
+
+    // SAFETY: this must exist, as we added it as the only source to be analyzed
```
BKDaugherty: This is all pretty much copied from the `sprocket check` implementation.

See also my more meta comment here: https://github.com/stjude-rust-labs/sprocket/pull/633/changes#r3004976295

src/commands/submit.rs
BKDaugherty: > Local validation before submit. Sprocket already has a full WDL analyzer — this command could validate the WDL and inputs locally before sending to the server. That would catch errors early with better diagnostics than the server's HTTP error responses, and it's where the real value of a dedicated submit command over curl comes from.

Did this with the current patch. Is this how you were thinking it would be done? I'm happy to either refactor `check` / `validate` to better re-use the logic, or directly call those methods if that would be better?

See also: https://github.com/stjude-rust-labs/sprocket/pull/633#discussion_r3004976295

src/commands/submit.rs
```diff
@@ -0,0 +1,396 @@
...
+            source,
+            diagnostics,
+            &[],
+            args.run_request_args.report_mode.unwrap_or_default(),
+            colorize,
+        )
+        .context("failed to emit diagnostics")?;
+
+        return Err(anyhow::anyhow!(
+            "Failed to submit WDL document to server due to analysis errors."
```
claymcleod:
```suggestion
            "failed to submit WDL document to server due to analysis errors"
```

src/commands/submit.rs
```diff
@@ -0,0 +1,396 @@
...
+        target: args.run_request_args.target,
+        index_on: args.run_request_args.index_on,
+    };
+
+    let resp = reqwest::Client::new()
+        .post(url)
+        .json(&request)
+        .send()
+        .await
+        .context("Sending Request")?;
```
claymcleod:
```suggestion
        .context("sending request")?;
```

src/commands/submit.rs
```diff
@@ -0,0 +1,396 @@
...
+        Inputs::Workflow(inputs) => {
+            // SAFETY: we wouldn't have a workflow inputs if a workflow didn't
+            // exist that matched the user's criteria.
+            let workflow = document.workflow().unwrap();
+            inputs.validate(document, workflow, None)?
+        }
+    }
+
+    let target_json_inputs = serde_json::from_str(&inputs_to_json(&target, &inputs)?)
+        .context("Deserializing previously serialized inputs shouldn't fail")?;
```
claymcleod:
```suggestion
        .context("deserializing previously serialized inputs shouldn't fail")?;
```

src/commands/submit.rs
```diff
@@ -0,0 +1,396 @@
...
+        let msg = serde_json::from_str::<ErrorResponse>(&body)
+            .map(|e| format!("{}: {}", e.kind, e.message))
+            .unwrap_or_else(|_| format!("HTTP {status}: {body}"));
+        return Err(CommandError::Single(anyhow::anyhow!(msg)));
+    }
+
+    let submit_response: serde_json::Value = resp
+        .json()
+        .await
+        .context("Expected a response body for successful SubmitRunRequest")?;
```
claymcleod:
```suggestion
        .context("expected a response body for successful `SubmitRunRequest`")?;
```

src/commands/submit.rs
```diff
@@ -0,0 +1,396 @@
...
+
+impl SprocketClientConnectionArgs {
+    fn base_url(&self, config: &Config) -> String {
+        let host = self.host.as_deref().unwrap_or(&config.server.host);
+        let port = self.port.unwrap_or(config.server.port);
+        format!("http://{host}:{port}")
+    }
+}
+
+/// CLI Arguments for specifying the body of the SubmitRunRequest.
```
claymcleod:
```suggestion
/// CLI arguments for specifying the body of the `SubmitRunRequest`.
```

src/commands/validate.rs
```diff
@@ -48,39 +49,27 @@ pub struct Args {
...
-
-    // SAFETY: this must exist, as we added it as the only source to be analyzed
-    // above.
-    let document = results.filter(&[&args.source]).next().unwrap().document();
-
-    let (target, inputs) = match Invocation::coalesce(&args.inputs, args.target.clone())
+/// Accepts an optional target and set of inputs, and resolves them for
+/// validation.
+pub async fn resolve_target_and_inputs(
+    input_args: &Vec<String>,
```
claymcleod: We should use `&[String]` instead of `&Vec<String>` here—Clippy's `ptr_arg` lint flags this.

```suggestion
    input_args: &[String],
```

claymcleod (changes requested):

src/commands/submit.rs
claymcleod: This is fine for now—we'll plan a consolidation of the shared analysis→validate pipeline in a follow-up.

src/commands/submit.rs
claymcleod: This trade-off is fine—keeping the target-prefix logic in one place is worth the extra roundtrip.

src/commands/submit.rs
claymcleod: Yes, this is what I had in mind. All three points are addressed—thanks for working through these.

claymcleod:
@BKDaugherty just a heads up, I plan to make these final fixes, nothing for you to do.

BKDaugherty:
> @BKDaugherty just a heads up, I plan to make these final fixes, nothing for you to do.

Thanks @claymcleod!

src/commands/submit.rs
BKDaugherty: Sweet! Sounds good!

claymcleod (approved):

Issue #278, opened 2025-01-15
more sophisticated line breaking in `wdl-format`


a-frantz:
stjude-rust-labs/wdl#242 introduces a `MaxLineLength` configuration for `wdl-format`. This is a great improvement, but leaves some more to be desired. It leads to some awkward line breaks. This can be improved by making the algorithm a little more sophisticated.

Potential improvements:
- a priority list of potential line breaks for each line we can backtrack through and select "the most desirable" token to line break on.
- a stack for "matching tokens" that should be line broken in tandem (open/close parenthesis/brackets etc.)

mcovarr:
On sprocket 0.27.0, this input:

```
version 1.3

workflow ArrayFormat {
  Array[Int] nums = [1, 2, 3, 4, 5]
}
```

run through `sprocket format view --max-line-length 240` renders as
```
version 1.3

workflow ArrayFormat {
    Array[Int] nums = [
        1,
        2,
        3,
        4,
        5,
    ]
}
```

i.e. array elements seem to always be put on their own line regardless of the line length allowance. Would it be possible to consider the maximum line length allowance when deciding where to line break collections?

Serial-ATA:
@mcovarr see also: #666

mcovarr:
@Serial-ATA Thanks!

Looking into this further, the "single element per line regardless of width" behavior appears to be shared by a small family: `Array [ … ]`, `Map { … }`, struct literals, and call input blocks all seem to expand to one element per line regardless of `--max-line-length`.

But personally it's only the arrays that seem off, as those tend to be short, homogeneous scalar lists where having a single element per line feels jarring, as in the example above. At least to me, maps, struct literals, and call-input blocks are keyed / record-like and usually read well (or at least okay) with a single element per line.

Serial-ATA:
Yeah, all those things fall into the same behavior unfortunately. I never got around to looking into it, but based on https://github.com/stjude-rust-labs/sprocket/issues/666#issuecomment-3962316367 it looks like it'd be non-trivial to fix.

Issue #232, opened 2025-07-02
format oddities


adthrasher:
I applied the `format` command to the SEAseq repo and found several odd outputs.
```wdl
         # create bowtie index on spikein genome
-        call bowtie.index as spikein_bowtie_idx {
-            input :
-                reference=select_first([spikein_reference, string_spikein])
-        }
+        call bowtie.index as spikein_bowtie_idx { input: reference = select_first([
+            spikein_reference,
+            string_spikein,
+        ]) }
     }
```
```wdl
     # FASTA faidx and chromsizes and effective genome size
     call samtools.faidx as samtools_faidx {
         # create FASTA index and chrom sizes files
-        input :
-            reference=reference
-    }
+        input: reference = reference }
     call util.effective_genome_size as egs {
```
```wdl
         Array[File] sample_R2_fastqfile = s_R2_fastq
 
         # Order FASTQs
         if (length(sample_R1_fastqfile) > 1) {
-            call peaseq_util.sortfiles as R1_sorted { input: fastqfiles=sample_R1_fastqfile }
-            call peaseq_util.sortfiles as R2_sorted { input: fastqfiles=sample_R2_fastqfile }
+            call peaseq_util.sortfiles as R1_sorted { input: fastqfiles = sample_R1_fastqfile
+                }
+            call peaseq_util.sortfiles as R2_sorted { input: fastqfiles = sample_R2_fastqfile
+                }
         }
```
```wdl
 
     # transpose to paired-end tuples
-    Array[Pair[File, File]] original_sample_fastqfiles = zip(original_sample_R1, original_sample_R2)
+    Array[Pair[File, File]] original_sample_fastqfiles = zip(original_sample_R1, original_sample_R2
+        )
 ```

a-frantz:
closed by #641 which resolved each of these examples

PR #641, opened 2026-02-17
feat(format): tandem line breaking on matching tokens

a-frantz:

tandem line breaks for matching tokens. This handles a bunch of the cases present in #232 and is part of the long awaited #278

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-format/src/token/post.rs
+++ b/crates/wdl-format/src/token/post.rs
@@ -297,6 +297,21 @@ fn tandem_line_break(kind: SyntaxKind) -> Option<SyntaxKind> {
     }
 }
 
+/// Tracks a tandem break.
+struct TandemBreak {
+    /// The [`SyntaxKind`] which opened this tandem break.
+    pub open: SyntaxKind,
+    /// The [`SyntaxKind`] which will close this tandem break.
+    pub close: SyntaxKind,
+    /// Token depth since opening the break.
+    ///
+    /// The close break is only added when `depth == 0`.
+    /// This is incremented by one for every token matching `open` after the
+    /// break is initiated. It is decremented by one for every token
+    /// matching `close` after the break is initiated.
+    pub depth: usize,
+}
+
 /// Current position in a line.
 #[derive(Default, Eq, PartialEq)]
 enum LinePosition {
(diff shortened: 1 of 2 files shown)


crates/wdl-format/src/token/post.rs
```diff
@@ -284,6 +284,19 @@ fn can_be_line_broken(kind: SyntaxKind) -> Option<LineBreak> {
...
 
+/// Gets the corresponding [`SyntaxKind`] that should be line broken in tandem
+/// with the provided [`SyntaxKind`].
+fn tandem_line_break(kind: SyntaxKind) -> Option<SyntaxKind> {
+    match kind {
+        SyntaxKind::OpenBrace => Some(SyntaxKind::CloseBrace),
+        SyntaxKind::OpenBracket => Some(SyntaxKind::CloseBracket),
+        SyntaxKind::OpenParen => Some(SyntaxKind::CloseParen),
+        SyntaxKind::OpenHeredoc => Some(SyntaxKind::CloseHeredoc),
+        SyntaxKind::PlaceholderOpen => Some(SyntaxKind::CloseBracket),
```
claymcleod: This should be `CloseBrace`, not `CloseBracket`.

```suggestion
        SyntaxKind::PlaceholderOpen => Some(SyntaxKind::CloseBrace),
```

crates/wdl-format/src/token/post.rs
```diff
@@ -603,6 +624,12 @@ impl Postprocessor {
                     pre_buffer.peek().map(|(_, v)| &**v),
                     &mut post_buffer,
                 );
+
+                if let Some(also_break_on) =
+                    tandem_line_break(*potential_line_breaks.get(&i).unwrap())
```
claymcleod: nit: This unwrap() is safe because `i` is guaranteed to be in `potential_line_breaks` when `self.interrupted` is true. A `// SAFETY:` comment documenting that invariant would help future readers.

crates/wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -1780,7 +1784,8 @@ workflow chip {
     }
 
     Boolean has_all_input_of_choose_ctl = length(select_all(ta_)) == num_rep && length(
-        select_all(ctl_ta_)) == num_ctl && num_ctl > 0
+        select_all(ctl_ta_
```
claymcleod: This one doesn't look better than the previous one to my eye. Is there any way to keep the one `)` at the end of this and just pull the `)` from `length(` down to the next line?

I won't hold the PR up because of this, but it would be ideal.

claymcleod: I think the reason is because the concept of matching token depth is not accounted for here. We could solve this by making `TandemEntry` also store a `depth` (incrementing `depth` on an unbroken open of the same kind, decrementing on close). And only tandem-break when `depth == 0`.

a-frantz: nice catch! I missed this case. Should be simple to tack onto this PR

claymcleod (approved):
The depth tracking fix looks great — exactly what was needed. Nice clean implementation with the TandemBreak struct. All CI green, approving.

PR #648, opened 2026-02-20
refactor(wdl-lsp): switch to `tower-lsp-server`

### LINKED ISSUE #586: LSP shutdown request returns error instead of `null`
While working on the IntelliJ plugin for Sprocket, I noticed that the language server returns an error when handling the `shutdown` request. According to the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#shutdown):

> The shutdown request is sent from the client to the server. It asks the server to shut down, but to not exit... If a server receives requests after a shutdown request those requests should error with `InvalidRequest`.
>
> **Response:** result: `null`

Instead, Sprocket responds with an error: "Unexpected params: null".

This doesn't break anything during normal use—diagnostics and all other features work fine. It only shows up as an exception in the IDE logs when the server is being shut down or restarted.

Here's the relevant part of the stack trace:

```
Caused by: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: Unexpected params: null
    at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleResponse(RemoteEndpoint.java:209)
```

I'm testing with the latest Sprocket via `sprocket analyzer --stdio` and LSP4IJ as the client in IntelliJ.

Serial-ATA:
`tower-lsp-server`/`ls-types` are maintained forks of `tower-lsp`/`lsp-types`. Lot of noise in here, since the LSP types now use `fluent_uri::Uri` instead of `Url`.

closes #586

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lsp/src/server.rs
```diff
@@ -881,21 +928,39 @@ impl<S: 'static> LanguageServer for Server<S> {
...
+                        debug!(
+                            "document `{uri}` has been changed",
+                            uri = event.uri.as_str()
+                        );
+
+                        let url = match uri_to_url(&event.uri) {
+                            Ok(url) => url,
+                            Err(e) => {
+                                error!("{}", e.message);
+                                return;
```
claymcleod: This `return` aborts the entire `did_change_watched_files` handler. If a URI conversion fails for one `CHANGED` event, any subsequent `DELETED` events in the same batch are silently dropped — those documents will never be removed from the analyzer. Should be `continue`.

crates/wdl-lsp/src/proto.rs
```diff
@@ -117,15 +120,16 @@ pub fn document_diagnostic_report(
     results: Vec<AnalysisResult>,
     source: &str,
 ) -> Option<DocumentDiagnosticReportResult> {
+    let url = params.text_document.uri.try_into_url().ok()?;
```
claymcleod: This `.ok()?` silently swallows the conversion error — if it fails, the entire diagnostic report is dropped with no indication of what went wrong. Consider logging the error before returning `None`, or propagating it. This is a new failure path that didn't exist when the URI was already a `Url`.

claymcleod (changes requested):
Every handler in `server.rs` calls `uri_to_url` individually, and `normalize_uri_path` already roundtrips through `Url` internally. IMO this would be a lot cleaner with a small wrapper type (e.g., `NormalizedUri`) that stores both the `Uri` and the pre-converted `Url` together, constructed once during normalization. Something like:

```rust
struct NormalizedUri {
    uri: Uri,
    url: Url,
}

impl NormalizedUri {
    fn normalize(mut uri: Uri) -> Result<Self> {
        // normalization logic from normalize_uri_path
        let url = Url::from_str(uri.as_str())?;
        Ok(Self { uri, url })
    }
}
```

That way each handler normalizes the URI once at the top and has both types available without repeated conversions. It would also eliminate the `uri_to_url` helper and the ~15 match/`?` sites.

I checked whether `ls-types` has a feature flag to use `url::Url` directly — it doesn't (though `ls-types#27` and `tower-lsp-server#34` are open issues requesting exactly that). So the conversion layer is unavoidable for now, which makes a wrapper type even more worthwhile.

claymcleod (approved):

PR #650, opened 2026-02-22
perf(test): silence logging

a-frantz:

this silences logging during test eval and also refactors the code to avoid using a semaphore, instead only pushing onto the fututres up to the `parallelism` limit and then waiting on something else to complete if that limit is reached. This removes some overhead.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/src/commands/test.rs
+++ b/src/commands/test.rs
@@ -359,7 +359,7 @@ async fn run_tests(
     engine: wdl::engine::Config,
     handle: FilterReloadHandle,
     errors: &mut Vec<Arc<anyhow::Error>>,
-    parallelism: usize,
+    permits: usize,
     should_filter: impl Fn(&TestDefinition) -> bool,
 ) -> Result<IndexMap<String, DocumentResults>> {
     let current = handle.clone_current().expect("should have filter");
@@ -367,7 +367,7 @@ async fn run_tests(
 
     let fixtures = Arc::new(fixtures);
     let engine = Arc::new(engine);
-    let mut permits = parallelism;
+    let mut permits = permits;
     let mut futures = JoinSet::new();
     let mut all_results = IndexMap::new();
     for (analysis, tests) in documents {
@@ -515,7 +515,11 @@ async fn run_tests(
                             };
                             (doc_name, target, test_name, result)
                         });
-                    } else if let Some(result) = futures.join_next().await {
+                    } else {
+                        let result = futures
+                            .join_next()
+                            .await
+                            .expect("futures should not be exhausted");
                         let (prior_doc_name, prior_target, prior_test_name, prior_test_iteration) =
                             result.with_context(|| "joining futures")?;
                         all_results
@@ -547,8 +551,6 @@ async fn run_tests(
                             };
                             (doc_name, target, test_name, result)
                         });
-                    } else {
-                        unreachable!("futures should not be exhasuted");
                     }
                 }
                 target_results.insert(test_name.to_string(), test_iterations);
@@ -573,7 +575,7 @@ async fn run_tests(
     Ok(all_results)
 }
 
-async fn process_tests(
+async fn process_results(
     results: IndexMap<String, DocumentResults>,
     root: &Path,
     clean: bool,
@@ -581,12 +583,12 @@ async fn process_tests(
 ) -> Result<BTreeMap<String, (usize, usize, usize)>> {
     let mut all_results = BTreeMap::new();
     for (document_name, target_results) in results {
-        info!("evaluating document: `{document_name}`");
+        info!("processing document: `{document_name}`");
         for (target_name, results) in target_results {
-            info!("evaluating target: `{target_name}`");
+            info!("processing target: `{target_name}`");
             let target_dir = root.join(&target_name);
             for (test_name, test_results) in results {
-                info!("evaluating test: `{test_name}`");
+                info!("processing test: `{test_name}`");
                 let mut success_counter = 0usize;
                 let mut fail_counter = 0usize;
                 let mut err_counter = 0usize;
@@ -742,7 +744,7 @@ pub async fn test(args: Args, config: Config, handle: FilterReloadHandle) -> Com
     )
     .await?;
 
-    let results = process_tests(results, &test_dir, !args.no_clean, &mut errors).await?;
+    let results = process_results(results, &test_dir, !args.no_clean, &mut errors).await?;
     print_all_results(results);
 
     if args.clean_all {


src/commands/test.rs
```diff
@@ -347,209 +346,253 @@ enum IterationResult {
...
-                        "skipping test `{}` due to problem with input matrix",
-                        test.name
-                    );
+) -> Result<IndexMap<String, DocumentResults>> {
+    let current = handle.clone_current().expect("should have filter");
+    handle.reload(LevelFilter::OFF).expect("should reload");
+
+    let fixtures = Arc::new(fixtures);
+    let engine = Arc::new(engine);
+    let mut permits = parallelism;
```
peterhuene: Should we just rename the argument to `permits`?

src/commands/test.rs
```diff
@@ -347,209 +346,253 @@ enum IterationResult {
...
+            let target = Arc::new(target);
+            let (is_workflow, outputs) =
+                match (wdl_document.task_by_name(&target), wdl_document.workflow()) {
+                    (Some(task), _) => (false, task.outputs()),
+                    (None, Some(wf)) if wf.name() == *target => (true, wf.outputs()),
+                    (..) => {
+                        errors.push(Arc::new(anyhow!(
+                            "no target named `{}` in `{}`",
+                            target,
+                            wdl_document.path()
```
peterhuene: Nit:

```suggestion
                            "no target named `{target}` in `{path}`",
                            path = wdl_document.path()
```

src/commands/test.rs
```diff
@@ -347,209 +346,253 @@ enum IterationResult {
...
-                .assertions
-                .parse(is_workflow, outputs)
-                .with_context(|| {
+                let matrix = match test.parse_inputs().with_context(|| {
                     format!(
-                        "parsing assertions of test `{}` for WDL document `{}`",
-                        &test.name,
+                        "parsing input matrix of test `{}` for WDL document `{}`",
+                        test.name,
                         wdl_document.path()
```
peterhuene: Nit:

```suggestion
                        "parsing input matrix of test `{name}` for WDL document `{path}`",
                        name = test.name,
                        path = wdl_document.path()
```

src/commands/test.rs
```diff
@@ -347,209 +346,253 @@ enum IterationResult {
...
+                let test_name = Arc::new(test.name.clone());
+                let assertions =
+                    match test
+                        .assertions
+                        .parse(is_workflow, outputs)
+                        .with_context(|| {
+                            format!(
+                                "parsing assertions of test `{}` for WDL document `{}`",
+                                &test.name,
+                                wdl_document.path()
```
peterhuene: Nit:

```suggestion
                                "parsing assertions of test `{name}` for WDL document `{path}`",
                                name = test.name,
                                path = wdl_document.path()
```

src/commands/test.rs
```diff
@@ -347,209 +346,253 @@ enum IterationResult {
...
+                                    RunResult::Task(Box::new(
+                                        evaluator.evaluate_task(cancellation, events).await,
+                                    ))
+                                },
+                                assertions,
+                                run_dir,
+                            };
+                            (doc_name, target, test_name, result)
+                        });
+                    } else if let Some(result) = futures.join_next().await {
```
peterhuene: Rather than the `else` that panics below, I think we could probably just do an `else` here with a `let result = futures.join_next().await.expect("futures should not be exhausted");`

src/commands/test.rs
```diff
@@ -347,209 +346,253 @@ enum IterationResult {
...
+                                } else {
+                                    RunResult::Task(Box::new(
+                                        evaluator.evaluate_task(cancellation, events).await,
+                                    ))
+                                },
+                                assertions,
+                                run_dir,
+                            };
+                            (doc_name, target, test_name, result)
+                        });
```
peterhuene, line 553: Can we extract out this common code with the above `spawn` call into a "evaluate_test` method, perhaps?

src/commands/test.rs
```diff
@@ -347,209 +346,253 @@ enum IterationResult {
...
         }
-        results.insert(target.to_string(), tests);
+        all_results.insert(doc_name.to_string(), document_results);
     }
-
-    Ok(results)
+    while let Some(result) = futures.join_next().await {
+        let (doc_name, target, test_name, test_iteration) =
+            result.with_context(|| "joining futures")?;
+        all_results
```
peterhuene, line 565: One of the things I think we should be doing here (and in the above `join_next` call) in response to a test completing is to print out a one-line result immediately (e.g. a line with the test name followed by a green checkmark or red X) rather than buffering all of the results and then printing out all of the results at once in `print_test_results`.

I think some of the logic done in `process_tests` (which, in my opinion, should probably be called `process_results` in its current form) should be done in response to a test completing: update some statistics and record information from a failing test to display as part of the end-of-run summary (plus whatever is necessary to clean up a test run, perhaps).

Thus `print_test_results` would become `print_test_run_summary` or some such and only concern itself with printing out failure details and counts, not a line for each test.

src/commands/test.rs
```diff
@@ -347,209 +346,253 @@ enum IterationResult {
...
     let mut all_results = BTreeMap::new();
-    for (document_name, target_results) in tests {
+    for (document_name, target_results) in results {
         info!("evaluating document: `{document_name}`");
         for (target_name, results) in target_results {
             info!("evaluating target: `{target_name}`");
             let target_dir = root.join(&target_name);
-            for (test_name, mut test_results) in results {
+            for (test_name, test_results) in results {
                 info!("evaluating test: `{test_name}`");
```
peterhuene: This message isn't accurate as the test has already been evaluated, we're just aggregating some result information in this method.

src/commands/test.rs
peterhuene: This suggestion means that we would have one line per test and input permutation (e.g. `test foo::bar::baz (input X of Y): ✅`) instead of what we have now, which is a single output line per test that aggregates the inputs.

I personally prefer one line for each test input permutation and we can perhaps even display what inputs lead to the failure as well as part of the result summary.

peterhuene (commented):
Looking good! Some feedback comments and a few nits.

src/commands/test.rs
a-frantz: IMO "parallelism" is a lot clearer to users than "permits"

peterhuene: I don't think users are going to be interacting with this function, plus it's a semaphore they're passing, not the semaphore's count.

a-frantz: o, yes ok. I mistook your suggestion being changing the `--parallelism` flag to `--permits`. I can make that change

src/commands/test.rs
a-frantz: I've been noodling at this refactor this morning and it is a relatively large undertaking to get right. How would you feel about merging this PR with the current logging and then having a follow up PR that restructures things?

src/commands/test.rs
a-frantz: Not easily, due to how many different variables are being referenced/passed into the `spawn()`. I don't like the duplication present here, and can add this to my restructuring in the follow up PR (if you approve the current state of things)

src/commands/test.rs
peterhuene: That's fine. We can follow-up with a change to make the output of the command more immediate.

PR #654, opened 2026-02-23
feat: implement a task monitor for the Slurm backend.

peterhuene:
This PR implements a task monitor for the Slurm backend in the
spirit of the one already implemented for the LSF backend.

A new tokio task is created which periodically runs `sacct` to track
known tasks that have been queued by the backend.

Additionally, tasks are now only queued with `sbatch` rather than
waiting for the task to complete.

The Slurm backend will now properly cancel tasks when interrupted.

The following new settings have been added to the Slurm backend
configuration:

* `interval` - the interval, in seconds, that the task manager will poll
  for task statuses.
* `max_concurrency` - the maximum number of `sbatch` and and `scancel`
  tasks to run at a time.
* `job_name_prefix` - a prefix to put before any Slurm job names.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/src/commands/test.rs
+++ b/src/commands/test.rs
@@ -521,7 +521,7 @@ async fn run_tests(
                             .await
                             .expect("futures should not be exhausted");
                         let (prior_doc_name, prior_target, prior_test_name, prior_test_iteration) =
-                            result.with_context(|| "joining futures")?;
+                            result.context("joining futures")?;
                         all_results
                             .get_mut(prior_doc_name.as_str())
                             .unwrap_or(&mut document_results)
@@ -560,8 +560,7 @@ async fn run_tests(
         all_results.insert(doc_name.to_string(), document_results);
     }
     while let Some(result) = futures.join_next().await {
-        let (doc_name, target, test_name, test_iteration) =
-            result.with_context(|| "joining futures")?;
+        let (doc_name, target, test_name, test_iteration) = result.context("joining futures")?;
         all_results
             .get_mut(doc_name.as_str())
             .unwrap()
@@ -750,7 +749,7 @@ pub async fn test(args: Args, config: Config, handle: FilterReloadHandle) -> Com
     if args.clean_all {
         remove_dir_all(test_dir.join(RUNS_DIR))
             .await
-            .with_context(|| "cleaning the file system of all test exections")?;
+            .context("cleaning the file system of all test executions")?;
     }
 
     if let Some(errors) = NonEmpty::from_vec(errors) {
--- a/src/inputs/file.rs
+++ b/src/inputs/file.rs
@@ -63,7 +63,7 @@ pub async fn read_input_file(path: &EvaluationPath) -> Result<JsonInputMap> {
     ) -> Result<JsonInputMap> {
         match format {
             Format::Json => match serde_json::from_str::<JsonValue>(contents)
-                .with_context(|| "failed to deserialize JSON inputs file `{path}`")?
+                .with_context(|| format!("failed to deserialize JSON inputs file `{path}`"))?
             {
                 JsonValue::Object(object) => Ok(json_map_to_inputs(object, &origin)),
                 _ => bail!(
@@ -72,10 +72,10 @@ pub async fn read_input_file(path: &EvaluationPath) -> Result<JsonInputMap> {
             },
             Format::Yaml => {
                 if let YamlValue::Mapping(mapping) = serde_yaml_ng::from_str::<YamlValue>(contents)
-                    .with_context(|| "failed to deserialize YAML inputs file `{path}`")?
+                    .with_context(|| format!("failed to deserialize YAML inputs file `{path}`"))?
                 {
                     let value = serde_json::to_value(mapping)
-                        .with_context(|| "invalid YAML input file `{path}`")?;
+                        .with_context(|| format!("invalid YAML input file `{path}`"))?;
 
                     if let JsonValue::Object(map) = value {
                         return Ok(json_map_to_inputs(map, &origin));
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,7 +108,7 @@ async fn real_main() -> CommandResult<()> {
             )?;
             config
                 .validate()
-                .with_context(|| "validating provided configuration")?;
+                .context("validating provided configuration")?;
             config
         }
     };


crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let (exit_code, signal) = s
+            .split_once(':')
+            .with_context(|| format!("invalid Slurm exit code `{s}`"))?;
+        Ok(Self {
+            exit_code: exit_code
+                .parse()
+                .with_context(|| "invalid exit code `{exit_code}`")?,
```
claymcleod: I think these `with_context` closures need `format!`—right now `{exit_code}` and `{signal}` print as literal text.

```suggestion
                .with_context(|| format!("invalid exit code `{exit_code}`"))?,
```

Same issue on line 226 with `{signal}`.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+            "COMPLETED" => Ok(Self::Completed),
+            "DEADLINE" => Ok(Self::Deadline),
+            "FAILED" => Ok(Self::Failed),
+            "NODE_FAIL" => Ok(Self::NodeFail),
+            "OUT_OF_MEMORY" => Ok(Self::OutOfMemory),
+            "PENDING" => Ok(Self::Pending),
+            "PREEMPTED" => Ok(Self::Preempted),
+            "RUNNING" => Ok(Self::Running),
+            "SUSPENDED" => Ok(Self::Suspended),
+            "TIMEOUT" => Ok(Self::Timeout),
```
claymcleod: I think this is missing several valid `sacct` states—at minimum `COMPLETING` (very common), `CONFIGURING`, `REQUEUED`, `RESIZING`, `REVOKED`, `SPECIAL_EXIT`, `SIGNALING`, `STAGE_OUT`, `STOPPED`, `RESV_DEL_HOLD`, `REQUEUE_FED`, `REQUEUE_HOLD`. An unrecognized state currently causes `update_jobs` to return `Err`, which triggers `fail_all_jobs`—so one job in `COMPLETING` state would kill every monitored job.

See: https://slurm.schedmd.com/job_state_codes.html

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+}
+
+impl FromStr for JobState {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        // See: https://slurm.schedmd.com/job_state_codes.html
+        match s {
+            "BOOT_FAIL" => Ok(Self::BootFail),
+            "CANCELLED" => Ok(Self::Canceled),
```
claymcleod: Slurm can append suffixes to certain states (e.g., `CANCELLED by 1000`). Consider using `starts_with` matching here instead of exact equality so these don't fall through to the unknown state error.

For example, [LUMC/slurm-cluster-status](https://github.com/LUMC/slurm-cluster-status/blob/master/slurm-cluster-status.py) handles this with `job_status.startswith("CANCELLED by")`.

```suggestion
            s if s.starts_with("CANCELLED") => Ok(Self::Canceled),
```

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+            "DEADLINE" => Ok(Self::Deadline),
+            "FAILED" => Ok(Self::Failed),
+            "NODE_FAIL" => Ok(Self::NodeFail),
+            "OUT_OF_MEMORY" => Ok(Self::OutOfMemory),
+            "PENDING" => Ok(Self::Pending),
+            "PREEMPTED" => Ok(Self::Preempted),
+            "RUNNING" => Ok(Self::Running),
+            "SUSPENDED" => Ok(Self::Suspended),
+            "TIMEOUT" => Ok(Self::Timeout),
+            _ => bail!("unknown Slurm job state `{s}"),
```
claymcleod:
```suggestion
            _ => bail!("unknown Slurm job state `{s}`"),
```

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+                .parse()
+                .with_context(|| "invalid signal number `{signal}`")?,
+        })
+    }
+}
+
+impl fmt::Display for JobExitCode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.signal > 0 {
+            write!(f, "signal number `{signal}`", signal = self.signal)
```
claymcleod: Nit: `code()` and `into_exit_status()` both mask the signal with `& 0x7F`, but `Display` uses `self.signal` unmasked.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+                        }
+
+                        state.jobs.keys().join(",")
+                    };
+
+                    match Self::read_jobs(&jobs).await.context("failed to query jobs with `sacct`") {
+                        Ok(output) => {
+                            let mut state = state.lock().expect("failed to lock state");
+                            if let Err(e) = state.update_jobs(&output, &events) {
+                                state.fail_all_jobs(&e);
```
claymcleod: I think the blast radius here is disproportionate—a single malformed job record (or an unrecognized state like `COMPLETING`) causes `fail_all_jobs` to fail *every* monitored job. Consider isolating per-job parse errors so that one bad record doesn't take down unrelated jobs.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -298,176 +1012,88 @@ impl TaskExecutionBackend for SlurmApptainerBackend {
...
+            let exit_code = tokio::select! {
                 _ = task_token.cancelled() => {
-                    crankshaft::events::send_event!(
-                        self.events.crankshaft(),
-                        crankshaft::events::Event::TaskCanceled {
-                            id: crankshaft_task_id
-                        },
-                    );
+                    if let Err(e) = cancelled.await {
+                        error!("failed to cancel task `{name}` (LSF job `{job_id}`): {e:#}");
```
claymcleod:
```suggestion
                        error!("failed to cancel task `{name}` (Slurm job `{job_id}`): {e:#}");
```

Same on line 1062.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+
+    /// Reads the current jobs using `sacct`.
+    ///
+    /// Returns the stdout of `sacct`.
+    async fn read_jobs(jobs: &str) -> Result<Vec<u8>> {
+        let mut command = Command::new("sacct");
+        let command = command
+            .arg("-P") // parseable
+            .arg("-n") // no header
+            .arg("--format") // JSON output
```
claymcleod:
```suggestion
            .arg("--format") // column format (pipe-delimited with -P)
```

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+        let mut timer = tokio::time::interval(interval);
+        timer.set_missed_tick_behavior(MissedTickBehavior::Delay);
+
+        loop {
+            select! {
+                _ = &mut drop => break,
+                _ = timer.tick() => {
+                    let jobs = {
+                        // If there are no jobs to monitor, do nothing
+                        let state = state.lock().expect("failed to lock state");
```
claymcleod, line 699: I think `std::sync::Mutex` with `.expect()` here means a panic while holding the lock poisons the mutex, which cascades to crash the entire monitor loop. Consider either using `tokio::sync::Mutex` or handling the poison case gracefully (e.g., `.lock().unwrap_or_else(|e| e.into_inner())`).

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+}
+
+/// The monitor is responsible for periodically querying Slurm for job state and
+/// sending task events.
+#[derive(Debug, Clone)]
+struct Monitor {
+    /// The state of the monitor.
+    state: Arc<Mutex<MonitorState>>,
+    /// A sender for notifying that the monitor has been dropped.
+    _drop: Arc<oneshot::Sender<()>>,
```
claymcleod, line 515: Consider noting that the `Arc` here means the drop signal is only sent when the last `Monitor` clone is dropped—important for understanding the shutdown semantics when `Monitor` is `Clone`.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,676 @@ use crate::v1::requirements::ContainerSource;
...
+    id: u64,
+    /// The task name for Crankshaft events.
+    ///
+    /// Note: this name differs from the job name used in `sbatch`.
+    task_name: String,
+    /// The receiver for when the job completes.
+    completed: oneshot::Receiver<Result<JobExitCode>>,
+}
+
+/// The monitor is responsible for periodically querying Slurm for job state and
```
claymcleod, line 507: The monitor/job/state lifecycle here closely mirrors the LSF backend. A shared trait or abstraction could reduce the ~400 lines of structural duplication—though I understand this may be intentional for now while both backends stabilize.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -484,7 +1110,8 @@ impl TaskExecutionBackend for SlurmApptainerBackend {
                 )
                 .into(),
             }))
-        }
+
+                }
```
claymcleod:
```suggestion
        }
```

claymcleod (changes requested):

crates/wdl-engine/src/backend/slurm_apptainer.rs
peterhuene: Man I'd love a lint for that somehow (I know it's `anyhow` specific, but that's an easy one to accidentally get wrong.

crates/wdl-engine/src/backend/slurm_apptainer.rs
peterhuene: Good point, I missed the part where if `sacct` recognizes one of the state "flags" it'll replace the base state with it in its output, so we'll have to recognize the flags too. I missed that they don't report this with a separate dedicated field like I expected it would.

crates/wdl-engine/src/backend/slurm_apptainer.rs
peterhuene: This makes me want to use the REST API rather than go through `sacct` if it's not being faithful to the underlying state values. Is this the only state that they do this to or are there other undocumented cases?

crates/wdl-engine/src/backend/slurm_apptainer.rs
peterhuene: If the mutex is poisoned there's no graceful recovery as we don't know a good known value to restore the state to (other than, perhaps, _an empty state_ meaning we would potentially lose track of what was being monitored previously with no way to abort those tasks).

Using `tokio::sync::Mutex` doesn't really help with that, it just automatically unlocks the mutex on panic so that the next time you lock it, it can't error; it doesn't guarantee that the previous owner of the mutex that panicked didn't leave the protected data in an inconsistent state.

Even so, we're not currently catching any panics here, so a panic while the mutex is held will be unwinding past this method to effectively error the monitor task completely (the effect of which may mean tasks never get any new Crankshaft events, thus stalling them).

We will need to harden all of the backend implementations were we to eventually share them between evaluators in a server context such that these sorts of monitoring tasks keep running despite panics. To be honest, it's really hard to retrofit server-orientated resiliency into what was previously a standalone client-only tool. It'll require a lot of careful inspection of the backend implementations in `wdl-engine` and also `crankshaft`.

crates/wdl-engine/src/backend/slurm_apptainer.rs
peterhuene: That could be a worthwhile refactoring. I think we can perhaps leave that to the future.

crates/wdl-engine/src/backend/slurm_apptainer.rs
claymcleod: Agreed, we can come back to this in another round of making the execution more panic-resistant as discussed.

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,694 @@ use crate::v1::requirements::ContainerSource;
...
+        let (exit_code, signal) = s
+            .split_once(':')
+            .with_context(|| format!("invalid Slurm exit code `{s}`"))?;
+        Ok(Self {
+            exit_code: exit_code
+                .parse()
+                .with_context(|| format!("invalid exit code `{exit_code}`"))?,
+            signal: signal
+                .parse()
+                .with_context(|| "invalid signal number `{signal}`")?,
```
claymcleod: This one still needs `format!` — the exit_code one above was fixed but this one was missed.

```suggestion
            signal: signal
                .parse()
                .with_context(|| format!("invalid signal number `{signal}`"))?,
```

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -54,12 +69,694 @@ use crate::v1::requirements::ContainerSource;
...
+        })
+    }
+}
+
+impl fmt::Display for JobExitCode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.signal > 0 {
+            // Mask the lower 7 bits of the signal for display
+            write!(f, "signal number `{signal}`", signal = self.signal & 0xF7)
```
claymcleod: `0xF7` is `11110111` — this clears bit 3 instead of bit 7. Should be `0x7F` to match `code()` and `into_exit_status()`.

```suggestion
            write!(f, "signal number `{signal}`", signal = self.signal & 0x7F)
```

PR #656, opened 2026-02-23
feat(wdl-doc): extended doc comment support

Serial-ATA:
Adds support for doc comments on tasks and workflows as well as their inputs/outputs

closes #577

Before submitting this PR, please make sure:

For external contributors:

- [x] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [x] You have not used AI on any parts of this pull request.

For all contributors:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added tests (when appropriate).
- [x] You have added an entry in the CHANGELOG (when appropriate).
- [x] You have updated the README or other documentation to account for these changes (when appropriate).
- [x] You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).

For PRs containing lint rule changes:

- [x] You have updated any and all effected entries within `RULES.md`.
- [x] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

PR #659, opened 2026-02-24
refactor(wdl-doc): move doc comment extraction to `wdl-ast`

Serial-ATA:
Simplifies working with doc comments outside of `wdl-doc`

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-ast/src/v1/decls.rs
+++ b/crates/wdl-ast/src/v1/decls.rs
@@ -826,7 +826,10 @@ impl<N: TreeNode> AstNode<N> for BoundDecl<N> {
 impl Documented<SyntaxNode> for BoundDecl<SyntaxNode> {
     fn doc_comments(&self) -> Option<Vec<Comment<<SyntaxNode as TreeNode>::Token>>> {
         let parent = self.inner().parent()?;
-        if parent.kind() != SyntaxKind::OutputSectionNode {
+        if !matches!(
+            parent.kind(),
+            SyntaxKind::InputSectionNode | SyntaxKind::OutputSectionNode
+        ) {
             return None;
         }
 


peterhuene (approved):
Looks great, thanks for refactoring this!

crates/wdl-ast/src/v1/decls.rs
```diff
@@ -756,6 +760,20 @@ impl<N: TreeNode> AstNode<N> for UnboundDecl<N> {
     }
 }
 
+impl Documented<SyntaxNode> for UnboundDecl<SyntaxNode> {
+    fn doc_comments(&self) -> Option<Vec<Comment<<SyntaxNode as TreeNode>::Token>>> {
+        let parent = self.inner().parent()?;
+        if parent.kind() != SyntaxKind::StructDefinitionNode {
```
a-frantz: `UnboundDecl` is also valid with an `InputSection` parent

crates/wdl-ast/src/lib.rs
```diff
@@ -61,6 +61,37 @@ mod element;
...
+/// Shared doc comment extraction logic.
+pub fn doc_comments<N: TreeNode>(
+    preceding_trivia: impl IntoIterator<Item = N::Token>,
+) -> impl Iterator<Item = Comment<N::Token>> {
+    preceding_trivia
+        .into_iter()
+        .take_while(|token| {
+            token.kind() == SyntaxKind::Whitespace
+                || (token.kind() == SyntaxKind::Comment
+                    && token.text().starts_with(DOC_COMMENT_PREFIX))
```
a-frantz: this will stop collecting at the first comment not starting with `##` which may be premature. This should collect _all_ comments, then filter to just those starting with `##`

Serial-ATA: That makes sense for directives, but are you saying we should also support this?

```wdl
## This is a foo struct
# random comment
## Docs continue...
struct Foo {}
```

crates/wdl-ast/src/v1/decls.rs
Serial-ATA: Was saving that extended support for #656, but I can add it here

crates/wdl-ast/src/lib.rs
a-frantz: yes, because we should definitely support:
```wdl
## This is a valid doc comment
# even if this is in the middle
struct Foo {}
```
and I think it's fine to support your case as well. I don't think it's likely to actually occur, whereas the case I point out is likely to occur

crates/wdl-ast/src/v1/decls.rs
a-frantz: if we are moving the doc comment support to `wdl-ast` now, I think it makes the most sense to just add all the supported cases in this PR instead of doing it piecemeal

peterhuene: ~~Would that also be true of output sections?~~

Erm, never mind, those are always bound decls.

peterhuene: That said, we probably do need a corresponding impl for `BoundDecl` then.

crates/wdl-ast/src/v1/decls.rs
```diff
@@ -802,6 +823,20 @@ impl<N: TreeNode> AstNode<N> for BoundDecl<N> {
     }
 }
 
+impl Documented<SyntaxNode> for BoundDecl<SyntaxNode> {
+    fn doc_comments(&self) -> Option<Vec<Comment<<SyntaxNode as TreeNode>::Token>>> {
+        let parent = self.inner().parent()?;
+        if parent.kind() != SyntaxKind::OutputSectionNode {
```
a-frantz: This can also be in an input section

a-frantz (approved):

PR #658, opened 2026-02-24
feat: add cache_excluded_keys option to skip inputs from cache check (#657)

akashchakrabortymsc-cmd:
Fixes #657 

Added a new `cache_excluded_keys` option to `TaskConfig` that lets users 
specify input keys that should be ignored when checking the cache.

Changes 

1. cache_excluded_keys added to TaskConfig in config.rs
2.  excluded_keys added to KeyRequest in cache.rs
3.  Filtering logic added in cache.rs
4.  Test added
5. CHANGELOG updated

- [x]  You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [x] You have not used AI on any parts of this pull request.

For all contributors:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [ ] Your code builds clean without any errors or warnings.
- [x] You have added tests (when appropriate).
- [x] You have added an entry in the CHANGELOG (when appropriate).
- [ ] You have updated the README or other documentation to account for these changes (when appropriate).
- [ ] You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate)

For PRs containing lint rule changes:

- [ ] You have updated any and all effected entries within `RULES.md`.
- [ ] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

Issue #657, opened 2026-02-23
Add configurable options for cacheable keys


adthrasher:
As a user, I'd like to mark certain keys as excepted from the cache equality check. For example, `requirements.memory` and `requirements.disks`, I'd like to optionally be able to exclude changes to those fields from the cache check. This may apply to inputs as well; for example, we often use an input key `modify_disk_size_gb` to add additional disk space at runtime. I wouldn't want changes to that field to cause a rerun of a previously successful task.

PR #662, opened 2026-02-24
feat: add exceptable hints/inputs/requirements to ignore when checking cache

adthrasher:
This adds new configuration options that allow a user to specify hints, inputs, and/or requirements that should be ignored when checking that a cache entry matches. Addresses #657, conflicts with #658.

For all contributors:

### DIFF
--- a/crates/wdl-engine/src/cache.rs
+++ b/crates/wdl-engine/src/cache.rs
@@ -2,6 +2,7 @@
 
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::fmt;
 use std::io::BufReader;
 use std::io::BufWriter;
@@ -16,6 +17,7 @@ use arrayvec::ArrayString;
 use serde::Deserialize;
 use serde::Serialize;
 use tokio::fs;
+use tracing::debug;
 use tracing::info;
 use url::Url;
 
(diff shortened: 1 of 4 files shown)


crates/wdl-engine/src/cache.rs
```diff
@@ -403,7 +446,13 @@ impl CallCache {
     ///
     /// Returns an error if the entry could not be read or if the entry is no
     /// longer valid.
-    pub async fn get(&self, key: &Key) -> Result<Option<TaskExecutionResult>> {
+    pub async fn get(
+        &self,
+        key: &Key,
+        excluded_hints: Vec<String>,
+        excluded_inputs: Vec<String>,
+        excluded_requirements: Vec<String>,
```
claymcleod: Consider storing the exclusion lists on the `CallCache` itself at construction time rather than threading them through every `key()` and `get()` call. The exclusions come from `TaskConfig`, which is set once for the evaluator — they don't vary per call. This would simplify both signatures:

```rust
// Before:
cache.key(request, excluded_inputs.clone())
cache.get(key, excluded_hints.clone(), excluded_inputs.clone(), excluded_requirements.clone())

// After:
cache.key(request)
cache.get(key)
```

If you do keep them as parameters (e.g., for future per-task config), bundling them into a struct would prevent transposition errors — all three are `Vec<String>`, so swapping them compiles silently.

crates/wdl-engine/src/cache.rs
```diff
@@ -195,13 +195,31 @@ impl Key {
...
+            kind: &str,
+            excluded: &[String],
+        ) -> Result<()>
         where
             K: std::hash::Hash + fmt::Display + Eq,
             V: Eq,
         {
             for (k, v) in a {
+                // Skip excluded keys
+                let key_str = k.to_string();
```
claymcleod, line 236: nit: excluded.contains() is an O(n) linear scan per map entry. Consider storing the exclusion lists as `HashSet<String>` to get O(1) lookups — this would be straightforward if exclusions are moved to `CallCache` at construction time.

crates/wdl-engine/src/cache.rs
```diff
@@ -195,13 +195,31 @@ impl Key {
...
+                if excluded.contains(&key_str) {
+                    info!("{} `{}` is excluded from cache checking, skipping", kind, k);
```
claymcleod: nit: Consider `debug!` instead of `info!` here — with multiple excluded keys across many cache checks, this could get noisy in production logs.

crates/wdl-engine/src/cache.rs
adthrasher: Initially, I was thinking of this as not a feature of the cache itself. For example, `hints` and `requirements` are not components of the cache key, so a user can dynamically change which ones are considered for matching without changing the cache itself. With the addition of `inputs`, this does become a feature of the cache since it alters the cache key. I'll take another look at it.

PR #668, opened 2026-02-26
feat: add configurable newline style to wdl-format

### LINKED ISSUE #667: Configurable newline style in `wdl-format`
`wdl-format` currently uses the native newline style of the platform it's built for:

https://github.com/stjude-rust-labs/sprocket/blob/36c67422bc719185d67fc7d85c09e8503cd58267/crates/wdl-format/src/lib.rs#L17-L22

Which would cause unnecessary changes in the case of reformatting on different platforms.

The current behavior can be the default, but it may be helpful to let users force a specific newline style. The options would be `auto`, `unix`, and `windows`.

akashchakrabortymsc-cmd:
Fixes #667 

**### Added a `NewlineStyle` enum with three options:**

1. `Auto` — uses the platform native newline (default, preserves existing behavior)
2.  `Unix` — always uses `\n`
3.  `Windows` — always uses `\r\n`

### **Changes**

1. Added `crates/wdl-format/src/config/newline.rs` with `NewlineStyle` enum
2. Added `newline_style` field to `Config`
3. Updated `token/post.rs` to use `config.newline_style.as_str()` instead of hardcoded `NEWLINE`
4. Added unit tests for all three newline styles

_All_ test passed 

For all contributors:

      is implemented.


crates/wdl-format/src/config/newline.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// Use Unix-style newlines (`\n`).
+    Unix,
+    /// Use Windows-style newlines (`\r\n`).
+    Windows,
+}
+impl Default for NewlineStyle {
+    fn default() -> Self {
+        DEFAULT_NEWLINE_STYLE
+    }
+}
```
Serial-ATA, line 22:
```suggestion

/// The default newline style.
pub const DEFAULT_NEWLINE_STYLE: NewlineStyle = NewlineStyle::Auto;

/// The newline style to use when formatting.
#[derive(Clone, Copy, Debug)]
pub enum NewlineStyle {
    /// Use the native newline style of the platform.
    Auto,
    /// Use Unix-style newlines (`\n`).
    Unix,
    /// Use Windows-style newlines (`\r\n`).
    Windows,
}

impl Default for NewlineStyle {
    fn default() -> Self {
        DEFAULT_NEWLINE_STYLE
    }
}

```

Spacing the items out

crates/wdl-format/src/config.rs
```diff
@@ -18,14 +20,16 @@ pub struct Config {
     pub max_line_length: MaxLineLength,
     /// Whether to sort input sections.
     pub sort_inputs: bool,
+    /// The newline style.
+    pub newline_style: NewlineStyle,
 }
```
Serial-ATA, line 36:
```suggestion
}

```

Serial-ATA (changes requested):
Looking good so far! Can you add a test for the config option? See https://github.com/stjude-rust-labs/sprocket/blob/main/crates/wdl-format/TESTING.md and https://github.com/stjude-rust-labs/sprocket/tree/main/crates/wdl-format/tests/format/trailing-commas for an example

crates/wdl-format/src/config/newline.rs
```diff
@@ -0,0 +1,60 @@
+//! Newline style within formatting configuration.
```
claymcleod, line 1: The migration from `NEWLINE` to `NewlineStyle` is partially complete — `pre.rs:197` still uses the old `NEWLINE` constant, and the constant in `lib.rs` isn't removed. This isn't a correctness bug (`.lines()` normalizes both styles), but would you mind completing the migration so the old constant can be removed?

claymcleod (changes requested):
Nice work on the `NewlineStyle` enum — it's looking solid. A couple of things to address before this is ready:

1. The migration from `NEWLINE` to `NewlineStyle` is incomplete — `pre.rs` still uses the old constant, and the `NEWLINE` constant in `lib.rs` isn't removed.
2. The newline style needs to be wired into sprocket's `FormatConfig` (`src/config.rs`) and the format command's CLI args (`src/commands/format.rs`) so users can actually configure it.

claymcleod (changes requested):
Hey — a few things still need to be addressed here:

1. **The code doesn't compile.** The `Config` struct in `wdl-format/src/config.rs` is missing a `newline_style: NewlineStyle` field — the builder method references `self.newline_style`, but the field doesn't exist. You also need to add `pub use newline::NewlineStyle;` to that file so the type is accessible from outside the module.

2. **The test config file is misnamed.** The test harness looks for `config.toml` (see `TESTING.md` and the `trailing-commas` example Serial-ATA linked), but your file is named `cargo.toml`. You'll also need a `source.default.formatted.wdl` for the default-config run, as described in the testing docs.

3. **Please don't reformat the CHANGELOG bullets.** The existing convention uses `*`, and this diff changes them to `-` throughout the file. Please revert those formatting changes and only add your new entry.

Please make sure all of these are addressed before requesting re-review.

src/commands/format.rs
```diff
@@ -52,6 +53,10 @@ pub struct Args {
     #[arg(long, value_name = "LENGTH", global = true)]
     pub max_line_length: Option<usize>,
 
+    /// The newline style to use (auto, unix, or windows).
```
Serial-ATA:
```suggestion
    /// The newline style to use.
```

No need to mention the variants here

crates/wdl-format/src/token/pre.rs
```diff
@@ -194,7 +193,7 @@ impl TokenStream<PreToken> {
                         // do not `trim()` the token as the whitespace may
                         // have syntactical meaning in markdown
                         documentation.push_str(t);
-                        documentation.push_str(NEWLINE);
+                        documentation.push('\n');
```
Serial-ATA, line 196: This isn't using the right newline style

Serial-ATA (changes requested):

crates/wdl-format/src/config/newline.rs
akashchakrabortymsc-cmd: Migration is complete — the NEWLINE constant has been removed from lib.rs and replaced with '\n' directly in pre.rs.

crates/wdl-format/src/token/pre.rs
akashchakrabortymsc-cmd: Should I thread the config through `push_preceding_trivia` to use `newline_style.as_str()` here, or is `'\n'` acceptable since `.lines()` normalizes both styles?

crates/wdl-format/src/config/newline.rs
```diff
@@ -0,0 +1,73 @@
...
+
+
+/// The default newline style.
+pub const DEFAULT_NEWLINE_STYLE: NewlineStyle = NewlineStyle::Auto;
+
+
+/// The default newline style.
+pub const DEFAULT_NEWLINE_STYLE: NewlineStyle = NewlineStyle::Auto;
```
claymcleod, line 4: The constant `DEFAULT_NEWLINE_STYLE` is defined twice.

src/commands/format.rs
```diff
@@ -131,11 +134,18 @@ pub async fn format(args: Args, config: Config, colorize: bool) -> CommandResult
             .unwrap_or(config.format.max_line_length),
     )
     .context("failed to create max line length configuration")?;
+    let newline_style = match args.newline_style.as_deref() {
+    Some("unix") => wdl::format::NewlineStyle::Unix,
+    Some("windows") => wdl::format::NewlineStyle::Windows,
+    _ => config.format.newline_style,
+};
```
claymcleod: The match arms here need to be indented to match the surrounding code.

src/commands/format.rs
```diff
@@ -52,6 +52,9 @@ pub struct Args {
     #[arg(long, value_name = "LENGTH", global = true)]
     pub max_line_length: Option<usize>,
 
+    /// The newline style to use 
+    #[arg(long, value_name = "STYLE", global = true, value_parser = ["auto", "unix", "windows"])]
+    pub newline_style: Option<String>,
```
claymcleod: Consider deriving `clap::ValueEnum` on `NewlineStyle` so the arg can be typed as `Option<NewlineStyle>` directly, eliminating the string match.

claymcleod (changes requested):
A few more things to address:

- The test fixture for Windows newlines (`source.formatted.wdl`) needs `\r\n` line endings to actually validate the feature.
- Several extra blank lines in `newline.rs` that `cargo fmt` should catch.

Re your serde question: keeping `serde` on `NewlineStyle` is the right call since it needs to deserialize from `sprocket.toml`.

crates/wdl-format/src/token/pre.rs
claymcleod: You're right that `.lines()` normalizes here, so this works today. That said, consider using the configured newline style even in intermediate representations as a defensive measure — if the output path changes later, this could silently produce wrong newlines. It would require threading the config through `push_preceding_trivia`, which is a bit more work but would complete the migration.

akashchakrabortymsc-cmd (commented):
Hey @claymcleod  here I addressed the latest feedback!

- Added `clap::ValueEnum` to `NewlineStyle` so `--newline-style` is properly typed as `Option<NewlineStyle>` 
- Fixed blank lines in `newline.rs`
- Updated `tests/format.rs` with a `preserve_line_endings` flag 
- Added `.gitattributes` exception 

Kindly check and let me know if further development is required ..

crates/wdl-format/tests/format.rs
```diff
@@ -157,15 +169,28 @@ fn run_test(test: &Path) -> Result<(), anyhow::Error> {
             config_path.display()
         ))?;
 
-        run_test_inner(config, &source, &path, &formatted_path)?;
+        run_test_inner(
+            FormatConfig::default(),
```
Serial-ATA:
```suggestion
            config,
```

crates/wdl-format/tests/format.rs
```diff
@@ -129,13 +141,13 @@ fn run_test_inner(
     source: &str,
     original_doc: &Path,
     formatted_doc: &Path,
+    preserve_line_endings: bool,
 ) -> anyhow::Result<()> {
     let formatted = format(config, source, original_doc)?;
-    compare_result(formatted_doc, &formatted)?;
```
Serial-ATA: This needs to be added back. Documents are formatted twice intentionally.

crates/wdl-format/tests/format.rs
```diff
@@ -157,15 +169,28 @@ fn run_test(test: &Path) -> Result<(), anyhow::Error> {
...
+            &path,
+            &formatted_path,
+            false,
+        )?;
         run_test_inner(
             FormatConfig::default(),
             &source,
             &path,
             &path.with_extension("default.formatted.wdl"),
+            true,
```
Serial-ATA: I think this is backwards. You're preserving the line endings of the default config, which is `auto`. So this will fail on Windows. Instead, you should preserve it in `source.formatted.wdl` which is forced to be `windows`.

Serial-ATA (changes requested):

PR #675, opened 2026-02-27
refactor: sprocket.toml parsing and merging

### LINKED ISSUE #457: TOML errors have poor error reporting
**Describe the bug**

had a wonky `sprocket.toml` and it took me a bit to figure out what the issue was just because the reported error was unhelpful.

**Expected behavior**

better reporting

**Terminal Output**
Please copy and paste the command used, logs generated, and the error produced in the markdown block below. This can also be attached to to the issue as a file.

relevant bit of `sprocket.toml`:
```toml
[run.backends.hpc]
type = "lsf_apptainer"
default_lsf_queue.name = "compbio"
max_scatter_concurrency = 200
apptainer_images_dir = "/scratch_space/$USER/sprocket-apptainer-images"
```
leads to:
```console
$ sprocket config resolve
error: failed to merge configuration

Caused by:
    invalid type: found map, expected a string for key "default.run" in sprocket.toml TOML file
```
the actual problem is an extra `.name` after `default_lsf_queue`. Removing that fixes it

### LINKED ISSUE #618: feat: add the ability to hide warnings
I would like the ability to run `sprocket check` and hide everything but errors. Currently you can hide notes but warnings are always printed.

`sprocket check --hide-warnings` checks a wdl document and only produces errors. Similar to and would imply  `--hide-notes` but also excludes warnings.

Currently we can do this by filtering sprocket output (`sprocket check | grep error`) but it would be nice if we could run this without extra boilerplate and processing of sprocket output.

**Additional context**
I have many wdl workflows using `version development` for running via miniwdl. I would like to check these using sprocket, which produces very nice diagnostic messages. When relying on a fallback version (fallback version branch [here](https://github.com/stjude-rust-labs/sprocket/tree/fallback-version)) we get many warnings in that case for each file:
```
warning: unsupported WDL version `development`; interpreting document as version `1.2`                                          
    ┌─ path/to/workflow.wdl:1:9                                                                        
    │                                                                                                                             
  1 │ version development                                                                                                         
    │         ^^^^^^^^^^^ this version of WDL is not supported
```

~Unlike other warnings, these cannot currently be silenced.~

`--hide-warnings` seems consistent with the existing flags for modifying the strictness of document checking: `--deny-notes` and `--deny-warnings` for more stringent checks, `--hide-notes` for less stringent check.

### LINKED ISSUE #673: Allow sprocket format config option to disable adding commas after last item in list
`sprocket format` inserts a comma after the last element in a list, e.g.

```wdl
    meta {
        description: "A simple task that prints a greeting message."
        outputs: {
            out: "The output file containing the greeting message.",
        }
    }
```

This fails in `miniwdl check`:

```log
WDL.Error.SyntaxError: Unexpected token Token('STRING1_FRAGMENT', '}\n    }\n\n    call t\n\n    output {\n        File out = t.out\n    }\n}\n\ntask t {\n    meta {\n        description: "A simple task that prints a greeting message."\n        outputs: {\n            out: "The output file containing the greeting message.",\n        }\n    }\n\n    parameter_meta {\n        name: "The name to greet."\n        punctuation: "The punctuation to use at the end of the greeting."\n    }\n\n    input {\n        String name = "world"\n        String punctuation = "!"\n    }\n\n    command <<<\n        echo "Hello, ') at line 8, column 9.
Expected one of: 
        * CNAME
Previous tokens: [Token('COMMA', ',')]
```

I'd like the ability to disable this format behavior in `sprocket.toml`.

No other option except writing a script to undo this.

**Additional context**

a-frantz:

Lots of changes in this PR, although the only change that would break existing user configs is removing the `format.with_tabs` field. All other user configs should continue to work.

The main goal of this PR is to de-obfuscate Sprocket configuration. Currently, there are many config fields which are effectively hidden from users due to "skip serializing" logic. I would argue that `sprocket config init` should expose every config field and its default value with very few exceptions. The only exceptions in this PR are the `container` and `shell` fields of `TaskConfig`. Every other field is serialized and emitted for fine tuning.

To enable this, many fields had to replace their `Option<_>` values with something that could be nulled in TOML. That is done via sentinel values specific to each field.

closes #457

For external contributors:

For all contributors:

      is implemented.


src/config.rs
```diff
@@ -129,24 +130,34 @@ pub struct FormatConfig {
     /// Use tabs for indentation (default is spaces).
     pub with_tabs: bool,
     /// The number of spaces to use for indentation levels (default is 4).
+    ///
+    /// This is ignored if `with_tabs = true`.
     pub indentation_size: usize,
-    /// The maximum line length (default is 90).
+    /// The maximum line length (default is 90). 0 means do not use a maximum
+    /// line length.
     pub max_line_length: usize,
```
claymcleod: In my opinion, this should be an `Option<usize>` rather than using `0` as a sentinel for "no limit". `MaxLineLength` already models this as `Option<usize>` internally, and the Sprocket config should mirror that. With `Option<usize>`, users can simply omit the field in their `sprocket.toml` rather than setting it to `0`, and the `Default` impl can use `max_line_length.get()` directly without the `unwrap_or(0)`.

crates/wdl-format/src/config/indent.rs
```diff
@@ -44,8 +41,7 @@ impl Indent {
     /// Attempts to create a new indentation level configuration.
     pub fn try_new(tab: bool, num_spaces: Option<usize>) -> Result<Self, IndentError> {
         match (tab, num_spaces) {
-            (true, None) => Ok(Indent::Tabs),
-            (true, Some(_)) => Err(IndentError::InvalidConfiguration),
+            (true, _) => Ok(Indent::Tabs),
```
claymcleod: Silently ignoring `num_spaces` when `tab=true` could be surprising — previously this was an explicit error. Consider logging a warning if `num_spaces` is `Some` when tabs are selected, so users know their `indentation_size` setting is being ignored.

claymcleod (changes requested):
A couple of thoughts on this one — mostly around the `max_line_length` representation.

src/config.rs
```diff
@@ -129,24 +130,34 @@ pub struct FormatConfig {
...
-            indentation_size: config.indent.num(),
-            max_line_length: config
-                .max_line_length
-                .get()
-                .expect("should have a max line length"),
-            sort_inputs: config.sort_inputs,
+            with_tabs: matches!(indent, Indent::Tabs),
+            indentation_size: indent.num(),
+            // MaxLineLength::get() returns `None` if no maximum is configured.
+            // setting max_line_length to `0` disables maximum line lengths.
```
claymcleod: This needs to be capitalized.

src/config.rs
claymcleod: Alternatively, you could set this field to be `MaxLineLength` directly, but then you'd need to consider adding validation to the `Deserialize` impl for `MaxLineLength` — right now, it seems that the deserialization logic is incorrect because it derives `Deserialize` transparently on the inner `Option<usize>`, which means a value like `30` would bypass the 60-240 range check in `try_new`. We'd have to fix that before using the type directly here.

a-frantz: The problem with using `Option<_>` anywhere in our configs is that there's no way to null a TOML field in `sprocket.toml`. Anywhere with `Option<_>` _needs_ to have a default of `None` or else there's no way for users to set it to `None`. It also becomes impossible for an earlier config in the search to set an `Option<>` field to `Some()` and then a later config in the search to override the earlier by unsetting it

crates/wdl-format/src/config/indent.rs
a-frantz: This is a bit of a bandaid over the fact that in TOML users can't specify `None` for `num_spaces`. The value will always be `Some` if it is coming from a `sprocket.toml`, so we wouldn't want to log a warning

src/config.rs
claymcleod: That's a fair point about TOML. I still don't love `0` as a sentinel value though. One option worth considering: accept either a number, the string `"none"`, or `null` (for formats that support it) via a custom deserializer. That way the Rust type can be `Option<usize>` and TOML users get an explicit way to disable the limit. We handle `null` to be defensive in case we add support for JSON/YAML in the future.

```rust
impl<'de> Deserialize<'de> for MaxLineLength {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where D: Deserializer<'de>
    {
        #[derive(Deserialize)]
        #[serde(untagged)]
        enum Value {
            Num(usize),
            Str(String),
            Null,
        }

        match Value::deserialize(deserializer)? {
            Value::Num(n) => Ok(MaxLineLength(Some(n))),
            Value::Str(s) if s == "none" => Ok(MaxLineLength(None)),
            Value::Str(s) => Err(de::Error::custom(
                format!("expected a number or \"none\", got \"{s}\"")
            )),
            Value::Null => Ok(MaxLineLength(None)),
        }
    }
}
```

crates/wdl-format/src/config/indent.rs
claymcleod: That brings up a broader question — it seems better to capture the semantic of whether a user explicitly set a value or not by defaulting config fields to `None` and resolving to concrete values at the point of use. If `indentation_size` were `Option<usize>` defaulting to `None`, then `Some(n)` with `with_tabs = true` would mean the user deliberately set a conflicting option, and the warning would be meaningful.

src/commands/check.rs
```diff
@@ -159,6 +163,7 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
     let deny_notes = args.common.deny_notes || config.check.deny_notes;
     let deny_warnings = args.common.deny_warnings || config.check.deny_warnings || deny_notes;
     let hide_notes = args.common.hide_notes || config.check.hide_notes;
+    let hide_warnings = args.common.hide_warnings || config.check.hide_warnings;
```
claymcleod: The doc on `--hide-warnings` says it hides "warning or note severity", but `hide_warnings` doesn't chain into `hide_notes`. Would you mind reordering so `hide_warnings` is computed first, then folding it into `hide_notes`?

```suggestion
    let hide_warnings = args.common.hide_warnings || config.check.hide_warnings;
    let hide_notes = args.common.hide_notes || config.check.hide_notes || hide_warnings;
```

src/config.rs
```diff
@@ -212,15 +253,15 @@ pub struct RunConfig {
     /// made by the events channel.
     ///
     /// The default is `5000`.
-    pub events_capacity: Option<usize>,
+    pub events_capacity: usize,
```
claymcleod, line 231: Setting `events_capacity = 0` in `sprocket.toml` will panic at runtime since [`tokio::sync::broadcast::channel`](https://docs.rs/tokio/latest/tokio/sync/broadcast/fn.channel.html) requires capacity >= 1 (see the **Panics** section). Consider adding validation in `Config::validate()`:

```rust
if self.run.events_capacity == 0 {
    bail!("`events_capacity` must be at least 1");
}
```

crates/wdl-engine/src/config.rs
```diff
@@ -838,10 +911,57 @@ pub struct TaskConfig {
...
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum Value {
+            Num(u8),
+            Str(String),
+            Null,
+        }
+
+        match Value::deserialize(deserializer)? {
+            Value::Num(n) => Ok(Retries(Some(n))),
```
claymcleod: The other custom types in this PR (`Parallelism`, `MaxConcurrentRuns`) reject out-of-range values at deser time. `Retries` accepts any `u8` here and relies on `TaskConfig::validate()`, but I don't think `EngineConfig::validate()` is called from the top-level `Config::validate()` — so values 101–255 could slip through. Consider adding a range check here, or ensuring the validation chain reaches `TaskConfig`.

```suggestion
            Value::Num(n) if n <= MAX_RETRIES => Ok(Retries(Some(n))),
            Value::Num(n) => Err(serde::de::Error::custom(format!(
                "retries cannot exceed {MAX_RETRIES}, got {n}"
            ))),
```

crates/wdl-engine/src/config.rs
```diff
@@ -838,10 +911,57 @@ pub struct TaskConfig {
     pub excluded_cache_inputs: HashSet<String>,
 }
 
+/// The default maximum number of retries to attempt if a task fails.
+#[derive(Debug, Clone, Default)]
+pub struct Retries(Option<u8>);
```
claymcleod: Six new types in this PR have custom Serialize/Deserialize impls (`FallBackVersion`, `MaxConcurrentRuns`, `Parallelism`, `Retries`, `MaxLineLength`, `Indent`) but no dedicated roundtrip tests. The 3 removed `max_concurrent_runs` validation tests weren't replaced with deser-level equivalents either. I think it'd be worth adding tests for at least the edge cases: sentinel strings (`"none"`, `"unlimited"`, `"available"`, `"default"`, `"tabs"`), zero-rejection, boundary values, and invalid strings.

crates/wdl-engine/src/config.rs
```diff
@@ -376,30 +384,86 @@ impl Config {
...
+pub struct Parallelism(Option<usize>);
+
+impl Parallelism {
+    /// Get the inner `Option<usize>`.
+    pub fn inner(&self) -> Option<usize> {
+        self.0
+    }
+}
+
+impl Serialize for Parallelism {
```
claymcleod: All six nullable config types in this PR follow the same serialize/deserialize pattern — the only differences are the inner type, sentinel string, and validation predicate. I think a declarative macro like `nullable_config_type!` should be used to consolidate the ~300 LOC of boilerplate and ensure consistent error messages across all six types.

crates/wdl-engine/src/backend/docker.rs
```diff
@@ -231,13 +231,11 @@ impl<'a> DockerTask<'a> {
...
                             .to_string(),
                     )
-                    .program(
-                        self.config
-                            .task
-                            .shell
-                            .as_deref()
-                            .unwrap_or(DEFAULT_TASK_SHELL),
-                    )
+                    .program(if self.config.task.shell.is_empty() {
```
claymcleod: This `is_empty()` pattern for `shell` and `container` appears 9 times across 6 backend files. IMO `TaskConfig` should expose helper methods like `shell_or_default() -> &str` and `container() -> Option<&str>` to centralize the empty-string-as-None semantics and eliminate the repetition.

tests/cli/config-resolve/multi/inputs/third.toml
```diff
@@ -1,5 +1,5 @@
 [run]
-backend = "third"
+backend = ""
```
claymcleod: The other nullable types in this PR use explicit sentinel strings (`"none"`, `"unlimited"`, `"default"`), but the `String` fields (`backend`, `shell`, `container`) use empty-string-as-unset semantics. `backend = ""` silently clearing an earlier config's backend setting could be surprising to users. Would you consider using a consistent approach — e.g., `"default"` or `"none"` sentinels — for these fields as well?

crates/wdl-engine/src/config.rs
```diff
@@ -376,30 +384,86 @@ impl Config {
...
+    pub retries: usize,
     /// The maximum parallelism for file transfers.
     ///
     /// Defaults to the host's available parallelism.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub parallelism: Option<usize>,
+    pub parallelism: Parallelism,
+}
+
+/// Maximum concurrent workflows.
```
claymcleod: This doc comment looks copy-pasted from `MaxConcurrentRuns` — `Parallelism` controls HTTP transfer parallelism, not concurrent workflows.

```suggestion
/// HTTP transfer parallelism.
```

crates/wdl-format/src/config/indent.rs
```diff
@@ -45,15 +47,15 @@ impl Indent {
     pub fn try_new(tab: bool, num_spaces: Option<usize>) -> Result<Self, IndentError> {
         match (tab, num_spaces) {
             (true, None) => Ok(Indent::Tabs),
-            (true, Some(_)) => Err(IndentError::InvalidConfiguration),
+            (true, Some(_)) => Err(IndentError::InvalidTabConfiguration),
             (false, Some(n)) => {
                 if n > MAX_SPACE_INDENT {
                     Err(IndentError::TooManySpaces(n))
                 } else {
                     Ok(Indent::Spaces(n))
```
claymcleod, line 55: `indent = 0` in `sprocket.toml` would produce zero-width indentation here — are we intentionally allowing that? If not, we should probably reject `n == 0` alongside the `n > MAX_SPACE_INDENT` check.

crates/wdl-engine/src/config.rs
```diff
@@ -376,30 +384,86 @@ impl Config {
...
+    }
+}
+
+impl<'de> Deserialize<'de> for Parallelism {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(untagged)]
```
claymcleod: If a user writes `parallelism = true` or `parallelism = 3.14`, none of the `Num/Str/Null` variants match and serde produces the generic "data did not match any variant of untagged enum Value" error. The custom error messages only trigger when a `Str` matches with a wrong value. Not blocking, but a `Visitor`-based deserializer could produce actionable errors for all TOML value types — worth considering if this gets consolidated into a macro (cf. the duplication comment).

crates/wdl-format/src/config/max_line_length.rs
```diff
@@ -53,3 +49,39 @@ impl Default for MaxLineLength {
...
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum Value {
+            Num(usize),
+            Str(String),
+            Null,
+        }
+
+        match Value::deserialize(deserializer)? {
+            Value::Num(n) => MaxLineLength::try_new(Some(n)).map_err(serde::de::Error::custom),
```
claymcleod, line 81: `MaxLineLength` and `Indent` validate via `try_new()` called from their `Deserialize` impl (like here), while `Parallelism` and `MaxConcurrentRuns` do inline validation in the match arms. Standardizing on the `try_new()` pattern would improve consistency and make these types easier to test independently of serde.

claymcleod (changes requested):
A few things to address. Two correctness issues (`--hide-warnings` not hiding notes, `events_capacity = 0` panic) and a validation gap in `Retries`. The rest are consistency and duplication suggestions.

crates/wdl-format/src/config/indent.rs
a-frantz: I think we should allow zero-width indentation. It has been possible since indentation configuration was first introduced and I don't see any reason to remove it now.

tests/cli/config-resolve/bad-values/parallelism/stderr
a-frantz, line 1: I don't know why the key being reported by `config-rs` is wrong. I'm fairly confident it is an upstream bug, not caused by anything we're doing here.

tests/cli/config-resolve/bad-values/task-retries/stderr
a-frantz, line 1: same here

tests/cli/config-resolve/multi/inputs/third.toml
claymcleod: This inconsistency remains—`backend`, `shell`, and `container` use empty-string-as-unset semantics while the `nullable_config_type!` types use explicit sentinels (`"none"`, `"unlimited"`, etc.). In particular, `backend = ""` silently clearing a previous config's backend setting is surprising for users. Not blocking if there's a plan to address this separately, but reopening to keep it tracked.

crates/wdl-engine/src/config.rs
claymcleod: Now that the macro is consolidated, a `Visitor`-based deserializer would be a relatively contained change in one place. The generic "data did not match any variant of untagged enum Value" errors for wrong TOML value types (e.g., `parallelism = true`) still persist. Reopening to keep this tracked—still not blocking.

src/commands/format.rs
```diff
@@ -116,26 +116,26 @@ fn format_document(
...
+            .context("failed to create max line length configuration")?
+    } else {
+        config.format.max_line_length
+    };
 
     let config = FormatConfig::default()
         .indent(indent)
         .max_line_length(max_line_length)
+        .sort_inputs(config.format.sort_inputs)
         .sort_imports(config.format.sort_inputs)
```
claymcleod:
```suggestion
        .sort_inputs(config.format.sort_inputs)
        .sort_imports(config.format.sort_imports)
```

claymcleod (changes requested):
The `nullable_config_type!` macro consolidation looks good—most of my earlier feedback is addressed. Two prior threads remain (string sentinel consistency and untagged enum error messages), plus one new inline comment on `format.rs`.

tests/cli/config-resolve/multi/inputs/third.toml
a-frantz: They have been updated to use sentinels now. `backend = "default"`, `shell = "bash"`, and `container = "ubuntu:latest"`. There are also two `cache` configurations that use the sentinel `system` to refer to the XDG locations

crates/wdl-engine/src/config.rs
```diff
@@ -148,6 +184,86 @@ impl<'de> serde::Deserialize<'de> for SecretString {
...
+    (
+        $name:ident,
+        $inner:ty,
+        $sentinel:literal,
+        $value:ident,
+        $validation:expr,
+        $expected:literal,
+        $default:expr
+    ) => {
+        #[doc = concat!("Configuration for ", stringify!($name), ".")]
```
claymcleod:
```suggestion
        #[doc = concat!("Configuration for [`", stringify!($name), "`].")]
```

crates/wdl-engine/src/config.rs
```diff
@@ -148,6 +184,86 @@ impl<'de> serde::Deserialize<'de> for SecretString {
...
+        #[derive(Clone, Debug)]
+        pub struct $name(Option<$inner>);
+
+        impl $name {
+            #[doc = concat!("Get the innner ", stringify!($inner), ".")]
```
claymcleod:
```suggestion
            #[doc = concat!("Get the inner [`", stringify!($inner), "`].")]
```

crates/wdl-engine/src/config.rs
```diff
@@ -237,19 +347,33 @@ pub struct Config {
...
     pub async fn validate(&self) -> Result<()> {
         self.http.validate()?;
         self.workflow.validate()?;
         self.task.validate()?;
 
-        if self.backend.is_none() && self.backends.len() < 2 {
-            // This is OK, we'll use either the singular backends entry (1) or
-            // the default (0)
+        if self.backends.is_empty() && self.backend == DEFAULT_BACKEND_NAME {
+            // we'll use the default
```
claymcleod, line 372: I think this is a regression, right? If there's only one backend defined, we want to use it, regardless of whether it is named as the default backend.

src/commands/format.rs
```diff
@@ -116,27 +116,28 @@ fn format_document(
...
+        config.format.max_line_length
+    };
 
     let config = FormatConfig::default()
         .indent(indent)
         .max_line_length(max_line_length)
-        .sort_imports(config.format.sort_inputs)
+        .sort_inputs(config.format.sort_inputs)
+        .sort_inputs(config.format.sort_inputs)
+        .sort_imports(config.format.sort_imports)
```
a-frantz, line 145:
```suggestion
        .sort_inputs(config.format.sort_inputs)
        .sort_imports(config.format.sort_imports)
```

crates/wdl-engine/src/config.rs
```diff
@@ -376,36 +493,63 @@ impl Config {
...
+    value > 0,
+    "a positive number",
+    None
+);
+
+impl Default for HttpConfig {
+    fn default() -> Self {
+        Self {
+            cache_dir: get_sentinel_cache_dir(),
+            retries: 5, // default as defined in cloud_copy
```
claymcleod:
```suggestion
            retries: 5, // Default as defined in `cloud_copy`.
```

crates/wdl-engine/src/config.rs
```diff
@@ -755,26 +908,22 @@ pub enum ContentDigestMode {
...
+    pub retries: Retries,
     /// The default container to use if a container is not specified in a task's
     /// requirements.
-    ///
-    /// Defaults to `ubuntu:latest`.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub container: Option<String>,
+    #[serde(
+        default = "get_default_container",
+        skip_serializing_if = "is_default_container"
```
claymcleod: Do we still want this considering the changes to make `sprocket config` print them?

crates/wdl-engine/tests/common/mod.rs
```diff
@@ -71,32 +71,19 @@ pub fn find_tests(
...
     if config_override_path.exists() {
         for config in base_configs.values_mut() {
-            let combined = Figment::from(Serialized::defaults(&config))
-                .merge(Yaml::file_exact(&config_override_path))
-                .extract()?;
+            let combined = config::Config::builder()
+                .add_source(config::Config::try_from(config).unwrap())
+                .add_source(config::File::from(config_override_path.as_path()).required(true))
+                .build()
+                .unwrap()
```
claymcleod: I wonder if these should not bubble their errors up so that the proper error reporting can be done, same as what used to happen.

crates/wdl-format/src/config/max_line_length.rs
```diff
@@ -2,14 +2,13 @@
...
 use serde::Deserialize;
 use serde::Serialize;
-use thiserror::Error;
 
 /// Error while creating a max line length configuration.
-#[derive(Error, Debug)]
+#[derive(thiserror::Error, Debug)]
 pub enum MaxLineLengthError {
-    /// Suppplied value outside allowed range.
+    /// Suppplied number outside allowed range.
```
claymcleod:
```suggestion
    /// Supplied number outside allowed range.
```

crates/wdl-format/src/config/max_line_length.rs
```diff
@@ -53,3 +49,39 @@ impl Default for MaxLineLength {
...
+        #[serde(untagged)]
+        enum Value {
+            Num(usize),
+            Str(String),
+            Null,
+        }
+
+        match Value::deserialize(deserializer)? {
+            Value::Num(n) => MaxLineLength::try_new(Some(n)).map_err(serde::de::Error::custom),
+            Value::Str(s) if s == "none" => Ok(MaxLineLength(None)),
```
claymcleod: nit: suggest consting `"none"`.

src/config.rs
```diff
@@ -30,14 +28,43 @@ const DEFAULT_HOST: &str = "127.0.0.1";
 const DEFAULT_PORT: u16 = 8080;
 
 /// Default database filename.
-pub const DEFAULT_DATABASE_FILENAME: &str = "sprocket.db";
+pub const DEFAULT_DATABASE_NAME: &str = "sprocket.db";
```
claymcleod: Shouldn't we keep `FILENAME` here? I don't see why we would change it.

src/config.rs
```diff
@@ -9,17 +9,15 @@ use anyhow::Context;
...
-use figment::providers::Toml;
 use serde::Deserialize;
 use serde::Serialize;
 use tracing::debug;
 use tracing::warn;
 use url::Url;
 use wdl::ast::SupportedVersion;
 use wdl::engine::Config as EngineConfig;
+use wdl::engine::nullable_config_type;
+use wdl::format::Config as FormatConfig;
```
claymcleod, line 21: The old `FormatConfig` wrapper had `deny_unknown_fields`, but `wdl::format::Config` does not. This means old `sprocket.toml` files with `with_tabs` or `indentation_size` will silently ignore those fields rather than erroring, and typos in `[format]` keys won't be caught. Would you mind adding `deny_unknown_fields` to `wdl::format::Config`?

crates/wdl-engine/src/config.rs
```diff
@@ -148,6 +184,86 @@ impl<'de> serde::Deserialize<'de> for SecretString {
...
+
+        impl<'de> Deserialize<'de> for $name {
+            fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+            where
+                D: serde::Deserializer<'de>,
+            {
+                #[derive(Deserialize)]
+                #[serde(untagged)]
+                enum Value {
+                    Inner($inner),
```
claymcleod, line 246: IMO it's worth noting that this macro should not be instantiated with `String` as `$inner` — serde's `untagged` enum tries variants in order, so `Inner(String)` would match before `Str(String)` and the sentinel would never be recognized. A doc comment on the macro would prevent a subtle future misuse.

crates/wdl-engine/src/config.rs
a-frantz: Yes, this is a behavior change (that should be surfaced in the changelog if we keep it).

The old behavior (if only one defined backend, use it) only worked when `backend.is_none()`, but we no longer have a "none" representation for `backend`. I do not think we should allow mismatches between a _specified_ backend (even if what's specified is `default`) and the selected backend. Imagine a user sets `backend = "some_HPC"` but they've only defined a single backend using a Docker backend. We should not ignore the `some_HPC` specification and silently choose the only defined backend (Docker).
Similarly, if they've defined one backend and named it something other than `"default"`, I think the most transparent behavior is to error when `backend = "default"` as there is a mismatch in the configuration.

crates/wdl-engine/src/config.rs
a-frantz: I think we should "hide" `task.shell`, as it really should not be modified unless the user has a very specific intention and knows what they're doing. I feel similarly but not as strongly about `container`. Really, users shouldn't be relying on default containers. Setting `task.container` to something other than `ubuntu:latest` is just asking for WDL that is non-portable, although I guess that is also true of relying on an ubuntu container...

I'm ok with either exposing or hiding the container field if you have an opinion either way

src/config.rs
a-frantz: I thought it was more consistent with the other consts we have, but on another look we also have a `CONFIG_FILE_NAME` which conflicts with either convention in this diff. I have a slight preference for simply "name" over "filename", but not a strong one. I do not like "file name" as present in the CONFIG const, so I would like to change that to either `CONFIG_NAME` or `CONFIG_FILENAME` for consistency.

src/config.rs
```diff
@@ -359,106 +392,208 @@ impl Default for TestConfig {
...
     /// supplied, the default Sprocket logo will be used.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub alt_light_logo: Option<PathBuf>,
+    #[serde(default = "get_sentinel_doc_config_value")]
+    pub alt_light_logo: String,
     /// An optional link to the project's homepage.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub homepage_url: Option<Url>,
+    #[serde(default = "get_sentinel_doc_config_value")]
+    pub home_url: String,
```
Serial-ATA: IMO this should stay `homepage_url`

If not though, the name needs to be updated to be consistent in both `sprocket dev doc --homepage-url` and `ExternalUrls::homepage`.

a-frantz: I changed this to differentiate it from _the docs site "homepage"_ which is populated with a user provided md file. Maybe this change should be reverted and we can change the name of the current "plain" `homepage` field? Any ideas as to what that should be called?

Serial-ATA: Yeah, maybe `index` or `index_page`?

Serial-ATA (approved):
LGTM, make sure to note the `homepage` -> `index_page` change in the changelog

crates/wdl-doc/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Changed
+
+* `homepage` has been renamed to `index_page` to disambiguate it from `homepage_url` ([#675](https://github.com/stjude-rust-labs/sprocket/pull/675)).
+* `additional_javascript` has been generalized to `additional_html` and now supports inserting HTML in multiple locations at once ([#675](https://github.com/stjude-rust-labs/sprocket/pull/675)).
+
+#### Removed
```
claymcleod: I don't think we should have a "removed" section, this can probably just be under changed.

crates/wdl-engine/src/config.rs
claymcleod: I guess my feeling is slightly to keep this one visible. I agree on `shell`.

src/config.rs
```diff
@@ -32,11 +30,40 @@ const DEFAULT_PORT: u16 = 8080;
 /// Default database filename.
 pub const DEFAULT_DATABASE_FILENAME: &str = "sprocket.db";
 
+/// Sentinel value for using a local database.
+const SENTINEL_DATABASE_FILENAME: &str = "localized";
```
claymcleod: "Localized" typically means "adapted for a locale" (i.e., internationalization). Since the intent is "resolve to a default local path in the output directory," I'd suggest renaming this to `"default"` — consistent with how `backend = "default"` works elsewhere in this PR.

CHANGELOG.md
```diff
@@ -53,10 +56,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Removed
+
+* Removed the `--prioritize-workflows-view` flag to `doc` subcommand ([#675](https://github.com/stjude-rust-labs/sprocket/pull/675)).
+
 ## 0.22.0 - 2026-03-12
 
 ### Added
 
+* New `--hide-warnings` flag to `check`/`lint` with a corresponding `check.hide_warnings` option in `sprocket.toml` ([#675](https://github.com/stjude-rust-labs/sprocket/pull/675)).
```
claymcleod: Should this go under 0.22.0?

CHANGELOG.md
```diff
@@ -66,6 +74,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Changed how defaults for many `sprocket.toml` entries are serialized and deserialized; non-default config value serialization is unchanged ([#675](https://github.com/stjude-rust-labs/sprocket/pull/675)).
```
claymcleod: Same here.

claymcleod (commented):
Just a few thoughts.

src/config.rs
```diff
@@ -531,14 +683,18 @@ impl Config {
...
             Some(&mut self.run.output_dir),
-            self.run.engine.task.cache_dir.as_mut(),
-            self.run.engine.http.cache_dir.as_mut(),
+            self.run.engine.task.cache_dir().as_mut(),
+            self.run.engine.http.cache_dir().ok().as_mut(),
             Some(&mut self.server.output_dir),
-            self.server.engine.task.cache_dir.as_mut(),
-            self.server.engine.http.cache_dir.as_mut(),
+            self.server.engine.task.cache_dir().as_mut(),
+            self.server.engine.http.cache_dir().ok().as_mut(),
```
claymcleod: The four `cache_dir()` / `cache_dir().ok()` calls here return owned values. Calling `.as_mut()` on those temporaries means the shell expansion never reaches the underlying config fields—this is silently a no-op.

The old code worked because it called `.as_mut()` directly on `Option<PathBuf>` struct fields. With the sentinel-based accessors returning owned values, we need a different approach—either expand the underlying string fields directly, or add setter methods and write the expanded paths back.

src/config.rs
```diff
@@ -359,106 +392,209 @@ impl Default for TestConfig {
...
+        } else {
+            match Url::from_str(&self.github_url) {
+                Ok(url) => Some(url),
+                Err(e) => {
+                    tracing::warn!("error while parsing configured GitHub URL: {e}");
+                    None
+                }
+            }
+        }
+    }
```
claymcleod, line 516: `DocConfig` and `DocExtraHtmlConfig` have 8 accessor methods that each repeat the same sentinel-check-and-convert pattern. Consider using `nullable_config_type!` to define `NullablePathBuf` and `NullableUrl` types—this would replace the manual accessors with the same infrastructure used for `FallBackVersion` and `MaxConcurrentRuns`, keeping the sentinel approach consistent across the config layer.

claymcleod, line 516: `homepage_url()` and `github_url()` silently swallow parse errors—a typo like `htp://example.com` logs a warning but returns `None`, and the link quietly vanishes from the docs. Since `Config::validate()` exists for catching config mistakes, it'd be more robust to validate these URLs there and surface an error early.

src/commands/check.rs
```diff
@@ -293,8 +298,12 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
...
                             }
 
-                            counts.warnings += 1;
-                            true
+                            if !hide_warnings {
+                                counts.warnings += 1;
+                                true
+                            } else {
+                                false
+                            }
```
claymcleod, line 306: Warnings are only counted when `!hide_warnings`, which means `--deny-warnings` has no effect when `--hide-warnings` is also set. A user setting both flags expects the command to suppress warning output but still fail if warnings exist.

```suggestion
                            counts.warnings += 1;
                            if !hide_warnings {
                                true
                            } else {
                                false
                            }
```

src/commands/format.rs
```diff
@@ -116,27 +116,27 @@ fn format_document(
...
+        config.format.indent
     };
 
-    let max_line_length = MaxLineLength::try_new(
-        args.max_line_length
-            .unwrap_or(config.format.max_line_length),
-    )
-    .context("failed to create max line length configuration")?;
+    let max_line_length = if let Some(max) = args.max_line_length {
+        MaxLineLength::try_new(if max == 0 { None } else { Some(max) })
```
claymcleod: The CLI uses `0` as the "no limit" sentinel for `--max-line-length`, but the config layer uses the string `"none"`. These should be consistent—the CLI should accept `"none"` (by changing the arg type to a string and parsing it) so that users don't have to remember two different conventions for the same setting.

src/config.rs
```diff
@@ -359,106 +392,209 @@ impl Default for TestConfig {
...
 }
 
 /// `doc.scripts` command configuration.
```
claymcleod:
```suggestion
/// `doc.extra_html` command configuration.
```

crates/wdl-doc/src/config.rs
```diff
@@ -18,18 +16,44 @@ pub struct ExternalUrls {
 
 /// The location to embed an arbitrary JavaScript `<script>` tag into each HTML
 /// page.
```
claymcleod: This doc comment still references JavaScript `<script>` tags, but the struct is now `AdditionalHtml` for arbitrary HTML content.

```suggestion
/// Additional HTML to embed into each generated page.
```

src/config.rs
```diff
@@ -110,74 +137,56 @@ pub struct CommonConfig {
     /// The report mode.
     pub report_mode: Mode,
     /// WDL-specific configuration.
-    #[serde(default)]
     pub wdl: WdlConfig,
 }
 
+nullable_config_type!(
+    FallBackVersion,
```
claymcleod: `FallBackVersion` should be `FallbackVersion`—"fallback" is a single compound word.

```suggestion
    FallbackVersion,
```

src/config.rs
```diff
@@ -359,106 +392,209 @@ impl Default for TestConfig {
     }
 }
 
+/// Sentinel value used throughout DocConfig.
```
claymcleod:
```suggestion
/// Sentinel value used throughout `DocConfig`.
```

claymcleod (changes requested):

src/commands/check.rs
a-frantz: > A user setting both flags expects the command to suppress warning output but still fail if warnings exist.

IMO this should be an error instead. I plan update the `clap` derives to deny this possibility instead

claymcleod: I considered that but thought there could be use cases where they should both be allowed. More than happy to mutually exclude them though. There are other arguments that need to be treated similarly if we do that.

src/config.rs
a-frantz: This is being pushed to a follow up PR immediately after the upcoming release

crates/wdl-doc/CHANGELOG.md
a-frantz: We've used `Removed` before, but inconsistently. I can bump it up to Changed, but noting there's precedent

src/commands/format.rs
```diff
@@ -116,27 +116,33 @@ fn format_document(
...
-        args.max_line_length
-            .unwrap_or(config.format.max_line_length),
-    )
-    .context("failed to create max line length configuration")?;
+    let max_line_length = if let Some(max) = args.max_line_length {
+        let max = match max.as_str() {
+            "none" => None,
+            _ => Some(
+                max.parse::<usize>()
+                    .context("parsing `--max-line-length`")?,
```
claymcleod:
```suggestion
                    .context("`--max-line-length` must be an integer")?,
```

src/config.rs
```diff
@@ -217,46 +228,62 @@ pub struct RunConfig {
...
 /// Server database configuration.
-#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case", deny_unknown_fields)]
 pub struct ServerDatabaseConfig {
-    /// Database URL (e.g., `sqlite://sprocket.db`).
-    /// If not provided, defaults to `sprocket.db` in the output directory.
-    #[serde(default)]
-    pub url: Option<String>,
+    /// Database URL (e.g., `sqlite://sprocket.db`). Defaults to `sprocket.db`
```
claymcleod:
```suggestion
    /// Database URL (e.g., `sqlite://sprocket.db`). Defaults to `sprocket.db` in the output directory.
```

claymcleod (approved):

PR #678, opened 2026-02-28
improve format line breaking

a-frantz:

another pass at improving the line breaking logic. There were some tandem breaks not being respected by the prior PR adding them ( #641 ) that this fixes. This also has better support for `if...then...else` clauses.

Additionally, it fixes a subtle bug in "temporary indent" logic that would trip if multiple placeholders were on the same Bash-line.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* More informative errors for invalid WDL ([#705](https://github.com/stjude-rust-labs/sprocket/pull/705)).
+* Fixed a bug that prevented `sprocket format` from working `--with-tabs` ([#678](https://github.com/stjude-rust-labs/sprocket/pull/678)).
+* More informative errors for invalid WDL while running `test` ([#705](https://github.com/stjude-rust-labs/sprocket/pull/705)).
 * Fixed a regression where nested inputs and task requirements/hints were not
   being respected on the `sprocket run` CLI ([#642](https://github.com/stjude-rust-labs/sprocket/pull/642)).
 


crates/wdl-format/src/token/post.rs
```diff
@@ -594,38 +576,35 @@ impl Postprocessor {
...
                             self.end_line(&mut post_buffer);
                         }
+                    } else if *break_kind == top_of_stack.open {
+                        top_of_stack.depth += 1;
                     }
-                } else if post_buffer.last_line_width(config) > max_length {
-                    // The line is already too long, and taking the next step
-                    // can only make it worse. Insert a line break here.
-                    self.interrupted = true;
-                    self.end_line(&mut post_buffer);
```
a-frantz, line 621: this branch is never hit in my test corpus (including additional repos not included in the test suite) so it was removed. Technically, it is possible to hit _iff_ there are two "potential line break" tokens in a row and the first one of those tokens spans an entire line by itself (never seen this happen, maybe possible with a super long string literal without placeholders, but this would then look funny I imagine? Theoretically anyway).

I _could_ add it back if we want to ensure handling of the above case in bizarro land. But the logic has to be expanded and IMO it just makes this tougher to parse through for a case that is never hit. I spent a non-trivial amount of time mucking around with this logic before I removed the branch entirely and noticed no changes in my test corpus.

The "expanded logic" for the branch to match the rest of this PR can be seen in this commit - https://github.com/stjude-rust-labs/sprocket/pull/678/changes/26f2c1294412746b7910467eebe43fb7b646d546

a-frantz:
dropped back into draft state while me and Andrew discuss https://github.com/stjudecloud/workflows/pull/303

crates/wdl-format/src/token/post.rs
```diff
@@ -418,40 +431,26 @@ impl Postprocessor {
...
-                if kind == SyntaxKind::LiteralCommandText {
-                    self.temp_indent = Rc::new(
-                        value
-                            .chars()
-                            .take_while(|c| matches!(c.to_string().as_str(), SPACE | crate::TAB))
-                            .collect(),
-                    );
+                    let popped = stream.0.pop().unwrap();
+                    // We don't actually want to pop the TempIndent token,
+                    // but rather a regualar Indent token before the temp indent.
```
claymcleod:
```suggestion
                    // but rather a regular Indent token before the temp indent.
```

crates/wdl-format/src/token/post.rs
```diff
@@ -646,19 +632,25 @@ impl Postprocessor {
...
                     };
                     break_stack.push(tandem_break);
+                    self.indent_level += 1;
                 }
             }
         }
 
+        // reduce indent for breaks never added
+        for _ in break_stack {
+            self.indent_level -= 1;
```
claymcleod: Consider using `saturating_sub(1)` here to guard against underflow if `indent_level` is already 0.

```suggestion
        for _ in break_stack {
            self.indent_level = self.indent_level.saturating_sub(1);
        }
```

crates/wdl-format/src/v1/task.rs
```diff
@@ -234,25 +236,49 @@ pub fn format_command_section(
...
+                                if lines.peek().is_none() {
+                                    // save the leading whitespace to use as temporary indent
+                                    bash_indent = Some(
+                                        line.chars()
+                                            .take_while(|c| {
+                                                matches!(
+                                                    c.to_string().as_str(),
+                                                    crate::SPACE | crate::TAB
+                                                )
+                                            })
```
claymcleod: This allocates a `String` per character. Consider matching the `char` directly:

```suggestion
                                            .take_while(|c| matches!(c, ' ' | '\t'))
```

claymcleod: You could const them if you want.

crates/wdl-format/src/v1/expr.rs
```diff
@@ -965,8 +965,17 @@ pub fn format_if_expr(
     config: &Config,
 ) {
     for child in element.children().expect("if expr children") {
+        match child.element().kind() {
+            SyntaxKind::ThenKeyword => {
```
claymcleod, line 983: Two things worth discussing about this change.

First, nested `else if` chains produce a staircase, which seems bad. Because each `IfExpr` applies its own `increment_indent`, chained `else if` expressions accumulate indentation.

Most formatters treat `else if` as a flat construct to keep all branches at the same indent:

```wdl
Array[Array[File]] fastqs_R1 = if length(fastqs_rep10_R1) > 0
    then [...]
else if length(fastqs_rep9_R1) > 0
    then [...]
else if length(fastqs_rep8_R1) > 0
    then [...]
else []
```

To achieve this, `format_if_expr` could check whether the else-branch child is itself an `IfExprNode` and, if so, call `decrement_indent()` before recursing into it. That way the nested `if` starts at the same level as `else`, and the staircase is avoided.

Second, all `if...then...else` are forced multi-line, even trivial ones. Would it make sense to gate the multi-line behavior on whether the expression exceeds `max_line_length`, so short expressions stay on one line?

claymcleod: Here's a sketch of what the flattening could look like.

```rust
pub fn format_if_expr(
    element: &FormatElement,
    stream: &mut TokenStream<PreToken>,
    config: &Config,
) {
    let mut saw_else = false;
    let mut flattened = false;

    for child in element.children().expect("if expr children") {
        match child.element().kind() {
            SyntaxKind::ThenKeyword => {
                stream.increment_indent();
            }
            SyntaxKind::ElseKeyword => {
                stream.end_line();
                saw_else = true;
            }
            SyntaxKind::IfExprNode if saw_else => {
                stream.decrement_indent();
                flattened = true;
            }
            _ => {}
        }

        saw_else = false;
        (&child).write(stream, config);
        stream.end_word();
    }

    if !flattened {
        stream.decrement_indent();
    }
}
```

Each level's `decrement_indent()` in the `IfExprNode` arm undoes the outer `then`'s indent bump, so the recursive call's own `increment_indent` keeps everything at the same level. The `flattened` flag prevents the final `decrement_indent()` from double-decrementing.

crates/wdl-format/src/v1/expr.rs
```diff
@@ -964,9 +964,29 @@ pub fn format_if_expr(
     stream: &mut TokenStream<PreToken>,
     config: &Config,
 ) {
+    let in_chain = matches!(
+        element.element().inner().parent().map(|p| p.kind()),
+        Some(SyntaxKind::IfExprNode)
+    );
```
claymcleod: The `in_chain` check looks at whether the parent node is an `IfExprNode`, but this doesn't distinguish which child position we're in. Consider `if a then if b then c else d else e` — the grammar accepts arbitrary `expr` in all three positions, so the inner if parses as the **then-body**:

```
IfExprNode (outer)
  IfKeyword, a, ThenKeyword,
  IfExprNode (inner),  ← then-body, not an else-if chain
  ElseKeyword, e
```

The inner IfExpr's parent is IfExprNode, so `in_chain = true`, but it shouldn't be flattened. You could check the preceding sibling instead of the parent to distinguish else-if chains from other nesting:

```suggestion
    let in_chain = element
        .element()
        .inner()
        .prev_sibling_or_token()
        .is_some_and(|prev| prev.kind() == SyntaxKind::ElseKeyword);
```

This is admittedly an unusual pattern in practice, but I think we should go ahead and fix.

crates/wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -1508,8 +1605,12 @@ workflow chip {
                 aligner = aligner_,
                 mito_chr_name = mito_chr_name_,
                 custom_align_py = custom_align_py,
-                idx_tar = if aligner == "bwa" then bwa_idx_tar_ else if aligner == "bowtie2"
-                    then bowtie2_idx_tar_ else custom_aligner_idx_tar,
+                idx_tar = if aligner == "bwa"
+                    then bwa_idx_tar_
+                    else if aligner == "bowtie2"
+                    then bowtie2_idx_tar_
+                    else custom_aligner_idx_tar
```
claymcleod, line 1612: Should this comma be on this line? Perhaps an improvement for a future PR?

a-frantz: We had a few of these "dangling commas" appear on the workflows repo. Agreed that this is not ideal. I improved the situation by adding parantheses so the comma appears as `),` instead, but I think it can be improved further

https://github.com/stjudecloud/workflows/pull/303/changes/fb8a88cbaf92c08330cc89e63dee80af05e99fff

PR #682, opened 2026-03-05
feat: make Apptainer executable path configurable

claymcleod:
This was a request from @williamrowell: apparently there is a project called [SingularityCE](https://sylabs.io/singularity/) that some folks use that is basically identical in the commands we use from Apptainer (see [this comment](https://github.com/sylabs/singularity/discussions/2948#discussioncomment-9578289) for a bit more info).

Adds an `executable` field to `ApptainerConfig` so users can specify a custom path to the Apptainer (or Singularity) binary. The field defaults to `"apptainer"` and is threaded through `ApptainerRuntime` to the script generation and image pull sites. The environment variable prefix is derived from the executable name (`SINGULARITYENV_` vs `APPTAINERENV_`).

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/src/backend/apptainer.rs
+++ b/crates/wdl-engine/src/backend/apptainer.rs
@@ -29,7 +29,7 @@ use tracing::warn;
 
 use crate::Value;
 use crate::backend::ExecuteTaskRequest;
-use crate::config::Config;
+use crate::config::ApptainerConfig;
 use crate::config::DEFAULT_TASK_SHELL;
 use crate::v1::requirements::ContainerSource;
 
@@ -57,9 +57,6 @@ const SINGULARITY_ENV_PREFIX: &str = "SINGULARITYENV";
 /// Represents the Apptainer container runtime.
 #[derive(Debug)]
 pub struct ApptainerRuntime {
-    /// Path to the container runtime executable (e.g., `"apptainer"` or
-    /// `"singularity"`).
-    executable: String,
     /// The cache directory for `.sif` images.
     cache_dir: PathBuf,
     /// The map of container source to `.sif` path.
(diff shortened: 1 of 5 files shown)


crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -152,12 +163,18 @@ impl ApptainerRuntime {
                 )
             })?;
 
+        let env_prefix = if self.executable.contains("singularity") {
```
claymcleod: Personally, I think this is fine. If you disagree, let me know.

crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -48,21 +48,32 @@ const GUEST_STDOUT_PATH: &str = "/mnt/task/stdout";
...
+
+/// The environment variable prefix for Singularity.
+const SINGULARITY_ENV_PREFIX: &str = "SINGULARITYENV";
+
 /// Represents the Apptainer container runtime.
 #[derive(Debug)]
 pub struct ApptainerRuntime {
+    /// Path to the container runtime executable (e.g., `"apptainer"` or
+    /// `"singularity"`).
+    executable: String,
```
peterhuene: Does `ApptainerRuntime` need to store the executable? Seems like `generate_apptainer_script` could perhaps instead have a signature of:

```rust
async fn generate_apptainer_script(
    &self,
    config: &ApptainerConfig,
    shell: Option<&str>,
    container_sif: &Path,
    request: &ExecuteTaskRequest<'_>,
) -> Result<String>
```

and by extension `generate_script` would have a signature of:

```rust
pub async fn generate_script(
    &self,
    config: &ApptainerConfig,
    shell: Option<&str>,
    request: &ExecuteTaskRequest<'_>,
    token: CancellationToken,
) -> Result<Option<String>>
```

which would allow the backends to pass in both the executable and the extra options in a single arg.

We don't need to pass in a `Config` like we're doing now just to get the shell out of it, either.

peterhuene (commented):
Just the one comment relating to arguments to the apptainer runtime.

Otherwise, looks good 👍

peterhuene (approved):

PR #683, opened 2026-03-05
fix: (Bug )avoid LSP crash when switching git branches (#680)

### LINKED ISSUE #680: LSP panic when switching branches
**Describe the bug**
Occasionally when switching branches, the `sprocket analyzer` process will immediately crash.

**Terminal Output**

```bash
thread '<unnamed>' (16018) panicked at crates/wdl-analysis/src/graph.rs:346:30:
cannot apply edits to a document that was not previously parsed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Rayon: detected unexpected panic; aborting
```

**Additional context**
Experiencing this while using the `sprocket-intellij` plugin.

akashchakrabortymsc-cmd:
Fixes #680

## As Solution 
Instead of panicking, I added a fallback — if the document was never parsed, we just do a fresh full parse from disk and recover quietly. Also removed the leftover `use std::panic;` .

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/src/graph.rs
```diff
@@ -343,9 +342,13 @@ impl DocumentGraphNode {
...
                         ),
-                        _ => panic!(
-                            "cannot apply edits to a document that was not previously parsed"
-                        ),
+                        _ => {
+                            tracing::warn!(
+                            uri = %self.uri,
+                                "document not yet parsed, falling back to full parse"
+                            );
+                            return self.full_parse(tokio, client);
```
Serial-ATA: `self.change` and `self.parse_state` aren't changed prior to this point, so calling `full_parse` again will just end up in an infinite loop.

What really needs to happen here is a re-fetch of the source. Any previous incremental changes likely can't be trusted and we need to start fresh.

Serial-ATA (changes requested):

crates/wdl-analysis/src/graph.rs
akashchakrabortymsc-cmd: Updated the fix

 instead of calling `full_parse` again, we now directly re-fetch the source from disk and return an error state if that fails. This way we start completely fresh without any risk of looping.

akashchakrabortymsc-cmd (commented):
Let me know if  there is still something to change

PR #685, opened 2026-03-05
fix: remap symlinks before saving call cache entry.

### LINKED ISSUE #684: Guest symlink updating can break cache checking
**Describe the bug**
The functionality in #580 and #621 introduces rewriting of symlinks in the container outputs. This updates the timestamp for the symlink. The caching should check the timestamp of the underlying file, but this doesn't seem be happening.

For St. Jude users, the context is this Slack thread: https://stjude.slack.com/archives/C065YHBRNBS/p1772724498213549

**Expected behavior**
Sprocket should rewrite the symlinks prior to calculating the digests for the cache. It also needs to resolve symlinks and check the timestamp on the underlying target.

peterhuene:
This PR ensures that remapping of local work directory symlinks occurs before saving the call cache entry.

By doing so, the digest of the work directory is calculated with updated symlinks rather than potentially broken symlinks.

Fixes #684.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-engine/CHANGELOG.md
+++ b/crates/wdl-engine/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Fixed
 
 * Fixed erroneous trimming of container extensions ([#690](https://github.com/stjude-rust-labs/sprocket/pull/690)).
+* Fixed a regression caused by PR 621 where remapping symlinks in a work
+  directory would invalidate the task's call cache entry ([#685](https://github.com/stjude-rust-labs/sprocket/pull/685)).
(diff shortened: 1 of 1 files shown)


crates/wdl-engine/src/eval/v1/task.rs
adthrasher, line 1120: This gets triggered on a rerun of the cached task.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -956,6 +956,16 @@ impl Evaluator {
                 continue;
             }
 
+            // Remap any guest symbolic links to the corresponding host paths
+            // This must occur *before* we put the result in the cache to ensure consistent
+            // work directory digesting
+            if let Err(e) = self.remap_links(&state, &result.work_dir) {
```
adthrasher:
```suggestion
            if !cached && let Err(e) = self.remap_links(&state, &result.work_dir) {
```

PR #686, opened 2026-03-05
feat: `wdl-diagnostics` crate

Serial-ATA:
Split from #628

Before submitting this PR, please make sure:

For external contributors:

- [ ] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [ ] You have not used AI on any parts of this pull request.
- [ ] You have added a few sentences describing the PR here.
- [ ] Your code builds clean without any errors or warnings.

For all contributors:

- [x] You have added tests (when appropriate).
- [x] You have added an entry in the CHANGELOG (when appropriate).
- [x] You have updated the README or other documentation to account for these changes (when appropriate).
- [x] You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).

For PRs containing lint rule changes:

- [x] You have updated any and all effected entries within `RULES.md`.
- [x] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

PR #692, opened 2026-03-05
refactor(sprocket): move doc tests to `wdl-doc`

Serial-ATA:
* Moved the `custom-logo` test to `wdl-doc`
* Added `sprocket dev doc --check` to analyze documents without producing output
  * Used in the `ignorefile` and `invalid_doc` CLI tests

Depends on #686

For external contributors:

For all contributors:

      is implemented.


crates/wdl-doc/tests/ui/custom_logo/custom_logo.rs
```diff
@@ -0,0 +1,85 @@
...
+            let generated_image_path;
+
+            let src = image.attr("src").await?;
+            match src.as_deref() {
+                Some(EXPECTED_IMAGE_DARK) => {
+                    generated_image_path = docs_path.join(EXPECTED_IMAGE_DARK);
+                    found_dark = true;
+                }
+                Some(EXPECTED_IMAGE_LIGHT) => {
+                    generated_image_path = docs_path.join(EXPECTED_IMAGE_DARK);
```
claymcleod: This looks like a copy-paste leftover from the `EXPECTED_IMAGE_DARK` arm above — the light branch is reading the dark path (`docs_path.join(EXPECTED_IMAGE_DARK)`). The sibling `custom_logo_alt.rs:57` has the right shape. Should be `EXPECTED_IMAGE_LIGHT` here.

src/commands/doc.rs
```diff
@@ -33,6 +33,9 @@ use crate::commands::CommandResult;
 pub struct Args {
     /// Path to the local WDL workspace to document.
     pub workspace: Option<Source>,
+    /// Analyze the documents without producing an output.
+    #[arg(long, conflicts_with_all = ["output", "open"])]
```
claymcleod: `--overwrite` isn't in this conflicts list, which means `sprocket dev doc --check --overwrite` will `remove_dir_all(&docs_dir)` at line 189 before `document_workspace` short-circuits on `config.check` — silent data loss against a flag that's documented as "analyze without producing an output". Broadening the list to include every output-affecting flag (`overwrite`, `logo`, `alt_light_logo`, `homepage_url`, `github_url`, `light_mode`, `index_page`, `html_head`, `html_body_open`, `html_body_close`, `theme`, `install`) would also make findings about pre-short-circuit work (theme build, HTML embeds) moot.

crates/wdl-doc/tests/ui/custom_logo/custom_logo_alt.rs
```diff
@@ -0,0 +1,85 @@
+//@args:--logo assets/test-logo.svg --alt-light-logo assets/test-logo.light.svg
```
claymcleod: This file duplicates ~85 lines of `custom_logo.rs` — same navigation, same `#logo img` query, same loop, same `found_dark`/`found_light` pair. Only differences are the extra light-SVG fixture and the slightly different branch structure. Consider extracting a shared `assert_logo_variants(driver, docs_path, category_dir, light_asset: Option<&str>)` helper in `mod.rs`; both test cases collapse to a few lines and the copy-paste bug at `custom_logo.rs:58` can't recur.

crates/wdl-doc/tests/ui.rs
```diff
@@ -98,35 +102,73 @@ pub type TestMap = HashMap<&'static str, Arc<dyn UiTest>>;
...
+                break;
+            };
+
+            let Some((field, value)) = meta.split_once(':') else {
+                bail!("Malformed meta line in '{}': {line}", path.display());
+            };
+
+            match field {
+                "args" => {
+                    metadata.wdl_doc_args = value.split(' ').map(ToString::to_string).collect()
```
claymcleod: `split(' ')` on the raw arg string means any path with a space in it silently breaks. A future contributor adding `--logo assets/my logo.svg` will get a confusing failure. `shlex::split(value)` would handle this, or this is moot if `//@args:` is replaced by a trait method.

crates/wdl-doc/tests/ui.rs
```diff
@@ -136,150 +178,125 @@ async fn find_tests(tmp_dir: PathBuf) -> anyhow::Result<Vec<Trial>> {
...
-
-                    // Cleanup
-                    driver.close_window().await?;
-                    driver.switch_to_window(ctx.primary_window.clone()).await?;
-
-                    result
-                };
+            let metadata = match TestMetadata::load(&test_path) {
+                Ok(m) => m,
+                Err(e) => {
```
claymcleod: On `TestMetadata::load` failure this logs an error and `continue`s — which silently drops the test from the discovered set. A malformed `//@args:` line in a test file means that test just doesn't run, with only a log line to notice. Prefer `bail!` (or push the error into `found` as a failing placeholder) so the harness exits non-zero on malformed metadata.

crates/wdl-doc/bin/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+        match e.kind() {
+            DocErrorKind::AnalysisFailed(analysis_results) => {
+                for result in analysis_results {
+                    let path = result.document().path().to_string();
+                    let source = result.document().root().text().to_string();
+
+                    emit_diagnostics(
+                        &path,
+                        source,
+                        result.document().diagnostics().filter(|d| {
```
claymcleod: The filter has a side effect (mutating `counts.errors` inside the predicate). Readable enough here but fragile — if the iterator is ever consumed speculatively (e.g. by a size hint or adapter), the count silently drifts. Prefer counting explicitly after filtering, or using `inspect`.

crates/wdl-doc/tests/ui.rs
```diff
@@ -399,36 +456,94 @@ async fn main() -> Result<(), anyhow::Error> {
...
     let args = libtest_mimic::Arguments::from_args();
-    let tests = find_tests(tmp.path().to_path_buf()).await?;
+    let tests = Arc::new(find_tests()?);
 
-    let conclusion = libtest_mimic::run(&args, tests);
+    let trials = create_trials(tmp.path().to_path_buf(), tests);
+    let conclusion = libtest_mimic::run(&args, trials);
 
+    cleanup_global_state().await;
+    conclusion.exit()
```
claymcleod: `conclusion.exit()` calls `std::process::exit`, which bypasses destructors. `tmp: TempDir` at line 466 is still in scope here, so the tempdir (and anything inside — generated HTML docs plus potentially the built `wdl-doc` binary output) leaks on every successful run, not just on panic. Cleanup relies on OS tempdir rotation. Either drop explicitly before exit (`let code = conclusion.exit_code(); drop(tmp); std::process::exit(code)`) or restructure so `main` returns `Result` and lets destructors run.

claymcleod (changes requested):
Wanted to flag this before merge: the previous revision of this harness called `wdl_doc::document_workspace()` directly as a library function. This PR replaces that with a new `wdl-doc-bin` crate + a `//@args:` mini-DSL parsed at runtime + an in-test `cargo build -p wdl-doc-bin` + a subprocess invocation per test. All of that machinery exists only to re-serialize `Config` through a second CLI surface.

An alternative: add a `configure(ConfigBuilder) -> ConfigBuilder` method to the `UiTest` trait, default-implemented to identity. `custom_logo` would then express its needs as:

```rust
fn configure(&self, cfg: ConfigBuilder) -> ConfigBuilder {
    cfg.custom_logo(Some(asset("test-logo.svg")))
}
```

That deletes the `bin/` crate, `TestMetadata`, `//@args:` parsing, the `"Malformed meta line"` / `"Unexpected meta line"` error paths, the `cargo build` inside tests, the `target/debug` lookup, and restores structured error propagation.

I saw quite a few more things, but I'm going to stop here pending the high-level approach question.

crates/wdl-doc/bin/src/main.rs
Serial-ATA: `wdl-doc-bin` was removed anyway, but I don't see this being an actual problem. This was a direct copy of other `sprocket` subcommands: https://github.com/stjude-rust-labs/sprocket/blob/85d3943b70a3c37b6c5f9e0b8b7221b50f656258/src/commands/check.rs#L293

CHANGELOG.md
```diff
@@ -140,6 +140,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   sharing pulled `.sif` images across runs ([#693](https://github.com/stjude-rust-labs/sprocket/pull/693)).
 * Intermediate `test` results are logged as they complete ([#674](https://github.com/stjude-rust-labs/sprocket/pull/674)).
 * New lint rule `DocCommentTabs` to ensure doc comments do not contain tab characters ([#664](https://github.com/stjude-rust-labs/sprocket/pull/664)).
+* `sprocket dev doc --check` to analyze documents without producing an output ([#691](https://github.com/stjude-rust-labs/sprocket/pull/691))
```
peterhuene: This CHANGELOG update appears to be in the wrong location.

tests/cli/doc/custom_logo/outputs/docs/pagefind/filter/en_65e219d.pf_filter
peterhuene, line 1: Should this file still exist after these changes?

Same comment for this and every other white-space only change to a `.../pagefind/...` file.

peterhuene (commented):
Looks good, I have just two comments.

tests/cli/doc/custom_logo/outputs/docs/pagefind/filter/en_65e219d.pf_filter
Serial-ATA: that whole directory doesn't exist anymore: https://github.com/Serial-ATA/sprocket/tree/doc-cli-tests/tests/cli/doc

Don't know why it doesn't show that correctly

peterhuene: Can confirm the files don't exist despite what PR UI says (although in the treeview it does show as deleted :upside_down_face:).

peterhuene (approved):

peterhuene:
Given @claymcleod's comments on this PR, we can wait until he returns to ensure his concerns are addressed.

src/commands/doc.rs
claymcleod: This is closer, but `--check` still accepts `--theme` and `--install`, and those paths run before `DocConfig::check(true)` gets a chance to short-circuit. I verified `sprocket dev doc --check --theme /tmp/sprocket-missing-theme .` fails while trying to build the theme, so this mode can still run generation-only subprocesses instead of only analyzing. Let's make `--check` conflict with the remaining generation-only flags, at least `theme` and `install`, and ideally the rest of the output customization flags too.

crates/wdl-doc/tests/ui.rs
```diff
@@ -399,36 +347,98 @@ async fn main() -> Result<(), anyhow::Error> {
...
+                        tracing::warn!("Failed to close tab: {e}");
+                    }
+                    if let Err(e) = driver.switch_to_window(ctx.primary_window.clone()).await {
+                        tracing::warn!("Failed to switch to primary window: {e}");
+                    }
 
+                    test_result.map_err(|e: anyhow::Error| e.into())
+                };
+
+                block_on(task)
```
claymcleod: `#[tokio::main]` already puts `libtest_mimic::run` on a Tokio runtime, so this nested `block_on` panics when the trials execute on that runtime thread. `IN_CI=1 cargo test -p wdl-doc --test ui --locked -- --test-threads 1` currently fails every test with `Cannot start a runtime from within a runtime`. One option is to make `main` synchronous again and use `block_on` only at the async boundaries, or otherwise avoid calling `block_on` from inside the runtime created for `main`.

claymcleod (changes requested):
Thanks for iterating on the harness. The original architecture concern is addressed; I left one remaining test-runner issue inline.

claymcleod (approved):

PR #698, opened 2026-03-06
feat: support container image fallback lists

### LINKED ISSUE #691: Support fallback container image pulling from arrays
Currently, when a WDL task specifies an array of container images via the `container` runtime key, Sprocket only uses the first element in the array. If that image fails to pull, the task simply fails—there's no attempt to fall back to subsequent entries. Sprocket should iterate through the array and try each image in order—for instance, so that a workflow can specify a preferred private image with a public fallback (e.g., `["ghcr.io/org/custom-image:v1.0", "ubuntu:22.04"]`). There's already a [NOTE in the source](https://github.com/stjude-rust-labs/sprocket/blob/b6e145b0a3fa13874b797b6a3e0bd059c2e3a1eb/crates/wdl-engine/src/eval/v1/task/requirements.rs#L166-L167) acknowledging this needs work, and it sounds like some upstream changes in Crankshaft may be needed to enable container resolution and fallback.

claymcleod:
When a WDL task specifies an array of container images in its `container` requirement, each candidate is now tried in order until one pulls successfully. This lets task authors provide fallback images for environments where a preferred registry might not be reachable. The WDL wildcard value `*` resolves to the configured default container (e.g., `ubuntu:latest`) at parse time rather than being modeled as a separate enum variant.

The actual container that was used during execution is now tracked via `TaskExecutionResult` and stored in `TaskPostEvaluationData`, so downstream consumers always know which image ran—not just which candidates were offered.

Closes #691

For all contributors:


crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -96,14 +84,13 @@ impl ApptainerRuntime {
...
-        shell: Option<&str>,
+        config: &Config,
         request: &ExecuteTaskRequest<'_>,
+        extra_args: impl Iterator<Item = &str>,
         token: CancellationToken,
-    ) -> Result<Option<String>> {
-        let path = match self
-            .pull_image(
-                &config.executable,
+    ) -> Result<Option<(String, ContainerSource)>> {
```
claymcleod, line 106: Here, we probably need to update the docstring to indicate that the returned container source is the one selected by Apptainer.

crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -493,11 +495,10 @@ mod tests {
         env.insert("FOO".to_string(), "bar".to_string());
         env.insert("BAZ".to_string(), "\"quux\"".to_string());
 
-        let runtime = ApptainerRuntime::new(&root.path().join("runs"), None).unwrap();
-        let script = runtime
+        let runtime = ApptainerRuntime::new(&root.path().join("runs")).unwrap();
+        let (script, _container) = runtime
```
claymcleod:
```suggestion
        let (script, _) = runtime
```

crates/wdl-engine/src/backend/slurm_apptainer.rs
```diff
@@ -923,16 +918,19 @@ impl TaskExecutionBackend for SlurmApptainerBackend {
...
 
-        let container =
+        let containers =
             requirements::container(inputs, requirements, self.config.task.container.as_deref());
-        if let ContainerSource::Unknown(_) = &container {
-            bail!(
-                "Slurm Apptainer backend does not support unknown container source `{container:#}`"
-            )
+        for container in &containers {
+            if let ContainerSource::Unknown(_) = container {
```
claymcleod: Do we really want to fail if we come across _any_ source that is unknown? Or only if we get to the list in line and it's unknown? I tend to think the latter.

crates/wdl-engine/src/backend/tes.rs
```diff
@@ -472,6 +478,12 @@ impl TaskExecutionBackend for TesBackend {
                 work_dir_url.path_segments_mut().unwrap().push("");
 
                 return Ok(Some(TaskExecutionResult {
+                    container: request
+                        .constraints
+                        .container
+                        .as_ref()
+                        .and_then(|c| c.first())
```
claymcleod, line 475: I wonder if we need an endpoint for seeing if we can access a container from within TES? Something to consider for the future. cc: @peterhuene

crates/wdl-engine/src/eval/v1/task/requirements.rs
```diff
@@ -150,51 +153,59 @@ impl std::fmt::Display for ContainerSource {
...
+        // SAFETY: `FromStr` for `ContainerSource` is infallible.
+        return vec![fallback.parse().unwrap()];
+    };
+
+    if let Some(array) = value.as_array() {
+        array
+            .as_slice()
+            .iter()
+            .map(|v| {
+                let s = v.as_string().expect("type should be string");
```
claymcleod: This probably needs to be coercion to string (and needs a test to that effect).

crates/wdl-engine/src/value.rs
```diff
@@ -3310,10 +3311,7 @@ impl TaskPostEvaluationValue {
             name: Arc::new(name.into()),
             id: Arc::new(id.into()),
             data: Arc::new(TaskPostEvaluationData {
-                container: constraints
-                    .container
-                    .as_ref()
-                    .map(|c| Arc::new(c.to_string())),
+                container: None,
```
claymcleod, line 3381: Add a note here why this is okay.

Cargo.toml
```diff
@@ -40,7 +40,7 @@ cloud-copy = { version = "0.7.0", features = ["cli"] }
 codespan-reporting = "0.13.1"
 colored = "3.1.1"
 convert_case = "0.11.0"
-crankshaft = { version = "0.9.0", features = ["docker"] }
+crankshaft = { path = "../crankshaft/crankshaft", features = ["docker"] }
```
claymcleod: Need to update this before we merge.

crates/wdl-engine/src/backend/tes.rs
claymcleod: I've opted for this instead: https://github.com/ga4gh/task-execution-schemas/issues/249.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -767,6 +767,7 @@ impl Evaluator {
                         container: &constraints
                             .container
                             .as_ref()
+                            .and_then(|c| c.first())
```
peterhuene: So I think we'll need to propagate `container` in the cache key request to be a slice otherwise modification of only the first one would trigger a cache miss, right?

Additionally, I'd probably have `ContainerSource` implement `Hashable` and make the field in `KeyRequest` a `&[ContainerSource]` rather than convert them all to string.

crates/wdl-engine/src/backend.rs
```diff
@@ -116,13 +119,68 @@ impl Input {
     }
 }
 
+/// The result of attempting to pull a single container image.
+pub type PullResult<T> = anyhow::Result<T>;
```
peterhuene: I personally feel like this type alias is unnecessary as these types shouldn't be public to the crate, but can definitely live with it.

crates/wdl-engine/src/backend.rs
```diff
@@ -116,13 +119,68 @@ impl Input {
...
+
+    /// Iterates over the failed pull attempts.
+    pub fn failures(&self) -> impl Iterator<Item = (&ContainerSource, &anyhow::Error)> {
+        self.0
+            .iter()
+            .filter_map(|(source, result)| result.as_ref().err().map(|e| (source, e)))
+    }
+}
+
+impl<T> Deref for PullResultMap<T> {
```
peterhuene: Do we actually use the `Deref` and `DerefMut` implementations of `PullResultMap` other than calling `insert`? That is to say, do we need this type to be an associative collection? It feels like it's used more like a `Vec`.

If not, I would consider changing the internal data type to `Vec<(ContainerSource, Result<T>)>` and perhaps renaming it to `PullResults` or something.

crates/wdl-engine/src/backend.rs
```diff
@@ -263,3 +323,81 @@ pub(crate) trait TaskExecutionBackend: Send + Sync {
...
+    #[test]
+    fn empty_pull_result_map_has_no_successful_container() {
+        let map: PullResultMap<String> = PullResultMap::default();
+        assert!(map.successful_container().is_none());
+    }
+
+    #[test]
+    fn empty_pull_result_map_reports_all_failed() {
+        let map: PullResultMap<String> = PullResultMap::default();
+        assert!(map.all_failed());
```
peterhuene: Looks like the `all_failed` method's implementation is missing.

peterhuene (commented):
A few comments to go over.

crates/wdl-engine/src/cache.rs
```diff
@@ -319,10 +320,10 @@ pub struct KeyRequest<'a> {
     ///
     /// This field contributes to the digests stored in a cache entry.
     pub command: &'a str,
-    /// The container used by the task.
+    /// The container candidates for the task.
     ///
     /// This field contributes to the digests stored in a cache entry.
-    pub container: &'a str,
+    pub container: &'a [ContainerSource],
```
peterhuene: Unfortunately, my previous review comment was in error.

I forgot that this field is the container *used by* (i.e. previously the first in the array but now the first to successfully pull with these changes) the task and not the `container` requirement itself.

The requirement gets digested and stored in the cache entry. So any change to the requirement would already have caused a cache miss.

Thus we should revert these changes and the `container` field on `KeyRequest` and `Key` both should be removed.

Instead, the `container` field of `CallCacheEntry` should be changed to be `Option<ContainerSource>` (with `ContainerSource` being made serializable).

The `CallCache::put` method should just store the `container` from the `TaskExecutionResult` in the call cache entry (and, correspondingly, the `get` method should return it).

peterhuene: However, that leaves us with a problem that the previous code was solving with the `container` field in `KeyRequest` and `Key`, namely if the task does not have a `container` requirement, we'll use the configured default container for the task's execution.

We expect a change to the default container to cause a cache miss when the task does not have a `container` requirement.

The difference in the old code is that we (albeit it with the limitation fixed by this PR) calculated the container that was "used" prior to calculating the cache key. With these changes, it's the backend that determines what container is used and that's performed after we calculate the cache key.

So perhaps we do need a `String` field on both `KeyRequest` and `Key` after all, not for the "selected container", but for the _default_ container that is only checked for changes if and only if there was no `container` (or alias name `docker`) requirement present for the task.

claymcleod: Addressed in 2ebfeb69. Reverted `container` on `Key`/`KeyRequest`, added `CallCacheEntry.container: Option<ContainerSource>` populated from `TaskExecutionResult` to track what actually ran, and added a narrowly-scoped `default_container: Option<String>` field on `Key`/`KeyRequest` that is only set when the task has no `container` requirement — so a change to the configured default container still triggers a cache miss in that case. Updated the test accordingly (`modified_default_container`).

peterhuene (approved):
Looks great!

PR #719, opened 2026-03-12
Include current document in reference search scope

vshnvii:
### Problem

The reference search currently determines the search scope using:

`graph.transitive_dependents(doc_index)`

This returns all documents that depend (directly or indirectly) on the current document. The system then iterates through these documents and collects references from them.

However, the current document itself may not always be included in the returned set of transitive dependents. Because of this, references that exist **inside the same document** might not be searched.

This can lead to incomplete reference results when the reference appears in the same file where the symbol is defined.

---

### Solution

This change ensures that the current document is explicitly included in the search scope.

After collecting the transitive dependents, the current document index is checked and added to the search scope if it is not already present.

This guarantees that references within the same document are also included during the search.

---

### Before

The search scope only included documents that depend on the current document.

If the current document was not part of the returned set, references within that document could be skipped.

```
search scope = transitive_dependents(doc_index)
```

---

### After

The search scope now includes:

* all transitive dependents
* the current document itself

```
search scope = transitive_dependents(doc_index) + current document
```

This ensures that reference collection also checks the document where the symbol originates.

---

### Impact

* Improves correctness of reference search
* Ensures references within the same document are not missed
* Maintains existing search behavior while making the search scope more complete


claymcleod:
Thanks for the contribution, @vshnvii! I think this change is actually a no-op. `graph.transitive_dependents(doc_index)` delegates to `Bfs::new(&self.inner, doc_index).iter(&self.inner)`, and petgraph's `Bfs` always yields the starting node first — so `doc_index` is always present in `search_scope`, and the `if !search_scope.contains(&doc_index)` branch will never fire.

If you're observing references within the same document being missed in practice, I'd love to see a reproducer — the root cause would be somewhere else (e.g. `collect_references_from_document` or `goto_definition`). And if it's just a defensive guard against the method's name-vs-behavior mismatch, then the cleaner fix is probably in `graph.rs`: either rename `transitive_dependents` or update its doc comment to reflect that the starting node is included.

For now, I'm going to close this PR, but feel free to open up another issue/PR with a fix if you find one!

PR #724, opened 2026-03-12
fix: adds enum type coercion support

### LINKED ISSUE #706: enum with default interpreted as option enum
**Describe the bug**
With the new enum functionality, if a task provides a default value, the `call` considers the type to be optional and declares a type mismatch

**Expected behavior**
The default value should be fine.

**Terminal Output**
```bash
% sprocket check enum_test.wdl 
error: type mismatch: expected type `type?`, but found type `type`
   ┌─ enum_test.wdl:25:23
   │
25 │             my_enum = type.B
   │             -------   ^^^^^^ this is type `type`
   │             │          
   │             this expects type `type?`

error: failing due to 1 error
```

Here is a WDL toy example that demonstrates the issue.

```wdl
version 1.3

enum type {
    A,
    B,
    C
}

task hello {
    input {
        String? name
        type my_enum = type.A
    }

    command <<<
        echo "Hello, ~{name} of type ~{my_enum}!"
    >>>
    
}

workflow main {
    call hello {
        input:
            name = "Alice",
            my_enum = type.B
    }
}
```

**Screenshots**
N/A

 - OS: Linux & mac
 - Version: sprocket 0.21.1 (2026-02-16)
 - Shell: bash
 - Toolchain version: rustc 1.93.0-nightly (83e49b75e 2025-12-03)

**Additional context**
N/A

claymcleod:
`EnumType::is_coercible_to` always returned `false`, and `CompoundType::is_coercible_to` had no match arm for enums. Together, this meant that passing a non-optional enum value to a task input with a default (which makes the expected type optional in WDL >= 1.2) produced a false type mismatch diagnostic.

This adds the missing coercion logic for enums, mirroring how struct coercion is already handled.

Closes #706.

For all contributors:

### DIFF
--- a/crates/wdl-analysis/CHANGELOG.md
+++ b/crates/wdl-analysis/CHANGELOG.md
@@ -9,8 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
 
-* Added enum type coercion support so that enum values can be passed to
-  call inputs where the expected type is optional due to a default value
+* Added enum type coercion support
   ([#706](https://github.com/stjude-rust-labs/sprocket/issues/706),
   [#724](https://github.com/stjude-rust-labs/sprocket/pull/724)).
 


crates/wdl-analysis/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Fixed
+
+* Added enum type coercion support so that enum values can be passed to
+  call inputs where the expected type is optional due to a default value
```
a-frantz: per my comment, I'm confused by this phrasing. Code and tests LGTM though, so 👍

a-frantz:
```suggestion
* Added enum type coercion support
```
I don't think we need to elaborate beyond this. The original phrasing made me do a double-take, as "the expected type is optional due to a default value" is not an intuitive or well-known semantic of the specification. The specifics here don't matter beyond the coercion support IMO

claymcleod:
> > before these clarifications, I would describe the expected behavior in these cases as undefined
> 
> I think this behavior clarification should be back-ported in point-release clarifications to WDL v1.0 and v1.1. It was a clarification that arrived in the v1.2 revision, but to make it >=v1.2 specific and undefined in earlier revisions feels confusing and an odd special case.
> 
> I don't think this behavior is intuitive, but the conversation is resolved at this point so there's not really anything to do besides clarifying the behavior in earlier versions. I do see the use cases this behavior enables, so, not trying to make a big stink out of a debate resolved before I joined the conversation 😅
> 
> The thread suggests that cromwell and miniwdl already have this behavior regardless of WDL version, so it seems Sprocket is the odd-one-out.
> 
> See comment for my suggested rephrase of the changelog.
> 
> EDIT: IMO the paragraph I linked to in the 1.3 spec should also be updated. I think the phrasing present is misleading. The table below does make the behavior explicit and clearly defined, but the text preceding the table reads strangely to me.

I'm happy to take this on, but that work is in the spec. Given that, I'm going to merge this work and create an issue.

PR #727, opened 2026-03-13
refactor: shrinks `Value` from 96 to 24 bytes

claymcleod:
While investigating Sprocket's baseline memory usage, I noticed that certain workloads—particularly those involving deeply nested compound types like `Map[String, Map[String, ...]]`—scaled poorly. The root cause was the `Value` enum in `wdl-engine`, which occupied 96 bytes per element due to large inline types (`Type` at 56 bytes, `TaskPostEvaluationValue` at 96 bytes) inflating the enum even though the common case (`PrimitiveValue`) is only 16 bytes.

This refactor pushes `Arc` wrapping down into the types that inflate `Value`. Each compound value type (`Array`, `Map`, `Pair`, `Struct`, `EnumVariant`) and each large hidden value type (`TaskPreEvaluation`, `TaskPostEvaluation`) becomes a thin wrapper around `Arc<Inner>`, where the inner struct owns all fields directly—eliminating nested `Arc` layering. Two new wrapper types (`NoneValue` and `TypeNameRefValue`) handle variants that previously stored a bare `Type`. `CallValue` wraps its `CallType` field behind `Arc` as well.

The tradeoff is an additional pointer indirection when accessing fields of compound and hidden values, since their data now lives behind `Arc` on the heap rather than inline in the enum. In practice, this cost is negligible: these types were already heap-allocated internally (e.g., `Array` stored an `Arc<Vec<Value>>`), so the refactor consolidates rather than adds indirection. Clone remains cheap—a refcount bump on one `Arc` rather than multiple.

A const assertion prevents `Value` from exceeding 24 bytes in the future. The `as_map` standard library function was also updated to use the `IndexMap` entry API, eliminating a redundant key clone.

Benchmarked on an Apple M3 Max against the prior version across five synthetic workloads:

| Benchmark | Before RSS | After RSS | Memory reduction | Before time | After time |
|-|-|-|-|-|-|
| Nested arrays (1K, 2^10) | 36 MB | 34 MB | -6% | 0.05s | 0.04s |
| Nested arrays (59K, 3^10) | 300 MB | 243 MB | -19% | 0.63s | 0.58s |
| Flat array (1M strings) | 549 MB | 423 MB | -23% | 4.50s | 3.52s |
| Nested arrays (1M, 4^10) | 3,578 MB | 2,980 MB | -17% | 8.93s | 8.45s |
| Nested maps (1M, 32^4) | 3,731 MB | 2,457 MB | -34% | 11.76s | 11.06s |

The improvement is most pronounced for nested maps because each map entry stores a `Value` for the key and a `Value` for the nested map, so the 4x per-element reduction compounds across all four nesting levels. Flat arrays benefit less because only one `Array` container exists—the savings come from shrinking each element `Value` within its backing `Vec`.

For all contributors:


crates/wdl-engine/src/stdlib/as_map.rs
```diff
@@ -59,12 +59,17 @@ fn as_map(context: CallContext<'_>) -> Result<Value, Diagnostic> {
...
-            ));
+        match elements.entry(key) {
+            indexmap::map::Entry::Occupied(e) => {
+                return Err(function_call_failed(
+                    FUNCTION_NAME,
+                    DuplicateKeyError(e.key().clone()),
+                    context.arguments[0].span,
+                ));
+            }
+            indexmap::map::Entry::Vacant(e) => {
```
claymcleod, line 70: This is more of a correctness sidequest rather than a performance improvement. In the case where the key is not in the map, we can avoid the preemptive clone of the key.

crates/wdl-engine/src/value.rs
```diff
@@ -2076,7 +2126,7 @@ impl Struct {
             .collect::<Result<IndexMap<_, _>>>()?;
 
         for (name, ty) in ty.members().iter() {
-            // Check for optional members that should be set to `None`
+            // Check for optional members that should be set to none
```
claymcleod, line 2110: I took the approach that WDL `None` values should be referred to as "none" to disambiguate it from Rust's `None` value.

crates/wdl-engine/src/value.rs
```diff
@@ -2102,58 +2152,58 @@ impl Struct {
     pub(crate) fn new_unchecked(
         ty: impl Into<Type>,
         name: Arc<String>,
-        members: Arc<IndexMap<String, Value>>,
+        members: impl Into<IndexMap<String, Value>>,
```
claymcleod, line 2136: Unlike `name`, every caller of this just constructed an `IndexMap` and immediately wrapped it with an `Arc`. So the original `Arc` was truly just for making this cheaply clonable.

crates/wdl-engine/src/eval/v1/expr.rs
```diff
@@ -1320,7 +1319,8 @@ impl<C: EvaluationContext> ExprEvaluator<C> {
                 // Evaluate the argument expressions
                 let mut count = 0;
                 let mut types = [const { Type::Union }; MAX_PARAMETERS];
-                let mut arguments = [const { CallArgument::none() }; MAX_PARAMETERS];
+                let mut arguments: [CallArgument; MAX_PARAMETERS] =
+                    std::array::from_fn(|_| CallArgument::none());
```
peterhuene: I think this could just be a call to `repeat` as `CallArgument` should probably be `Clone`:


```suggestion
                let mut arguments: [CallArgument; MAX_PARAMETERS] =
                    std::array::repeat(CallArgument::none()));
```

peterhuene (commented):
Looks great, just one minor nit!

PR #728, opened 2026-03-13
feat(sprocket): shell-expand config paths

### LINKED ISSUE #391: Path values in `wdl-engine` configuration are not being shell expanded
Since we already have a dependency on `shellexpand` to handle this in input files, we should be shell expanding any path value from the configuration as well.

For example, specifying the HTTP cache directory or the apptainer image cache directory should support `~` and environment variables.

Serial-ATA:
closes #391

For external contributors:

For all contributors:

      is implemented.


src/commands/run.rs
```diff
@@ -527,6 +527,32 @@ fn inputs_to_json(target: &str, inputs: &Inputs) -> Result<String> {
...
+        config.run.engine.http.cache_dir.as_mut(),
+    ]
+    .into_iter()
+    .flatten()
+    {
+        match shellexpand::path::full(path.as_path()) {
+            Ok(expanded) => *path = PathBuf::from(expanded),
+            Err(e) => {
+                return Err(anyhow!(
+                    "failed to expand '{}' in path '{path:?}': {}",
```
peterhuene: Nit: we tend to use backticks as delimiters in our error messages.

```suggestion
                    "failed to expand `{}` in path `{path:?}`: {}",
```

src/commands/test.rs
```diff
@@ -789,18 +790,16 @@ pub async fn test(
 
     let test_dir = workspace.join(WORKSPACE_TEST_DIR);
     let fixture_origins = EvaluationPath::from(test_dir.join(FIXTURES_DIR).as_path());
-    let engine = {
```
peterhuene, line 792: This was a weird block of code. Thanks for fixing it.

src/commands/run.rs
```diff
@@ -532,6 +532,32 @@ fn inputs_to_json(target: &str, inputs: &Inputs) -> Result<String> {
     Ok(serde_json::to_string(&map)?)
 }
 
+/// Shell expand directory paths in `config.run`.
```
claymcleod: The `server` subcommand has an identical `output_directory` and `engine` config with `task.cache_dir` / `http.cache_dir` — these paths won't get shell-expanded. One way to cover both would be to move this logic into `Config::validate()`, which already exists as a normalization point and where `ServerConfig::validate()` already canonicalizes paths. That would also eliminate the `pub(crate)` cross-module import.

claymcleod:
```suggestion
/// Shell-expands directory paths in `config.run`.
```

src/commands/run.rs
```diff
@@ -532,6 +532,32 @@ fn inputs_to_json(target: &str, inputs: &Inputs) -> Result<String> {
...
+    .flatten()
+    {
+        match shellexpand::path::full(path.as_path()) {
+            Ok(expanded) => *path = PathBuf::from(expanded),
+            Err(e) => {
+                return Err(anyhow!(
+                    "failed to expand `{}` in path `{path:?}`: {}",
+                    e.var_name.to_string_lossy(),
+                    e.cause
+                )
```
claymcleod: Use `path.display()` instead of `{path:?}` for consistency with the rest of the codebase.

```suggestion
                return Err(anyhow!(
                    "failed to expand `{}` in path `{}`: {}",
                    e.var_name.to_string_lossy(),
                    path.display(),
                    e.cause
                )
```

claymcleod (changes requested):
The `server` subcommand has the same `output_directory` and `engine` config with identical `task.cache_dir` / `http.cache_dir` fields, but they won't get shell-expanded. Moving the expansion logic into `Config::validate()` would cover all subcommands in one place.

claymcleod (approved):
Looks good — thanks!

PR #729, opened 2026-03-14
refactor: shrinks `Type` from 56 to 24 bytes

claymcleod:
Follow-up to #727, which shrunk `Value` from 96 to 24 bytes. This PR applies the same `Arc<Inner>` pattern to the `Type` enum, reducing it from 56 bytes to 24 bytes. This is more of a PR centered around consistency of these types with the `value.rs` types and just overall correctness than a performance argument.

Each compound type (`ArrayType`, `StructType`, `EnumType`, `CallType`) now wraps its fields in an `Arc<Inner>` struct, and `CompoundType` is Arc-wrapped at the `Type` enum level as `Compound(Arc<CompoundType>, bool)`. Inner Arcs were audited and removed where they existed only for cheap cloning—they're retained only where there is genuine sharing (e.g., `StructType`'s name is shared with the engine's `Struct` value type) or where copy-on-write via `Arc::make_mut` is used (e.g., `CallType`'s outputs in `optional()`/`promote_scatter()`). Each retained Arc field is documented with the rationale.

A const assertion prevents `Type` from exceeding 24 bytes in the future, matching the assertion added for `Value` in #727.

Benchmarked on an Apple M3 Max against #727 across the same five synthetic workloads:

| Benchmark | #727 RSS | This PR RSS | Memory reduction | Time delta |
|-|-|-|-|-|
| Nested arrays (1K, 2^10) | 34 MB | 34 MB | ~0% | ~0% |
| Nested arrays (59K, 3^10) | 243 MB | 219 MB | -10% | ~0% |
| Flat array (1M strings) | 423 MB | 427 MB | ~0% | ~0% |
| Nested arrays (1M, 4^10) | 2,980 MB | 2,797 MB | -6% | ~0% |
| Nested maps (1M, 32^4) | 2,457 MB | 2,384 MB | -3% | ~0% |

The improvement is most pronounced for deeply nested array workloads, where many intermediate `Array`/`Map`/`Pair` types are constructed—each carrying a `Type` that is now 32 bytes smaller.

For all contributors:


crates/gauntlet/src/repository.rs
```diff
@@ -233,6 +233,7 @@ impl Repository {
     }
 }
 
+/// The maximum depth to which submodules are recursively initialized.
```
claymcleod, line 236: I'm not sure why this wasn't firing earlier, but I've fixed it anyway here.

crates/wdl-analysis/src/types.rs
```diff
@@ -173,7 +173,7 @@ pub enum Type {
     /// The type is a compound type.
     ///
     /// The second field is whether or not the compound type is optional.
-    Compound(CompoundType, bool),
+    Compound(Arc<CompoundType>, bool),
```
peterhuene: Is this `Arc` necessary? Every variant of `CompoundType` itself is wrapped in an `Arc` internally and `CompoundType` is 16 bytes in size, so this variant remains the same size.

peterhuene: Additionally, reverting this would also allow us to revert a bunch of changed lines back to nested matching which the `Arc` prevented.

claymcleod: Good catch—this was just a holdout from earlier work I had done but forgot to come back to.

PR #733, opened 2026-03-16
fix: improve ShellCheck diagnostic span for WDL interpolations

### LINKED ISSUE #605: No clear syntax to avoid shellcheck SC2086 for expression substitutions
**Describe the bug**
`sprocket lint` + shellcheck complains about unquoted placeholder expressions that are ignored by `miniwdl check` + shellcheck.  Some of these are valid complaints, I think, but others don't have clear workarounds.

Example: `{if threads > 1 then "--threads " + (threads - 1) else ""}`

Because the leading `--` is missing, it isn't treated as an option and shellcheck (when called by `sprocket lint`) flags it as SC2086.

Incidentally, `miniwdl check` + shellcheck also doesn't flag unquoted substitutions in filenames with SC2086, but I think the correct behavior is to flag those.  I'm not sure what the appropriate solution for the longer expression substitutions should be, though.

**Expected behavior**
Ideally we'd have some workaround to handle command substitutions that wouldn't require suppressing the rule for the entire command.

**Terminal Output**

```bash
$ cat << EOF > samtools_index.wdl
version 1.0

task samtools_index {
  meta {
    description: "index a bam with samtools"
    outputs: {
      index: {
        description: "bam index"
      }
    }
  }

  parameter_meta {
    bam: {
      description: "aligned, sorted bam"
    }
  }

  input {
    File bam
  }

  Int threads = 4
  Int mem_gb = 8

  command <<<
    set -euo pipefail

    samtools index --bai \
      ~{if threads > 1 then "--threads " + (threads - 1) else ""} \
      --output ~{bam}.bai \
      ~{bam}
  >>>

  output {
    File index = "${bam}.bai"
  }

  runtime {
    docker: "quay.io/pacbio/pb_wdl_base@sha256:4b889a1f21a6a7fecf18820613cf610103966a93218de772caba126ab70a8e87"
    cpu: threads
    memory: mem_gb + " GiB"
  }
}

EOF

$ shellcheck --version

ShellCheck - shell script analysis tool
version: 0.9.0
license: GNU General Public License, version 3
website: https://www.shellcheck.net

$ sprocket --version

sprocket 0.20.1 (2026-02-04)

$ sprocket lint --report-mode one-line samtools_index.wdl

samtools_index.wdl:30:7: note[ShellCheck]: Double quote to prevent globbing and word splitting.
samtools_index.wdl:31:16: note[ShellCheck]: Double quote to prevent globbing and word splitting.
samtools_index.wdl:32:7: note[ShellCheck]: Double quote to prevent globbing and word splitting.

$ miniwdl version

miniwdl v1.13.1

$ miniwdl check --debug --strict samtools_index.wdl 

samtools_index.wdl
    task samtools_index
```

 - OS: linux, Rocky 9
 - Version: WDL version 1.0, sprocket 0.20.1
 - Shell: bash
 - Toolchain version: rustc 1.93.0 (254b59607 2026-01-19)

**Additional context**

kris70lesgo:

ShellCheck diagnostics for WDL command placeholders currently highlight the entire `~{...} `placeholder even when only a specific expression inside the interpolation is problematic.
This pr improves diagnostic precision by tracking the span of the interpolated expression and mapping ShellCheck warnings back to that span instead of the full placeholder.
Changes:

- Refactor `to_bash_var()` to return the span of the interpolated expression.
- Track placeholder -> expression span mapping during command sanitization.
- Update shellcheck_lint() to use the expression span when emitting diagnostics.
Example

Before:
```
~{if threads > 1 then "--threads " + (threads - 1) else ""}
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
After:

```
"--threads " + (threads - 1)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lint/src/rules/shellcheck.rs
```diff
@@ -502,23 +527,25 @@ fn evaluates_to_bash_literal(expr: &Expr) -> bool {
 /// it is replaced with a literal value.
 /// If it is a string, then the string is checked to see if it evaluates to a
 /// literal. Otherwise, it is replaced with a bash variable.
-fn to_bash_var(placeholder: &Placeholder, ty: Option<Type>) -> (String, bool) {
+fn to_bash_var(placeholder: &Placeholder, ty: Option<Type>) -> (String, bool, Option<Span>) {
```
a-frantz: can we define a new struct for this return instead of returning a tuple?

a-frantz (commented):
this looks like a great start! Can you fix all the issues reported by CI?

claymcleod:
Tightening the span to the inner expression is a clear win for complex placeholders like `~{if threads > 1 then "--threads " + (threads - 1) else ""}` — the fix genuinely lives inside the expression, and @a-frantz's workaround in #605 makes that clear. But for the simple `~{bam}` case (which accounts for most of the `Arena.toml` column shifts in this PR), the canonical bash fix is `"~{bam}"` — quote the whole placeholder. Pointing at just `bam` arguably hides that signal and reads as "quote `bam` somehow" rather than "wrap the substitution".

One option is conditional tightening: keep the full `~{...}` span when the expression is a bare identifier, and only narrow when the fix lives inside (addition, `if/then/else`, etc.).

@kris70lesgo @a-frantz — what do you think?

kris70lesgo:
@claymcleod thanks for calling it out
rn this always narrows the span to the inner expression, which works well for compound expressions where the fix clearly lives inside the interpolation, but for simple identifiers like ~{bam}, I agree this loses context 
the correct fix is to quote the entire substitution ("~{bam}"), and highlighting only `bam` can be misleading.
I think a conditional approach is the right direction:
- keep the full ~{...} span for simple identifiers
- narrow the span only when the expression is non-trivial

I can update the implementation to distinguish these cases based on the expression AST and adjust the span

kris70lesgo:
@claymcleod @a-frantz can u review the pr again ?
Thanks

kris70lesgo:
@claymcleod It looks like several CI failures are coming from changes outside the scope of this PR (in wdl-lsp/server.rs).
My changes are limited to ShellCheck span handling, and all related tests (including arena) are passing locally.
Would you prefer:
1. keeping this PR scoped and waiting for the upstream failures to be resolved, or  
2. including a minimal fix here to unblock CI?
Happy to proceed either way.

Serial-ATA:
@kris70lesgo your branch has https://github.com/kris70lesgo/sprocket/commit/85d3943b70a3c37b6c5f9e0b8b7221b50f656258 which was removed from `main`, you'll have to drop that commit

kris70lesgo:
@Serial-ATA thanks for pointing that out 
Addressed the remaining CI issues ,all checks are now passing
@claymcleod implemented conditional span handling:
- full ~{...} span for simple identifiers
- inner expression span for non-trivial expressions

Would appreciate your review

crates/wdl-lint/src/rules/shellcheck.rs
```diff
@@ -330,13 +330,40 @@ fn shellcheck_lint(
...
+        let end = diagnostic.end_column + shift_tree.prefix_sum(diagnostic.end_line - 1, 0) - 1;
+        (start, end)
+    };
+
+    // Check if this diagnostic falls within a placeholder's expression span.
+    // If so, use the expression span instead of the full placeholder span.
+    let refined_expr_span = expr_spans
+        .iter()
+        .find_map(|((var_start, var_end), expr_span)| {
+            if diagnostic_span_in_cmd.0 >= *var_start && diagnostic_span_in_cmd.1 <= *var_end {
```
peterhuene: Span ends are exclusive, so should this be `diagnostic_span_in_cmd.1 < *var_end`?

crates/wdl-lint/src/rules/shellcheck.rs
```diff
@@ -330,13 +330,40 @@ fn shellcheck_lint(
...
     line_map: &HashMap<usize, Span>,
     shift_tree: &FenwickTree<usize>,
+    expr_spans: &HashMap<(usize, usize), Span>,
 ) -> Diagnostic {
     let label = format!(
         "SC{}[{}]: {}",
         diagnostic.code, diagnostic.level, diagnostic.message
     );
+    // This span is relative to the command text.
+    let diagnostic_span_in_cmd = {
```
peterhuene: Personally, I think the use of the `tuple` here is unnecessary and makes the code harder to read rather than just having a `start` and `end` variable.

crates/wdl-lint/src/rules/shellcheck.rs
```diff
@@ -523,31 +548,58 @@ fn evaluates_to_bash_literal(expr: &Expr) -> bool {
     }
 }
 
+/// Result of converting a WDL placeholder to bash.
+struct BashVar {
+    /// The bash substitution string.
+    value: String,
+    /// Whether this is a literal (true) or bash variable (false).
+    quoted: bool,
```
peterhuene: This field name is confusing. I'd just just calling this `literal`.

peterhuene (changes requested):
Hi @kris70lesgo.

Sorry for the delay in reviewing this!

I'm not sure how these changes are supposed to work as you've described.

Because we're not actually evaluating the WDL expressions during linting, we can't know which branch of an `if` expression would be taken. As such, we treat the entire placeholder expression as a replacement using the placeholder's expression type.

In your example, the placeholder expression `~{if threads > 1 then "--threads " + (threads - 1) else ""}` has type `String`, so we replace the entire placeholder with a dummy Bash variable to represent it.

There simply is no general solution by which we can narrow the span to the "problematic" branch of `"-- threads" + (threads - 1)`. At best, we could apply a specific "use the span for the alternative branch for an `if` expression when one branch is an empty string literal" heuristic to solve this.

Having a test that covers such a case would illustrate the problem and one should be added to the PR for it to be accepted.

I don't particularly find it useful to point at the placeholder's expression vs. the placeholder itself given the limitation I just mentioned.

As such I'm inclined to close this PR as "won't fix", but I'll leave this open for a little longer in case further discussion is needed.

crates/wdl-lint/src/rules/shellcheck.rs
kris70lesgo: Yes, I updated this to use the exclusive end bound: `diagnostic_end < *var_end`.

kris70lesgo:
@peterhuene Thanks for the detailed explanation. I adjusted the implementation to avoid generally narrowing diagnostics to arbitrary placeholder expressions.

The updated behavior now keeps the full placeholder span by default, and only applies the specific heuristic you described: for an `if` expression where exactly one branch is an empty string literal, the diagnostic span is scoped to the non-empty branch.

I also added focused unit tests for:
- bare identifier placeholders keeping the full placeholder span
- `if` expressions with an empty false branch
- `if` expressions with an empty true branch
- `if` expressions without an empty branch falling back to the full placeholder span

a-frantz:
I'm still OOO ( :face_with_thermometer: ) and haven't reviewed this PR, but I did see @peterhuene 's comment which misunderstands what I intended for this lint improval. I'll briefly clarify what I envisioned so this can get back on track!

The issue with the current implementation of `ShellCheck` is that while it correctly identifies placeholders with potentially problematic word splitting at any depth of the WDL expression, it will always highlight the entire placeholder even in cases where only a specific branch of the expression is the problem.

```
note[ShellCheck]: Double quote to prevent globbing and word splitting.
   ┌─ tmp.wdl:30:7
   │
30 │       ~{if threads > 1 then "--threads " + (threads - 1) else ""} \
   │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   │       │
   │       SC2086[info]: Double quote to prevent globbing and word splitting.
   │       more info: https://www.shellcheck.net/wiki/SC2086
   │
   = fix: address the diagnostic as recommended in the message
```
the current diagnostic above suggests that the whole expression (both `if` branches) should be quoted despite that naive solution (which would silence the lint) leading to incorrect evaled Bash at runtime.

Instead, we should narrow the scope of the span to only "the problem branch" (`"--threads " + (threads -1)`) to make the correct Bash easier to spot. **In theory, there could be multiple "problem branches" within one placeholder.** No heuristics are needed, although the `ShellCheck` implementation will have to be updated to potentially return multiple diagnostics for any singular WDL placeholder

although I expect multiple problem branches in "real" WDL being unlikely, the correct solution will involve parsing the whole expression and highlighting multiple spans without a short circuit once a problem is detected

Hope this clarifies things!

peterhuene:
@a-frantz I don't think I misunderstood what was intended, but that certainly could be the case.

That said, let's first see if my understanding of how the Shellcheck lint works now and the desired functionality matches reality.

The lint applies a (not always correct) heuristic to determine if a placeholder expression is a "literal" and thus could be substituted with a dummy literal value or, failing that, a Bash variable. I say "not always correct" as there are any number of WDL expressions that could be substituted with a literal, but aren't by this heuristic; for example, the placeholder `~{"~{1}"}` would be substituted with a Bash variable and not a literal.

In the above example placeholder expression, the lint will end up substituting a Bash variable for the placeholder because the subexpression `(threads - 1)` is not considered literal (correctly so).

The "sanitized" command that the lint feeds Shellcheck ends up with (effectively) a `${foo}` in it where the placeholder would be.

Shellcheck rightfully doesn't like that because the variable is not quoted in the sanitized command, so it emits a diagnostic saying so with the span of `${foo}` in the sanitized command. The lint fixes up the Shellcheck-reported span of the diagnostic based off of some data structures used to map Shellcheck spans back to the original source spans.

This results in highlighting the entire span of the placeholder in the lint's diagnostic and it is misleading to users who might attempt to fix the issue by putting quotes around the placeholder, but that would not be the correct thing to do in this case. So instead the lint should report the span of the "incorrect" branch of the `if` expression for users to quote.

Here is where I run into issues thinking about a general solution to knowing the "incorrect" branch to highlight as Shellcheck reported spans will never map to something _inside_ of a placeholder as we're substituting for the entire placeholder and not _parts_ of the placeholder.

Perhaps I am missing something in how to solve this _without_ a heuristic, an example being "if the placeholder is an `if` expression with a branch being an empty string literal, then the _alternative_ branch is likely the problem so use its span instead". Although, for what Shellcheck diagnostics is that always true? Does this span fix-up need to be limited to specific Shellcheck diagnostic codes?

Could you walk me through the general solution you're envisioning?

claymcleod:

The underlying work remains tracked in #605, so it won't be lost. You're welcome to reopen this PR if you'd like to pick it back up.

claymcleod:
Actually, looking at this more closely, it looks like we're waiting on a response from @a-frantz. I'll reopen and remove the label.

a-frantz:
this totally fell off my radar, thanks for the ping @adthrasher ! I'm going to take a closer look at this soon. Apologies for the delay here :)

a-frantz:
I've realized that relying on ShellCheck to emit this diagnostic is an unnecessary complication.

Through static analysis, we have all the information needed to determine when a WDL placeholder may introduce problematic shell splitting. We're already doing this on `main`, but it is obfuscated by a pass through shellcheck; essentially we assume every WDL placeholder will introduce problematic splitting, unless we can determine the final substitution will be some form of "literal". We control what shellcheck sees with this "literal or Bash var" logic and as such its our WDL-specific-logic that is ultimately toggling whether SC2086 fires or not.

We should remove this redirection: we can simply silence SC2086 in all cases, and have a new lint rule, which I'll call `ShellSplitting` for the rest of this comment. A lot of the logic already in this PR could be pulled out and refactored into this new rule.

`ShellSplitting` should warn users if they are inserting an unquoted user-supplied WDL String, File, or Directory. A single WDL placeholder may emit any number of these diagnostics through conditional branching. `ShellSplitting` will need to look at the `LiteralCommandText` as well as the placeholder contents. If a placeholder occurs during a quoted span (either single or double), we can move on and assume the author has intentionally handled the quoting. This would only need to fire for placeholders that are not embedded within Bash quotes already.

a-frantz:
@kris70lesgo are you interested in this new direction for the issue? I know we left you hanging for a while.

If not, I can take this over sometime soon when my queue opens up :)
I'll make sure to credit you as a co-author if I do!

claymcleod:
Hi there @kris70lesgo,

Thanks for the contribution! Unfortunately, you haven't responded to issue in a while, and it looks like it's been superseded by another PR. I'm going to close this, but feel free to reopen if you want to take a look before the other PR merges. Thanks!

PR #737, opened 2026-03-19
Implemented the formatter change so short if/then/else can stay inline

### LINKED ISSUE #721: Allow short if/then/else expressions to remain on a single line
When formatting if/then/else expressions, the formatter currently forces every if/then/else onto multiple lines—even trivially short ones like `if x then a else b`. It would be nice to keep simple expressions inline when they fit within the configured max line length.

The main challenge is that at PreToken time, we don't know how far into the line we already are (e.g., the expression could be inside a declaration or a command placeholder). A rough heuristic using a fraction of `max_line_length` works for many cases but produces inconsistent results at the boundary. The more robust approach would be to handle this in the postprocessor, which knows the actual column position and can make a precise fit-or-split decision—though that requires buffering and collapsing multi-line structures, which adds complexity to the postprocessor.

This should probably be a configurable formatter option, since some people prefer the consistency of always splitting while others prefer the compactness of keeping short expressions inline.

Hemanntnagar:
Fixes #721 

Implemented the formatter change so short if/then/else can stay inline when enabled, while wrapping still happens in postprocessing when line length exceeds the configured max.

 For external contributors:
 
 * [x]  You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
 * [x]  You have not used AI on any parts of this pull request.
 * [x]  You have added a few sentences describing the PR here.
 * [x]  Your code builds clean without any errors or warnings.
 
 For all contributors:
 
 * [x]  You have added tests (when appropriate).
 * [x]  You have added an entry in the CHANGELOG (when appropriate).
 * [x]  You have updated the README or other documentation to account for these changes (when appropriate).
 * [x]  You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).
 
 * [x]  You have updated any and all effected entries within `RULES.md`.
 * [x]  You have added a test case in `crates/wdl-lint/tests/lints` that covers every
   is implemented.


claymcleod:
Unfortunately, I don't think this approach is correct either. `render_pre_tokens` re-runs the whole step loop a second time just to measure width—it clones the pre-token stream, spins up a fresh `Postprocessor`, and manually copies five fields of state (`position`, `indent_level`, `interrupted`, `line_spacing_policy`, `temp_indent`) to do so. That's 2x the rendering cost on every candidate expression, and the manual state copy is a drift hazard: the moment someone adds a field to `Postprocessor`, collapse measurement silently diverges from the real render with no compiler error.

A better approach would be to do the collapse entirely on the already-rendered `post_buffer` in `crates/wdl-format/src/token/post.rs`. After the existing step loop finishes assembling `post_buffer`, scan the `TokenStream<PostToken>` for a simple `if … then … else` span, measure its joined width against `config.max_line_length`, and if it fits, strip the intervening line-break and indent post-tokens in place before appending to `out_stream`. That lets you drop `render_pre_tokens` entirely and replace the pre-token index surgery in `collapsed_simple_if_then_else` with a single pass over post-tokens. Could you try that instead?

peterhuene:
Hi @Hemanntnagar. Thanks again for this contribution!

I'm going to close this PR for inactivity given the outstanding unaddressed feedback comment by @claymcleod.

Please feel free to reopen the PR if you push up additional changes to review that address that comment.

We appreciate your help!

PR #738, opened 2026-03-19
switch API pagination to cursor tokens

### LINKED ISSUE #581: Replace offset-based pagination with cursor-based pagination
The list operations in the database layer currently use `OFFSET` clauses for pagination. This works fine for now, but doesn't scale well for large datasets since the database has to scan through all preceding rows to reach the offset. We should eventually switch to cursor-based (keyset) pagination using a column like `created_at` as the cursor. The API already uses a `next_token`, so this can be changed without breaking the API contract.

See https://use-the-index-luke.com/no-offset for more background.

From https://github.com/stjude-rust-labs/sprocket/pull/540#discussion_r2743144387.

raman976:
Fixes #581 

Replaces offset-based pagination with cursor-based (keyset) pagination across all list operations. Instead of OFFSET clauses , queries now use composite WHERE (created_at, tiebreaker) conditions , which use indexes directly and scale with dataset size. The next_token field now encodes an opaque base64 cursor instead of a numeric offset ,existing clients sending numeric tokens will receive 400 Bad Request. The total field is retained for compatibility.

For external contributors:

- [X] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [X] You have added a few sentences describing the PR here.
- [X] Your code builds clean without any errors or warnings.

For all contributors:

- [X] You have added tests (when appropriate).
- [X] You have added an entry in the CHANGELOG (when appropriate).
- [X] You have updated the README or other documentation to account for these changes (when appropriate).
- [X] You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).

- [X] You have updated any and all effected entries within `RULES.md`.
- [X] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      is implemented.

### DIFF
--- a/src/server/api/v1/pagination.rs
+++ b/src/server/api/v1/pagination.rs
@@ -1,0 +1,21 @@
+//! Helpers for opaque pagination tokens.
+
+use base64::Engine;
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use serde::Serialize;
+use serde::de::DeserializeOwned;
+
+/// Encodes a pagination cursor into an opaque token.
+pub fn encode_token<T: Serialize>(cursor: &T) -> Result<String, serde_json::Error> {
+    let json = serde_json::to_vec(cursor)?;
+    Ok(URL_SAFE_NO_PAD.encode(json))
+}
+
+/// Decodes a pagination cursor from an opaque token.
+pub fn decode_token<T: DeserializeOwned>(token: &str) -> Result<T, String> {
+    let decoded = URL_SAFE_NO_PAD
+        .decode(token)
+        .map_err(|_| format!("invalid `next_token`: `{token}`"))?;
+
+    serde_json::from_slice(&decoded).map_err(|_| format!("invalid `next_token`: `{token}`"))
+}
--- a/src/server/api/v1/runs.rs
+++ b/src/server/api/v1/runs.rs
@@ -17,9 +17,22 @@ use uuid::Uuid;
 use super::AppState;
 use super::RunStatus;
 use super::error::Error;
+use super::pagination::decode_token;
+use super::pagination::encode_token;
 use super::send_command;
+use crate::system::v1::db::RunCursor;
 use crate::system::v1::exec::svc::RunManagerCmd;
 use crate::system::v1::exec::svc::run_manager::commands;
+
+/// Cursor payload used for run pagination tokens.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct RunsCursorToken {
+    /// Creation time of the last run on the previous page.
+    created_at: DateTime<Utc>,
+    /// UUID of the last run on the previous page.
+    uuid: Uuid,
+}
 
 /// Request to submit a new run.
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
@@ -57,7 +70,8 @@ pub struct ListRunsQueryParams {
     #[serde(default)]
     pub limit: Option<i64>,
     /// Token for pagination. It is expected that clients pass the value from a
-    /// previous response to retrieve the next page.
+    /// previous response to retrieve the next page. Legacy numeric tokens are
+    /// not supported.
     #[serde(default)]
     pub next_token: Option<String>,
 }
@@ -161,7 +175,9 @@ impl From<commands::RunResponse> for RunResponse {
 pub struct ListRunsResponse {
     /// The runs.
     pub runs: Vec<Run>,
-    /// Total count before pagination.
+    /// Total count matching filters before applying cursor pagination.
+    ///
+    /// This is retained for compatibility with existing clients.
     pub total: i64,
     /// Next token for pagination. Pass this value as `next_token` in the next
     /// request to retrieve the next page of results.
@@ -277,25 +293,47 @@ pub async fn list_runs(
         _ => Error::BadRequest("invalid query parameters".to_string()),
     })?;
 
-    let offset = match query.next_token.as_deref() {
-        Some(t) => t
-            .parse::<i64>()
-            .map_err(|_| Error::BadRequest(format!("invalid `next_token`: `{}`", t)))?,
-        None => 0,
+    let cursor = match query.next_token.as_deref() {
+        Some(token) => {
+            let parsed: RunsCursorToken = decode_token(token).map_err(Error::BadRequest)?;
+            Some(RunCursor {
+                created_at: parsed.created_at,
+                uuid: parsed.uuid,
+            })
+        }
+        None => None,
     };
-    let limit = query.limit.unwrap_or(100);
 
-    let response = send_command(&state.run_manager_tx, |rx| RunManagerCmd::List {
+    let limit = query.limit.unwrap_or(100);
+    let db_limit = if limit > 0 {
+        Some(limit.saturating_add(1))
+    } else {
+        Some(limit)
+    };
+
+    let mut response = send_command(&state.run_manager_tx, |rx| RunManagerCmd::List {
         status: query.status,
-        limit: query.limit,
-        offset: Some(offset),
+        limit: db_limit,
+        cursor,
         rx,
     })
     .await?;
 
-    let next_offset = offset + limit;
-    let next_token = if next_offset < response.total {
-        Some(next_offset.to_string())
+    let has_more = if limit > 0 && response.runs.len() > limit as usize {
+        response.runs.truncate(limit as usize);
+        true
+    } else {
+        false
+    };
+
+    let next_token = if has_more {
+        response.runs.last().and_then(|run| {
+            encode_token(&RunsCursorToken {
+                created_at: run.created_at,
+                uuid: run.uuid,
+            })
+            .ok()
+        })
     } else {
         None
     };
--- a/src/server/api/v1/sessions.rs
+++ b/src/server/api/v1/sessions.rs
@@ -16,9 +16,22 @@ use uuid::Uuid;
 use super::AppState;
 use super::SprocketCommand;
 use super::error::Error;
+use super::pagination::decode_token;
+use super::pagination::encode_token;
 use super::send_command;
+use crate::system::v1::db::SessionCursor;
 use crate::system::v1::exec::svc::RunManagerCmd;
 use crate::system::v1::exec::svc::run_manager::commands;
+
+/// Cursor payload used for session pagination tokens.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct SessionsCursorToken {
+    /// Creation time of the last session on the previous page.
+    created_at: DateTime<Utc>,
+    /// UUID of the last session on the previous page.
+    uuid: Uuid,
+}
 
 /// Query parameters for listing sessions.
 #[derive(Debug, Clone, Serialize, Deserialize, IntoParams, ToSchema)]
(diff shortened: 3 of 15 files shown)


peterhuene:
Hi @raman976. Thank you for your patience with getting this PR reviewed.

I've pushed up a rebase to your branch so that it is based off of the current in `main` and only contains your single commit.

I'll review this shortly.

src/server/api/v1/runs.rs
```diff
@@ -277,25 +293,47 @@ pub async fn list_runs(
...
 
-    let offset = match query.next_token.as_deref() {
-        Some(t) => t
-            .parse::<i64>()
-            .map_err(|_| Error::BadRequest(format!("invalid `next_token`: `{}`", t)))?,
-        None => 0,
+    let cursor = match query.next_token.as_deref() {
+        Some(token) => {
+            let parsed: RunsCursorToken = decode_token(token).map_err(Error::BadRequest)?;
+            Some(RunCursor {
```
peterhuene, line 299: Could we implement `From<RunsCursorToken> for RunCursor` alongside the definition of `RunsCursorToken` such that this becomes `decode_token(token).map(|t| Some(t.into()).map_err(Error::BadRequest)?`?

src/server/api/v1/runs.rs
```diff
@@ -277,25 +293,47 @@ pub async fn list_runs(
...
+                created_at: parsed.created_at,
+                uuid: parsed.uuid,
+            })
+        }
+        None => None,
     };
+
     let limit = query.limit.unwrap_or(100);
+    let db_limit = if limit > 0 {
```
peterhuene, line 308: Could we add a comment here explaining why the limit is being adjusted (i.e. we need to fetch one more than is required to indicate if more rows are available for pagination).

src/server/api/v1/pagination.rs
```diff
@@ -0,0 +1,21 @@
...
+use base64::Engine;
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use serde::Serialize;
+use serde::de::DeserializeOwned;
+
+/// Encodes a pagination cursor into an opaque token.
+pub fn encode_token<T: Serialize>(cursor: &T) -> Result<String, serde_json::Error> {
+    let json = serde_json::to_vec(cursor)?;
+    Ok(URL_SAFE_NO_PAD.encode(json))
+}
```
peterhuene, line 12: I feel like this function should probably return `Result<String, Error>`; if we fail to serialize the cursor, we should log an error and return `Error::Internal`.

src/server/api/v1/sessions.rs
```diff
@@ -106,24 +122,46 @@ pub async fn list_sessions(
...
+        false
+    };
+
+    let next_token = if has_more {
+        response.sessions.last().and_then(|session| {
+            encode_token(&SessionsCursorToken {
+                created_at: session.created_at,
+                uuid: session.uuid,
+            })
+            .ok()
```
peterhuene, line 163: From the previous comment, we should just `?` it rather than swallowing any serialization errors and returning a `None` next token.

src/server/api/v1/pagination.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Decodes a pagination cursor from an opaque token.
+pub fn decode_token<T: DeserializeOwned>(token: &str) -> Result<T, String> {
```
peterhuene, line 15: For consistency, we could also probably have this return `Result<T, Error>` and internally map it to:

```rust
Error::BadRequest("invalid `next_token`".into())
```

src/server/api/v1/tasks.rs
```diff
@@ -180,26 +210,48 @@ pub async fn list_tasks(
...
 
-    let offset = match query.next_token.as_deref() {
-        Some(t) => t
-            .parse::<i64>()
-            .map_err(|_| Error::BadRequest(format!("invalid `next_token`: `{}`", t)))?,
-        None => 0,
+    let cursor = match query.next_token.as_deref() {
+        Some(token) => {
+            let parsed: TasksCursorToken = decode_token(token).map_err(Error::BadRequest)?;
+            Some(TaskCursor {
```
peterhuene, line 216: Same comment as before: probably should have a `From<TasksCursorToken> for TaskCursor` implementation near the definition of `TasksCursorToken`.

src/server/api/v1/tasks.rs
```diff
@@ -180,26 +210,48 @@ pub async fn list_tasks(
...
+                created_at: parsed.created_at,
+                name: parsed.name,
+            })
+        }
+        None => None,
     };
+
     let limit = query.limit.unwrap_or(100);
+    let db_limit = if limit > 0 {
```
peterhuene, line 225: Same comment as before: explanation for the adjustment.

src/server/api/v1/tasks.rs
```diff
@@ -180,26 +210,48 @@ pub async fn list_tasks(
...
+        false
+    };
+
+    let next_token = if has_more {
+        response.tasks.last().and_then(|task| {
+            encode_token(&TasksCursorToken {
+                created_at: task.created_at,
+                name: task.name.clone(),
+            })
+            .ok()
```
peterhuene, line 253: Same comment as before: we should not be silently discarding any serialization errors.

src/server/api/v1/tasks.rs
```diff
@@ -263,26 +315,48 @@ pub async fn get_task_logs(
...
 
-    let offset = match query.next_token.as_deref() {
-        Some(t) => t
-            .parse::<i64>()
-            .map_err(|_| Error::BadRequest(format!("invalid `next_token`: `{}`", t)))?,
-        None => 0,
+    let cursor = match query.next_token.as_deref() {
+        Some(token) => {
+            let parsed: TaskLogsCursorToken = decode_token(token).map_err(Error::BadRequest)?;
+            Some(TaskLogCursor {
```
peterhuene, line 321: Same comment as before: probably should have a `From<TaskLogsCursorToken> for TaskLogCursor` implementation near the definition of `TaskLogsCursorToken`.

src/server/api/v1/tasks.rs
```diff
@@ -263,26 +315,48 @@ pub async fn get_task_logs(
...
+                created_at: parsed.created_at,
+                id: parsed.id,
+            })
+        }
+        None => None,
     };
+
     let limit = query.limit.unwrap_or(100);
+    let db_limit = if limit > 0 {
```
peterhuene, line 330: Same comment as before: an explanation for the adjustment.

src/server/api/v1/tasks.rs
```diff
@@ -263,26 +315,48 @@ pub async fn get_task_logs(
...
+        false
+    };
+
+    let next_token = if has_more {
+        response.logs.last().and_then(|log| {
+            encode_token(&TaskLogsCursorToken {
+                created_at: log.created_at,
+                id: log.id,
+            })
+            .ok()
```
peterhuene, line 358: Same comment as before: don't silently discard serialization errors.

src/system/v1/db/sqlite.rs
```diff
@@ -173,18 +174,37 @@ impl Database for SqliteDatabase {
...
-        )
-        .bind(limit)
-        .bind(offset)
-        .fetch_all(&self.pool)
-        .await?;
+        let sessions: Vec<Session> = if let Some(cursor) = cursor {
+            sqlx::query_as(
+                "select uuid, subcommand, created_by, created_at from sessions
+                  where datetime(created_at) < datetime(?)
+                  or (datetime(created_at) = datetime(?) and uuid < ?)
```
peterhuene, line 188: I think we have a problem here. The `uuid` column, while it is text, is not actually ordered. We can't use the `<` operator like we're doing here (it'll appear to work because it's using an ordinal comparison of the strings, but they're UUIDs and are in no way sequential).

Instead, the cursor should probably be using `id` and not `uuid`.

If that's the case, I'm not sure if using `created_at` is helpful for this particular cursor (i.e. wouldn't it just be `id < ?`?).

We then wouldn't have to use JSON for round-tripping this cursor, just the integer.

src/system/v1/db/sqlite.rs
```diff
@@ -335,34 +355,77 @@ impl Database for SqliteDatabase {
...
+        let runs: Vec<Run> = match (status, cursor) {
+            (Some(status), Some(cursor)) => {
+                sqlx::query_as(
+                    "select r.uuid, s.uuid as session_uuid, r.name, r.source, r.target, r.status,
+                     r.inputs, r.outputs, r.error, r.directory, r.index_directory, r.started_at,
+                     r.completed_at, r.created_at from runs r join sessions s on r.session_id = \
+                     s.id
+                     where r.status = ?
+                     and (datetime(r.created_at) < datetime(?)
+                     or (datetime(r.created_at) = datetime(?) and r.uuid < ?))
```
peterhuene, line 371: Same comment as above re: using `<` for `uuid`. The cursor should also be using `id` and likely doesn't need `created_at` at that point.

src/system/v1/db/sqlite.rs
```diff
@@ -335,34 +355,77 @@ impl Database for SqliteDatabase {
...
+                .await?
+            }
+            (None, Some(cursor)) => {
+                sqlx::query_as(
+                    "select r.uuid, s.uuid as session_uuid, r.name, r.source, r.target, r.status,
+                     r.inputs, r.outputs, r.error, r.directory, r.index_directory, r.started_at,
+                     r.completed_at, r.created_at from runs r join sessions s on r.session_id = \
+                     s.id
+                     where datetime(r.created_at) < datetime(?)
+                     or (datetime(r.created_at) = datetime(?) and r.uuid < ?)
```
peterhuene, line 405: Same comment as above re: using `uuid` and `<`.

src/system/v1/db/sqlite.rs
```diff
@@ -578,7 +640,13 @@ impl Database for SqliteDatabase {
         if status.is_some() {
             query.push_str(" and t.status = ?");
         }
-        query.push_str(" order by t.created_at desc limit ? offset ?");
+        if cursor.is_some() {
+            query.push_str(
+                " and (datetime(t.created_at) < datetime(?) or (datetime(t.created_at) = \
+                 datetime(?) and t.name < ?))",
```
peterhuene, line 646: The task names are also not ordered and thus we can't use `<` on them.

This one is trickier to fix because the `tasks` table unfortunately is using the name as the primary key.

To fix this, I think we'll need to change the `tasks` table definition such that we add an integer `id` column like we have on the rest of the tables and instead use a unique constraint on the name (and an index too).

The cursor could then be just the `id` column.

src/system/v1/db/sqlite.rs
```diff
@@ -656,15 +729,27 @@ impl Database for SqliteDatabase {
         if source.is_some() {
             query.push_str(" and source = ?");
         }
-        query.push_str(" order by created_at asc limit ? offset ?");
+        if cursor.is_some() {
+            query.push_str(
+                " and (datetime(created_at) > datetime(?) or (datetime(created_at) = datetime(?) \
+                 and id > ?))",
```
peterhuene, line 735: Here we're already using the id in the cursor, so do we need the timestamp? That is to say, would `id > ?` be sufficient on its own?

peterhuene (changes requested):
Thanks again for your patience with me on this PR.

I've reviewed the PR and think may need some additional changes before we can merge it.

Please review my feedback and let me know if you have any questions.

claymcleod:

The underlying work remains tracked in #581, so it won't be lost. You're welcome to reopen this PR if you'd like to pick it back up.

Issue #184, opened 2025-09-20
enhanced logging


a-frantz:
writing down some thoughts about logging improvements I'd like to see; a wishlist of sorts that could be tackled piecemeal

parent issue for #183 and #196

1. persist logs in addition to STDERR stream
    - DEBUG logs should be written to a file in the execution directory, regardless of CL verbosity
2. logs should be structured in a way they can be searched easily, via tools like `grep`
    - if I'm interrogating logs from a complex workflow, I should be able to easily narrow down the scope with something like `grep <task name>`
3. performance metrics are valuable, but their availability is backend dependent, so this gets tricky
    - WDL authors are primarily concerned with memory and disk space requirements, so Sprocket providing insight into these would be awesome
    - LSF does not have any (good) way to track disk utilization, so this feature may not be within grasp
    - this information should probably be machine readable. JSON maybe? This would allow an external script (or even Sprocket itself) to aggregate metrics across many executions and calculate some stats for tailoring requests
    - time information is also valuable (see #183) but has a fair amount of nuance. Would be great if we could break this information down into categories like "networking", "executing", etc

j-andrews7:
FWIW, I am coming largely from Nextflow/nf-core, which for all of their complexity, do have really excellent performance reports that may be worth emulating in certain ways. I've attached an example of the full HTML report as well as the trace log that just snags all the metrics in text form.

[execution_trace_2024-09-05_08-48-52.txt](https://github.com/user-attachments/files/22447582/execution_trace_2024-09-05_08-48-52.txt)
[execution_report_2024-09-05_08-48-52.html](https://github.com/user-attachments/files/22447581/execution_report_2024-09-05_08-48-52.html)

j-andrews7:
I would also add capturing tool versions to this wishlist, which is preferably done upstream in tool modules/task definitions and may require some additional thought. Not really sure this is a `sprocket` task so much as something that should be standard practice and perhaps urged in the WDL docs.

But as a user, that's info I need to publish and gleaning it from container names in task definitions is gonna put me in a raw mood.

a-frantz:
> I've attached an example of the full HTML report as well as the trace log that just snags all the metrics in text form.

Wow, I'm really impressed by this report! Seems like a great goal for Sprocket. But as mentioned in the issue, we will have to work with what information is provided by each backend (which of course is not standardized). What infrastructure was this report generated from? It's very detailed!

[...]

PR #743, opened 2026-03-20
feat(sprocket): show task stderr output

### LINKED ISSUE #196: feat: streaming STDERR from tasks to the main log
I think our logging in general could be beefed up, but a specific logging feature I'd like to see us add in the near-ish future is STDERR streaming. Many of our tasks run for several hours, to several days for extreme cases, and being able to peek at the STDERR logs is often helpful to ensure that the process is still doing something instead of hanging.

The simplest UX for this is to stream the STDERR from running tasks back to the main process' logging. 

I believe the current alternative involves opening a second shell and tailing the logs of the running container, which is not trivial to track down the right file.

**Additional context**
miniwdl has this (enabled by passing `--verbose` to `run`) and I've found it helpful.

I have some other ideas for logging features I'd find helpful, but maybe I should write those up in a separate issue (also this is the most concrete and fleshed out request I have)

### LINKED ISSUE #840: dev test should print test output with a flag
When running `sprocket dev test` in CI (e.g. GitHub Actions), the final report shows only failed tests. However, the test outputs (e.g. stdout and stderr) are inaccessible for debugging.

I would like a flag (or some other method, e.g. verbosity) to include the tail or possibly the full files in the logging output so that users can debug.

N/A

**Additional context**
N/A

Serial-ATA:
Adds `sprocket run --stderr` to print the stderr of the task while it executes.

Like `miniwdl run --verbose`, lines are limited to 4KiB.

closes #196

For external contributors:

For all contributors:

      is implemented.


src/commands/run.rs
```diff
@@ -409,6 +430,50 @@ async fn progress(
                             state.canceled += 1;
                             Some(id)
                         }
+                        CrankshaftEvent::TaskStderr { id, message } if show_stderr => {
```
claymcleod, line 445: Bollard's `NewlineLogOutputDecoder` delivers one Docker multiplexed frame per `TaskStderr` event. Docker frames contain whatever the daemon has buffered at flush time—and since stderr is line-buffered by default on most systems, each frame almost always contains one or more complete lines. Mid-line splits would only occur if a single line exceeds Docker's ~8KiB frame limit or the process uses unusual write patterns, both of which are unlikely for the target use case (long-running tasks logging progress).

Given that, the line-extraction buffer, 4KiB limit, and `shrink_to` bookkeeping here are probably more sophistication than we need. Consider just sending each frame directly through `tracing` as it arrives, without buffering.

Additionally, consider dropping the custom `target:` and the dedicated `--stderr` flag, and just using `tracing::debug!(...)`. Users who want to see task stderr can pass `-vv` for debug verbosity. This would eliminate the runtime `EnvFilter` modification, the logging architecture restructure in `lib.rs`, and the `test.rs` changes—significantly reducing the scope of this PR.

claymcleod (dismissed):
The buffering and logging architecture changes can likely be simplified substantially—see inline comment for details.

src/commands/run.rs
Serial-ATA: > Bollard's NewlineLogOutputDecoder delivers one Docker multiplexed frame per TaskStderr event

I was under the assumption that `TaskStderr` could _potentially_ come from another backend, not assuming Docker. Regardless though, you mentioned cases where the buffering still makes sense

> consider dropping the custom target: and the dedicated --stderr flag, and just using tracing::debug!(...). Users who want to see task stderr can pass -vv for debug verbosity.

I added `--stderr` and a separate target since it's probably useful to isolate compared to all the other debug output from `wdl-engine`. Any thoughts on that @a-frantz?

a-frantz: relevant (old) issue with some thoughts I had on the topic - #184 
Specifically, bullet point 2 is relevant here. I'd love to see our logging get more structured (by taking advantage of `tracing` features) so that users (me) can `grep` logs to filter things down to only lines relevant to specific tasks. Which is to say I agree with Alex that isolating logs is important for this feature and should be accounted for in other places too.

As for the `--stderr` flag, I don't feel strongly either way. Either a dedicated flag or just always emitting stderr streams at `-vv` level seems fine to me (particularly now that we've reduced the `debug` noise from `wdl-analysis`)

a-frantz: > I was under the assumption that could come from another backend, not assuming Docker.

agree with this BTW. This should work on backends other than just docker (I would use it on the HPC)

claymcleod: I think, at a minimum, the flag should be something like `--show-task-stderr`. I'm fine with that though, and I think we still have buffering work to do here.

claymcleod:
I came to review this one, but it's still waiting on buffering. See the end of [this comment](https://github.com/stjude-rust-labs/sprocket/pull/743#discussion_r3076025186).

Additionally, the `CHANGELOG.md` needs to be updated to use the new flag name.

Serial-ATA:
@claymcleod what buffering work do you mean? Your suggestions were Docker-specific, I built buffering on *top* of that since `crankshaft` is generic. https://github.com/stjude-rust-labs/sprocket/pull/743#discussion_r2976899214

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * New `--hide-warnings` flag to `check`/`lint` with a corresponding `check.hide_warnings` option in `sprocket.toml` ([#675](https://github.com/stjude-rust-labs/sprocket/pull/675)).
 * Added support for configuring the `doc` subcommand via `sprocket.toml` ([#773](https://github.com/stjude-rust-labs/sprocket/pull/773))
+* Added `--stderr` option to the `run` subcommand to show task stderr during execution ([#743](https://github.com/stjude-rust-labs/sprocket/pull/743)).
```
Serial-ATA:
```suggestion
* Added `--show-task-stderr` option to the `run` subcommand to show task stderr during execution ([#743](https://github.com/stjude-rust-labs/sprocket/pull/743)).
```

adthrasher: This appears to still be needed.

src/commands/run.rs
```diff
@@ -189,6 +191,10 @@ pub struct Args {
     #[clap(long)]
     pub no_call_cache: bool,
 
+    /// Show task stderr during execution.
```
peterhuene, line 194:
```suggestion
    /// Show task stderr during execution.
    ///
    /// Not all task execution backends support streaming task output.
```

src/commands/run.rs
```diff
@@ -281,6 +287,20 @@ struct Task {
...
+}
+
+impl Task {
+    const STDERR_BUFFER_SIZE: usize = 4096;
+
+    fn new(name: Arc<String>, token: CancellationToken) -> Self {
+        Self {
+            name,
+            token,
+            stderr_buffer: Vec::with_capacity(Self::STDERR_BUFFER_SIZE),
```
peterhuene: Could we delay this allocation until we actually write to it for the first time? i.e. `Vec::reserve` if the capacity is 0.

CHANGELOG.md
```diff
@@ -83,6 +83,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * New `--hide-warnings` flag to `check`/`lint` with a corresponding `check.hide_warnings` option in `sprocket.toml` ([#675](https://github.com/stjude-rust-labs/sprocket/pull/675)).
 * Added support for configuring the `doc` subcommand via `sprocket.toml` ([#773](https://github.com/stjude-rust-labs/sprocket/pull/773))
+* Added `--show-task-stderr` option to the `run` subcommand to show task stderr during execution ([#743](https://github.com/stjude-rust-labs/sprocket/pull/743)).
```
peterhuene: `CHANGELOG` update is in the wrong place.

peterhuene (commented):
Looks good, just minor comments!

src/commands/run.rs
peterhuene: Resolving this comment as the option has been renamed and I don't think we can rely on the backend to line buffer the output.

peterhuene (approved):

PR #744, opened 2026-03-20
fix: preserves evaluation results for in-flight tasks during lazy cancellation

claymcleod:
Previously, when cancellation was configured in Slow mode and the user pressed ctrl+c once, running tasks were allowed to finish but their results were stored as `Canceled` in the database—evaluation outputs were lost even though the tasks completed successfully. Additionally, the `Ok(None)` handler in `execute_target` was a no-op, which left lazy-canceled runs stuck in `Canceling` status indefinitely when some tasks couldn't start.

This changes the cancellation semantics so that only a force cancellation (second ctrl+c in Slow mode, or any cancel in Fast mode) overrides task results with `Canceled`. A lazy cancellation lets in-flight tasks complete and stores their actual results while preventing new tasks from being dispatched. Notably, workflows that are on their last task (or single-task workflows/task targets) where all tasks complete successfully are now correctly stored as `Completed` with their outputs, since there is nothing left to cancel.

The following table summarizes the expected outcomes for each scenario:

| Target | Scenario | Cancel mode | Target run status | Outputs |
|-|-|-|-|-|
| Workflow (1 task) | Task in-flight, completes | Slow (1x) | Completed | Yes |
| Workflow (1 task) | Task in-flight, killed | Slow but forced (2x) | Canceled | No |
| Workflow (1 task) | Task in-flight, killed | Fast (1x) | Canceled | No |
| Workflow (3 tasks) | 2 complete, 1 never starts | Slow (1x) | Canceled | No |
| Task | Task in-flight, completes | Slow (1x) | Completed | Yes |
| Task | Task in-flight, killed | Slow but forced (2x) | Canceled | No |
| Task | Task in-flight, killed | Fast (1x) | Canceled | No |

- **Evaluator state checks.** Task and workflow evaluators now check for the `Canceling` state specifically rather than calling `user_canceled()`, which returned `true` for both `Waiting` and `Canceling` states. Since we no longer need to override results in the `Waiting` state, that method is removed.
- **Run finalization.** The `Ok(None)` handler in `execute_target` now calls `cancel_run` to properly finalize lazy-canceled runs that couldn't complete all tasks. Previously this was a no-op, which left runs stuck in `Canceling` status indefinitely.
- **Test coverage.** Seven cancellation test scenarios covering both workflow and task targets, matching the table above. All cancel tests now wait for the specific task to reach `Running` status in the database before issuing the cancel, avoiding a race where the cancel arrives before the task evaluator dispatches to Docker.

For all contributors:


crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -598,7 +598,7 @@ impl Evaluator {
             .perform_task_evaluation(document, task, inputs, eval_root_dir.as_ref(), task.name())
             .await;
 
-        if self.cancellation.user_canceled() {
+        if self.cancellation.state() == CancellationContextState::Canceling {
```
peterhuene: A cancellation may also be caused by an error being encountered in a task, which triggers either a slow or fast cancellation. We store a flag to denote if the user initiated the cancellation or if it was caused by an error.

We want to return the result when the cancellation is the result of an error and _not_ user initiated.

Take for example this workflow:

```wdl
version 1.3

task t {
    input {
        Int i
    }

    command <<<
        exit ~{if i == 5 then 1 else 0}
    >>>
}


workflow w {
    scatter (i in range(10)) {
        call t { i }
    }
}
```

Running this, by default, will produce the desired result:

```
$ sprocket run repro.wdl -t w
2026-03-23T14:57:30.547546Z ERROR wdl_engine::eval: an evaluation error occurred: waiting for any executing tasks to complete: task execution failed for task `t` (id `t-5`): process terminated with exit code 1: see `/Users/phuene/src/peterhuene/sprocket/out/runs/w/2026-03-23_145727821845000/calls/t-5/attempts/0/stdout` and `/Users/phuene/src/peterhuene/sprocket/out/runs/w/2026-03-23_145727821845000/calls/t-5/attempts/0/stderr` for task output
error: task execution failed for task `t` (id `t-5`): process terminated with exit code 1: see `/Users/phuene/src/peterhuene/sprocket/out/runs/w/2026-03-23_145727821845000/calls/t-5/attempts/0/stdout` and `/Users/phuene/src/peterhuene/sprocket/out/runs/w/2026-03-23_145727821845000/calls/t-5/attempts/0/stderr` for task output
   ┌─ test.wdl:3:6
   │
 3 │ task t {
   │      ^ this task failed to execute
   │
   ┌─ test.wdl:16:9
   │
16 │         call t { i }
   │         ---- called from this location

error: aborting due to evaluation error
```

this is because the default failure mode is `slow` and so it wasn't hitting this conditional and returning `EvaluationError::Canceled`.

However, if we execute it under `fast` fail mode:

```toml
[run]
fail = "fast"
```

We see the problem:

```
$ sprocket run repro.wdl -t w
2026-03-23T14:59:19.595058Z ERROR wdl_engine::eval: an evaluation error occurred: waiting for any executing tasks to cancel: task execution failed for task `t` (id `t-5`): process terminated with exit code 1: see `/Users/phuene/src/peterhuene/sprocket/out/runs/w/2026-03-23_145917333784000/calls/t-5/attempts/0/stdout` and `/Users/phuene/src/peterhuene/sprocket/out/runs/w/2026-03-23_145917333784000/calls/t-5/attempts/0/stderr` for task output
error: evaluation was interrupted
```

We're masking the encountered error with a cancellation error.

peterhuene: This comment also applies to the similar change in `workflow.rs`.

crates/wdl-engine/src/eval.rs
```diff
@@ -222,12 +222,6 @@ impl CancellationContext {
         self.second.clone()
     }
 
-    /// Determines if the user initiated the cancellation.
-    pub(crate) fn user_canceled(&self) -> bool {
```
peterhuene, line 226: Re: the previous comments, we should still want some way of determining when a cancellation is due to an error or when it was user initiated so that the former isn't treated like the latter.

peterhuene (commented):
A few feedback comments to go over.

crates/wdl-engine/src/eval/v1/task.rs
peterhuene: So I think we'll need something here rather than reverting the change that was made previously.

Perhaps something like `if self.cancellation.user_canceled() && self.cancellation.state() == CancellationContextState::Canceling` or some such.

peterhuene (approved):

PR #745, opened 2026-03-21
fix: improves target resolution and input key prefixing

claymcleod:
Two issues with how `sprocket run` resolves targets and handles input keys:

**`--target` didn't prefix file input keys.** When rerunning a failed task with `--target my_task` and an input file containing unprefixed keys (e.g., `{"str": "hello"}`), sprocket errored with "expected the key to be prefixed with the workflow or task name." CLI key-value pairs were prefixed but file inputs were not. Now both are handled uniformly—if `--target` is set and a key doesn't already start with the target prefix, it gets prepended automatically. Keys that already carry the prefix are left alone.

**No-inputs case required `--target` unconditionally.** If you ran `sprocket run foo.wdl` on a document with a single workflow (or single task), it would fail asking for `--target` even though the target is unambiguous. Now the no-inputs path falls through to the same inference chain: single workflow > single task > error asking for `--target`.

For all contributors:

### DIFF
--- a/src/inputs.rs
+++ b/src/inputs.rs
@@ -146,17 +146,20 @@ impl Invocation {
     ///
     /// * If the key already starts with `"{target}."`, it is returned as-is.
     /// * Otherwise the key is prefixed with `"{target}."`.
-    fn prefix_key(&self, key: String) -> Result<String> {
+    fn prefix_key(&self, key: String) -> String {
         match &self.target {
             Some(target) => {
-                let dot_prefix = format!("{target}.");
-                if key.starts_with(&dot_prefix) {
-                    Ok(key)
+                if key
+                    .strip_prefix(target)
+                    .and_then(|r| r.strip_prefix('.'))
+                    .is_some()
+                {
+                    key
                 } else {
-                    Ok(format!("{dot_prefix}{key}"))
+                    format!("{target}.{key}")
                 }
             }
-            None => Ok(key),
+            None => key,
         }
     }
 
@@ -166,14 +169,14 @@ impl Invocation {
             Input::File(url) => {
                 let inputs = file::read_input_file(&url).await?;
                 for (key, value) in inputs {
-                    self.inputs.insert(self.prefix_key(key)?, value);
+                    self.inputs.insert(self.prefix_key(key), value);
                 }
             }
             Input::Pair { key, value } => {
                 let cwd = std::env::current_dir()
                     .context("failed to determine the current working directory")?;
 
-                let key = self.prefix_key(key)?;
+                let key = self.prefix_key(key);
                 self.inputs.insert(
                     key,
                     LocatedJsonValue {


src/inputs.rs
```diff
@@ -536,4 +600,122 @@ mod tests {
...
+        assert!(invocation.inputs.contains_key("foo.new.key"));
+        assert!(invocation.inputs.contains_key("foo.baz"));
+        assert!(invocation.inputs.contains_key("foo.quux"));
+    }
+
+    #[tokio::test]
+    async fn dotted_key_conflicts_with_different_target() {
+        // `inputs_two.json` contains `new.key`. When `new` IS a known target
+        // and `--target` is `foo`, this should error because the key is
+        // prefixed with a conflicting target name.
```
adthrasher: Tagging @a-frantz since I think this is the case she was concerned with via Slack.

If `foo` is a workflow and `new` is a task that is called by `foo`, then this shouldn't be an error. The full key should end up being `foo.new.key`.

a-frantz: > Tagging @a-frantz since I think this is the case she was concerned with via Slack.

yes!

> If `foo` is a workflow and `new` is a task that is called by `foo`, then this shouldn't be an error. The full key should end up being `foo.new.key`.

agree

claymcleod: Okay! I'm fine with that. The tradeoff is that we get a slightly less good error message (instead of being able to recognize that this is another target in the file and pointing this specific case out, we just get a generic error message saying that `foo.bar.name` is not an input). But I suppose that is a fine tradeoff for more straightforward behavior in these cases.

src/inputs.rs
```diff
@@ -140,22 +140,40 @@ pub struct Invocation {
...
+    /// * Otherwise the key is prefixed with `"{target}."`.
+    fn prefix_key(&self, key: String) -> Result<String> {
+        match &self.target {
+            Some(target) => {
+                let dot_prefix = format!("{target}.");
+                if key.starts_with(&dot_prefix) {
+                    Ok(key)
+                } else {
+                    Ok(format!("{dot_prefix}{key}"))
+                }
```
peterhuene: Nit:

```suggestion
                if key.strip_prefix(target).and_then(|r| r.strip_prefix('.')).is_some() {
                    Ok(key)
                } else {
                    Ok(format!("{target}.{key}"))
                }
```

src/inputs.rs
```diff
@@ -140,22 +140,40 @@ pub struct Invocation {
 }
 
 impl Invocation {
+    /// Prefixes a key with the target name if needed.
+    ///
+    /// When `self.target` is set:
+    ///
+    /// * If the key already starts with `"{target}."`, it is returned as-is.
+    /// * Otherwise the key is prefixed with `"{target}."`.
+    fn prefix_key(&self, key: String) -> Result<String> {
```
peterhuene: It appears this doesn't need to return `Result`.

PR #748, opened 2026-03-21
fix: dev test silently accepts bad paths

### LINKED ISSUE #742: `dev test` silently accepts bad paths
**Describe the bug**

Sprocket test translates non-existent paths to `File?` inputs as `None`.

**Expected behavior**

`test` should fail these tests when the supplied path is wrong.

**Terminal Output**
Please copy and paste the command used, logs generated, and the error produced in the markdown block below. This can also be attached to to the issue as a file.

```terminal
~/tmp [12:02:42]
$  cat test.wdl
version 1.1

task foo {
    input {
        File? optional
    }

    command <<<
        echo "~{optional} might exist"
    >>>
}

~/tmp [12:04:34]
$  cat test.yml
foo:
  - name: bar
    input:
      optional:
        - this/path/does/not/exist.txt
~/tmp [12:04:40]
$  sprocket dev test
test::foo::bar (iteration #1): ✅
Sprocket test result summary:
✅ `test::foo::bar` success! (1 successful test execution)
```

Hemanntnagar:
Fix #742 

## Cause
Value::resolve_paths treated optional File / Directory types (File?, Directory?) by turning any missing path into None, as long as optional was true (from the declared type). That’s wrong for explicit inputs (JSON/YAML/CLI): a bogus path is not the same as “input omitted,” but it was silently turned into None.

Expression/evaluation paths still need the old behavior (e.g. tasks/test-join-paths where an optional file from join_paths doesn’t exist on disk yet and should become None on outputs).

## Fix

1. Value::resolve_paths now takes a leading flag strict_explicit_files: bool (after &self):
- true: optional File? / Directory? with a concrete path must point at an existing file/dir; otherwise error (no silent None).
- false: keeps the previous behavior: optional + missing path → None (for evaluated WDL / backend path resolution).

2. TaskInputs::join_paths and WorkflowInputs::join_paths pass strict_explicit_files: true, so all merged user inputs (including sprocket dev test via src/eval.rs) get strict checks.

3. Task/workflow evaluation (add_backend_inputs, workflow decls, task outputs) pass false so expression semantics stay the same.

4. CompoundValue::resolve_paths forwards the same strict_explicit_files flag into nested values (structs, arrays, etc.).

5. Unit tests in crates/wdl-engine/src/value.rs cover:
- strict + missing path → error
- non-strict + missing path → None
- strict + truly omitted (Value::None) → still None
- strict + existing file → OK

Files touched

- crates/wdl-engine/src/value.rs — resolve_paths + CompoundValue::resolve_paths + tests
- crates/wdl-engine/src/inputs.rs — join_paths (task + workflow) uses true
- crates/wdl-engine/src/eval/v1/task.rs — false for backend + output resolution
- crates/wdl-engine/src/eval/v1/workflow.rs — false for evaluated decls/outputs

For external contributors:
 
 * [x]  You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
 * [x]  You have not used AI on any parts of this pull request.
 * [x]  You have added a few sentences describing the PR here.
 * [x]  Your code builds clean without any errors or warnings.
 
 For all contributors:
 
 * [x]  You have added tests (when appropriate).
 * []  You have added an entry in the CHANGELOG (when appropriate).
 * []  You have updated the README or other documentation to account for these changes (when appropriate).
 * [x]  You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).
 
 * [x]  You have updated any and all effected entries within `RULES.md`.
 * [x]  You have added a test case in `crates/wdl-lint/tests/lints` that covers every
   is implemented.


crates/wdl-engine/src/value.rs
```diff
@@ -732,6 +742,7 @@ impl Value {
     /// `wdl_engine`. Expect this interface to change soon!
     pub(crate) async fn resolve_paths<F>(
         &self,
+        strict_explicit_files: bool,
```
a-frantz: nit
```suggestion
        must_exist: bool,
```

a-frantz (changes requested):
Looks like a good start! Can you address the issues found by the CI checks?

peterhuene (changes requested):
Hi @Hemanntnagar.

Unfortunately I don't think this is the right approach to fix the issue.

The existing behavior of the task/workflow inputs collections and evaluation of tasks/workflows is intentionally tied to the WDL specification and shouldn't be altered to address the issue.

Instead, I would expect a fix _scoped only to_ the test command. Specifically, prior to evaluating the task or workflow, the test command should find the target task or workflow in the document, iterate through the target's inputs looking for any of type `File` or `Directory` (optional or not), finding the corresponding input in the inputs collection and then check to see if the input value (when joined with the fixtures directory as the base) exists as a file or directory depending on the type and error if it does not.

I would greatly prefer that solution over altering `wdl-engine` to fix this.

claymcleod:

The underlying work remains tracked in #742, so it won't be lost. You're welcome to reopen this PR if you'd like to pick it back up.

PR #751, opened 2026-03-23
feat(wdl-grammar)!: error on keyword identifiers

### LINKED ISSUE #275: Identifiers that are reserved keywords should be disallowed.
The specification says ([link](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#reserved-keywords)):

> The following (case-sensitive) language keywords are reserved and cannot be used to name declarations, calls, tasks, workflows, import namespaces, struct types, or aliases.

However, `wdl` is completely fine with reserved keywords being used as identifiers.

```wdl
version 1.2

task run {
    input {
        String in
    }
}
```

This should be a validation error instead.

### LINKED ISSUE #813: bug: `env` should not be a reserved keyword in WDL 1.0
`env` is not a reserved keyword in the WDL 1.0 specification, but sprocket currently rejects it as an identifier in WDL 1.0 documents. This prevents valid WDL 1.0 files from linting cleanly.

For example, in [this WDL 1.0 file](https://github.com/Ultimagen/healthomics-workflows/blob/main/workflows/trim_align_sort/tasks/trimming_tasks.wdl#L257), the declaration `String env='cutadaptenv'` produces:

```
error: expected declaration name, but found `env` keyword
    ┌─ workflows/trim_align_sort/tasks/trimming_tasks.wdl:257:12
    │
257 │     String env='cutadaptenv'
    │            ^^^ unexpected `env` keyword
```

The `env` keyword was introduced in WDL 1.2 as part of the `env { }` block for environment variable declarations. It should be valid as an identifier in WDL versions prior to 1.2.

Serial-ATA:
closes #275
closes #813

For external contributors:

For all contributors:

      is implemented.


Serial-ATA:
@claymcleod this isn't comparing `Token`s though, it's comparing values of `Ident`s

Serial-ATA:
This is ready for review, but I've got it blocked till it's determined how those keywords should be replaced in the workflows repo.

This is really noisy since `wdl-grammar` wasn't version-aware before.

crates/wdl-engine/tests/workflows/test-meta-values/source.wdl
```diff
@@ -3,7 +3,7 @@ version 1.3
 workflow test_meta_values {
   meta {
     authors: ["Jim", "Bob"]
-    version: 1.1
```
peterhuene, line 6: So `meta` keys are one place that I think we need to continue to allow "any" identifier (i.e. accept keywords) as `version` especially is one that's commonly used as far as I'm aware.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -587,39 +687,26 @@ fn enum_definition(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marke
...
-        }
-        found => {
-            let (found, span) = found
-                .map(|(t, s)| (Some(t.describe()), s))
-                .unwrap_or_else(|| (None, parser.span()));
-            return Err((
-                marker,
-                expected_one_of(ENUM_ITEM_EXPECTED_NAMES, found, span),
-            ));
+    ident!(parser, marker, "enum variant declaration");
```
peterhuene: For consistency with `enum_definition` above:

```suggestion
    ident!(parser, marker, "enum variant name");
```

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -1133,7 +1219,7 @@ fn runtime_section(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marke
 
 /// Parses an item in a runtime section.
 fn runtime_item(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marker, Diagnostic)> {
-    expected_in!(parser, marker, ANY_IDENT, "runtime key");
```
peterhuene, line 1321: Related to my first comment, I think we should probably accept "any" identifier (i.e. reserved names) for:

* `runtime` section keys
* `requirements` section keys
* `hints` section keys
* `meta` section keys
* `parameter_meta` section keys; technically we could limit this to identifiers given the requirement that it match an input/output, but probably best for consistency with the `meta` section; it's a linting warning for any key that don't match an input/output anyway.  
* "meta" object keys
* workflow hints object keys (workflow hints don't support expression evaluation, hence why there are separate parser rules)
* literal hints, literal inputs, and literal output keys
* `Object` literal keys

Here's the relevant section from the 1.3 spec:

> The following (case-sensitive) language keywords are reserved and cannot be used to name declarations, calls, tasks, workflows, import namespaces, struct types, or aliases...

Basically the restriction applies to anything that introduces a _name_ into a [namespace](https://github.com/openwdl/wdl/blob/wdl-1.3/SPEC.md#namespaces) and therefore should not accept a reserved name. Keys, however, should accept reserved names.

By extension, the access expression identifier should accept reserved names unconditionally because what is being accessed may have a key with a reserved name.

peterhuene (commented):
This looks good! I just have one particular comment that needs to be addressed relating to "keys" accepting "any" identifier.

crates/wdl-grammar/src/grammar/v1.rs
Serial-ATA: @peterhuene from the comments on https://github.com/openwdl/wdl/issues/763, keywords in meta sections is really just an oversight. It might not even be something that appears that often? For gauntlet at least it only showed up in the workflows repo

I can drop the parser back to any ident for now if you wanna put those thoughts in that issue and wait on it

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -2419,8 +2518,7 @@ fn object(
 
 /// Parses a single item in a literal object.
 fn object_item(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marker, Diagnostic)> {
-    expected_in!(parser, marker, ANY_IDENT, "object key");
-    parser.update_last_token_kind(SyntaxKind::Ident);
+    ident!(parser, marker, "object key");
```
peterhuene: I think object literal keys should allow for keywords and thus be `ANY_IDENT`.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -2503,7 +2599,7 @@ fn literal_hints(
 
 /// Parses a literal hints item.
 fn literal_hints_item(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marker, Diagnostic)> {
-    expected_in!(parser, marker, ANY_IDENT, "hint key");
+    ident!(parser, marker, "hint key");
```
peterhuene: Literal hints items are part of hint sections and also function like map keys. I think this should be `ANY_IDENT`.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -2603,10 +2695,19 @@ fn index_expr(
...
+    any_ident: bool,
 ) -> Result<CompletedMarker, (Marker, Diagnostic)> {
     parser.require(Token::Dot);
-    expected_in!(parser, marker, ANY_IDENT, "name");
-    parser.update_last_token_kind(SyntaxKind::Ident);
+
+    // In the case of `task` member accesses, we can have multiple
+    // keywords (e.g. `task.meta`). So we just allow anything through
+    // and leave any further errors to the validation pass.
+    if any_ident {
```
peterhuene: Personally I think we should just always accept `ANY_IDENT` here rather than special casing it via the atom expression `task`, as one may also access `task.meta.version`, for example; the result of the expression `task.meta` is an `Object` and we should be able to get the `version` key from it.

With what we have now, `any_ident` will only be true if the preceding expression of the `.` operator is the name reference expression `task`. But in the above, it would be an access expression (`task.meta`) and thus `.version` wouldn't be allowed.

peterhuene (commented):
Looks great!

Just three additional comments to go over.

peterhuene (approved):
Approving in advance of requested changes.

PR #758, opened 2026-03-25
fix: implement `is_ident` utility function correctly.

peterhuene:
This PR fixes the implementation of `is_ident`. The previous implementation incorrectly only checked if the given string could be parsed as a single _token_ and not as a single _identifier token_.

The fix is to properly ensure the token that was read is a `Token::Ident`.

Added a unit test to properly cover that function's implementation.

We had some test coverage for the stdlib functions that use `is_ident` but they used "invalid" identifiers with a hyphen in them which ended up being rejected because they didn't lex as a single token, thus masking the bug.

This also removes the `filter` field from Gauntlet as it was no longer used (analysis now uses ignore files for repositories).

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/gauntlet/Gauntlet.toml
+++ b/crates/gauntlet/Gauntlet.toml
@@ -9193,6 +9193,11 @@ message = "ww-delly.wdl:36:10: warning[UnusedInput]: unused input `reference_fas
 permalink = "https://github.com/getwilds/wilds-wdl-library/blob/cf23adebc923524c531f6182cc0e97c29d8a4f23/modules/ww-delly/ww-delly.wdl/#L36"
 
 [[diagnostics]]
+document = "getwilds/wilds-wdl-library:/modules/ww-diamond/testrun.wdl"
+message = "testrun.wdl:7:20: error: unknown task or workflow `download_diamond_data`"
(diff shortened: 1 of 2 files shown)


crates/gauntlet/Gauntlet.toml
```diff
@@ -9192,6 +9192,11 @@ document = "getwilds/wilds-wdl-library:/modules/ww-delly/ww-delly.wdl"
 message = "ww-delly.wdl:36:10: warning[UnusedInput]: unused input `reference_fasta_index`"
 permalink = "https://github.com/getwilds/wilds-wdl-library/blob/cf23adebc923524c531f6182cc0e97c29d8a4f23/modules/ww-delly/ww-delly.wdl/#L36"
 
+[[diagnostics]]
+document = "getwilds/wilds-wdl-library:/modules/ww-diamond/testrun.wdl"
+message = "testrun.wdl:7:20: error: unknown task or workflow `download_diamond_data`"
+permalink = "https://github.com/getwilds/wilds-wdl-library/blob/cf23adebc923524c531f6182cc0e97c29d8a4f23/modules/ww-diamond/testrun.wdl/#L7"
```
peterhuene, line 9198: This diagnostic is the result of their WDL importing a file from GitHub using `main` and recently (28 minutes ago) it looks like this task was renamed or removed.

PR #759, opened 2026-03-26
feat(wdl-lint): add OptionalInputSafety rule

### LINKED ISSUE #715: New `wdl-lint` rule: `OptionalInputSafety`
## Rule: `OptionalInputSafety`

- **Tags**: `Correctness`
- **Applies to**: `command` block placeholders in all task definitions.
- **RFC**: [sprocket#707](https://github.com/stjude-rust-labs/sprocket/discussions/707)

### Description

Flag any placeholder expression in a `command` block whose type is optional (`T?`) and is not explicitly guarded by `defined()`, `select_first()`, `select_all()`, or an `if`/`else` expression.

Even when an unguarded optional interpolates harmlessly as an empty string, requiring explicit handling makes the intent clear to future readers and prevents a class of subtle bugs where an empty string produces a malformed command.

### Good examples

```wdl
command <<<
    set -euo pipefail
    samtools sort ~{if defined(output_path) then "-o " + output_path else ""} ~{input_bam}
>>>
```

```wdl
command <<<
    set -euo pipefail
    samtools sort -o ~{select_first([output_path, "default.bam"])} ~{input_bam}
>>>
```

### Bad example

```wdl
input {
    String? optional_flag
}

command <<<
    set -euo pipefail
    samtools sort ~{optional_flag} ~{input_bam}
>>>
```

**Diagnostic**: optional value `optional_flag` used in command placeholder without a guard; use `if defined()`, `select_first()`, or an `if`/`else` expression to handle the `None` case explicitly.

### Edge cases or exceptions

- Non-optional types are not flagged.
- Placeholders outside command blocks (e.g., in string literals) are not flagged by this rule.
- The WDL spec defines that `None` interpolates as the empty string in command blocks — this is valid WDL. The rule enforces explicitness, not correctness per se.

### Implementation notes

Visit every `Placeholder` node within a `command` block. Evaluate the type of the placeholder expression. If the type is optional and the expression is not:
- a call to `defined()`, `select_first()`, or `select_all()`
- an `if`/`else` expression (which inherently handles both cases)
- a string concatenation involving the above

then emit a warning. The type information is available from the analysis layer via `evaluate_expr()`.

Veercodeprog:
closes #715 

i implemented this in a pretty small and focused way — it only checks placeholders inside command blocks, gets the expression type from the analysis layer, skips anything that isn’t optional, and only warns when an optional value isn’t clearly guarded. first pass is intentionally simple and just covers the obvious safe cases instead of trying to handle every deeply nested pattern.

I kept the first pass intentionally simple and focused on the most obvious safe patterns rather than trying to reason about every possible nested expression shape.

To check `source.wdl` vs `source.errors`:

```bash
cargo test -p wdl-lint optional-input-safety --locked
```

To refresh `source.errors` after changing diagnostics:

```bash
BLESS=1 cargo test -p wdl-lint optional-input-safety --locked
```

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lint/src/rules/optional_input_safety.rs
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        None
+    }
+
+    fn resolve_type_name(
+        &mut self,
+        name: &str,
+        span: Span,
+    ) -> std::result::Result<Type, Diagnostic> {
```
peterhuene, line 71: Nit:


```suggestion
    ) -> Result<Type, Diagnostic> {
```

crates/wdl-lint/src/rules/optional_input_safety.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// Creates a new context for evaluating types in a command section.
+    fn new(document: Document, scope: ScopeRef<'a>) -> Self {
+        Self { document, scope }
+    }
+}
+
+/// Returns `true` if the expression uses an optional value in a way that should
+/// trigger a warning (e.g. bare optional or optional under `+` even when the
+/// overall expression type is non-optional).
+fn optional_use_is_unsafe<C: EvaluationContext>(
```
peterhuene, line 101: I think perhaps we could implement this another way that might prove more robust and give more specific diagnostics.

For example, it should emit a diagnostic for the expression `if foo then my_optional_string else "bar"` since the type of this expression would be `String?` and given there was no check for `my_optional_string` being defined. Ideally the diagnostic would point directly at `my_optional_string` instead of the entire `if` expression as being the problem.

However, we also shouldn't flag any name reference expression for which we've seen a call to `defined` in an `if` conditional expression prior, e.g. `if defined(my_optional_string) then my_optional_string else "bar"` should not generate a diagnostic.

Given that only an `if` expression my introduce a "scope" where a reference to name that is an optional type is "safe", we might want this lint to implement its own visitor to `Expr` to solve this.

That visitor could maintain state such as:
* `checked: Vec<HashSet<String>>`: the stack of "checked" names (i.e. okay to have a reference to from either the left or right expression of an `if` expression)
* `counter: usize`: counter to maintain when a call to `defined` should add a name to the top of the checked stack

* Upon visiting an `if` expression:
  * Push a new empty `HashSet` to the checked stack
  * Increment the counter
  * Visit the conditional expression
  * Decrement the counter
  * Visit the left expression
  * Visit the right expression
  * Pop the top of the checked stack
* Upon visiting a `call` expression:
  * If the counter is greater than 0, the call is to `defined`, the call has exactly one argument, and the argument is a name reference expression, add the name to the top checked set.
* Upon visiting a name reference expression:
   * Lookup the name in the command's scope to get its type
   * If the type is optional, look through the checked stack to see if any set contains the name; if it does not, emit a diagnostic for the name reference expression's span informing that the use is missing a `defined` guard.
* For all other expressions, recurse on any contained expressions.

What do you think about this approach?

peterhuene (changes requested):
HI @Veercodeprog.

I apologize for taking so long to review your PR; this completely fell off my radar.

Overall, this looks very promising and would be an excellent lint to have.

I added a detailed comment regarding the approach of the lint itself. Let me know your thoughts.

Given it's been so long since you received a review and you don't have the time to further address the feedback, let us know and one of the maintainers will take over.

Thanks!

claymcleod:

The underlying work remains tracked in #715, so it won't be lost. You're welcome to reopen this PR if you'd like to pick it back up.

PR #796, opened 2026-04-08
fix: narrow find-all-references search scope in WDL analysis

Mahathi-s154:
## Summary

This narrows the document search scope used by `textDocument/references` and `textDocument/rename` in `wdl-analysis`.

Previously, `find_all_references` always scanned every transitive dependent of the definition document. That was broader than necessary for symbols that are only visible within the defining document, such as call aliases, import namespace aliases, import alias targets, and task/workflow-local declarations.

This change classifies the resolved definition token first, then limits the candidate document set accordingly:

  - Local-only definitions search only the defining document.
  - Exported definitions such as tasks, workflows, task/workflow inputs and outputs, structs, struct members, enums, and enum variants still search transitive importer documents.

The matching logic after scope selection is unchanged: references are still confirmed by resolving each candidate identifier back through `goto_definition`.

  ## Testing

  - `cargo test -p wdl-analysis find_all_references`
  - `cargo test -p wdl-lsp --test find_all_references`
  - `cargo test -p wdl-lsp --test rename`

  ## Notes

  Added unit coverage for scope classification in `crates/wdl-analysis/src/handlers/find_all_references.rs`.

  Added LSP integration coverage for imported type aliases and call aliases in:

  - `crates/wdl-lsp/tests/find_all_references.rs`
  - `crates/wdl-lsp/tests/rename.rs`

  Added workspace fixtures:

  - `crates/wdl-lsp/tests/workspace/find_references/aliases.wdl`
  - `crates/wdl-lsp/tests/workspace/rename/aliases.wdl`

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -32,6 +36,79 @@ struct TargetDefinition {
...
+    } else {
+        vec![definition_doc]
+    }
+}
+
+/// Returns `true` when a definition can be referenced from importer documents.
+///
+/// Local-only symbols, like call aliases and declarations inside task/workflow
+/// bodies, should remain within the defining document.
+fn needs_transitive_importers(token: &SyntaxToken) -> bool {
```
peterhuene: I think we might be able to improve this implementation such that it more often returns `false`.

If start by looking up the _name scope_ in the token's document with `find_scope_by_position` using the token's start. If there is a name scope, look up the name. If there is a name, check to see if it a "local" name (i.e. it's not a name introduced by an input or output; this could be done simply by adding more information to the `Name` struct) and return false if it is a "local" name.

The above will help return false when requested to find all references to specifically the `x` in the output section:

```wdl
workflow example {
    String x = "hi"

    output {
        String out = x
    }
}
```

Otherwise, do the specialized checks for names that aren't inside of a task/workflow name scope that are considered "local" and return false, such as you've already implemented:

* Explicit import namespace aliases
* Import alias target names

Note: call alias nodes would be handled by the above name scope lookup as they would always be within the name scope of a workflow.

Otherwise, we can assume that transitive dependents will need to be searched. 

What are your thoughts on this approach?

peterhuene (commented):
Hi @Mahathi-s154.

Thank you for your patience for a review on this PR!

The implementation looks good and the added test coverage is great.

I just have one feedback comment on the implementation of `needs_transitive_importers` for discussion.

crates/wdl-analysis/src/handlers/find_all_references.rs
Mahathi-s154: Thanks, @peterhuene.  That approach makes sense.

I agree that the analyzed name scope should be the primary signal here. I'll update the implementation to look up the scope at the definition token position, resolve the name from that scope, and return `false` for names that are local to a task or workflow body. I'll also add metadata to `Name` so that inputs and outputs can continue to be treated as exported.

For the remaining cases, I'll keep the explicit checks for import namespace aliases and import alias target names, and default to searching transitive dependents otherwise.

I'll also add test coverage for the output-section example you pointed out.

Should I go ahead with this?

peterhuene: Hi @Mahathi-s154.

That all sounds good to me, please proceed if you have the time. We really appreciate your contribution!

PR #795, opened 2026-04-07
feat(wdl-format)!: configurable newline style

Serial-ATA:
Continuation of #668

closes #667

Before submitting this PR, please make sure:

For external contributors:

- [ ] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [ ] You have not used AI on any parts of this pull request.
- [ ] You have added a few sentences describing the PR here.
- [ ] Your code builds clean without any errors or warnings.

For all contributors:

- [x] You have added tests (when appropriate).
- [x] You have added an entry in the CHANGELOG (when appropriate).
- [ ] You have updated the README or other documentation to account for these changes (when appropriate).
- [ ] You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).

For PRs containing lint rule changes:

- [ ] You have updated any and all effected entries within `RULES.md`.
- [ ] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

PR #807, opened 2026-04-10
feat(tests): test wdl-{lint, analysis} rule examples

### LINKED ISSUE #771: wdl-lint/analysis: test lint rule examples
It's easy to write invalid WDL in the lint rule examples, as they're just strings:

https://github.com/stjude-rust-labs/sprocket/blob/7ba6e214516698fe47ab5af5f3641410098238ad/crates/wdl-lint/src/rules/pascal_case.rs#L50-L69

We should have a test that goes through all the rules (both `wdl-lint` and `wdl-analysis`) and their examples, makes sure that they all parse fine, and actually trigger the rule.

Serial-ATA:
closes #771

For external contributors:

For all contributors:

      is implemented.


tests/lint_examples.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// The base temp directory.
+    tmp: TempDir,
+}
+
+impl TestContext {
+    /// Create a new `TestContext`.
+    fn new() -> anyhow::Result<Self> {
+        let tmp = tempdir()?;
+        let tests = HashMap::new();
+        let formatter = Formatter::new(FormatConfig::default());
```
Serial-ATA: Windows tests are failing right now, would be nice to have #795 rather than doing a manual replacement

tests/lint_examples.rs
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+        // TODO: Actually emit the diagnostics, https://github.com/stjude-rust-labs/sprocket/pull/686
+        if result
+            .document()
+            .parse_diagnostics()
+            .iter()
+            .filter(|d| d.severity().is_error())
+            .count()
+            > 1
```
claymcleod: This threshold looks off — `> 1` means a document with exactly one error-severity parse diagnostic passes silently. The format command uses `!is_empty()` for the same check (i.e., zero tolerance). Should this be `> 0`?

claymcleod (changes requested):
Nice work getting test coverage for rule examples, this is a solid addition.

One thing worth noting beyond the inline comment: the `examples` field in `sprocket explain --format json` changes from a list of strings to a list of `Example` objects. That's a breaking change to the JSON schema, so it'd be good to call it out in a CHANGELOG entry (or at least the PR description) so consumers know to expect it.

tests/lint_examples.rs
Serial-ATA: Oops, typo. Good catch

CHANGELOG.md
```diff
@@ -32,6 +32,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Input files (JSON/YAML) must now be prefixed with `@`
   (e.g., `@inputs.json`) to disambiguate them from bare array values
   ([#820](https://github.com/stjude-rust-labs/sprocket/pull/820)).
+* The `examples` field of `sprocket explain --format json` has changed from Markdown codeblocks to the following
```
claymcleod: This needs to go into the latest set of entries, not under 0.24.0.

crates/wdl-analysis/src/rules.rs
```diff
@@ -117,18 +138,20 @@ impl Rule for UnusedImportRule {
...
 
 workflow example {
-    meta {}
-
-    output {}
 }
-```"#]
+"#,
+            },
+            revised: None,
```
adthrasher: What criteria are you using to decide when to provide a revised block? This seems similar to the UnusedInputRule below, but that has a revised block and this does not.

crates/wdl-analysis/src/rules.rs
```diff
@@ -171,20 +194,30 @@ impl Rule for UnusedInputRule {
...
-    meta {}
-
     input {
         String unused
     }
+}
+"#,
+            },
+            revised: Some(LabeledSnippet {
+                label: Some("Consider removing the input entirely"),
```
adthrasher, line 218: It may be worth naming the input, since `input` is a keyword and also appears in the example.

```suggestion
                label: Some("Consider removing the input `unused` entirely"),
```

crates/wdl-analysis/src/rules.rs
Serial-ATA: Yeah, this could use one. I leave them off for lints like `HostPathLiterals` where the solution would vary between projects. This was just a copy and paste mistake, probably

src/commands/explain.rs
```diff
@@ -137,7 +138,7 @@ pub struct Rule {
     /// An extended description of the rule (possibly Markdown formatted).
     pub explanation: &'static str,
     /// Markdown-formatted examples that would trigger the rule.
-    pub examples: &'static [&'static str],
+    pub examples: &'static [Example],
```
claymcleod, line 142: Now that `examples` is structured, it'd be nice to render the negative snippet (and revision) as fenced WDL from `Display for Rule` so the default `sprocket explain <RULE>` output gains the benefit of the refactor too.

src/commands/explain.rs
```diff
@@ -137,7 +138,7 @@ pub struct Rule {
     /// An extended description of the rule (possibly Markdown formatted).
     pub explanation: &'static str,
     /// Markdown-formatted examples that would trigger the rule.
```
claymcleod:
```suggestion
    /// Structured examples that would trigger the rule, each paired with an optional revision.
```

crates/wdl-lint/CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `UnnecessaryFunctionCall` lint rule
   ([#842](https://github.com/stjude-rust-labs/sprocket/pull/842)).
 
+#### Removed
```
claymcleod, line 25: The CHANGELOG also lists the `examples()` shape change but not the new `DynClone` supertrait on `Rule` or the `Send + Sync` widening of `wdl_lint::rules()`, both of which are breaking for downstream implementers. We should update the `CHANGELOG.md` to reflect this.

claymcleod (approved):
Approved with the following comments that may need to be addressed first.

PR #810, opened 2026-04-10
feat: diagnostic baseline support

claymcleod:
One thing I've noticed is that it's a bit difficult for users to wholesale commit to sprocket checking their codebase because often there are thousands of existing diagnostics. They may want to use `sprocket check` for all future work to make sure they don't introduce any issues, but they don't want to put in the effort to go back and fix their whole codebase first.

This PR adds a `sprocket-baseline.toml` file that keeps a baseline of expected lints/warnings and excludes them from both `sprocket check`'s output and its exit code calculation. Users run `sprocket lint --generate-baseline --deny-warnings` (or `--deny-notes`) and Sprocket writes all current diagnostics that would cause failure into the baseline. Then they can put `sprocket check` in CI to catch new issues while cleaning up existing ones on their own schedule.

Baseline entries are matched using a hash of the source content at the diagnostic span rather than line numbers. This means adding or removing lines elsewhere in the file won't invalidate unrelated baseline entries, which should produce significantly less churn than a line-number-based approach. If the actual flagged code is edited, the hash changes and the entry stops matching, so the diagnostic surfaces again.

The baseline is sensitive to the flags used to generate it. If you generate with `--deny-notes` (which baselines everything) and then run with only `--deny-warnings`, the note entries won't be relevant to the run, so they'll be reported as stale and the command will fail with a suggestion to regenerate with matching flags. Similarly, if you fix a diagnostic and its baseline entry becomes stale, the command fails with a suggestion to regenerate.

The `Baseline` type lives in `wdl-lint` so both the CLI and the LSP can use it. The LSP filters diagnostics against the baseline before publishing them to the editor.

This PR also fixes a pre-existing bug in the LSP where the exception filters for `Linter::new` and `DiagnosticsConfig::new` were inverted. Both used `.filter(|r| exceptions.contains(...))` which kept only excepted rules instead of excluding them. In practice this meant lint rules never ran in the LSP unless you happened to except them, and analysis diagnostics like `UnusedInput` were silently disabled. The fix negates both filters.

For all contributors:


crates/wdl-lsp/src/server.rs
```diff
@@ -384,7 +389,7 @@ impl ServerOptions {
             .with_diagnostics_config(DiagnosticsConfig::new(
                 wdl_analysis::rules()
                     .iter()
-                    .filter(|r| exceptions.contains(&r.id().into())),
+                    .filter(|r| !exceptions.contains(&r.id().into())),
```
claymcleod, line 392: This and the line like it below were pre-existing bugs.

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+    Parse {
+        /// The path that was being parsed.
+        path: PathBuf,
+        /// The underlying deserialization error.
+        #[source]
+        source: toml::de::Error,
+    },
+    /// Failed to serialize a baseline to TOML.
+    #[error("failed to serialize baseline")]
+    Serialize(#[source] toml::ser::Error),
```
peterhuene: `from` implies `source` and also adds the `From` impl, which is nice to have, so I tend to use it over `source` for newtype variants.

```suggestion
    Serialize(#[from] toml::ser::Error),
```

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+/// The default baseline file name.
+pub const DEFAULT_BASELINE_FILENAME: &str = "sprocket-baseline.toml";
+
+/// A single baselined diagnostic entry.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct BaselineEntry {
+    /// The rule ID that produced the diagnostic.
+    rule: String,
+    /// The file path recorded for the diagnostic (typically relative to the
+    /// project root).
```
peterhuene: Shouldn't the path be relative to the baseline file?

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+    ///
+    /// Path matching is suffix-based so that a baseline generated with
+    /// relative paths (e.g., `tools/foo.wdl`) still matches when the document
+    /// path is absolute (e.g., `/home/user/project/tools/foo.wdl`) or vice
+    /// versa.
+    fn matches_entry(&mut self, rule: &str, path: &str, span_text: &str) -> bool {
+        let hash = blake3::hash(span_text.trim().as_bytes());
+        let hash_hex = hash.to_hex();
+        let mut any = false;
+        for (i, entry) in self.baseline.diagnostic.iter().enumerate() {
```
peterhuene: Assuming the repository has a large number of diagnostics in the baseline (which is likely why they're using the baseline feature in the first place), this linear scan seems rather inefficient.
If instead we sorted the data after load (since it's immutable anyway), could we at least perform a binary search by path (scan forward for the next upper bound entry with a different path), then by rule id (scan forward for the next upper bound entry with a different rule id), then by hash?

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+            let span = label.span();
+            let root = document.root();
+            let syntax = wdl_ast::AstNode::inner(&root);
+            let span_text = syntax
+                .text()
+                .slice(rowan::TextRange::new(
+                    rowan::TextSize::from(span.start() as u32),
+                    rowan::TextSize::from(span.end() as u32),
+                ))
+                .to_string();
```
peterhuene: Rather than calling `to_string` here, we could use [`for_each_chunk`](https://docs.rs/rowan/latest/rowan/struct.SyntaxText.html#method.for_each_chunk) for hashing each chunk of text in the span.

Ideally this could be performed after we've performed a binary search by path (i.e. URI) and then by rule, but before we do a binary search by hash.

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+    }
+}
+
+/// Returns `true` if two paths refer to the same file.
+///
+/// Uses suffix matching so that a relative path like `tools/foo.wdl` matches
+/// an absolute path like `/home/user/project/tools/foo.wdl`. The shorter path
+/// must be a suffix of the longer path, aligned on a path separator boundary
+/// (either `/` or `\`).
+fn paths_match(a: &str, b: &str) -> bool {
```
peterhuene: This seems like a bit too fuzzy a match to me.

If we store paths relative to the baseline file and turn them into URIs upon loading the baseline file, then we should be able to match on the document's URI with direct string matches.

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// project root).
+    path: String,
+    /// A `blake3` hash of the trimmed source text at the diagnostic's primary
+    /// label span.
+    ///
+    /// Because this is based on content rather than line numbers, the entry
+    /// survives insertions or deletions elsewhere in the file. If the flagged
+    /// code itself is edited, the hash changes and the entry no longer
+    /// matches.
+    source_hash: String,
```
peterhuene: This could probably be `ArrayString<64>` to save having to `to_string` the `to_hex` calls.

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+
+impl BaselineEntry {
+    /// Creates a new baseline entry by hashing the given source content.
+    pub fn new(
+        rule: impl Into<String>,
+        path: impl Into<String>,
+        source_content: &str,
+        message: impl Into<String>,
+    ) -> Self {
+        let hash = blake3::hash(source_content.trim().as_bytes());
```
peterhuene: Could this `trim` call potentially not match what we do in `BaselineMatcher::suppresses`? From what I see in that implementation, we always hash the diagnostic's span of text without trimming.

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+/// Create one with [`Baseline::matcher`]. Use [`suppresses`](Self::suppresses)
+/// while iterating over diagnostics, then consult
+/// [`stale_entries`](Self::stale_entries) once iteration is complete to find
+/// entries that no longer correspond to any active diagnostic.
+#[derive(Debug)]
+pub struct BaselineMatcher<'a> {
+    /// The immutable baseline this matcher borrows from.
+    baseline: &'a Baseline,
+    /// Indices of entries matched by at least one diagnostic during the run.
+    matched: HashSet<usize>,
```
peterhuene: This feels like a great candidate to be a `BitSet` rather than a `HashSet<usize>`.

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,552 @@
...
+}
+
+impl<'a> BaselineMatcher<'a> {
+    /// Returns `true` if the given diagnostic is suppressed by any entry in
+    /// the underlying baseline.
+    ///
+    /// This also records which entries matched. Multiple diagnostics are
+    /// allowed to match the same entry.
+    pub fn suppresses(
```
peterhuene: nit: given what the function does, perhaps `is_suppressed` is a better name?

crates/wdl-lsp/src/server.rs
peterhuene: Good eye!

src/commands/check.rs
```diff
@@ -261,6 +288,83 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
...
+                    warn!(
+                        "skipping source `{uri}` with unknown scheme `{scheme}` during baseline \
+                         generation"
+                    );
+                    continue;
+                }
+            }
+
+            let path = result.document().path().to_string();
+            let source = result.document().root().text().to_string();
```
peterhuene: Not that it matters _too_ much, but we could also use `SyntaxText::slice` and `SyntaxText::for_each_chunk` to hash the source text here and pass that to `BaselineEntry::new` (probably as an `ArrayString<64>`?).

Perhaps we need a `fn hash_span(&self, span: Span) -> Result<ArrayString<64>>` (errors on span out of range) on `wdl_analysis::Document` that abstracts the `slice` and `for_each_chunk` calls?

src/commands/check.rs
```diff
@@ -273,51 +377,61 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
...
         let mut diagnostics = result.document().diagnostics().peekable();
 
         if diagnostics.peek().is_some() {
-            let path = result.document().path().to_string();
-            let source = result.document().root().text().to_string();
-
             emit_diagnostics(
                 &path,
-                source,
+                source.clone(),
```
peterhuene: Why the `clone` here? I couldn't see another use the borrow checker would complain about.

Additionally (and unrelated to these changes), I think `emit_diagnostics` should take this by `&str` and internally wrap the source with `Cow::Borrowed` (and `Cow::Owned` for the backtrace source).

src/commands/check.rs
```diff
@@ -273,51 +377,61 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
...
             }
-            v => todo!("unhandled uri scheme: {v}"),
+            scheme => {
+                warn!("skipping source `{uri}` with unknown scheme `{scheme}`");
+                continue;
+            }
         };
 
+        let path = result.document().path().to_string();
+        let source = result.document().root().text().to_string();
```
peterhuene: Do these need to be moved out from the `if diagnostics.peek().is_some()` conditional?

src/commands/check.rs
```diff
@@ -273,51 +377,61 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
...
+
+                    if hidden {
+                        return false;
+                    }
+
+                    if would_fail && matched {
+                        return false;
+                    }
+
+                    if would_fail {
```
peterhuene: This seems wrong to me. Should we not be incrementing the counts regardless of `would_fail`? At this point we're including the diagnostic in the output.

tests/cli/check/baseline/custom-baseline-path/inputs/my-baselines/custom.toml
```diff
@@ -0,0 +1,17 @@
+[[diagnostic]]
+rule = "InputName"
+path = "source.wdl"
+source_hash = "3ae7d805f6789a6402acb70ad4096a85a56bf6804eaf25c0493ac697548d30b5"
+message = "declaration identifier must be at least 3 characters"
```
peterhuene: Is `message` going to be a helpful field when reviewing diffs to a baseline file?

Without the span in the file, it seems not terribly useful to include the message to me as users can't inspect the relevant source; it only serves to increase the baseline file size.

tests/cli/check/baseline/stale-after-fix/stderr
```diff
@@ -0,0 +1,5 @@
+error: the baseline contains 3 stale entries for diagnostics that no longer exist:
+  - InputName at `source.wdl`: declaration identifier must be at least 3 characters
```
peterhuene: Likewise to the previous comment, I don't think messages relating to text that is likely no longer in the source file is terribly useful here.

Perhaps something like this?

```
error: the baseline contains 3 stale entries for diagnostics that no longer exist:
  - 1 stale `InputName` diagnostic from `source.wdl`
  - 1 stale `MetaSection` diagnostic from `source.wdl`
  - 1 stale `UnusedInput` diagnostic from `source.wd`
rerun with `--generate-baseline` to update it (if this is unexpected, make sure you run the command with the same arguments you used to generate the baseline)
```

where we aggregate the counts by rule id.

peterhuene (commented):
Here are some review comments to go over. 

Generally, this looks good and I think this will be a really useful feature!

For my understanding, could you explain the rational behind only baselining denied warnings and notes?

It feels like the baseline should be for every diagnostic regardless of hiding/denying warnings and notes.

The `--deny-warnings` and `--deny-notes` CLI options should only affect the process exit code. Accordingly, hiding warnings/notes should always exclude them from the output.

So then we would baseline all of the diagnostics and only check the baseline matcher when we would otherwise display the diagnostic (i.e. it's not otherwise hidden), right?

We would only exit non-zero on an error or on a displayed (i.e. not hidden or unmatched) warning/note when denying warnings/notes.

I feel it a little bit odd that the baseline may differ on CLI options; I feel like it leads to a confusing UX.

claymcleod:
> It feels like the baseline should be for every diagnostic regardless of hiding/denying warnings and notes.

You make a good point! I've updated this in the latest commit.

tests/cli/check/baseline/custom-baseline-path/inputs/my-baselines/custom.toml
claymcleod: Agree, it was an idea but I think we should just drop it.

crates/wdl-lint/src/baseline.rs
claymcleod: Paths are now stored relative to the baseline file and resolved once at load to canonical absolute paths for in-memory comparison. Remote sources compare as URL strings. `paths_match` is gone.

src/commands/submit.rs
```diff
@@ -317,7 +317,7 @@ command <<<>>>
             let poll_url = format!("{base_url}/api/v1/runs/{uuid}");
             let mut status = String::new();
 
-            for _ in 0..50 {
+            for _ in 0..600 {
```
claymcleod, line 320: I'm not sure how this got here, but I remember the fix, so I'm going to leave it.

src/commands/check.rs
claymcleod: Good point. Today there's no behavior change since `counts` is only read via `verify_no_warnings`/`verify_no_notes` under the matching `--deny-*` flag, but the field should reflect "what we emitted" regardless of why. Fixed in `40077f69`.

crates/wdl-analysis/src/document.rs
```diff
@@ -850,6 +853,33 @@ impl Document {
...
+    /// Computes the `blake3` hash of the document's source text over the
+    /// given span and returns the hex form.
+    ///
+    /// Uses `rowan::SyntaxText::for_each_chunk` so the span's text is never
+    /// materialized as a `String`.
+    ///
+    /// Returns `None` if `span` falls outside the document's source text.
+    pub fn hash_span(&self, span: Span) -> Option<ArrayString<64>> {
+        let root = self.root();
+        let syntax = wdl_ast::AstNode::inner(&root);
```
peterhuene: Is there a reason for the more explicit form of this call rather than just `let text = self.root().inner().text();`?

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,736 @@
...
+    base_dir: Option<PathBuf>,
+}
+
+impl BaselineIndex {
+    /// Recomputes `resolved` and `sorted` from the given entries.
+    fn rebuild(&mut self, entries: &[BaselineEntry]) {
+        self.resolved = entries
+            .iter()
+            .map(|entry| resolve_entry_ref(entry, self.base_dir.as_deref()))
+            .collect();
```
peterhuene: Nitty nit:

Could we reuse the capacity of these two `Vec`s, such as by calling `clear` followed by `extend` rather than dropping?

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,736 @@
...
+    ///
+    /// `None` for entries whose path could not be resolved (e.g., the
+    /// referenced file does not exist); such entries will never match and
+    /// will be reported as stale.
+    ///
+    /// # Invariant
+    ///
+    /// `resolved.len()` equals the length of the backing entries vector at
+    /// all times. Maintained by [`Self::rebuild`].
+    resolved: Vec<Option<String>>,
```
peterhuene: I feel like if we simply stored `Option<Url>` here rather than `Option<String>` we could do direct comparisons against a document's `uri()` without having to have `canonicalize_document_ref` on every lookup.

`resolve_entry_ref` could then basically convert the `path` from the baseline file to a `Url` by checking if the string contains `://` and, if so, parse it as a URL or otherwise call `Url::from_file_path` after joining it to the base path.

crates/wdl-lint/src/baseline.rs
```diff
@@ -0,0 +1,736 @@
...
+            let doc_ref = canonicalize_document_ref(document.uri());
+            return self.matches_entry(rule, &doc_ref, &hash);
+        }
+
+        false
+    }
+
+    /// Returns baseline entries that were never matched by any diagnostic
+    /// during this run.
+    pub fn stale_entries(&self) -> Vec<&'a BaselineEntry> {
```
peterhuene: This could probably return `impl Iterator<Item = &'a BaselineEntry>` so we don't have to `collect` (let the caller do so if they really desire it).

Looking at where this is really used in `check.rs`, it doesn't seem a collection is warranted.

src/commands/check.rs
```diff
@@ -261,6 +289,81 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
...
+
+            for d in result.document().diagnostics() {
+                if matches!(d.severity(), Severity::Warning | Severity::Note)
+                    && args.common.suppress_imports
+                    && !provided_source_uris.contains(uri)
+                {
+                    continue;
+                }
+
+                if let Some(rule) = d.rule()
```
peterhuene, line 349: So in an earlier comment I said "all diagnostics" when I meant "all warning and note diagnostics (regardless of denying)". I think error diagnostics should not be included in the baseline as these will be parsing or analysis errors that are never hidden and shouldn't otherwise match a baseline.

src/commands/check.rs
```diff
@@ -284,40 +390,38 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
...
-
-                            if !hide_notes {
-                                counts.notes += 1;
-                                true
-                            } else {
-                                false
-                            }
-                        }
+                    let matched = matcher
+                        .as_mut()
```
peterhuene: Likewise, I think we should skip checking for a match for error diagnostics.

src/commands/check.rs
```diff
@@ -326,6 +430,37 @@ pub async fn check(args: CheckArgs, config: Config, colorize: bool) -> CommandRe
...
+                ));
+            }
+            return Err(anyhow!(
+                "the baseline contains {count} stale entr{plural} for diagnostics that no longer \
+                 exist:{listing}\nrerun with `--generate-baseline` to update it",
+                count = stale.len(),
+                plural = if stale.len() == 1 { "y" } else { "ies" },
+            )
+            .into());
+        }
```
peterhuene: Given the requested change above for `stale_entries` to return iterator, I think this section could be better expressed by:


```suggestion
        let mut groups = std::collections::BTreeMap::new();
        for entry in matcher.stale_entries() {
            debug!(
                rule = entry.rule(),
                path = entry.path(),
                "stale baseline entry"
            );

            *groups.entry((entry.rule(), entry.path())).or_insert(0usize) += 1;
        }

        if !groups.is_empty() {
            let mut listing = String::new();
            for ((rule, path), count) in &groups {
                listing.push_str(&format!(
                    "\n  - {count} stale `{rule}` diagnostic{plural} from `{path}`",
                    plural = if *count == 1 { "" } else { "s" },
                ));
            }
            return Err(anyhow!(
                "the baseline contains {count} stale entr{plural} for diagnostics that no longer \
                 exist:{listing}\nrerun with `--generate-baseline` to update it",
                count = stale.len(),
                plural = if stale.len() == 1 { "y" } else { "ies" },
            )
            .into());
        }
```

peterhuene (commented):
A few more minor comments to go over, but this is looking really good!

src/commands/check.rs
peterhuene: Nit: also, the `return Err(anyhow!(...))` could possibly be a `bail!`.

src/commands/check.rs
claymcleod: Done in 2b475a7f—errors are now skipped during baseline generation.

crates/wdl-lint/src/baseline.rs
claymcleod: Done in cca96e13—`stale_entries` now returns `impl Iterator<Item = &BaselineEntry>` and the caller in `check.rs` builds the `BTreeMap` in a single pass.

crates/wdl-lint/src/baseline.rs
claymcleod: A lot of the ceremony around `canonicalize_document_ref` was because I was storing filesystem paths as strings to avoid the `file://` prefix, which I thought looked ugly. However, I think it's fine to just store it in this case as well.

claymcleod: One thing that stands out as to why we can't compare directly against `document.uri()` is that the analysis engine constructs document URIs via `path-clean` rather than `std::fs::canonicalize`, so they don't resolve symlinks. The baseline side canonicalizes at load time, but the document side still needs a per-lookup `canonicalize_document_ref` call to match. We'd need to canonicalize at the analysis layer to fully eliminate it, which feels out of scope here.

peterhuene (approved):
Looks great! Thanks for addressing the feedback comments.

PR #820, opened 2026-04-15
feat: adds ergonomic array inputs on the CLI

claymcleod:
This introduces ergonomic ways to provide `Array` inputs on the command line without JSON syntax, along with supporting engine changes for correct path resolution and type coercion.

**Repeated keys.** Specifying the same key multiple times collects the values into an array (e.g., `task.files=a.txt task.files=b.txt`). A single occurrence remains scalar.

**Trailing bare arguments.** Values without an `=` that follow a key-value pair are appended to the preceding key's array (e.g., `task.files=a.txt b.txt c.txt`). This pairs naturally with shell glob expansion, where `task.files=*.txt` expands to individual positional arguments.

**Scalar-to-array auto-wrapping.** When a WDL input expects `Array[T]` and a single scalar value is provided (e.g., `task.items=hello`), the engine auto-wraps it into a single-element array rather than rejecting it with a type mismatch.

**Per-element path origins.** Each array element retains its own filesystem origin for resolving relative `File`/`Directory` paths. `join_paths` now accepts a per-key slice of origins (`&[EvaluationPath]`) instead of a single reference, so mixed-origin arrays (e.g., values from an input file combined with CLI arguments) resolve paths correctly.

**`@` prefix for input files.** Input files (JSON/YAML) must now be prefixed with `@` (e.g., `@inputs.json`) to disambiguate them from bare array values. This is a breaking change that follows the convention used by `curl` and other CLI tools.

For all contributors:


peterhuene (approved):
Looks good!

PR #824, opened 2026-04-16
fix(wdl-format): preserve comment order

### LINKED ISSUE #799: wdl-format: comments are reordered
**Describe the bug**
`wdl-format` moves all normal comments below doc comments:

```wdl
version 1.2

# TODO: Implement this
## This is a Foo struct
struct Foo {}
```

Becomes:

```wdl
version 1.2

## This is a Foo struct
# TODO: Implement this
struct Foo {}
```

This happens regardless of whitespace, so the following is also reodered:

```wdl
version 1.2

# TODO: Implement this

## This is a Foo struct
struct Foo {}
```

**Expected behavior**
The comments should keep their original order

alexnham:
bugfix: comments reordering closes #799

Keeps the order of comments as previously every comment would go below doc comments.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-format/src/token/pre.rs
```diff
@@ -177,9 +177,10 @@ impl TokenStream<PreToken> {
     fn push_preceding_trivia(&mut self, token: &wdl_ast::Token) {
         assert!(!token.inner().kind().is_trivia());
         let preceding_trivia = token.inner().preceding_trivia();
-        let mut documentation = String::new();
         let mut trivia = Vec::new();
         let mut exceptions = HashSet::new();
+        let mut doc_string = String::new();
+        let mut docs_index: i32 = -1;
```
claymcleod: We should use `Option<usize>` here instead of an `i32` sentinel—it expresses the "not yet set" state idiomatically and eliminates the `as i32` / `as usize` casts.

Rather than pushing and overwriting intermediate values, consider recording the insertion point and doing a single `Vec::insert` after the collection loop:

```rust
let mut doc_string = String::new();
let mut docs_insert_at: Option<usize> = None;

// In the doc comment branch—just accumulate and record position:
doc_string.push_str(t);
doc_string.push_str(NEWLINE);
if docs_insert_at.is_none() {
    docs_insert_at = Some(trivia.len());
}

// After the collection loop, before the emit loop:
if let Some(i) = docs_insert_at {
    trivia.insert(i, PreToken::Trivia(Trivia::Comment(
        Comment::Documentation(Rc::new(doc_string)),
    )));
}
```

This avoids all intermediate `String` clones and `Rc` allocations, and `Vec::insert` is fine here since the trivia vec is always small.

crates/wdl-format/src/token/pre.rs
```diff
@@ -193,8 +194,19 @@ impl TokenStream<PreToken> {
...
+                            let comment = PreToken::Trivia(Trivia::Comment(
+                                Comment::Documentation(Rc::new(doc_string.clone())),
+                            ));
+                            trivia.push(comment);
+                            docs_index = trivia.len() as i32 - 1;
+                        } else {
+                            trivia[docs_index as usize] = PreToken::Trivia(Trivia::Comment(
+                                Comment::Documentation(Rc::new(doc_string.clone())),
+                            ));
+                        }
```
claymcleod: When `##` doc comments are separated by `#` normal comments (e.g., `## a`, `# x`, `## b`), the code still consolidates them into one `Documentation` token placed at the first `##`'s position—moving `## b` before `# x`. This is the same class of reordering that this PR aims to fix, and the Bar and Baz test cases show it happening.

I think the fix is to flush `doc_string` whenever a regular `#` comment interrupts a `##` sequence, so each contiguous `##` run becomes its own `Documentation` token. You'd also want to skip pushing `BlankLine` tokens while `doc_string` is non-empty (since whitespace between `##` lines is part of the doc block, not a trivia separator). This eliminates the need for `docs_index` entirely—just push `Documentation` tokens inline as each doc block closes, and flush any remaining content after the loop.

The expected output for the Bar and Baz test cases would need to be updated to reflect the preserved comment ordering.

claymcleod (changes requested):
Thanks for tackling this, the core idea of inserting the `Documentation` token in-place rather than hoisting it is the right direction. The two inline comments cover the main suggestions: using `Option<usize>` instead of `i32` with a deferred allocation, and flushing doc blocks when a `#` comment interrupts a `##` sequence so non-contiguous doc comments don't get reordered either.

crates/wdl-format/src/token/pre.rs
alexnham: Just to confirm:

```
## 1

# 2

## 3

# 4

## 5
```

Would stay in that order of 1 2 3 4 5 ?

alexnham: Another example is in clays_complex_script

```
## # Header
# regular comment
#@ except: CommentWhitespace, DeprecatedObject, MetaDescription
#@ except: ParameterMetaMatched, MatchingOutputMeta

## part of preamble
```

It should go :

```
## # Header
# regular comment
## part of preamble
#@ except: CommentWhitespace, DeprecatedObject, MetaDescription
#@ except: ParameterMetaMatched, MatchingOutputMeta
```

claymcleod: I think both of these are correct

alexnham:
How does this implementation feel ? It's able to keep the order and removes doc_string and index entirely. It updates test cases `clays_complex_script` and `comment_normalization` to updated format

crates/wdl-format/tests/format/comments_ordering/source.formatted.wdl
```diff
@@ -0,0 +1,26 @@
+version 1.2
+
+# TODO: Implement this
+## This is a Foo struct
```
Serial-ATA, line 4: Ideally the newlines between the comments would be preserved as well, like the "hello world" comment on the `Baz` struct

claymcleod:

The underlying work remains tracked in #799, so it won't be lost. You're welcome to reopen this PR if you'd like to pick it back up.

PR #831, opened 2026-04-24
feat: add WDL 1.4 grammar and symbolic imports

claymcleod:
Adds WDL `1.4` to `wdl-grammar` and `wdl-ast`, along with the reserved `from` keyword and parsing for the three import forms from [`openwdl/wdl#765`](https://github.com/openwdl/wdl/pull/765). `1.4` is gated behind a new `wdl-analysis` feature flag `wdl_1_4`, which defaults to `false` during the unstable development cycle.

This is phase one of a multi-PR effort to land the WDL modules system in Sprocket. Subsequent phases will add the `wdl-modules` and `wdl-modules-resolver` crates, wire resolution into `wdl-analysis`, and ship the `sprocket module` CLI surface.

The three import forms we parse:

```wdl
# Form 1: namespace import. `as <alias>` is optional; without it the
# namespace is derived from the source (file stem for quoted, last path
# component for symbolic). UDTs are copied into the importing document;
# tasks and workflows are reached through the namespace.
import "csvkit.wdl"
import openwdl/csvkit as csv

# Form 2: wildcard. Brings every task, workflow, and UDT into scope.
# No namespace.
import * from openwdl/csvkit

# Form 3: selected members. Brings only the listed items into scope,
# with optional per-member \`as <Name>\` rename. No namespace.
import { sort, CsvSort as MySort } from openwdl/csvkit
```

Each `<source>` is either a quoted URI or an unquoted symbolic module path; once resolved, both behave identically. Form 1 is the only form that accepts `as <alias>` on the source. Forms 2 and 3 reject it. Member entries in form 3 are single identifiers with an optional `as <Name>` rename; dotted paths are not accepted. Trailing commas inside the braces are permitted.

### CST and AST

`ImportStatementNode` is a single unified node covering all three forms with both source styles. The shape is recovered through two enums on `ImportStatement`:

* `ImportForm::{Namespace, Wildcard, Selected}` — returned by `form()`, names what each form does to the importing scope rather than numbering it.
* `ImportSource::{Uri, ModulePath}` — returned by `source()`, distinguishes quoted URIs from symbolic module paths.

The braced selection clause is represented by `ImportMembers` and `ImportMember`, each of which exposes `name()` and `alias()`. The names are deliberately not prefixed with `Symbolic` because the syntax works for both quoted and symbolic sources.

### `from` keyword

The v1 lexer now recognizes `from` as `Token::FromKeyword` unconditionally. The parser accepts it anywhere an identifier is allowed (via `ANY_IDENT`), so pre-`1.4` documents that happen to use `from` as an identifier continue to parse without a regression. The import grammar widens its identifier-position slots — namespace alias, alias source/target, selected-member name/alias, and symbolic module path components — so `import "lib.wdl" as from`, `import "lib.wdl" alias from as MyFrom`, `import { from } from "csvkit.wdl"`, and `import openwdl/from/csvkit` all parse cleanly. The new `tests/parsing/from-as-ident-imports/` fixture locks this in.

### Version gating

`FeatureFlags::wdl_1_4` in `wdl-analysis/src/config.rs` defaults to `false`. When a document declares `version 1.4` without the flag set, `VersionVisitor` reports `` use of WDL version 1.4 is unstable; set \`feature_flags.wdl_1_4 = true\` in your configuration to opt in ``. Fixtures under `wdl-analysis/tests/validation/wdl-1-4-{with,without}-flag/` cover both paths. This pattern mirrors the `wdl_1_3` precedent from PRs #411 and #435.

A pre-`1.4` document using a form-2/3 import (or any symbolic source) gets a single, focused diagnostic — the wildcard/selected forms already require `from`, which subsumes the symbolic-path-only case.

### Symbolic-import resolution

When `version 1.4` is opted into via `feature_flags.wdl_1_4 = true`, symbolic imports parse and validate but cannot yet be resolved to documents (the resolver lands in a subsequent phase). Rather than silently dropping them, `ImportsVisitor` emits a clear `` symbolic module resolution is not yet implemented `` diagnostic on the path span. The new `wdl-1-4-symbolic-import-with-flag/` fixture locks the behavior.

### Formatting

The formatter renders form 1 and the wildcard form on a single line. Braced member lists stay inline when the full statement fits within `max_line_length`, and wrap to an indented multiline form with trailing commas when they do not. A trailing comma on a single-line form is stripped; a multiline form always carries one on the last member. A comment anywhere inside the braces forces the multiline form so the comment can keep its own line, exercised by `tests/format/symbolic_imports_comments/`.

### Scope of this PR

This PR lands grammar, AST, and format only. The analysis-level semantics for forms 2 and 3 (bringing tasks, workflows, and UDTs into the importing document's scope), the symbolic-source resolver, and the lockfile and trust-store machinery are deferred to subsequent phases of the modules effort.

For all contributors:

- [x] You have made a PR to the \`next\` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).


crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1630,7 +1636,14 @@ fn resolve_import(
     stmt: &ImportStatement,
     importer_index: NodeIndex,
 ) -> Result<(Arc<Url>, Document), Option<Diagnostic>> {
-    let uri = stmt.uri();
+    let uri = match stmt.uri() {
+        Some(uri) => uri,
+        None => {
+            // Symbolic module paths do not resolve through the quoted-URI
+            // graph; the module resolver handles them separately.
+            return Err(None);
```
claymcleod, line 1638: This will be wired up in a future PR.

crates/wdl-analysis/src/queue.rs
```diff
@@ -1105,7 +1106,13 @@ where
                 None | Some(Ast::Unsupported) => {}
                 Some(Ast::V1(ast)) => {
                     for import in ast.imports() {
-                        let text = match import.uri().text() {
+                        // Only quoted imports contribute dependency edges;
+                        // symbolic imports resolve through the module
+                        // resolver and do not add graph nodes here.
```
claymcleod, line 1111: TBD whether this will remain true in future versions.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -364,7 +365,6 @@ const ANY_IDENT: TokenSet = TokenSet::new(&[
     Token::OutputKeyword as u8,
     Token::ParameterMetaKeyword as u8,
     Token::RequirementsKeyword as u8,
-    Token::HintsKeyword as u8,
```
claymcleod, line 367: This was duplicated with line 355.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -479,28 +479,149 @@ fn item(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marker, Diagnost
...
+            expected_fn!(parser, marker, import_alias);
+        }
     }
 
     marker.complete(parser, SyntaxKind::ImportStatementNode);
     Ok(())
 }
 
+/// Parses the braced member-selection clause `{ <m>, ... }`.
+fn import_members(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marker, Diagnostic)> {
```
peterhuene, line 630: I think the body of this function could be replaced with:

```rust
braced_items!(
    parser,
    marker,
    Some(Token::Comma),
    IMPORT_MEMBER_RECOVERY_SET,
    import_member
);

 marker.complete(parser, SyntaxKind::ImportMembersNode);
```

with the recovery set being `TokenSet::new(&[Token::Ident as u8, Token::CloseBrace as u8]);`

`braced_items!` handles the looping, but it also recovers from errors by skipping unexpected tokens until it finds (in this case) another identifier, comma, or close brace.

As import statements are pretty important for code analysis, it'd be best to maximally recover from unexpected syntax encountered within the import member list.

crates/wdl-analysis/src/queue.rs
peterhuene: I think it'll likely need to change as we'll want evaluation edges between the importing document and the module's `index.wdl`, allowing us to then parse the `index.wdl` and the documents it imports.

I would imagine on startup of the engine we would want to basically check if that we're evaluating in the context of a module, and if so, resolve dependencies up front (generate or use lock file, download the module source, resolve versions, etc.).

At this point here, though, we should be asking the module resolver to simply give us the local path to the module's `index.wdl` and then treat it as if it had been imported by a string (i.e. the absolute path to the `index.wdl`).

crates/wdl-ast/src/v1/import.rs
```diff
@@ -105,7 +235,131 @@ impl<N: TreeNode> AstNode<N> for ImportStatement<N> {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SymbolicModulePath<N: TreeNode = SyntaxNode>(N);
+
+impl<N: TreeNode> SymbolicModulePath<N> {
+    /// The path components, in source order.
+    pub fn components(&self) -> impl Iterator<Item = Ident<N::Token>> + use<'_, N> {
+        self.0.children_with_tokens().filter_map(|c| match c {
+            NodeOrToken::Token(t) => Ident::cast(t),
+            NodeOrToken::Node(_) => None,
+        })
```
peterhuene: This can be simplified to:

```suggestion
        self.tokens()
```

crates/wdl-ast/src/v1/import.rs
```diff
@@ -105,7 +235,131 @@ impl<N: TreeNode> AstNode<N> for ImportStatement<N> {
...
+    pub fn alias(&self) -> Option<Ident<N::Token>> {
+        self.idents().nth(1)
+    }
+
+    /// Returns every `Ident` child token in source order.
+    fn idents(&self) -> impl Iterator<Item = Ident<N::Token>> + use<'_, N> {
+        self.0.children_with_tokens().filter_map(|c| match c {
+            NodeOrToken::Token(t) => Ident::cast(t),
+            NodeOrToken::Node(_) => None,
+        })
```
peterhuene: This can also be simplified to:


```suggestion
        self.tokens()
```

peterhuene (commented):
This looks great.

Just a few minor comments.

crates/wdl-grammar/src/grammar/v1.rs
claymcleod: I just want to point out that the above is not unequivocally better in all cases. For example, in this snippet:

```rust
import { sort, cut from openwdl/csvkit
```

The parse tree from the existing approach is much more palatable than the new one. In particular, everything from `from` and after is outside the `ImportMembersNode`. With the solution you suggest here, `openwdl` and `csvkit` are interpreted as members attempting to be imported.

I think what I'm going to go for is something with the semantics of `braced_items!` but perhaps a safety hatch to escape on the `from` keyword.

crates/wdl-analysis/src/queue.rs
claymcleod: Sounds good. I plan to implement the resolver in the next PR, so nothing to do here for now.

crates/wdl-grammar/src/grammar/v1.rs
peterhuene: We can improve this a bit as a follow-up. If the unexpected token happens to be the one that immediately follows the list (e.g. `from` in this case), we could also assume that we're missing a closing token, emit a "expected closing delimiter" error, and perhaps insert a "zero-length" closing token into the tree and attempt to parse the remainder of the import statement.

No error recovery heuristic is going to give desired results _all_ of the time, but given that import statements with errors can easily cascade errors throughout a document, it would be ideal if we can recover from errors in these statements especially.

peterhuene (approved):

PR #836, opened 2026-05-01
feat: add `wdl-modules` crate

claymcleod:
This PR introduces `wdl-modules`, a pure Rust reference implementation of the WDL module specification ([openwdl/wdl#765](https://github.com/openwdl/wdl/pull/765)). The crate covers every part of the spec that does not require networking or process spawning; resolution and registry concerns ship in a follow-up `resolver` module within the `wdl-modules` crate.

Includes

- `module.json` manifest parsing and validation.
- `module-lock.json` lockfile parsing.
- Deterministic SHA-256 content hashing.
- Ed25519 signing and verification of `module.sig` via OpenSSH-format keys.
- Symbolic-path, dependency-source, version-requirement, and SPDX license parsing.
- Strict JSON parsing (rejects duplicate object keys at any depth).
- Module file-tree validation (reserved-filename placement, NFC uniqueness).

For all contributors:


crates/wdl-modules/src/license.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(into = "String", try_from = "String")]
+pub struct LicenseExpression {
+    // NOTE: we store the canonical string rather than the parsed
+    // `spdx::Expression` because the upstream `Expression` only derives
+    // `Clone`—not `Debug`, `PartialEq`, `Eq`, or `Hash`—and those four
+    // derives are required for `LicenseExpression` to compose into
+    // `Manifest`, `Tool`, and other types that themselves derive them.
+    /// The canonical (parsed and re-rendered) form of the expression.
+    canonical: String,
```
Serial-ATA: `Expression` implements `AsRef<str>`, you can add manual impls that make use of that rather than reparsing in `as_expression()`

crates/wdl-modules/src/version_requirement.rs
```diff
@@ -0,0 +1,134 @@
...
+impl VersionRequirement {
+    /// Tests whether a [`Version`] satisfies this requirement.
+    pub fn matches(&self, version: &Version) -> bool {
+        self.0.matches(version)
+    }
+
+    /// Returns a reference to the underlying [`VersionReq`].
+    pub fn inner(&self) -> &VersionReq {
+        &self.0
+    }
```
Serial-ATA: Could implement `Deref<Target=VersionReq>` instead

crates/wdl-modules/src/version_requirement.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl TryFrom<String> for VersionRequirement {
+    type Error = VersionRequirementError;
+
+    fn try_from(s: String) -> Result<Self, Self::Error> {
+        let trimmed = s.trim();
+        if trimmed.is_empty() {
+            return Err(VersionRequirementError(s));
+        }
```
Serial-ATA:
```suggestion
```

Unnecessary guard

Serial-ATA (commented):
Haven't looked in depth, just a few nits

crates/wdl-modules/src/dependency/name.rs
peterhuene, line 1: Nit: I feel like this file and `dependency_source.rs` could be combined into a single `dependency` module.

crates/wdl-modules/src/dependency/name.rs
```diff
@@ -0,0 +1,120 @@
...
+use std::fmt;
+use std::str::FromStr;
+
+use serde::Deserialize;
+use serde::Serialize;
+use thiserror::Error;
+
+/// An error parsing a [`DependencyName`].
+#[derive(Debug, Error, PartialEq, Eq)]
+#[error("dependency name `{0}` does not match `[A-Za-z][A-Za-z0-9_]*`")]
```
peterhuene: I wonder if we should do what `cargo` does and allow `-` in dependency names, but normalize them to `_` for the purpose of registering them with the module resolver which will be based off of the WDL identifier.

crates/wdl-modules/src/dependency_name.rs
```diff
@@ -0,0 +1,120 @@
...
+/// governs the keys of `dependencies` in [`Manifest`](crate::Manifest), the
+/// top-level keys of [`Lockfile::dependencies`](crate::Lockfile), and the
+/// `<dep-name>` portion of a [`SymbolicPath`](crate::SymbolicPath).
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+#[serde(try_from = "String")]
+pub struct DependencyName(String);
+
+impl DependencyName {
+    /// Returns the dependency name as a string slice.
+    pub fn inner(&self) -> &str {
```
peterhuene: Nit: I'd suggest naming this `as_str` 

```suggestion
    pub fn as_str(&self) -> &str {
```

crates/wdl-modules/src/dependency_name.rs
```diff
@@ -0,0 +1,120 @@
...
+        &self.0
+    }
+
+    /// Consumes the [`DependencyName`] and returns the inner [`String`].
+    pub fn into_inner(self) -> String {
```
peterhuene: Nit: I'd recommend calling this `into_string` or instead implement `From<DependencyName> for String`.

```suggestion
    pub fn into_string(self) -> String {
```

crates/wdl-modules/src/dependency_source.rs
```diff
@@ -0,0 +1,317 @@
...
+        /// A short description of which validation rule was violated.
+        reason: &'static str,
+    },
+
+    /// A version requirement on a Git dependency was invalid.
+    #[error(transparent)]
+    VersionRequirement(#[from] VersionRequirementError),
+
+    /// The `git` URL did not parse.
+    #[error("invalid `git` URL: {0}")]
```
peterhuene: For consistency with other error messages that don't backtick `git`:

```suggestion
    /// The Git URL did not parse.
    #[error("invalid Git URL: {0}")]
```

crates/wdl-modules/src/license.rs
peterhuene: Seconded this feedback. I'd prefer a manual impl of `Debug`, `PartialEq` (can keep the `Eq` auto-impl), and `Hash` which make use of `as_ref`.

crates/wdl-modules/src/license.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+impl LicenseExpression {
+    /// Returns the canonical string form of the expression.
+    pub fn as_str(&self) -> &str {
+        &self.canonical
+    }
+
+    /// Re-parses the canonical string into a fresh [`spdx::Expression`].
+    pub fn as_expression(&self) -> spdx::Expression {
```
peterhuene: With the above comment re: wrapping `Expression`, we can then change this to return a reference to the inner.

crates/wdl-modules/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "wdl-modules"
+version = "0.1.0"
+description = "Manifest, lockfile, content hashing, and signing for the WDL module system"
```
peterhuene: I find this description a little too specific.

Perhaps simply `Implementation of the WDL module specification` or some such?

crates/wdl-modules/src/strict_json.rs
peterhuene, line 1: I feel like this implementation is more generic than is required for the crate and needs, effectively, two deserializations (the last being from a `Value`).

If this exists simply to deserialize a map to enforce key uniqueness then I think we could simplify it to just that use case.

I would instead suggest that we implement a specific deserialization function that we can use with `serde(deserialize_with = deserialize_unique_key_map)` on the requisite fields:

```rust
fn deserialize_unique_key_map<'de, D, K, V>(deserializer: D) -> Result<BTreeMap<K, V>, D::Error> where D: Deserializer<'de>, K: Deserialize, V: Deserialize {
   deserialize.deserialize_map(Visitor)
}
```

where `Visitor` could just be inlined into the function and specific to deserializing a map with unique keys.

crates/wdl-modules/src/lockfile.rs
```diff
@@ -0,0 +1,414 @@
...
+            r#"{
+                "version": 1,
+                "dependencies": {
+                    "spellbook": {
+                        "source": {
+                            "git": "https://github.com/openwdl/spellbook",
+                            "commit": "a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2"
+                        },
+                        "modules": {
+                            ".": {
```
peterhuene: This layout of using `.` to describe the dependency is a little strange to me rather than promoting this information to the dependency's entry itself.

I also don't quite understand the need to describe more than one module here as my understanding would be that a dependency can only ever be exactly one module.

For _repositories_ with more than one module, it's a matter of the resolver to figure out which module is being referenced for the dependency, but the dependency doesn't need to describe every module in the repository as that seems extraneous.

crates/wdl-modules/src/relative_path.rs
```diff
@@ -0,0 +1,241 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+use thiserror::Error;
+use unicode_normalization::UnicodeNormalization;
+
+/// An error constructing a [`RelativePath`].
+#[derive(Clone, Debug, Eq, Error, PartialEq)]
+pub enum RelativePathError {
+    /// The path contains non-UTF-8 bytes.
+    #[error("path contains non-UTF-8 bytes")]
```
peterhuene: Nit:

```suggestion
    #[error("path is not encoded as UTF-8")]
```

crates/wdl-modules/src/relative_path.rs
```diff
@@ -0,0 +1,241 @@
...
+    NonUtf8,
+    /// The path is empty.
+    #[error("path is empty")]
+    Empty,
+    /// The path contains a null byte.
+    #[error("path contains a null byte")]
+    NullByte,
+    /// The path uses Windows-style `\` separators.
+    #[error("path uses Windows-style `\\` separators")]
```
peterhuene: Nit:

```suggestion
    #[error("path contains Windows path separators (e.g. `\\`)")]
```

crates/wdl-modules/src/relative_path.rs
```diff
@@ -0,0 +1,241 @@
...
+use std::str::FromStr;
+
+use serde::Deserialize;
+use serde::Serialize;
+use thiserror::Error;
+use unicode_normalization::UnicodeNormalization;
+
+/// An error constructing a [`RelativePath`].
+#[derive(Clone, Debug, Eq, Error, PartialEq)]
+pub enum RelativePathError {
```
peterhuene, line 23: It may make sense to record and display the path in question for some of these error variants, I think.

crates/wdl-modules/src/relative_path.rs
```diff
@@ -0,0 +1,241 @@
...
+    #[error("path is empty")]
+    Empty,
+    /// The path contains a null byte.
+    #[error("path contains a null byte")]
+    NullByte,
+    /// The path uses Windows-style `\` separators.
+    #[error("path uses Windows-style `\\` separators")]
+    Backslash,
+    /// The path starts with `/`.
+    #[error("path starts with `/`")]
```
peterhuene: Nit:

```suggestion
    #[error("path must be relative and therefore cannot start with a `/`")]
```

crates/wdl-modules/src/relative_path.rs
```diff
@@ -0,0 +1,241 @@
...
+    Absolute,
+    /// The path uses a Windows-style drive letter.
+    #[error("path uses a Windows-style drive letter")]
```
peterhuene: Nit:

```suggestion
    #[error("path contains a Windows drive letter")]
```

To be honest, I feel like this and a few other variants could just be rolled into a "path cannot be absolute" error.

crates/wdl-modules/src/relative_path.rs
```diff
@@ -0,0 +1,241 @@
...
+    DriveLetter,
+    /// The path resolves to nothing after lexical cleanup.
+    #[error("path resolves to empty")]
+    ResolvesToEmpty,
+    /// The path escapes the module root via leading `..` segments.
+    #[error("path escapes the module root")]
```
peterhuene: Nit:

```suggestion
    #[error("path escapes the module root directory")]
```

crates/wdl-modules/src/relative_path.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    /// Returns the path as a [`Path`].
+    pub fn as_path(&self) -> &Path {
+        Path::new(&self.0)
+    }
+
+    /// Consumes the [`RelativePath`] and returns its inner string.
+    pub fn into_inner(self) -> String {
+        self.0
+    }
```
peterhuene: This feels redundant given the `From<RelativePath> for String` impl.

crates/wdl-modules/src/symbolic_path.rs
```diff
@@ -0,0 +1,162 @@
...
+     non-`.`/`..` segments"
+)]
+pub struct SymbolicPathError(String);
+
+/// A symbolic module path.
+///
+/// The string form is `<dep-name>[/<sub-path>]`. The `<dep-name>` is the
+/// key declared under `dependencies` in the consumer's `module.json`; the
+/// optional `<sub-path>` addresses a specific module within a multi-module
+/// dependency source. Path components are case-sensitive.
```
peterhuene:
```suggestion
/// optional `<sub-path>` addresses a specific document within a module.
/// 
/// Path components are case-sensitive.
```

I don't want us to confuse an individual document within a module with itself being a module; a module is a collection of WDL documents only.

peterhuene (commented):
This looks good.

I just have a few review comments to go over.

crates/wdl-modules/src/strict_json.rs
claymcleod: The spec requires rejecting duplicate keys at any nesting depth—not just in specific map fields—so a `deserialize_with` annotation on individual struct fields wouldn't cover top-level duplicates or keys inside nested structures we don't explicitly model (e.g., the `extra` catch-all maps). I renamed the internals to make the intent clearer, but the two-step structure (checked `Value` → typed `T`) is the simplest way I found to enforce the invariant globally.

crates/wdl-modules/src/dependency/name.rs
claymcleod: Good idea. Implemented in 2774356d.

crates/wdl-modules/src/lockfile.rs
claymcleod: Flattened in 9ecde114: `version`, `checksum`, `signer`, and `dependencies` now sit directly on `DependencyEntry`. Corresponding spec update pushed to `feat/dependency-name-hyphens` in `openwdl/wdl`.

crates/wdl-modules/src/relative_path.rs
claymcleod: Done in de0a18a5. Error variants now include the offending path, and `Absolute` and `DriveLetter` are consolidated into a single `Absolute` variant.

crates/wdl-modules/src/strict_json.rs
peterhuene: That feels overly restrictive for the modules proposal, to be honest. It makes a lot of sense to have a restriction for known fields like the dependency map, for example, but to do it document-wide, especially when we're expected to round-trip any unknown fields, could be potentially limiting given that JSON doesn't _prohibit_ duplicate keys, just recommends they shouldn't be used.

At any rate, I don't feel strongly enough about it to press the issue.

peterhuene (approved):
Looks good!

peterhuene (approved):
Reviewed the latest commits and those made just prior to my last review.

Looks good!

PR #838, opened 2026-05-07
feat: add resolver layer to `wdl-modules`

claymcleod:
Stacked on #836. Adds the `resolver` feature gate to `wdl-modules`, covering the `Resolver` trait, its Git-backed implementation (`GitResolver`), and the infrastructure for on-disk caching, version selection, lockfile generation, TOFU trust handling, and module materialization.

The resolver is the networking and process-spawning half of the module spec that `wdl-modules` deliberately excludes from its base feature set. The feature gate keeps `git2`, `tokio`, `globset`, `futures`, and friends out of consumers that only need manifest/lockfile/hashing types (e.g., `wdl-doc`). `GitResolver` is constructed via a `bon::Builder` so the binary owns the policy of where configuration lives and passes in the trust store, cache root, and `[modules]` config explicitly.

The trait surface has three methods. `resolve_tree` walks a consumer's dependency map recursively, materializing each dep into the on-disk cache via sparse checkout, hashing it, verifying its signature if present, and assembling a `ResolvedTree` that mirrors the lockfile shape. `materialize` resolves a single symbolic import (e.g., `import openwdl/csvkit/cut`) to a concrete file path, used by the analyzer at import time. `discover_versions` lists available semver tags for a dependency, used by CLI commands like `sprocket module versions`. All three are `#[async_trait]` for dyn-compatibility since the analyzer needs `Box<dyn Resolver>`.

Cycle detection threads a `Vec<(DependencyName, ResolvedSource)>` through the recursion, comparing resolved sources by equality. If the same `(url, commit)` or `(path)` appears twice in the chain, `ResolverError::Cycle` fires with the full path. The recursion uses `BoxFuture` from `futures` since async fn recursion requires boxing.

The `git` submodule wraps `git2` into composable pieces. `default_credentials` chains the user's Git credential helper, ssh-agent, and a fallback. `default_callbacks` and `default_fetch_options` build on it. `connect_remote` takes a `Direction` and returns a connected remote. `list_advertised_refs` sits on top and returns `(refname, oid_hex)` pairs. `clone_with_sparse_checkout` does a full clone then a path-filtered `CheckoutBuilder` (libgit2 does not honor porcelain sparse-checkout config, so we use `dry_run()` for the initial checkout and `CheckoutBuilder::path` for the filtered one). `ensure_materialized` is the top-level entry point: clone if the leaf doesn't exist, extend the sparse set if it does.

Version selection parses tag names through a `VersionTag` newtype (`FromStr` + `Display`) that handles both root tags (`v1.2.3`) and path-prefixed tags (`csvkit/v1.2.3`). `list_remote_refs` filters advertised refs to the `refs/tags/` namespace with annotated-tag peeling, returning a `RemoteRefs` type alias (`HashMap<String, GitCommit>`). `resolve_version_to_commit` picks the best match and returns both the `Version` and the `GitCommit` the tag points at so the caller can verify the manifest's version field agrees.

Git sub-paths are validated at parse time through a `GitModulePath` newtype that wraps `RelativePath` and additionally rejects `.` and the empty string. `DependencySource::Git { path }` is `Option<GitModulePath>` rather than `Option<PathBuf>`, so absolute paths, `..` segments, and other escape vectors are rejected before they reach the resolver. The resolver joins validated paths onto cache leaves and sparse-checkout targets without risk of directory traversal.

Both `resolve_tree` and `materialize` route through a shared `verify_materialized_dependency` helper that hashes the module root, applies the large-file warning, and reads/verifies `module.sig`. This prevents symbolic imports from bypassing the integrity policy that `resolve_tree` enforces. `materialize` additionally canonicalizes the resolved file path and rejects it if the canonical form escapes the canonical module root, guarding against symlink-based escapes even when the directory walk in `hash_directory` would also catch them.

The resolver enforces two transitive-dependency policies. First, `LocalPath` dependencies declared by a non-local (Git) parent are rejected to prevent a remote manifest from coercing reads of arbitrary local directories. Second, Git URL schemes are checked against a configurable policy before any `git2` call. The default permits `https` and `ssh` for top-level deps but only `https` for transitive deps, so a remote manifest cannot trigger SSH authentication against an attacker-controlled host. `file://` is blocked at both levels by default. `max_advertised_refs` (default 100,000) caps the number of refs the resolver accepts from a remote, bounding the work a malicious advertisement can cause.

Structured cache keys now include an 8-char URL hash suffix (`tasks-a1b2c3d4` instead of `tasks`) so URLs with shared leading path segments (e.g., `gitlab/x/y` and `gitlab/x/y/z`) produce distinct cache leaves. The human-readable `host/org/repo` prefix is preserved for easy navigation.

After a version selector resolves, the resolver compares the selected semver version against `module.json`'s `version` field. If they disagree, `ResolverError::TagManifestMismatch` fires. This prevents a malicious repository from publishing a `v2.0.0` tag whose manifest claims `0.0.1`.

`ResolvedSource::Git` gains an optional `path` field so the lockfile records which sub-path a dependency was locked against. `satisfies()` in `partial_relock` now compares paths and handles each selector type correctly: version requirements check against locked module versions, commit selectors compare SHAs directly, and tag/branch selectors always force re-resolution since they are mutable refs. `partial_relock` returns `Result` and errors when a consumer-declared dependency is absent from the freshly-resolved tree rather than silently dropping it. `LockfileDiff::compute` recurses through nested `LockedModule.dependencies` and records the full dep chain in `NewSigner.dep_chain` so transitive signer changes trigger the confirmation prompt.

`[modules]` is wired into `sprocket.toml` as a top-level config section. The `cache`, `git`, `types`, and `versions` submodules are `pub(crate)`. The public surface is `Resolver`, `GitResolver`, `ModulesConfig`, `ResolverError`, lock/relock result types, and trust-store types.

Full CLI commands (`sprocket module lock`, `update`, etc.) and `wdl-analysis` integration are deferred to follow-up PRs.

For all contributors:


crates/wdl-modules/src/resolver/config.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// URL schemes permitted for top-level Git dependencies. Defaults
+    /// to `["https", "ssh"]`.
+    #[serde(default = "default_top_level_schemes")]
+    pub allowed_schemes: Vec<String>,
+
+    /// URL schemes permitted for transitive Git dependencies. Defaults
+    /// to `["https"]` so remote manifests cannot silently trigger SSH
+    /// authentication against an attacker-controlled host.
+    #[serde(default = "default_transitive_schemes")]
+    pub allowed_transitive_schemes: Vec<String>,
```
claymcleod, line 46: This is something in particular I am curious about—is it a concern to allow transitive connections to unknown servers, especially since we allow the users to set their allowed hosts (though I doubt many will do that by default).

claymcleod: Additionally, we have a setting that you have to enable to allow credentials for transitive hosts _at all_. I'm starting to think allowing this particular setting to be configured is a bit much.

crates/wdl-modules/src/dependency/git_module_path.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn try_from(s: String) -> Result<Self, Self::Error> {
+        Self::try_from(PathBuf::from(s))
+    }
+}
+
+impl TryFrom<PathBuf> for GitModulePath {
+    type Error = GitModulePathError;
+
+    fn try_from(p: PathBuf) -> Result<Self, Self::Error> {
+        let s = p.to_str().ok_or(RelativePathError::NonUtf8)?.to_string();
```
peterhuene: Does this need the `to_string` call here as this allocation seems unnecessary? We'll alloc again as part of the `TryFrom<String>` implementation of `RelativePath` given how it only really works on `&Path`.

Personally, I'd refactor `RelativePath` such that it removes the `TryFrom<String>` and `TryFrom<PathBuf>` implementations, instead only implementing them for `&str` and `&Path` as the owned versions aren't useful since we'll always allocate as part of the conversion (i.e. the "sanitized" path).

Likewise, I'd move the implementation in `TryFrom<&Path>` to `FromStr` since it's really operating at the `String` level and the `TryFrom<&Path>` could just call the `FromStr` implementation after converting the `&Path` to `&str`. 

The same comment applies to these trait implementations of `GitModulePath`.

crates/wdl-modules/src/dependency/git_module_path.rs
peterhuene, line 1: Nit: I'm personally not a fan of underscored module names when something simply like `git` would do.

Additionally, I don't tend to like having 1:1 mapping between types and modules; instead, I'm of the opinion that modules should serve to group related types together.

crates/wdl-modules/src/dependency.rs
```diff
@@ -1,10 +1,13 @@
...
 
+mod git_module_path;
 mod name;
 mod source;
 
 pub use name::DependencyName;
 pub use name::DependencyNameError;
 pub use source::DependencySource;
 pub use source::DependencySourceError;
+pub use source::GitModulePath;
```
peterhuene, line 15: It's a little weird to me to re-export a re-export from another module. Perhaps we could just export them from their defining module (`git_module_path` in this case?).

crates/wdl-modules/src/dependency/source.rs
```diff
@@ -7,10 +7,10 @@ use serde::Serialize;
 use thiserror::Error;
 use url::Url;
 
+pub use super::git_module_path::GitModulePath;
+pub use super::git_module_path::GitModulePathError;
```
peterhuene: Nit: to my previous comment on the super module, it's a little strange for this module to export types from its sibling module.

crates/wdl-modules/src/relative_path.rs
```diff
@@ -76,7 +76,7 @@ impl AsRef<Path> for RelativePath {
 
 impl std::fmt::Display for RelativePath {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Display::fmt(&self.0, f)
+        write!(f, "{}", self.0)
```
peterhuene: Or alternatively (and without the macro machinery):

```suggestion
        self.0.fmt(f)
```

crates/wdl-modules/src/resolver.rs
```diff
@@ -0,0 +1,2186 @@
...
+        &self,
+        consumer: &Manifest,
+        path: &SymbolicPath,
+    ) -> Result<MaterializedFile, ResolverError>;
+
+    /// Resolves every transitive dependency declared by `consumer`.
+    ///
+    /// Walks the consumer's `dependencies` map, recurses into each dep's
+    /// own manifest, and records every module visited along the way.
+    /// Detects cycles.
```
peterhuene:
```suggestion
    ///   
    /// The implementation of this method is expected to return an error upon the detection of a cycle.
```

crates/wdl-modules/src/resolver.rs
```diff
@@ -0,0 +1,2186 @@
...
+    }
+
+    /// Returns the active trust store.
+    pub fn trust_store(&self) -> &TrustStore {
+        &self.trust
+    }
+
+    /// Returns a policy-enforcing Git fetcher.
+    fn fetcher(&self) -> GitFetcher {
+        GitFetcher::new(self.policy.clone())
```
peterhuene: Judging by the fields of `ResolverPolicy`, this seems like a pretty heavy-weight clone to perform for the construction of a dependent type.

~~Could `GitFetcher` take the policy by reference? That said, I haven't yet looked at the call sites for this method, though.~~

Upon seeing the call site, I think this field might justifiably be an `Arc<ResolverPolicy>`.

crates/wdl-modules/src/resolver.rs
```diff
@@ -0,0 +1,2186 @@
...
+        .await
+        // SAFETY: the closure performs only libgit2 work and
+        // does not panic; a `JoinError` would only fire on
+        // runtime shutdown.
+        .unwrap();
+
+        if let Err(err) = result {
+            if plan.leaf.starts_with(&self.cache_root)
+                && plan.leaf.exists()
+                && let Err(io_err) = std::fs::remove_dir_all(&plan.leaf)
```
peterhuene: Is this a safe operation for concurrent `sprocket` instances? On the surface it feels like it's missing acquiring an exclusive advisory lock.

crates/wdl-modules/src/resolver.rs
```diff
@@ -0,0 +1,2186 @@
...
+            module_root: plan.module_path.clone(),
+            cache_leaf: plan.leaf.clone(),
+        })
+    }
+
+    /// Recursively resolves a dependency map for `resolve_tree`.
+    ///
+    /// Each iteration: policy check, materialize, read manifest, cycle
+    /// check, verify, recurse into transitive deps, assemble result.
+    fn resolve_dependencies<'a>(
```
peterhuene, line 262: We should probably document why this method is returning a boxed future here (i.e. that it is recursive) given that it differs from the other methods.

crates/wdl-modules/src/resolver.rs
```diff
@@ -0,0 +1,2186 @@
...
+                        requirement,
+                        considered,
+                    },
+                })?;
+                Ok((Some(version), commit))
+            }
+            GitSelector::Tag(tag) => {
+                let dep = name.clone();
+                let url = url.clone();
+                let fetcher = self.fetcher();
```
peterhuene: I think we can probably move the outer `fetcher` into the closure?

```suggestion
```

crates/wdl-modules/src/resolver.rs
```diff
@@ -0,0 +1,2186 @@
...
+                            dep: name.manifest().to_string(),
+                            kind: GitRefKind::Tag,
+                            name: tag.clone(),
+                        })?;
+                Ok((None, commit))
+            }
+            GitSelector::Branch(branch) => {
+                let dep = name.clone();
+                let url = url.clone();
+                let fetcher = self.fetcher();
```
peterhuene: I think we can probably move the outer `fetcher` into the closure?

```suggestion
```

crates/wdl-modules/src/resolver.rs
```diff
@@ -0,0 +1,2186 @@
...
+    async fn plan_git_materialization(
+        &self,
+        name: &DependencyName,
+        url: &url::Url,
+        selector: &GitSelector,
+        path: &Option<GitModulePath>,
+        scope: DependencyScope,
+        mode: ResolutionMode,
+    ) -> Result<GitMaterializationPlan, ResolverError> {
+        let path_prefix = path.as_ref().map(GitModulePath::as_str).map(str::to_string);
```
peterhuene: We pass this as a reference below, do we need the `to_string` call here?

crates/wdl-modules/src/resolver.rs
```diff
@@ -0,0 +1,2186 @@
...
+            }
+            ResolutionMode::Fresh => {
+                self.resolve_git_selector(name, url, selector, path_prefix.as_deref(), scope)
+                    .await?
+            }
+        };
+
+        let key = CacheKey::from_git_url(url, &commit);
+        let leaf = key.absolute_path(&self.cache_root);
+        let sparse_path = path_prefix.clone().unwrap_or(".".to_string());
```
peterhuene: This always allocates due to the use of `unwrap_or`, so it otherwise should be using `unwrap_or_else`.

However, if we delay the allocation from above by not preemptively calling `to_string`, this would simply be `path_prefix.unwrap_or(".").to_string()`.

```suggestion
        let sparse_path = path_prefix.unwrap_or(".").to_string();
```

crates/wdl-modules/src/resolver/cache.rs
```diff
@@ -0,0 +1,285 @@
...
+        digest_hex: String,
+    },
+}
+
+impl CacheKey {
+    /// Derives a `CacheKey` from a Git URL and a commit SHA.
+    pub fn from_git_url(url: &Url, commit: &GitCommit) -> Self {
+        let prefix = match url.host_str() {
+            Some(host) => {
+                let segments: Vec<&str> = url.path().split('/').filter(|s| !s.is_empty()).collect();
```
peterhuene: Could we not `collect` here and instead perhaps just do a `splitn` rather than `split` with:

```rust
match (segments.next(), segments.next()) {
    (Some(org), Some(repo)) => {
        let repo = repo.strip_suffix(".git").unwrap_or(repo);
        let digest = hash_url(url);
        let repo_with_suffix = format!("{repo}-{}", &digest[..8]);
        PrefixKey::GitStructured {
            host: host.to_string(),
            org: org.to_string(),
            repo_with_suffix,
        },
    }
    _ => PrefixKey::GitOpaque {
        digest_hex: hash_url(url),
    }
}
```

crates/wdl-modules/src/resolver/config.rs
```diff
@@ -0,0 +1,424 @@
...
+        }
+        let bytes = s
+            .parse::<bytesize::ByteSize>()
+            .map_err(|_| LargeFileWarningError(s.to_string()))?
+            .as_u64();
+        Ok(Self::Threshold(bytes))
+    }
+}
+
+impl TryFrom<String> for LargeFileWarning {
```
peterhuene: Similar to other comments relating to trait impls, I don't feel like `TryFrom<String>` implementations that delegate to a `FromStr` implementation add anything other than potentially misleading API users that the ownership is being transferred instead of being dropped. I'd prefer callers simply call `parse` themselves instead to avoid cases where callers might clone the string just to convert it to this enum.

crates/wdl-modules/src/resolver/config.rs
```diff
@@ -0,0 +1,424 @@
...
+                || (v6.segments()[0] & 0xFE00) == 0xFC00
+                // fe80::/10 — link-local addresses
+                || (v6.segments()[0] & 0xFFC0) == 0xFE80
+        }
+    }
+}
+
+/// Threshold for the large-file warning emitted at sign- and fetch-time.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(into = "String", try_from = "String")]
```
peterhuene: I think `serde_with::DisplayFromStr` is a better alternative to using `into = "String", try_from = "String"`.

It's better to have `Display` and `FromStr` implementations on enums rather than `Into<String>` and `TryFrom<String>` impls.

crates/wdl-modules/src/resolver/policy.rs
```diff
@@ -0,0 +1,349 @@
+//! Network and resource enforcement policy for the module resolver.
+//!
+//! This module translates a [`ModulesConfig`](super::config::ModulesConfig)
+//! into a [`ResolverPolicy`] that is evaluated at fetch time. Each Git URL is
+//! checked against the policy before any network activity occurs: the URL
+//! scheme must appear in the configured allowlist, the hostname must not be on
```
peterhuene:
```suggestion
//! scheme must appear in the configured allow list, the hostname must not be on
```

crates/wdl-modules/src/resolver/policy.rs
```diff
@@ -0,0 +1,349 @@
...
+                return Err(ResolverError::GitHostPolicyViolation {
+                    dep: name.manifest().to_string(),
+                    url: url.to_string(),
+                    host: host.to_string(),
+                });
+            }
+            // Resolve the hostname to IP addresses and reject if any
+            // resolved address is non-public.
+            //
+            // Port 443 is passed only because `to_socket_addrs` requires
```
peterhuene: Nit: not that it matters much, but a value of `0` might better indicate that the port is insignificant for the purpose of the address lookup.

crates/wdl-modules/src/resolver/scope.rs
peterhuene, line 1: Does this need its own module given that it's only 23 lines?

crates/wdl-modules/src/dependency/name.rs
peterhuene, line 1: This module is also quite small and perhaps better off included in the parent?

crates/wdl-modules/src/lib.rs
```diff
@@ -34,7 +34,10 @@ pub mod hash;
 pub mod license;
 pub mod lockfile;
 pub mod manifest;
+pub mod module_walk;
 pub mod relative_path;
+#[cfg(feature = "resolver")]
+pub mod resolver;
```
peterhuene, line 40: If the module is public do we need to re-export so many of its types into the root?

crates/wdl-modules/src/resolver/config.rs
```diff
@@ -0,0 +1,424 @@
...
+}
+
+#[cfg(test)]
+use crate::resolver::scope::DependencyScope;
+
+#[cfg(test)]
+impl ModulesConfig {
+    /// Returns `true` if the given URL scheme is permitted for a
+    /// dependency at this level of the tree.
+    fn scheme_allowed(&self, scheme: &str, scope: DependencyScope) -> bool {
```
peterhuene: This method appears to be unused.

crates/wdl-modules/src/resolver/versions.rs
```diff
@@ -0,0 +1,371 @@
...
+pub fn parse_versions(refs: &RemoteRefs, path_prefix: Option<&str>) -> Vec<Version> {
+    let mut versions: Vec<Version> = refs
+        .keys()
+        .filter_map(|tag| tag.parse::<VersionTag>().ok())
+        .filter(|t| t.prefix() == path_prefix)
+        .map(VersionTag::into_version)
+        .collect();
+    versions.sort_by(|a, b| b.cmp(a));
+    versions
+}
```
peterhuene: I think this method should be refactored such that there's an internal method that returns `impl Iterator<Version>` instead of `Vec<Version>`.

Then `parse_versions` would call `collect` and `sort_by` returning the collection.

`filter_matching` would add the filter by matching requirement, call `collect` and `sort_by`, returning the collection.

Finally, `select_version` would call `find` on the iterator.

crates/wdl-modules/src/resolver/git.rs
```diff
@@ -0,0 +1,914 @@
...
+        count: usize,
+        /// The configured limit.
+        limit: usize,
+    },
+
+    /// A module subtree exceeds configured file or byte limits.
+    #[error(
+        "module subtree `{path}` exceeds tree limits (files: {files}, bytes: {bytes}, \
+         max_files: {}, max_bytes: {})",
+        max_files.map_or_else(|| "unlimited".to_string(), |v| v.to_string()),
```
peterhuene: Nit:

```suggestion
        max_files.map(|v| v.to_string()).as_deref().unwrap_or("unlimited"),
```

crates/wdl-modules/src/resolver/git.rs
```diff
@@ -0,0 +1,914 @@
...
+        max_bytes.map_or_else(|| "unlimited".to_string(), |v| v.to_string()),
```
peterhuene: Nit:

```suggestion
        max_bytes.map(|v| v.to_string()).as_deref().unwrap_or("unlimited"),
```

crates/wdl-modules/src/resolver/git.rs
```diff
@@ -0,0 +1,914 @@
...
+        path: parent.to_path_buf(),
+        source,
+    })?;
+    let name = leaf.file_name().unwrap().to_string_lossy();
+    let lock_path = parent.join(format!(".{name}{LOCK_EXT}"));
+    let file = File::create(&lock_path).map_err(|source| GitError::Io {
+        path: lock_path.clone(),
+        source,
+    })?;
+    file.lock().map_err(|source| GitError::Io {
```
peterhuene: Should we use `try_lock` to first detect that the lock will block and log the fact we're going to block waiting for the lock to be acquired? Generally a good idea to let users know why their sprocket might appear hung.

I do something to that effect in the file locks for the call cache.

peterhuene (commented):
This looks really great! I especially like the enforcement of various limits for the cloned modules and the implementation that tries to be as efficient as it can for materializing the module files locally.

I just have mostly minor feedback comments to go over.

Looking forward to landing this feature!

crates/wdl-modules/src/dependency/git_module_path.rs
claymcleod: Done. Validation moved into `FromStr`, `TryFrom<&Path>` reduced to a thin adapter, and the owned-`String`/`PathBuf` impls dropped on both types. Serde now goes through `serde_with::SerializeDisplay`/`DeserializeFromStr`. I swept the same pattern across the other validated newtypes too (`LicenseExpression`, `ContentHash`, `SymbolicPath`, `VersionRequirement`, `VerifyingKey`, `Signature`); `DependencyName` and `GitCommit` keep `TryFrom<String>` since they store the owned `String` directly rather than reparsing it.

crates/wdl-modules/src/resolver/config.rs
claymcleod: Dropped the `TryFrom<String>` impl; callers `parse` instead.

crates/wdl-modules/src/resolver/config.rs
claymcleod: Switched to the `serde_with::SerializeDisplay`/`DeserializeFromStr` derive macros, the type-level equivalent of `DisplayFromStr` for these standalone newtypes. Serde now goes through `Display`/`FromStr`.

crates/wdl-modules/src/lib.rs
claymcleod: Trimmed the crate-root re-exports to the primary entry points: the `Resolver` trait, `Manifest`, and `Lockfile`. Supporting types (errors, value newtypes, config, resolver result types) are reached through their defining module now.

crates/wdl-modules/src/resolver.rs
claymcleod: The exclusive advisory lock is acquired here. `ensure_materialized` calls `lock_cache_leaf` before the `leaf.exists()` check and the clone/extend (`resolver/git.rs`), so concurrent `sprocket` instances serialize on the per-leaf `.lock` file. A cache leaf is keyed by `(url, commit)` and its contents are immutable once written, so reads after the lock is released are safe. Happy to make the locking more visible at this call site if you'd prefer.

crates/wdl-modules/src/resolver/config.rs
claymcleod: Resolved by dropping the standalone `allow_transitive_credentials` toggle and gating transitive credentials on `allowed_transitive_hosts` instead. A transitive dependency may only be fetched from a host on that list, and credentials are presented only to those hosts, so a transitive manifest cannot direct the user's credentials at a host they have not vouched for. The list defaults to `["github.com", "gitlab.com"]`, and reaching a host outside it fails fast with a `GitHostNotAllowed` error that names the config key to add it to.

crates/wdl-modules/src/resolver/config.rs
```diff
@@ -0,0 +1,396 @@
...
+}
+
+/// Returns the default set of allowed hosts for transitive Git
+/// dependencies.
+///
+/// The two major public Git hosts are trusted by default so transitive
+/// dependencies resolve out of the box while still presenting
+/// credentials only to well-known hosts.
+fn default_allowed_transitive_hosts() -> Vec<String> {
+    vec!["github.com".into(), "gitlab.com".into()]
```
claymcleod, line 117: In particular, is this a good decision? I tend to think it will be annoying it you have to configure the standard source locations, but it does mean we're opting people in to sharing credentials when connecting to these hosts.

crates/wdl-modules/src/resolver.rs
peterhuene: I meant specifically that we're deleting the leaf directory not under the lock (as far as I can tell, at least) as `ensure_materialized` has already released it. That feels like a race with another call to `ensure_materialized` that has succeeded on its `leaf.exists()` check (under the lock) because it might be being deleted by this line.

peterhuene (approved):
Looks great! I just have that one additional comment regarding the clean-up of leaf directories on error relating to locking, but otherwise approved.

PR #839, opened 2026-05-09
fix(sprocket): cancel tests on CTRL+C

### LINKED ISSUE #837: dev test should cancel tasks on ctrl+c
**Describe the bug**
When exiting `sprocket dev test` with Ctrl+C, scheduled containers continue to run.

**Expected behavior**
Sprocket should cancel any scheduled containers.

**Terminal Output**
N/A

**Screenshots**
N/A

 - OS: MacOS
 - Version: 0.24.0
 - Shell: bash
 - Toolchain version: rustc 1.93.0-nightly (83e49b75e 2025-12-03)

**Additional context**
N/A

Serial-ATA:
closes #837 

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/src/commands/test.rs
+++ b/src/commands/test.rs
@@ -828,13 +828,13 @@ pub async fn test(
 
             _ = tokio::signal::ctrl_c() => {
                 if cancellation.state() == CancellationContextState::Canceling {
-                    continue
+                    return Err(anyhow!("evaluation was interrupted").into());
                 }
 
                 match cancellation.cancel() {
                     CancellationContextState::NotCanceled => unreachable!("should be canceled"),
                     CancellationContextState::Waiting | CancellationContextState::Canceling => {
-                        error!("evaluation was interrupted");
+                        error!("canceling tests: use Ctrl-C to immediately terminate Sprocket and skip cleanup");
                     },
                 }
             },
--- a/src/commands/test.rs
+++ b/src/commands/test.rs
@@ -21,8 +21,10 @@ use path_clean::PathClean;
 use regex::Regex;
 use serde_json::Value as JsonValue;
 use tokio::fs::remove_dir_all;
+use tokio::select;
 use tokio::task::JoinSet;
 use tracing::debug;
+use tracing::error;
 use tracing::info;
 use tracing::level_filters::LevelFilter;
 use wdl::analysis::AnalysisResult;
@@ -30,6 +32,7 @@ use wdl::ast::AstNode;
 use wdl::diagnostics::DiagnosticCounts;
 use wdl::diagnostics::emit_diagnostics;
 use wdl::engine::CancellationContext;
+use wdl::engine::CancellationContextState;
 use wdl::engine::EvaluatedTask;
 use wdl::engine::EvaluationError;
 use wdl::engine::EvaluationPath;
@@ -217,6 +220,7 @@ struct TestIteration {
     result: RunResult,
     assertions: Arc<ParsedAssertions>,
     run_dir: PathBuf,
+    cancellation: CancellationContext,
 }
 
 impl TestIteration {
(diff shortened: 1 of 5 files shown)


src/commands/test.rs
```diff
@@ -794,29 +800,62 @@ pub async fn test(
...
+                    continue
+                }
+
+                match cancellation.cancel() {
+                    CancellationContextState::NotCanceled => unreachable!("should be canceled"),
+                    CancellationContextState::Waiting | CancellationContextState::Canceling => {
+                        error!("evaluation was interrupted");
+                    },
+                }
+            },
```
claymcleod, line 874: The second Ctrl+C is silently swallowed here. The `run` command prints a message on first cancel and bails on the second. Let's match that pattern:

```suggestion
            _ = tokio::signal::ctrl_c() => {
                if cancellation.state() == CancellationContextState::Canceling {
                    return Err(anyhow!("testing was interrupted").into());
                }

                match cancellation.cancel() {
                    CancellationContextState::NotCanceled => unreachable!("should be canceled"),
                    CancellationContextState::Waiting => unreachable!("`FailureMode::Fast` does not use the `Waiting` state"),
                    CancellationContextState::Canceling => {
                        error!("canceling existing tasks: use Ctrl-C to immediately terminate Sprocket and skip directory cleanup");
                    },
                }
            },
```

src/commands/test.rs
```diff
@@ -794,29 +800,62 @@ pub async fn test(
...
+
+            res = &mut runner_task => {
+                drop(runner_task);
```
claymcleod, line 877: This `drop` is a no-op—the future has already resolved in the `select!` arm. We should just remove it.

```suggestion
```

claymcleod (changes requested):
Solid fix—the shared `CancellationContext` approach matches the existing `run` command pattern well. Two inline comments: the second Ctrl+C should escalate instead of being swallowed, and the `drop(runner_task)` is unnecessary.

src/commands/test.rs
Serial-ATA: The drop is necessary. The task mutably borrows `errors` which is also mutably borrowed in this arm

src/commands/test.rs
Serial-ATA: That's intentional. IMO `SIGINT` is an indication that you want the *tests* to stop, not the entire command. I don't think cleanup should be optional here, otherwise the user can just `SIGKILL`

claymcleod (approved):

adthrasher:
On my first attempt, it canceled the tasks, but I got an error when it was removing one of the test directories.

```
% sprocket dev test
^C2026-05-11T13:35:39.530249Z ERROR sprocket::commands::test: canceling tests: use Ctrl-C to immediately terminate Sprocket and skip cleanup
error: removing prior test dir: `/Users/athrashe/Library/CloudStorage/OneDrive-St.JudeChildrensResearchHospital/repo/workflows/test/runs/sort/works`

Caused by:
    Directory not empty (os error 66)
```

Otherwise, it seems to work as intended and addresses the issue.

Serial-ATA:
@adthrasher [`9c856d5` (this PR)](https://github.com/stjude-rust-labs/sprocket/pull/839/commits/9c856d5d5ed3619523c4c385a024247b4d1f37a8) should fix it. Can you try again? It isn't erroring for me anymore, but I wasn't able to consistently reproduce it before anyway

adthrasher:
> @adthrasher [`9c856d5` (this PR)](https://github.com/stjude-rust-labs/sprocket/pull/839/commits/9c856d5d5ed3619523c4c385a024247b4d1f37a8) should fix it. Can you try again? It isn't erroring for me anymore, but I wasn't able to consistently reproduce it before anyway

Similarly, I can't reproduce. It just so happened to occur on my first run, so I reported it.

PR #842, opened 2026-05-12
fix: honor `#@ except:` directives on ancestor nodes for `UnnecessaryFunctionCall`

claymcleod:
The `UnnecessaryFunctionCall` analysis rule checks `expr.inner().is_rule_excepted()`, which only examines preceding siblings of the expression node. When `#@ except: UnnecessaryFunctionCall` is placed on a containing declaration—the natural place to put it—the directive has no effect because the expression is a child of the declaration, not a sibling of the comment.

This changes the check to walk `expr.inner().ancestors()`, matching the pattern already used by `Diagnostics::exceptable_add` in `validation.rs:52`. A concrete case where this matters: the `giraffe.wdl` workflow in [vgteam/vg_wdl](https://github.com/vgteam/vg_wdl) uses an `if (false)` block to manufacture a `None`-valued `File?` (since WDL 1.0 has no `None` literal), and the `select_first` call inside is intentional dead code that cannot be simplified away.

The existing test case is extended to cover a `select_first` call on a declaration annotated with `#@ except: UnnecessaryFunctionCall`, confirming the warning is suppressed.

### DIFF
--- a/crates/wdl-analysis/src/types/v1.rs
+++ b/crates/wdl-analysis/src/types/v1.rs
@@ -1561,7 +1561,10 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
                         Ok(binding) => {
                             if let Some(severity) =
                                 self.context.diagnostics_config().unnecessary_function_call
-                                && !expr.inner().is_rule_excepted(UNNECESSARY_FUNCTION_CALL)
+                                && !expr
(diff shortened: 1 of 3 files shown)


peterhuene (approved):

peterhuene:
Needs a CHANGELOG entry but otherwise gtg.

PR #843, opened 2026-05-13
feat(wdl-lint): add `BashSetSyntax` lint rule

### LINKED ISSUE #316: Lint rule for ensuring `set -euo pipefail` or similar is set
Upgrading discussion #60 to an issue for visibility.

Repeating myself, but I would think there's an "off the shelf" solution we can pull for this. It might involve shelling out, but I think that would be ok. I just don't want us to get in the business of writing a Bash parser 😬

### LINKED ISSUE #708: New `wdl-lint` rule: `CommandShellSafety`
## Rule: `CommandShellSafety`

- **Tags**: `Correctness`
- **Applies to**: `command` blocks in all task definitions.
- **Related**: [sprocket#316](https://github.com/stjude-rust-labs/sprocket/issues/316)
- **RFC**: [sprocket#707](https://github.com/stjude-rust-labs/sprocket/discussions/707)

### Description

Require `set -euo pipefail` (or a configurable equivalent) as the first executable statement in every `command` block. Silent failures in bash are a very common source of incorrect results in bioinformatics pipelines.

### Configuration (`Sprocket.toml`)

```toml
[lint.rules.CommandShellSafety]
require = ["e", "u", "o pipefail"]
deny = []
```

Flags listed in `require` must be present in the `set` statement; flags in `deny` must not be present. Any flag not in either list is implicitly allowed. When this section is omitted, the default is `require = ["e", "u", "o pipefail"]` and `deny = []`.

### Good example

```wdl
command <<<
    set -euo pipefail
    samtools sort ~{input_bam} -o ~{sorted_bam}
>>>
```

### Bad example

```wdl
command <<<
    samtools sort ~{input_bam} -o ~{sorted_bam}
>>>
```

**Diagnostic**: missing `set -euo pipefail` as first statement in command block.

### Edge cases or exceptions

- Blank lines or comments before the `set` statement should still pass, as long as `set` appears before any executable statement.
- Additional flags beyond those in `require` are permitted unless explicitly listed in `deny` (e.g., `set -euxo pipefail` passes the default config because `x` is neither required nor denied).
- The [upstream issue (#316)](https://github.com/stjude-rust-labs/sprocket/issues/316) notes a concern about requiring this for single-command tasks. This RFC takes the position that always requiring it is correct: the cost is zero for single-command tasks (the flags are a no-op), and it prevents omission when a second command is added later. This sidesteps the bash parsing problem discussed in the issue.

### Implementation notes

In the AST, locate every `command` block node. Extract the text content and parse the first non-comment, non-blank line. Verify it is a `set` invocation, then extract the individual flags (single-char flags and `-o <option>` pairs). Check that every entry in `require` is present and no entry in `deny` is present. If the `set` statement is missing entirely, emit the diagnostic. Load `require` and `deny` from the workspace `Sprocket.toml`; fall back to the defaults when absent.

Serial-ATA:
Enforces a *minimum* set of options to the bash `set` builtin. By default it enforces `set -euo pipefail`, but can be configured to any combination of *relevant* options.

It also:

* Flags invalid `set` commands
* Flags interactive-only options
* Recognizes *unset* options (e.g., `set +e -uo pipefail`)

closes #708
closes #316

For external contributors:

For all contributors:

      is implemented.


claymcleod:
Since this is a new rule, we should work towards getting the test coverage higher here.

tests/cli/run/fallback-version/with-config/stderr
```diff
@@ -4,4 +4,4 @@ warning[UsingFallbackVersion]: unsupported WDL version `development`; interpreti
 1 │ version development
   │         ^^^^^^^^^^^ this version of WDL is not supported
 
-outputs were also written to `./out/runs/test/_TIMESTAMP_/outputs.json`
```
adthrasher, line 7: This needs to be switched back to the placeholder

crates/wdl-lint/tests/lints/bash-set-syntax/source.wdl
adthrasher, line 1: Should this trigger if someone does something like
```
set -euo pipefail
echo "Hello, World!"
set +e
echo "foo"
```

Serial-ATA: I don't think it's worthwhile to check multiple commands IMO. This is more of a reminder to keep your `set` flags consistent and placed at the top of all `command`s. If you go and unset options later on, you're actively sidestepping the lint and likely doing so intentionally anyway.

crates/wdl-lint/src/rules/bash_set_syntax.rs
```diff
@@ -0,0 +1,515 @@
...
+        // Since we know the command is `set` at the very least
+        let mut last_chunk_end = SET_COMMAND_NAME.len();
+
+        let opts_trimmed = opts.trim_start();
+        if opts_trimmed.is_empty() {
+            return (false, last_chunk_end);
+        }
+
+        let mut remaining_expected: HashSet<_> = self.expected_options.iter().copied().collect();
+        let mut chunks = opts_trimmed.split_whitespace();
```
claymcleod, line 182: This tokenizer treats shell control operators as part of the preceding word, so a valid prelude like `set -euo pipefail; echo hi` or `set -euo pipefail&& echo hi` gets parsed as `-o pipefail;`/`-o pipefail&&` and reported as bad. Could we split or stop parsing at shell control operators before validating the `set` options?

crates/wdl-lint/src/rules/bash_set_syntax.rs
```diff
@@ -0,0 +1,515 @@
...
+
+                    if INTERACTIVE_ONLY_SHORT.contains(&opt) {
+                        diagnostics.exceptable_add(
+                            interactive_only(opt_span, &format!("{mode}{opt}")),
+                            SyntaxElement::from(section.inner().clone()),
+                            &self.exceptable_nodes(),
+                        )
+                    }
+
+                    self.expected_options
```
claymcleod: We should reject unknown `set` options rather than ignoring them. As written, `set -euo pipefail -o nope` still clears all required options and passes, even though Bash rejects `-o nope`; validating every parsed short/long option against known `set` options would keep this lint from accepting invalid preludes.

crates/wdl-lint/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Added `HostPathLiterals` lint rule to flag absolute path defaults in `File` and `Directory` declarations ([#736](https://github.com/stjude-rust-labs/sprocket/pull/736)).
 * New lint rule `EmptyOutputs` to detect `task`s with no `output`s ([#846](https://github.com/stjude-rust-labs/sprocket/pull/846)).
+* Added `BashSetSyntax` lint rule to enforce consistent [bash set](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)
```
claymcleod: This entry is under `0.23.0`, which has already been released; that makes the changelog claim `BashSetSyntax` shipped in a past release. Let's move it to the `Unreleased` section instead.

crates/wdl-lint/tests/lints/bash-set-syntax/source.wdl
```diff
@@ -0,0 +1,140 @@
+#@ except: MetaSections, RequirementsSection, EmptyOutputs
+
+version 1.3
+
+task good {
+    command <<<
```
claymcleod, line 6: It might be worth adding two fixtures here that lock down important boundaries: braced `command {}` syntax, and a valid initial `set -euo pipefail` followed later by `set +e`. The latter documents the behavior from the open review thread so future parser changes do not accidentally turn the initial-prelude rule into a whole-command-state rule.

crates/wdl-lint/src/rules/bash_set_syntax.rs
```diff
@@ -0,0 +1,515 @@
...
+        "Bash has many silent failure cases, which can produce invalid results and be difficult to \
+        debug. The [set command](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) \
+        should be used in all `command` sections to enforce stricter behavior."
+    }
+
+    fn examples(&self) -> &'static [Example] {
+        &[Example {
+            negative: LabeledSnippet {
+                label: None,
+                snippet: r#"version 1.2
```
claymcleod: Let's use `version 1.3` in these examples so the rule docs match the current WDL version used by the new fixtures.

claymcleod (commented):
This looks good after these comments are addressed. Please pull in the latest `main`, and then we can merge.

claymcleod:
Thanks for addressing the parser edge cases here. Before this is merge-ready, please move the `BashSetSyntax` changelog entry out of the released `0.24.1` section and update the rule example to `version 1.3`.

For the `deny` side of #708, let's split that into follow-up work rather than block this PR on it. Could you open a separate issue for extending `BashSetSyntax` with a configurable deny-list for enabled `set` options, e.g., rejecting `set -euxo pipefail` when `xtrace` is denied?

Serial-ATA:
@claymcleod I don't see a deny list being particularly useful here. This is really just intended to enforce a minimum set of options for consistency. Do you have a usecase?

claymcleod:
The main use case I had in mind is site-level log hygiene around `xtrace`. Some environments may want to require `euo pipefail` while also disallowing `xtrace`, since `-x` prints expanded commands and arguments into task logs where secrets, signed URLs, or other sensitive command arguments could be exposed.

claymcleod (approved):

Issue #713, opened 2026-03-11
New `wdl-lint` rule: `ParameterizedResources`


claymcleod:
## Rule: `ParameterizedResources`

- **Tags**: `Portability`, `Correctness`
- **Applies to**: `cpu`, `memory`, and `disks` values in `runtime`/`requirements` sections.
- **RFC**: [sprocket#707](https://github.com/stjude-rust-labs/sprocket/discussions/707)

### Description

Flag bare literal values for `cpu`, `memory`, and `disks` in runtime/requirements sections. Encourage dynamic or parameterized values that callers can tune for their environment or that scale across retries.

Tasks that genuinely don't benefit from dynamic resources can suppress the diagnostic with an `#@ except` directive.

### Version-aware behavior

- **WDL 1.0**: flag bare literals, suggest referencing input parameters (callers have no other way to override).
- **WDL 1.1**: flag bare literals as a note (callers can override via input JSON, but dynamic scaling isn't possible).
- **WDL 1.2+**: flag bare literals, suggest using `task.attempt` for retry-aware scaling.

### Good examples

WDL 1.0/1.1 — parameterized via input:

```wdl
input {
    Int cpu = 4
    String memory = "8 GiB"
}

runtime {
    cpu: cpu
    memory: memory
}
```

WDL 1.2+ — dynamic retry-aware scaling:

```wdl
requirements {
    memory: if task.attempt == 0 then "8 GiB" else "~{8 * (task.attempt + 1)} GiB"
    cpu: 4
    disks: "20 GiB"
}
```

### Bad example

```wdl
runtime {
    cpu: 4
    memory: "8 GiB"
}
```

**Diagnostic**: runtime resource values are bare literals; consider parameterizing for caller tunability or retry-aware scaling.

### Edge cases or exceptions

- No other runtime/requirements/hints are checked by this rule.
- For WDL 1.2+, a bare literal for `cpu` or `disks` may be acceptable if only `memory` benefits from retry scaling. The rule flags but does not prescribe — the diagnostic suggests dynamic values, it doesn't require them. Tasks that don't benefit from dynamic resources can suppress the diagnostic with an `#@ except` directive.

### Implementation notes

[...]

PR #850, opened 2026-05-16
feat(wdl-lint): add `StructOverMap` rule

### LINKED ISSUE #717: New `wdl-lint` rule: `PreferStructOverMap`
## Summary

Add a new lint rule `PreferStructOverMap` that flags `Map[String, *]` types in declarations where a `Struct` would provide clearer semantics and better validation. This encourages use of WDL's type system for self-documenting interfaces.

## Details

- **Tags**: `Clarity`
- **Applies to**: all declarations (inputs, outputs, private) in tasks and workflows. WDL >= 1.1 only.
- **Default**: no-op (pedantic-only rule)
- **Pedantic**: emits `warning`

### Good example

```wdl
struct SampleInfo {
    String sample_id
    String library
    File fastq_r1
    File fastq_r2
}

task align {
    input {
        SampleInfo sample
        File reference
    }
    # ...
}
```

### Bad example

```wdl
task align {
    input {
        # lint: PreferStructOverMap — consider using a Struct instead of Map[String, String]
        Map[String, String] sample_info
        # lint: PreferStructOverMap — consider using a Struct instead of Map[String, File]
        Map[String, File] input_files
        File reference
    }
    # ...
}
```

### Edge cases

- Legitimate uses of `Map[String, *]` exist (e.g., arbitrary key-value metadata, environment variables, user-defined annotations where keys are not known at authoring time). Authors should suppress the diagnostic with `#@ except: PreferStructOverMap`.
- Only `Map` types with a `String` key type are flagged. `Map[Int, String]` or `Map[File, String]` are not flagged.
- Nested occurrences (e.g., `Array[Map[String, File]]`) are also flagged.
- Version-gated to WDL >= 1.1 since structs were introduced in that version.

### Implementation notes

Visit all `BoundDecl` and `UnboundDecl` nodes. Inspect the declared type: if it is a `Map` type whose key type is `String`, emit a diagnostic. Only emit when pedantic mode is enabled. Skip documents with WDL version < 1.1.

## Context

Part of the lint rule set proposed in https://github.com/stjude-rust-labs/sprocket/discussions/707.

Serial-ATA:
closes #717 

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lint/src/rules/struct_over_map.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    fn exceptable_nodes(&self) -> Option<&'static [SyntaxKind]> {
+        Some(&[
+            SyntaxKind::VersionStatementNode,
+            SyntaxKind::StructDefinitionNode,
+            SyntaxKind::TaskDefinitionNode,
+            SyntaxKind::WorkflowDefinitionNode,
+        ])
+    }
```
adthrasher, line 131: I feel like this is the type of lint you'd want to be able to disable at the declaration level as well.

crates/wdl-lint/src/rules/struct_over_map.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+"#,
+            },
+            revised: Some(LabeledSnippet {
+                label: Some("If possible, consider switching to a `struct`"),
+                snippet: r#"version 1.3
+
+struct Config {
+    Boolean verbose
+    Boolean dryRun
```
adthrasher: This should be snake_case to avoid tripping that lint rule.

crates/wdl-lint/tests/lints/struct-over-map/source.wdl
adthrasher, line 1: I can't recall how we handle `optional` declarations, but the tests should probably include that case as well.

crates/wdl-lint/src/rules/struct_over_map.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    fn description(&self) -> &'static str {
+        "Checks for usage of `Map[String, ...]` types."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "In cases where arbitrary key-value storage isn't necessary, `struct`s can be leveraged to \
+         provide clearer semantics and better validation.\n\nOf course, there are many valid use \
+         cases for `Map[String, ...]` (e.g. environment variables). Expect many false positives."
```
adthrasher, line 61: `Expect many false positives`

To me, this would be an odd thing to read in a lint rule explanation for a rule that is enabled by default.  Curious what @claymcleod thinks here.

Serial-ATA: I have the same warning in the description of #713, though that one is much more common with over 100 occurrences across the 2 arena repos.

As I suggested in https://github.com/stjude-rust-labs/sprocket/discussions/707#discussioncomment-16276647, these rules could be behind an allowed-by-default `Pedantic` tag. I wouldn't push to get these in until there's something like that in place

crates/wdl-lint/src/rules/struct_over_map.rs
Serial-ATA: oops, yeah

PR #858, opened 2026-05-18
feat(wdl-analysis)!: add `MeaninglessLintDirective` rule

### LINKED ISSUE #304: new diagnostic for meaningless lint directive
Lint directives in the preamble should **not** fire this diagnostic. This is to support the "boilerplate" use case where someone just wants to blanket `except` certain rules for any WDL document.

Other lint directives that **don't actually suppress anything** should emit a `note`. Detecting this will involve tracking what rules are being called with `exceptable_add()` and then looking through the CST.

Serial-ATA:
`Validator` now keeps track of all `#@ except:` comments in the document, marking the ones that are *actually* applied.

`Validator` is also now aware of all rules it may emit, so `KnownRules` was promoted to `wdl-analysis`.

This is noisy as it also moves the other analysis rules to use `Diagnostics` and `Diagnostics::exceptable_add()`, rather than `Vec<Diagnostic>`

closes #304

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/src/validation.rs
```diff
@@ -25,58 +40,158 @@ mod requirements;
...
+    type IntoIter = std::vec::IntoIter<Self::Item>;
+    type Item = Diagnostic;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.diagnostics.into_iter()
+    }
+}
+
+impl Default for Diagnostics {
+    fn default() -> Self {
```
claymcleod: This can just be derived. And probably `new()` can go away in favor of `default()`.

crates/wdl-analysis/src/validation.rs
```diff
@@ -89,68 +204,135 @@ impl Diagnostics {
...
+
+    /// Catch any unapplied lint exceptions.
+    fn check_meaningless_lint_directives(&self, diagnostics: &mut Diagnostics, severity: Severity) {
+        let mut meaningless_lint_directives = Diagnostics::new();
+
+        let visitor_known_rules = self.known_rules();
+        let invalid_directives = diagnostics
+            .iter()
+            .filter_map(|d| {
+                if d.rule() == Some("ExceptDirectiveValid") {
```
claymcleod, line 238: This would probably benefit from a comment explaining what you're doing here. I was able to work it out after a little bit of time, but would be helpful to have that information up front.

crates/wdl-analysis/tests/analysis/unused-import/source.diagnostics
```diff
@@ -4,3 +4,9 @@ warning[UnusedImport]: unused import namespace `bar`
 8 │ import "bar.wdl"
   │        ^^^^^^^^^
 
+note[MeaninglessLintDirective]: unnecessary `except` directive for lint rule `UnusedImport`
```
claymcleod: This fixture codifies a false positive. The `ok` namespace from `import "bar.wdl" as ok` is genuinely unused, so the `#@ except: UnusedImport` is doing its job by suppressing that warning, yet it gets flagged as meaningless.

The cause is that unused-import diagnostics are still emitted through `analysis_diagnostics.extend(...)` filtered on `Namespace::excepted` in `document.rs`, not through `exceptable_add`. That path never marks the exception applied in the ledger, so `check_meaningless_lint_directives` treats the directive as unused. Consider routing the unused-import emission through `exceptable_add` against the import node and retiring the `Namespace::excepted` field once it has no other consumers; this note should then disappear from the expected output.

crates/wdl-analysis/src/validation.rs
claymcleod, line 238: A few more things:

- The rule is matched with the bare string `"ExceptDirectiveValid"`. Using that rule's `ID` constant instead would keep this in sync if the rule is ever renamed.

claymcleod (changes requested):
I was in the middle of reviewing this but was not able to finish. Thus, I'm just adding my partial comments here for consideration.

crates/wdl-analysis/src/validation.rs
Serial-ATA: `ExceptDirectiveValid` is from `wdl-lint`, this is `wdl-analysis` so that constant isn't available

crates/wdl-analysis/src/rules.rs
```diff
@@ -480,6 +545,114 @@ workflow example {
...
+    fn deny(&mut self) {
+        self.0 = Severity::Error;
+    }
+
+    fn severity(&self) -> Severity {
+        self.0
+    }
+}
+
+/// Represents the using fallback version rule.
```
adthrasher: This looks like a copy/paste error.

crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -893,6 +895,16 @@ pub fn unnecessary_function_call(
         .with_label(label.to_string(), label_span)
 }
 
+/// Creates an "unnecessary function call" diagnostic.
```
adthrasher: Copy/paste error?

crates/wdl-grammar/src/diagnostic.rs
```diff
@@ -50,6 +49,20 @@ impl Span {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use wdl_grammar::Span;
+    /// let parent = Span::new(0, 10);
+    /// let child = Span::new(5, 5);
+    /// assert!(child.within(parent));
+    /// ```
+    pub fn within(&self, other: Self) -> bool {
```
adthrasher, line 76: Is this used anywhere?

crates/wdl-analysis/src/rules.rs
```diff
@@ -558,6 +737,83 @@ task greet {
...
+
+workflow example {
+}
+"#,
+            }),
+        }]
+    }
+
+    fn exceptable_nodes(&self) -> Option<&'static [wdl_ast::SyntaxKind]> {
+        Some(&[SyntaxKind::VersionStatementNode])
```
adthrasher: Should this be `Self::EXCEPTABLE_NODES`?

crates/wdl-grammar/src/diagnostic.rs
Serial-ATA: Not anymore, but it's still probably useful to have. Might be misremembering, but I may even have another branch that I copied this into

crates/wdl-analysis/src/validation.rs
```diff
@@ -25,58 +40,144 @@ mod requirements;
...
+    pub(crate) diagnostics: Vec<Diagnostic>,
+    /// `#@ except:` directives discovered during traversal.
+    ///
+    /// `HashMap<Rule, applied>`
+    exceptions: HashMap<ExceptRule, bool>,
+}
 
 impl Diagnostics {
     /// Adds a diagnostic to the collection.
     pub fn add(&mut self, diagnostic: Diagnostic) {
```
adthrasher, line 80: Should this function get a stronger comment to explain that `exceptable_add` is probably preferred?

crates/wdl-analysis/src/validation.rs
```diff
@@ -89,68 +190,144 @@ impl Diagnostics {
...
+            self.visitors.push(visitor);
+        }
+    }
+
+    /// Adds rule names to the validator's known rules set.
+    pub fn extend_known_rules(&mut self, rules: impl IntoIterator<Item = String>) {
+        self.known_rules.extend(rules);
+    }
+
+    /// Catch any unapplied lint exceptions.
```
adthrasher, line 209: I think this should include a description of how the function works.

adthrasher (approved):

crates/wdl-analysis/src/validation.rs
```diff
@@ -182,6 +371,37 @@ impl Visitor for Validator {
     }
 
     fn comment(&mut self, diagnostics: &mut Diagnostics, comment: &Comment) {
+        if let Some(Directive::Except(except)) = comment.directive() {
+            for rule in except {
+                if self.known_rules.contains(&rule.name) {
+                    continue;
+                }
+
+                let diagnostic =
```
claymcleod: This bypasses `DiagnosticsConfig::known_rules`, so users cannot disable or deny `KnownRules` even though `DiagnosticsConfig` now has a `known_rules: Option<Severity>` field. Could we gate this on `document.config().diagnostics_config().known_rules` and apply that severity before adding the diagnostic?

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -189,22 +193,41 @@ pub(crate) fn populate_document(
...
     // declarations might reference an imported or locally-defined struct or enum
+    let mut import_nodes_by_namespace = HashMap::new();
     for item in ast.items() {
         match item {
-            DocumentItem::Import(import) => match import.form() {
-                ImportForm::Namespace => {
-                    add_namespace(document, graph, &import, index);
+            DocumentItem::Import(import) => {
+                if let Some((ns, _span)) = import.namespace() {
+                    import_nodes_by_namespace.insert(ns, import.inner().clone());
```
claymcleod: This map can overwrite the node for a successfully inserted namespace when a later duplicate import uses the same namespace. The unused-import pass below iterates `document.namespaces`, which keeps the first successful import, but then looks up the node here and may attach suppression handling to the duplicate/conflicting import instead; using `entry(...).or_insert_with(...)`, or inserting only after `add_namespace` succeeds, would keep the node aligned with the namespace being checked.

tests/cli/explain/tag/Clarity/stdout
```diff
@@ -9,7 +9,6 @@ Rules with the tag `Clarity`:
   - ExceptDirectiveValid
   - HereDocCommands
   - ImportPlacement
-  - KnownRules
   - MetaSections
```
claymcleod, line 12: `KnownRules` is still listed by `sprocket explain --help`, but this removes it from every tag listing because analysis rules do not carry tags. Could we either preserve its previous tags in the explain metadata or intentionally remove it from `--help`/document that analysis rules are not tag-addressable? Otherwise `explain --tag Clarity` no longer helps users discover this rule even though it still exists.

crates/wdl-analysis/src/validation.rs
```diff
@@ -89,68 +193,153 @@ impl Diagnostics {
...
+            if *applied
+                // Try not to clash with `ExceptDirectiveValid`
+                || invalid_directives.iter().any(|label| label.span() == exception.span)
+                // If none of the visitors know the rule, it can't ever fire
+                || (!ALL_RULE_IDS.iter().any(|r| r == &exception.name) && !visitor_known_rules.contains(&exception.name))
+            {
+                continue;
+            }
+
+            meaningless_lint_directives.add(meaningless_lint_directive(
```
claymcleod: This uses `add`, so `MeaninglessLintDirective` itself cannot be suppressed even though the rule defines `EXCEPTABLE_NODES` and participates in `DiagnosticsConfig`. For example, `#@ except: MeaninglessLintDirective` before the version statement would still be reported as meaningless. Could we route these diagnostics through `exceptable_add` against the node the directive applies to, or make the rule explicitly unsuppressible and remove the exceptable/config surface?

claymcleod (commented):
The overall direction looks right, but I left a few comments on `KnownRules` configuration/discoverability and exception tracking for unused or meaningless directives.

tests/cli/explain/tag/Clarity/stdout
Serial-ATA: It's not different from any other analysis rule though. I don't see why we would make a special case for `KnownRules`?

adthrasher: I looked at this previously and came to the same conclusion as Alex. `KnownRules` loses its `clarity` tag, but that is consistent with the analysis rules.

crates/wdl-analysis/src/validation.rs
Serial-ATA: good catch, made it exceptable on all nodes

claymcleod (approved):
Good job!

PR #866, opened 2026-05-21
fix(wdl-analysis): allow enum types for struct members

### LINKED ISSUE #825: wdl-analysis: struct members can't be enum
**Describe the bug**
The following doesn't pass type checking:

```wdl
version 1.3

enum Foo {
    Bar
}

struct Baz {
    Foo foo
}
```

Since resolution only checks against other structs:

https://github.com/stjude-rust-labs/sprocket/blob/85d3943b70a3c37b6c5f9e0b8b7221b50f656258/crates/wdl-analysis/src/document/v1.rs#L1725-L1745

**Expected behavior**
The example should work.

**Terminal Output**

```bash
error: unknown type name `Foo`
  ┌─ tmp.wdl:8:5
  │
8 │     Foo foo
  │     ^^^
```

Serial-ATA:
closes #825

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/src/diagnostics.rs
```diff
@@ -395,6 +395,16 @@ pub fn recursive_struct(name: &str, span: Span, member: Span) -> Diagnostic {
...
 
+/// Creates a "recursive enum" diagnostic.
+pub fn recursive_enum(name: &str, span: Span, ty: wdl_ast::v1::Type) -> Diagnostic {
+    Diagnostic::error(format!("enum `{name}` has a recursive definition"))
+        .with_highlight(span)
+        .with_help(format!(
+            "the type `{}` participates in the recursion",
+            ty.inner().text()
+        ))
+}
```
claymcleod: The sibling `recursive_struct` right above this labels the offending member's span with `.with_label`, so the diagnostic points at the exact spot in source. `recursive_enum` instead reprints the type via `.with_help` and `ty.inner().text()`, which points nowhere in-source and, for a compound type parameter like `enum Foo[Array[Bar]]`, reports `Array[Bar]` rather than the recursive `Bar`. Consider taking the offending reference's `Span` and using `.with_label` so the two diagnostics behave consistently.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1805,141 +1805,176 @@ fn set_struct_types(document: &mut DocumentData) {
...
+                    && document.structs[to].namespace.is_none()
+                {
+                    TypeIndex::Struct(to)
+                } else if let Some(to) = document.enums.get_index_of(name.text())
+                    && document.enums[to].namespace.is_none()
+                {
+                    TypeIndex::Enum(to)
+                } else {
+                    continue;
+                };
```
claymcleod: This struct-then-enum lookup with the `namespace.is_none()` guard is duplicated verbatim in the enum loop below. It might be worth pulling it into a small `fn resolve_dep(document, name) -> Option<TypeIndex>` so both loops share one definition and the local-definition rule lives in a single place.

crates/wdl-analysis/tests/analysis/enum-struct-cycles/source.wdl
```diff
@@ -0,0 +1,7 @@
+version 1.3
```
claymcleod, line 1: A few paths go untested: a pure `enum`-> `enum` cycle (two enums whose type parameters reference each other), the non-cyclic `enum` -> `struct` direction, and an `enum`/`struct` nested inside a compound member type like `Array[Foo]` or `Map[String, Foo]`, since `find_type_refs` descends into those. I think a couple more fixtures here would be worthwhile.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1756,32 +1754,34 @@ fn set_struct_types(document: &mut DocumentData) {
...
 
-    if document.structs.is_empty() {
+    if document.structs.is_empty() && document.enums.is_empty() {
         return;
     }
 
-    /// Recursively finds all nested struct type dependencies to build
-    /// dependency graphs
+    /// Recursively finds all nested type dependencies to build dependency
+    /// graphs
```
claymcleod, line 1780: Leftover plural from the two-pass version; one graph now.

```suggestion
    /// Recursively finds all nested type dependencies to build a dependency
    /// graph
```

claymcleod (changes requested):

crates/wdl-analysis/src/diagnostics.rs
Serial-ATA: Fixed printing the whole type:

```
error: enum `CyclicEnum4` has a recursive definition
   ┌─ tests/analysis/enum-struct-cycles/source.wdl:20:6
   │
20 │ enum CyclicEnum4[Array[CyclicEnum3]] {
   │      ^^^^^^^^^^^
   │
   = help: the type `CyclicEnum3` participates in the recursion
```

I'm not sure what you mean by it "points nowhere in-source". The `struct` diagnostic labels specific members since one recursive member doesn't imply that they all are, unlike an `enum` which is entirely invalid

claymcleod: The idea is that it's just the name of the offending type and not a label pointing to it in source. Since this is so similar to the `recursive_struct` in functionality, we should mirror the two.

crates/wdl-analysis/tests/analysis/enum-struct-cycles/source.wdl
```diff
@@ -0,0 +1,36 @@
...
+struct CyclicStruct3 {
+    CyclicEnum5 cycle
+}
+
+# Non-cyclic
+enum NonCyclicEnum {}
+
+struct NonCyclicStruct {
+    NonCyclicEnum non_cyclic
+}
```
adthrasher: This should get a newline for consistency.

claymcleod (approved):

PR #867, opened 2026-05-21
feat(wdl-lsp)!: `textDocument/foldingRange` support

Serial-ATA:
Makes it possible to collapse:

* `task`s/`workflow`s and their individual sections
* imports
* comment blocks

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lsp/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 0.19.1 - 2026-05-14
 
+### Added
```
claymcleod, line 10: This should be added under unreleased.

crates/wdl-ast/src/lib.rs
```diff
@@ -90,6 +91,45 @@ pub fn doc_comments<N: TreeNode>(
...
+        .inner()
+        .siblings_with_tokens(Direction::Next)
+        .find_map(|sibling| {
+            let SyntaxElement::Token(token) = sibling else {
+                return None;
+            };
+
+            CloseDelimiter::cast(token)
+        })
+        .expect("should have a close brace");
```
claymcleod: Two issues. First, both `.expect(...)` calls panic on incomplete parses (missing brace), which is reachable on every keystroke through the LSP folding-range handler. Let's return `Option<(OpenDelimiter, CloseDelimiter)>` from `get_block` and propagate `Option<TextRange>` through `HasBlock::block_range`. Second, the messages say "brace" but the helper is parameterized on `OpenDelimiter`/`CloseDelimiter` and is called with heredoc tokens for `CommandSection`. Either make them delimiter-agnostic or drop the messages when returning `Option`.

crates/wdl-analysis/src/handlers/folding_range.rs
```diff
@@ -0,0 +1,235 @@
...
+                range = hints.block_range();
+            }
+            SyntaxKind::RuntimeSectionNode => {
+                let runtime = RuntimeSection::cast(node).expect("should cast");
+                range = runtime.block_range();
+            }
+            SyntaxKind::PlaceholderNode => {
+                let placeholder = Placeholder::cast(node).expect("should cast");
+                range = placeholder.expr().inner().text_range();
+            }
```
claymcleod: Single-line placeholders (the common case for `~{x}` or `~{sep=" " arr}`) produce one-line folds here that waste protocol bytes and can render as a confusing fold marker. I think we should drop any range whose `start_pos.line == end_pos.line` at the top of `folding_range()` before pushing, which handles this branch and any future arm with the same shape.

crates/wdl-analysis/src/queue.rs
```diff
@@ -138,6 +141,14 @@ pub struct NotifyChangeRequest {
     pub discard_pending: bool,
 }
 
+/// Represents a request to get all folding ranges in a document.
+pub struct FoldingRangeRequest {
+    /// The document to be formatted.
```
claymcleod:
```suggestion
    /// The document to get folding ranges for.
```

crates/wdl-analysis/src/handlers/folding_range.rs
```diff
@@ -0,0 +1,235 @@
...
+        let end_pos = position(&lines, range.end())?;
+
+        ranges.push(FoldingRange {
+            start_line: start_pos.line,
+            start_character: Some(start_pos.character),
+            end_line: end_pos.line,
+            end_character: Some(end_pos.character),
+            kind,
+            collapsed_text: None,
+        });
```
claymcleod, line 89: The fold here ends on the line containing the closing `}` (or `>>>`), so that line is hidden when the region is collapsed (clients hide `start_line + 1` through `end_line`). The result is `task foo { ...` with no closing-brace cue visible. The convention in polished LSPs (rust-analyzer is the reference) is to stop one line earlier so the brace stays visible. Could be fixed here by decrementing `end_line` for block-style folds (everything except `kind == Some(FoldingRangeKind::Comment)` and `Some(FoldingRangeKind::Imports)`), or upstream in `block_range()` by returning the range that ends at the line before the close-delimiter token.

crates/wdl-lsp/tests/workspace/folding_range/source.wdl
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    parameter_meta {
+        unused: "So is the parameter meta"
+    }
+
+    command <<<
+        echo "Commands too"
+    >>>
+}
```
claymcleod, line 53: The fixture covers about a third of the new behavior. It skips `InputSection`, `OutputSection`, `RuntimeSection`, `RequirementsSection`, both `*HintsSection` variants, a full `WorkflowDefinition`, the braced `command { ... }` branch of `CommandSection::block_range`, every `Placeholder`, comments inside a task body, a trailing/EOF comment, multi-group separated imports, and directive-comment groups (`#@`). The braced-command and `Placeholder` branches are particularly worth covering because they have distinct code paths in `block_range` and `determine_folding_range`. Let's extend `source.wdl` (or add a second fixture) to exercise these.

crates/wdl-analysis/src/analyzer.rs
```diff
@@ -620,6 +622,28 @@ where
...
+                    "failed to send folding range request to the queue because the channel has \
+                     closed"
+                )
+            })?;
+
+        rx.await.map_err(|_| {
+            anyhow!(
+                "failed to send folding range request to the queue because the channel has closed"
+            )
+        })
```
claymcleod, line 681:
```suggestion
        rx.await.map_err(|_| {
            anyhow!(
                "failed to receive folding range response from the queue because the channel has \
                 closed"
            )
        })
```

crates/wdl-ast/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Changed
+
+* Replaced `Comment::is_{directive,doc_comment}` with `Comment::Kind` ([#867](https://github.com/stjude-rust-labs/sprocket/pull/867)).
```
claymcleod: The new enum is `CommentKind`, not `Comment::Kind`. Also worth adding an `#### Added` section above this one for the new public API: `CommentKind` enum, `HasBlock` trait, and `Comment::kind()` accessor.

claymcleod (changes requested):
A few more things.

`HasBlock` and `get_block` reinvent the existing `AstNode::scope_span`/`braced_scope_span`/`heredoc_scope_span` family at `crates/wdl-ast/src/lib.rs:293-325`. Worth checking whether the existing primitives can be extended (or wrapped) to cover the inclusive-range case the folding handler wants, rather than introducing a parallel mechanism.

The `Request::FoldingRange` dispatch arm in `queue.rs` runs the handler inline. The analyze-node path wraps in `panic::catch_unwind(AssertUnwindSafe(...))` at `queue.rs:959` precisely so a panic does not tear down the worker. A panic in the folding-range handler (any of the eleven `expect("should cast")` calls, or `get_block` on an incomplete brace) would close the queue and break LSP for the session. Same five-line wrapper would isolate this path.

Serial-ATA:
Didn't know `braced_scope_span()` existed already... that simplifies things

crates/wdl-analysis/src/handlers/folding_range.rs
Serial-ATA: This was mimicking the behavior of RustRover, which I personally prefer:

image

image

Though I did forget to set the `collapsed_text` to show brackets, which looks like this:

image

claymcleod (approved):

PR #869, opened 2026-05-22
fix(wdl-engine): reopen the call cache lock file read-only before locking

### LINKED ISSUE #862: Call cache lock fails on shared HPC filesystem
**Describe the bug**
I was trying out the task call caching functionality on a HPC Linux node with:
```toml
[run.task]
cache = "on"
cache_dir = "sprocket_call_cache"
```
However `sprocket run` failed before workflow evaluation starts when task call caching is enabled.

Looking at the code path: https://github.com/stjude-rust-labs/sprocket/blob/df9a5275ae5d0f543047414792dfb58de733bac1/crates/wdl-engine/src/cache/lock.rs#L28-L39

the lock file is opened write-only, which is then used for a shared lock:

https://github.com/stjude-rust-labs/sprocket/blob/df9a5275ae5d0f543047414792dfb58de733bac1/crates/wdl-engine/src/cache/lock.rs#L61

Adding `.read(true)` in a local branch fixed the issue:
```rust
options.create(true).read(true).write(true);
```

**Expected behavior**
When call caching is enabled, Sprocket should create/open the call-cache .lock file and acquire the
shared lock successfully, then continue workflow evaluation.

**Terminal Output**
```bash
2026-05-20T08:54:05.398286Z  INFO wdl_engine::cache: using call cache directory `sprocket_call_cache`
error: failed to create workflow evaluator
Caused by:
    0: failed to acquire shared lock on cache entry file `sprocket_call_cache/.lock`
    1: Bad file descriptor (os error 9)
```

**Environment:**
- OS: Linux 
- Version: 0.25.0
- Shell: bash

tmokveld:
Fix #862

With `[run.task] cache = "on"`, `sprocket run` fails before workflow evaluation starts on NFS-backed filesystems:

```bash
error: failed to create workflow evaluator
Caused by:
    0: failed to acquire shared lock on cache entry file `sprocket_call_cache/.lock`
    1: Bad file descriptor (os error 9)
```

`acquire_shared(..., create: true)` created the global call-cache lock file with a write-only handle and then attempted to acquire a shared lock on that same handle. Apparently this works on some local filesystems, but can fail on NFS.

Fix:
When the lock file is missing, create it with `create_new(true)` using a temporary write handle and also deal with `AlreadyExists` in case of concurrent creators. The write handle is then explicitly dropped and the file is reopened read-only before acquiring the shared lock.

The `acquire_shared` test was renamed to `acquire_shared_existing_file ` to make it more explicit that an existing file is used, where the missing-file code-path is not followed. Added the `acquire_shared_creates_file_returns_readable_lock ` test to actually test the missing-file code-path.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Fixed
-
-* Fixed shared lock acquisition to reopen new lock files read-only before locking.
-
 ## 0.25.0 - 2026-05-14
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* Fixed shared lock acquisition to reopen new lock files read-only before locking.
+
 ## 0.25.0 - 2026-05-14
 
 ### Added
@@ -108,7 +112,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Increased SQLite `busy_timeout` from 5s to 30s and added retry with
   exponential backoff when opening the database
   ([#734](https://github.com/stjude-rust-labs/sprocket/pull/734)).
-* Fixed a bug where the `format`, `run`, `lock` and `inputs` commands would not 
+* Fixed a bug where the `format`, `run`, `lock` and `inputs` commands would not
   utilize the configured `fallback_version`
   ([#784](https://github.com/stjude-rust-labs/sprocket/pull/784)).
 
@@ -270,7 +274,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Apptainer-based backends now store converted container images within each run directory, rather than in a user-specified directory ([#463](https://github.com/stjude-rust-labs/sprocket/pull/463)).
 * `sprocket run` now writes a `.sprocketignore` file directly to the `runs/` directory instead of the `runs/<entrypoint>/` directory ([#481](https://github.com/stjude-rust-labs/sprocket/pull/481)).
 
-
 ### Fixed
 
 * Fixed a bug in `sprocket config init` where `sprocket.toml` was unnecessarily loaded and would fail if malformed ([#473](https://github.com/stjude-rust-labs/sprocket/pull/473)).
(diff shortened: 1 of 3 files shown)


crates/wdl-engine/src/cache/lock.rs
```diff
@@ -28,15 +28,41 @@ impl LockedFile {
...
-            // Create or open the file, but do not truncate it if it exists
-            let mut options = fs::OpenOptions::new();
-            options.create(true).write(true);
-            options.open(path).with_context(|| {
-                format!(
-                    "failed to create call cache entry file `{path}`",
-                    path = path.display()
-                )
-            })?
+            match fs::File::open(path) {
```
peterhuene: I think we could probably simplify the entire assignment of `file` to:

```rust
let file = match fs::File::open(path) {
    Ok(file) => file,
    Err(e) if e.kind() == io::ErrorKind::NotFound => {
        if !create {
            return Ok(None);
        }

        // Create the file, which requires writable access
        let mut options = fs::OpenOptions::new();
        options.create(true).write(true);
        options.open(path).with_context(|| {
            format!("failed to create file `{path}`", path = path.display())
        })?;

        // Re-open the file as readable as the lock is shared and we don't want the file
        // to be writable
        fs::File::open(path).with_context(|| {
            format!("failed to open file `{path}`", path = path.display())
        })?
    }
    Err(e) => {
        return Err(e).with_context(|| {
            format!("failed to open file `{path}`", path = path.display())
        })?;
    }
};
```

peterhuene (commented):
This looks good, thanks for fixing!

I have the one comment relating to simplifying the existing code along with this change.

Once the comment is addressed, I may push up a commit to your branch for unrelated changes to the error messages in this file as they are too specific (it's not always a call cache entry that is locked, it may also be the global cache lock).

claymcleod:
Also needs a `CHANGELOG.md` entry! Thanks for this

crates/wdl-engine/src/cache/lock.rs
tmokveld: Looks good to me! The only thing I kept was `create_new(true)` over `create(true)` to make the create path race-safe.

peterhuene: I think we would want the `create` race, though; if another sprocket instance creates the file before the current instance, `create` would end up opening the file instead of returning an error like `create_new` does.

We can't fix the other race, however, which is if someone deletes the file in-between the create and re-open calls (since Sprocket doesn't delete the file, that would be far less likely to be encountered).

tmokveld: Right, I understand now, I updated the PR

crates/wdl-engine/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Fixed
+
+* Fixed shared lock acquisition to reopen new lock files read-only before locking.
```
peterhuene: Nit: add link to the PR.

```suggestion
* Fixed shared lock acquisition to reopen new lock files read-only before locking ([#869](https://github.com/stjude-rust-labs/sprocket/pull/869)).
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* Fixed shared lock acquisition to reopen new lock files read-only before locking.
+
```
peterhuene: We don't need to duplicate the WDL crate changes into the CHANGELOG for `sprocket` itself.

```suggestion
```

peterhuene: Or we can just revert the changes to this file.

peterhuene (approved):
Looks good, just the nits on the CHANGELOG entries (link to the PR and the root `CHANGELOG.md` is for sprocket itself, which doesn't need to duplicate the sub-crate changes.

peterhuene:
Just two nit comments to resolve; reverting the root CHANGELOG.md should also resolve the merge conflict (or we can rebase).

Regarding pushing up the changes to the error messages, I have a PR that I'll be putting up shortly and I'll just include those changes there rather than as part of this PR.

Thanks again for fixing this!

peterhuene (approved):

PR #870, opened 2026-05-22
feat(wdl-lsp): improved semantic tokens

Serial-ATA:
Improves the output of [textDocument/semanticTokens/full](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_fullRequest), using more specific token types/modifiers to allow for better highlighting

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/src/handlers/semantic_tokens.rs
+++ b/crates/wdl-analysis/src/handlers/semantic_tokens.rs
@@ -77,6 +77,7 @@ pub const WDL_SEMANTIC_TOKEN_MODIFIERS: &[SemanticTokenModifier] = &[
     SemanticTokenModifier::DEFINITION,
     SemanticTokenModifier::DOCUMENTATION,
     SemanticTokenModifier::READONLY,
+    SemanticTokenModifier::STATIC,
 ];
 
 /// Handles a semantic token request for a full document.
@@ -334,6 +335,7 @@ fn resolve_identifier_ty(
         SyntaxKind::MetadataObjectItemNode => {
             let metadata_item = MetadataObjectItem::cast(parent.clone()).expect("should cast");
             if metadata_item.name().inner() == token {
+                add_modifier(modifiers, SemanticTokenModifier::STATIC);
                 add_modifier(modifiers, SemanticTokenModifier::READONLY);
                 return if metadata_item.parent::<ParameterMetadataSection>().is_some() {
                     Some(SemanticTokenType::PARAMETER)
--- a/crates/wdl-analysis/src/handlers/semantic_tokens.rs
+++ b/crates/wdl-analysis/src/handlers/semantic_tokens.rs
@@ -340,7 +340,6 @@ fn resolve_identifier_ty(
                 return if metadata_item.parent::<ParameterMetadataSection>().is_some() {
                     Some(SemanticTokenType::PARAMETER)
                 } else {
-                    // I guess?
                     Some(SemanticTokenType::PROPERTY)
                 };
             }
--- a/crates/wdl-analysis/src/handlers/semantic_tokens.rs
+++ b/crates/wdl-analysis/src/handlers/semantic_tokens.rs
@@ -28,7 +28,6 @@ use wdl_ast::TreeToken;
 use wdl_ast::v1::AccessExpr;
 use wdl_ast::v1::CallExpr;
 use wdl_ast::v1::CallTarget;
-use wdl_ast::v1::Decl;
 use wdl_ast::v1::EnumDefinition;
 use wdl_ast::v1::EnumVariant;
 use wdl_ast::v1::Expr;
@@ -37,6 +36,8 @@ use wdl_ast::v1::LiteralStruct;
 use wdl_ast::v1::LiteralStructItem;
 use wdl_ast::v1::MetadataObjectItem;
 use wdl_ast::v1::ParameterMetadataSection;
+use wdl_ast::v1::RequirementsItem;
+use wdl_ast::v1::RuntimeItem;
 use wdl_ast::v1::StructDefinition;
 use wdl_ast::v1::TaskDefinition;
 use wdl_ast::v1::TypeRef;
@@ -212,17 +213,9 @@ fn token_ty(token: &SyntaxToken, document: &Document) -> Option<(SemanticTokenTy
     }
 
     if let Some(version) = document.version()
-        && version == SupportedVersion::V1(V1::Two)
+        && version >= SupportedVersion::V1(V1::Two)
         && kind == SyntaxKind::RuntimeKeyword
         && parent.kind() == SyntaxKind::RuntimeSectionNode
-    {
-        add_modifier(&mut modifiers, SemanticTokenModifier::DEPRECATED)
-    }
-
-    if token.text() == "docker"
-        && parent
-            .ancestors()
-            .any(|n| n.kind() == SyntaxKind::RuntimeSectionNode)
     {
         add_modifier(&mut modifiers, SemanticTokenModifier::DEPRECATED)
     }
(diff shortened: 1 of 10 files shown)


crates/wdl-analysis/src/handlers/semantic_tokens.rs
```diff
@@ -319,58 +364,85 @@ fn resolve_identifier_ty(
...
+            }
+
+            return None;
+        }
+        SyntaxKind::EnumVariantNode => {
+            return Some(SemanticTokenType::ENUM_MEMBER);
+        }
+        SyntaxKind::BoundDeclNode => return Some(SemanticTokenType::VARIABLE),
+        _ => {}
+    }
```
claymcleod: This block is shadowed by the `Decl::cast` arm above. `Decl::cast` accepts both `BoundDeclNode` and `UnboundDeclNode`, so any struct field hits the `Decl` arm first and returns `VARIABLE` instead of reaching the `StructDefinitionNode => PROPERTY` branch here. I think we should hoist the struct-field check into the `Decl::cast` arm (e.g., return `PROPERTY` when `parent.parent().kind() == SyntaxKind::StructDefinitionNode`) and then delete this block.

crates/wdl-analysis/src/handlers/semantic_tokens.rs
```diff
@@ -319,58 +364,85 @@ fn resolve_identifier_ty(
...
-                {
-                    add_modifier(modifiers, SemanticTokenModifier::READONLY);
-                    Some(SemanticTokenType::PARAMETER)
-                } else {
-                    Some(SemanticTokenType::VARIABLE)
-                }
-            }
-        };
+        return Some(SemanticTokenType::VARIABLE);
     }
```
claymcleod, line 462: Narrowing the fallback to `Type::Call(_)` drops classification for every bare identifier reference whose parent is `NameRefExprNode`. That covers variable uses on the RHS of expressions, `if` conditions, function arguments, scatter exprs, and the LHS of a struct/enum access. The prior offset-based lookup mapped these to `PARAMETER` (for input-section defs) or `VARIABLE`. Let's restore that. At minimum, add a `SyntaxKind::NameRefExprNode => VARIABLE` arm in the `match parent.kind()` above, and re-resolve to `PARAMETER` via the scope lookup when the definition site is in an `InputSectionNode`.

crates/wdl-analysis/src/handlers/semantic_tokens.rs
```diff
@@ -319,58 +364,85 @@ fn resolve_identifier_ty(
...
+                    return Some(SemanticTokenType::ENUM);
+                } else if ident_targets_rhs
+                    && e.definition()
+                        .variants()
+                        .any(|v| v.name().text() == token.text())
+                {
+                    return Some(SemanticTokenType::ENUM_MEMBER);
+                }
+            }
+        }
```
claymcleod: When the LHS resolves to a struct, this returns `STRUCT` for the RHS member token too. The enum branch below correctly distinguishes LHS from RHS. Let's mirror that for the struct branch so the RHS falls through to the `PROPERTY` return at the bottom.

```suggestion
        if let Expr::NameRef(name_expr) = target.strip_parenthesized() {
            let ident_targets_lhs = token == name_expr.name().inner();
            if document.struct_by_name(name_expr.name().text()).is_some() {
                if ident_targets_lhs {
                    return Some(SemanticTokenType::STRUCT);
                }
            } else if let Some(e) = document.enum_by_name(name_expr.name().text()) {
                if ident_targets_lhs {
                    return Some(SemanticTokenType::ENUM);
                } else if ident_targets_rhs
                    && e.definition()
                        .variants()
                        .any(|v| v.name().text() == token.text())
                {
                    return Some(SemanticTokenType::ENUM_MEMBER);
                }
            }
        }
```

crates/wdl-analysis/src/handlers/semantic_tokens.rs
```diff
@@ -304,6 +337,18 @@ fn resolve_identifier_ty(
...
     }
 
+    if let Some(m) = MetadataObjectItem::cast(parent.clone())
+        && m.name().inner() == token
+    {
+        add_modifier(modifiers, SemanticTokenModifier::READONLY);
+        return if m.parent::<ParameterMetadataSection>().is_some() {
+            Some(SemanticTokenType::PARAMETER)
+        } else {
+            // I guess?
```
claymcleod: Let's drop this comment.

crates/wdl-ast/src/v1/expr.rs
```diff
@@ -907,11 +907,26 @@ impl<N: TreeNode> Expr<N> {
...
+        loop {
+            if let Self::Parenthesized(inner) = expr {
+                expr = inner.expr();
+                continue;
+            }
+
+            break;
+        }
+        expr
     }
```
claymcleod:
```suggestion
    pub fn strip_parenthesized(mut self) -> Self {
        while let Self::Parenthesized(inner) = self {
            self = inner.expr();
        }
        self
    }
```

crates/wdl-lsp/tests/semantic_tokens.rs
```diff
@@ -0,0 +1,530 @@
...
+            length: 3,
+            token_type: token_type_index(SemanticTokenType::TYPE),
+            token_modifiers_bitset: 0,
+        },
+        // Member name
+        SemanticToken {
+            delta_line: 0,
+            delta_start: 4,
+            length: 3,
+            token_type: 1,
```
claymcleod: A few hard-coded `token_type` indices in this file (here, and at lines 420, 462, 470) bypass `token_type_index(...)`. They tie the test to the current order of `WDL_SEMANTIC_TOKEN_TYPES` and will silently desync if the array is reordered. Let's swap each for the helper call.

crates/wdl-lsp/tests/workspace/semantic_tokens/source.wdl
```diff
@@ -0,0 +1,33 @@
...
+workflow wf {
+    call do_work
+    
+    call foo.bar
+
+    Hello h = Hello.World
+    Foo f = Foo {
+        bar: 1
+    }
+}
```
claymcleod, line 57: The fixture exercises some paths but skips most of the new modifier behavior this PR adds: `READONLY` on input parameters, `READONLY` + `DEFAULT_LIBRARY` on `task.*` access, `ASYNC` on `scatter`, `DEPRECATED` on `runtime`/`docker` (needs a v1.2 doc), `DECLARATION` on an aliased `import`, and metadata/`parameter_meta` sections. Let's extend the fixture (or add a second one) to cover these so future regressions get caught.

crates/wdl-lsp/tests/semantic_tokens.rs
```diff
@@ -0,0 +1,530 @@
...
+}
+
+fn modifiers(modifiers: &[SemanticTokenModifier]) -> u32 {
+    modifiers.iter().fold(0, |mut acc, modifier| {
+        if let Some(pos) = WDL_SEMANTIC_TOKEN_MODIFIERS
+            .iter()
+            .position(|m| m == modifier)
+        {
+            acc |= 1 << pos;
+        }
```
claymcleod: `token_type_index` panics on missing type but this returns `0` silently. If a modifier is renamed or removed, every test referencing it would silently compare against an empty modifier bitset rather than failing. Let's mirror `token_type_index` and panic with `` "modifier `{modifier:?}` not found in `WDL_SEMANTIC_TOKEN_MODIFIERS`" ``.

claymcleod (changes requested):
One note that doesn't fit on a single line. `resolve_identifier_ty` could be reshaped as a single `match parent.kind()` dispatch to match the convention used by `hover.rs::resolve_hover_by_context` and `goto_definition.rs::resolve_by_context`. Each arm casts the parent (via `T::cast(parent.clone()).unwrap()` since the kind was just matched), runs its check, and returns. That style would eliminate the chain of speculative casts, keep parent-kind dispatch in one column, and make the dead-code class of bug (see the inline comment on the `match parent.kind()` block) impossible by construction. The two cases that legitimately need an ancestor walk (`ImportStatement` namespace alias, `AccessExpr` via ancestor) can fall out into a small `or_else` chain after the match.

claymcleod (approved):

PR #872, opened 2026-05-23
feat: wire resolver into `wdl-analysis` and Sprocket binary

claymcleod:
This PR connects the `wdl-modules` resolver layer to the analysis paths that users hit through `sprocket`, the LSP server, and server execution. The resolver crate can already materialize module dependencies; this branch makes analysis consume those materialized files so symbolic WDL imports behave like concrete local imports once a `module.json` is present.

### Overview

The code now has resolver-aware analysis without making ordinary analysis depend on Git behavior. `wdl-analysis` carries a `ResolutionContext` with a `Resolver` and optional `module.json` path, while the default analyzer constructors still use a null resolver so existing non-module callers keep their current behavior. Resolver-aware constructors let the CLI, LSP server, and server execution paths opt into module resolution when configuration is available.

### Current behavior

Symbolic imports now resolve before validation continues, which lets downstream analysis work against concrete documents. The analyzer materializes `import owner/module/path`, `import * from owner/module`, and `import { task } from owner/module` through `wdl-modules`; wildcard and selected imports then populate imported task and workflow metadata with their original source document. Unqualified selected and wildcard imported calls resolve for validation and go-to-definition, and imported callable conflicts are diagnosed once without cascading into misleading follow-on call-input errors.

Analysis-time Git resolution disables ambient credentials so diagnostics do not unexpectedly depend on local SSH agents or credential helpers. Local path module roots are also normalized for resolver traversal, cycle detection, and trust lookup.

### Integration points

The branch wires the resolver through each surface that constructs analysis. The CLI builds a resolution context from discovered module manifests and `[modules]` configuration, the LSP server passes configured WDL feature flags and resolution context into `Analyzer`, and server source validation preserves allowed directories as `Source::Directory` before resolving module entrypoints at run submission.

For all contributors:


crates/gauntlet/Gauntlet.toml
```diff
@@ -4912,11 +4912,6 @@ document = "broadinstitute/warp:/deprecated/pipelines/dna_seq/germline/single_sa
 message = "BroadInternalUltimaGenomics.wdl:28:3: error: unknown type name `VcfPostProcessing`"
 permalink = "https://github.com/broadinstitute/warp/blob/54032fd9003183eb90029033fd2f88d98ac9a001/deprecated/pipelines/dna_seq/germline/single_sample/UltimaGenomics/BroadInternalUltimaGenomics.wdl/#L28"
 
-[[diagnostics]]
-document = "broadinstitute/warp:/deprecated/pipelines/dna_seq/germline/single_sample/UltimaGenomics/BroadInternalUltimaGenomics.wdl"
```
claymcleod, line 4736: Note that these still have errors for the import failing—now we're not double flagging the unknown namespace. I think that's reasonable: if an import failed, it's implied that the namespace is going to be unknown. FWIW, this is also how `rustc` behaves.

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -664,28 +671,19 @@ fn resolve_access_expression(
             }
         }
 
-        // Check for enum definition in local document.
```
claymcleod: Put all these comments back.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -464,13 +484,502 @@ fn are_enums_equal(a: &EnumDefinition, b: &EnumDefinition) -> bool {
...
+            document,
+            &imported,
+            &uri,
+            member_name.text(),
+            &local_name,
+            member_span,
+        ) {
+            found_any = true;
+        }
+        if import_selected_enum(
```
peterhuene: Should these be `else if`? First one should win, right? There shouldn't be duplicate names in the various data structures of the imported document.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -464,13 +484,502 @@ fn are_enums_equal(a: &EnumDefinition, b: &EnumDefinition) -> bool {
...
+        .or_else(|| {
+            document
+                .imported_workflows
+                .get(name)
+                .map(|workflow| workflow.span)
+        })
+}
+
+/// Creates a diagnostic for a wildcard import conflict.
+fn wildcard_import_conflict(name: &str, import_span: Span, prev_span: Span) -> Diagnostic {
```
peterhuene: Can we move this and the next two diagnostic functions into `diagnostics.rs`? Generally that's where all of the analysis diagnostics are kept, although some of the validation rules (as they came over from `wdl-ast` originally) still have their own local diagnostic functions.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -1738,7 +2342,7 @@ fn resolve_import(
     Ok((imported_node.uri().clone(), imported_document))
 }
 
-/// Sets struct and enum types in the document.
+/// Sets the struct types in the document.
```
peterhuene: Is the previous comment no longer acurate?

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -856,21 +888,12 @@ fn resolve_struct_literal_item(
...
     if let Some(struct_info) = analysis_doc.struct_by_name(struct_name.text()) {
-        let (uri, def_lines) = match struct_info.namespace() {
-            Some(ns_name) => {
-                // SAFETY: we just found a struct_info with this namespace name and the document
-                // guarantees that `analysis_doc.namespaces` contains a corresponding entry for
-                // `ns_name`.
-                let ns = analysis_doc.namespace(ns_name).unwrap();
+        let (uri, def_lines) = match struct_info.source() {
+            Some(source) => {
+                let imported_node = graph.get(graph.get_index(source).unwrap());
```
peterhuene, line 895: Should this have a corresponding `SAFETY` comment to the modified enum code above?

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -856,21 +888,12 @@ fn resolve_struct_literal_item(
...
 
-                // SAFETY: `ns.source` comes from a valid namespace entry which guarantees the
-                // document exists in the graph.
-                let imported_node = graph.get(graph.get_index(ns.source()).unwrap());
-
-                // SAFETY: we successfully got the document above, it's in
-                // `ParseState::Parsed` which always has a valid lines field.
                 let lines = imported_node.parse_state().lines().unwrap();
```
peterhuene, line 899: Could probably keep this `SAFETY` comment.

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -951,19 +974,13 @@ fn resolve_call_input_item(
...
 
-                // SAFETY: we know `get_index` will return `Some` as `ns.source` comes from
-                // `analysis_doc.namespaces` which only contains namespaces for documents that
-                // are guaranteed to be present in the graph.
                 let node = graph.get(graph.get_index(ns.source()).unwrap());
                 let Some(imported_doc) = node.document() else {
                     return Ok(None);
                 };
 
-                // SAFETY: we successfully got the document above, it's in
```
peterhuene, line 962: Could probably keep this `SAFETY` comment.

crates/wdl-analysis/src/handlers/goto_definition.rs
```diff
@@ -951,19 +974,13 @@ fn resolve_call_input_item(
                     return Ok(None);
                 };
 
-                // SAFETY: we know `get_index` will return `Some` as `ns.source` comes from
```
peterhuene, line 954: Could probably keep this `SAFETY` comment.

crates/wdl-analysis/src/handlers/hover.rs
```diff
@@ -223,25 +223,17 @@ fn resolve_hover_by_context(
     match parent_node.kind() {
         SyntaxKind::TypeRefNode | SyntaxKind::LiteralStructNode => {
             if let Some(s) = document.struct_by_name(token.text()) {
-                let root = if let Some(ns_name) = s.namespace() {
-                    // SAFETY: we just found a struct with this namespace name and the document
```
peterhuene, line 227: Should we leave these safety comment changes here but update them to reflect the changes?

Feels like we should be consistent in documenting these between these various handlers; either it's obvious why the `unwrap`s are safe and we can skip the comment or we should keep them.

crates/wdl-analysis/src/analyzer.rs
```diff
@@ -323,14 +323,106 @@ pub struct IncrementalChange {
 /// the queue thread to join.
 ///
 /// The type parameter is the context type passed to the progress callback.
-#[derive(Debug)]
```
peterhuene, line 328: Could this remain `derive(Debug)`? Seems like `ResolutionContext` impls `Debug`.

crates/wdl-analysis/src/analyzer.rs
```diff
@@ -417,6 +564,16 @@ where
     pub async fn add_directory(&self, path: impl AsRef<Path>) -> Result<()> {
         let path = path.as_ref().to_path_buf();
         let config = self.config.clone();
+        // When the scanned directory lies inside the active module, stop the
+        // walk at nested module boundaries: a subdirectory with its own
+        // `module.json` is a separate (local-path dependency) module whose WDL
+        // files reach the analyzer through symbolic-import materialization, not
+        // directory scanning. Outside an active module there is nothing to scope
+        // to, so scan everything.
+        let active_module_root = self.resolution.module_root().map(Path::to_path_buf);
```
peterhuene: On the surface, it seems like the `to_path_buf` call isn't necessary here and we can borrow it long enough to check (at least it seems valid to the borrow-checker in my brain).

crates/wdl-analysis/src/analyzer.rs
```diff
@@ -417,6 +564,16 @@ where
     pub async fn add_directory(&self, path: impl AsRef<Path>) -> Result<()> {
         let path = path.as_ref().to_path_buf();
```
peterhuene: Existing issue: feels like `path` should be `impl Into<PathBuf>` and not `impl AsRef<Path>`.

crates/wdl-analysis/tests/validation/wdl-1-4-symbolic-import-with-flag/source.errors
```diff
@@ -1,18 +1,18 @@
-error: symbolic module resolution is not yet implemented
+error: symbolic import resolution is enabled, but no module governs this document
```
peterhuene: Upon seeing this diagnostic, I think "no module governs this document" might be confusing to users.

To me, at least, the fact the importing document isn't part of a module (i.e. there is no `module.json`) and by extension its list of dependencies would be empty should be treated as if it _were_ in a module with and it didn't contain the imported module name in its dependencies.

That is to say that I would expect to see something like "`openwdl/csvkit` is not a dependency" as a diagnostic.

crates/wdl-modules/src/resolver.rs
```diff
@@ -519,14 +587,14 @@ impl GitResolver {
...
-                self.resolve_git_selector(name, url, selector, path_prefix, scope)
+                self.resolve_git_selector(name, url, selector, path_prefix.as_deref(), scope)
                     .await?
             }
         };
 
         let key = CacheKey::from_git_url(url, &commit);
         let leaf = key.absolute_path(&self.cache_root);
-        let sparse_path = path_prefix.unwrap_or(".").to_string();
+        let sparse_path = path_prefix.clone().unwrap_or(".".to_string());
```
peterhuene:
```suggestion
        let sparse_path = path_prefix.clone().unwrap_or_else(|| ".".to_string());
```

peterhuene (commented):
Looks good, just a few minor comments to go through.

peterhuene (approved):
Looks good!

PR #873, opened 2026-05-25
feat(wdl-lsp): hover support on doc comments

Serial-ATA:
Adds the ability to hover doc comments to get a rendered preview. Not sure if other editors typically do this, but I really like this feature in RustRover

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/src/handlers/hover.rs
+++ b/crates/wdl-analysis/src/handlers/hover.rs
@@ -19,17 +19,16 @@ use wdl_ast::AstNode;
 use wdl_ast::AstToken;
 use wdl_ast::Comment;
 use wdl_ast::Documented;
+use wdl_ast::Ident;
 use wdl_ast::SyntaxKind;
 use wdl_ast::SyntaxNode;
 use wdl_ast::SyntaxToken;
 use wdl_ast::TreeNode;
 use wdl_ast::TreeToken;
 use wdl_ast::v1::AccessExpr;
-use wdl_ast::v1::BoundDecl;
 use wdl_ast::v1::CallExpr;
 use wdl_ast::v1::CallTarget;
 use wdl_ast::v1::Decl;
-use wdl_ast::v1::EnumDefinition;
 use wdl_ast::v1::EnumVariant;
 use wdl_ast::v1::LiteralStruct;
 use wdl_ast::v1::LiteralStructItem;
@@ -37,9 +36,6 @@ use wdl_ast::v1::MetadataObject;
 use wdl_ast::v1::MetadataValue;
 use wdl_ast::v1::ParameterMetadataSection;
 use wdl_ast::v1::StructDefinition;
-use wdl_ast::v1::TaskDefinition;
-use wdl_ast::v1::UnboundDecl;
-use wdl_ast::v1::WorkflowDefinition;
 use wdl_grammar::SyntaxElement;
 
 use crate::Document;
@@ -98,7 +94,7 @@ pub fn hover(
     let hovered_token = root
         .token_at_offset(offset)
         .find(|t| t.kind() == SyntaxKind::Ident || t.kind() == SyntaxKind::Comment)
-        .ok_or_else(|| anyhow!("no token found at offset"))?;
+        .ok_or_else(|| anyhow!("no hoverable token found at offset"))?;
 
     let parent_node = hovered_token.parent().expect("token has no parent");
 
@@ -180,7 +176,7 @@ fn resolve_hover_by_context(
     document: &Document,
     graph: &DocumentGraph,
 ) -> Result<Option<String>> {
-    // Hovering doc comments of an item produces the same content as hoving the
+    // Hovering doc comments of an item produces the same content as hovering the
     // identifier of the item
     if token.kind() == SyntaxKind::Comment {
         let comment = Comment::cast(token.clone()).expect("should cast");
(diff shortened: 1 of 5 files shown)


crates/wdl-analysis/src/handlers/hover.rs
```diff
@@ -86,14 +95,15 @@ pub fn hover(
     };
 
     let offset = position_to_offset(&lines, position, encoding)?;
-    let Some(token) = find_identifier_token_at_offset(&root, offset) else {
-        bail!("no identifier found at position");
-    };
+    let hovered_token = root
+        .token_at_offset(offset)
+        .find(|t| t.kind() == SyntaxKind::Ident || t.kind() == SyntaxKind::Comment)
+        .ok_or_else(|| anyhow!("no token found at offset"))?;
```
claymcleod: The message reads "no token found" but a token may have been found and rejected because it wasn't an `Ident` or `Comment`. Consider `bail!("no hoverable token at position")` for symmetry with the surrounding `bail!` style and a more accurate description.

crates/wdl-analysis/src/handlers/hover.rs
```diff
@@ -170,6 +180,62 @@ fn resolve_hover_by_context(
...
+                        || token.kind() == SyntaxKind::Whitespace =>
+                {
+                    continue;
+                }
+                _ => break,
+            };
+
+            let item_name = match target.kind() {
+                SyntaxKind::VersionStatementNode => {
+                    return match document.root().doc_comments() {
```
claymcleod: Every other arm produces output that leads with a ```` ```wdl … ``` ```` signature followed by `---` and the doc text. This branch returns just the raw doc string, so hovering preamble docs looks inconsistent with hovering anything else. Consider:

```rust
SyntaxKind::VersionStatementNode => {
    let Some(comments) = document.root().doc_comments() else {
        return Ok(None);
    };
    if comments.is_empty() {
        return Ok(None);
    }
    let docs = comments_to_string(comments).unwrap_or_default();
    // SAFETY: kind was just checked to be `VersionStatementNode`.
    let version = VersionStatement::cast(target).unwrap().version().text();
    return Ok(Some(format!("```wdl\nversion {version}\n```\n---\n{docs}")));
}
```

crates/wdl-analysis/src/handlers/hover.rs
```diff
@@ -170,6 +180,62 @@ fn resolve_hover_by_context(
     document: &Document,
     graph: &DocumentGraph,
 ) -> Result<Option<String>> {
+    // Hovering doc comments of an item produces the same content as hoving the
+    // identifier of the item
```
claymcleod:
```suggestion
    // Hovering doc comments of an item produces the same content as hovering
    // the identifier of the item.
```

crates/wdl-lsp/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+* Added hover support for doc comments, allowing editors to render preamble comments and give more surface area
+  to render item docs ([#873](https://github.com/stjude-rust-labs/sprocket/pull/873)).
```
claymcleod:
```suggestion
* Extended hover to render docs when the cursor is over a doc comment or the file preamble ([#873](https://github.com/stjude-rust-labs/sprocket/pull/873)).
```

crates/wdl-lsp/tests/hover.rs
```diff
@@ -332,8 +332,26 @@ async fn should_hover_blank_line_doc_merges_paragraphs() {
...
 }
+
+#[tokio::test]
+async fn should_render_docs_on_comment_hover() {
+    let mut ctx = setup().await;
+    // Hovering over doc comments should produce the same content as hovering
+    // over the item name.
+    let response = hover_request(&mut ctx, "doc_comments.wdl", Position::new(5, 4)).await;
+    assert_hover_content(&response, "Greets someone by name.");
+    assert_hover_content(&response, "Used for hover doc tests.");
```
claymcleod, line 349: Consider adding `assert_hover_content(&response, "task doc_only");` so the test actually verifies the "same content as hovering over the item name" claim — right now the assertions would still pass if the implementation only returned the raw doc text.

claymcleod (changes requested):
Two follow-ups worth considering:

- The eight-arm cast/`name()` block could collapse to a single `target.descendants_with_tokens().find_map(...)` lookup for the first `Ident`—every supported item's `.name()` returns exactly that token—which also removes the per-kind enumeration as a maintenance hazard when new documentable item kinds get added later. As a bonus, promoting `VersionStatementNode` to a top-level arm of the main `match parent_node.kind()` would let hovering the `version` keyword itself render the preamble docs.
- Test coverage: negative test for plain `#` line comments, multi-line `##` block hovered on the second line, and a trailing doc comment with no following item.

crates/wdl-analysis/src/handlers/hover.rs
Serial-ATA: I feel like having the doc comments alone makes sense. Showing the `workflow`/`task`/`struct`/`enum` when hovering over their comments makes sense since they're logically attached to each other. But preambles are a description of the document as a whole, not the version statement.

image

Serial-ATA: Since this was inspired by the behavior in RustRover, here's what its previews look like:

image

image

claymcleod: Okay! That makes sense to me.

claymcleod (approved):

PR #874, opened 2026-05-25
feat(wdl-lsp): call hierarchy support

Serial-ATA:
Adds support for the `textDocument/prepareCallHierarchy`, `callHierarchy/incomingCalls`, and `callHierarchy/outgoingCalls` requests

For editors that support it, you can use it to visualize workflows:

```wdl
task first {
    command <<<>>>
}

workflow second {
    call first
}

workflow third {
    call second
}

workflow all_together {
    call first
    call second
    call third
}
```

image

Or reverse it to find the callers of a task/workflow:

image

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/src/handlers/call_hierarchy.rs
+++ b/crates/wdl-analysis/src/handlers/call_hierarchy.rs
@@ -344,20 +344,14 @@ pub fn outgoing_calls(
             &from_doc
         };
 
-        let source_index = graph.get_index(to_doc.uri()).ok_or_else(|| {
-            anyhow!(
-                "document `{uri}` not found in graph",
-                uri = to_doc.uri()
-            )
-        })?;
+        let source_index = graph
+            .get_index(to_doc.uri())
+            .ok_or_else(|| anyhow!("document `{uri}` not found in graph", uri = to_doc.uri()))?;
 
         let source_node = graph.get(source_index);
         let lines = match source_node.parse_state() {
             ParseState::Parsed { lines, .. } => lines.clone(),
-            _ => bail!(
-                "document `{uri}` has not been parsed",
-                uri = to_doc.uri()
-            ),
+            _ => bail!("document `{uri}` has not been parsed", uri = to_doc.uri()),
         };
 
         let Some(from_span) = scope.lookup(ident).map(|s| s.span()) else {
--- a/crates/wdl-analysis/src/handlers/call_hierarchy.rs
+++ b/crates/wdl-analysis/src/handlers/call_hierarchy.rs
@@ -1,0 +1,341 @@
+//! Handlers for `call hierarchy` requests.
+//!
+//! This module implements the LSP "textDocument/prepareCallHierarchy"
+//! functionality for WDL files.
+//!
+//! See: [LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareCallHierarchy)
+
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+use std::sync::Arc;
+
+use anyhow::Result;
+use anyhow::anyhow;
+use anyhow::bail;
+use line_index::LineIndex;
+use lsp_types::CallHierarchyIncomingCall;
+use lsp_types::CallHierarchyItem;
+use lsp_types::CallHierarchyOutgoingCall;
+use lsp_types::Range;
+use lsp_types::SymbolKind;
+use url::Url;
+use wdl_ast::AstNode;
+use wdl_ast::AstToken;
+use wdl_ast::v1::TaskDefinition;
+use wdl_ast::v1::WorkflowDefinition;
+use wdl_grammar::Span;
+use wdl_grammar::SyntaxNode;
+
+use crate::Document;
+use crate::SourcePosition;
+use crate::SourcePositionEncoding;
+use crate::document::Task;
+use crate::document::Workflow;
+use crate::graph::DocumentGraph;
+use crate::graph::ParseState;
+use crate::handlers::EnclosingScopeKind;
+use crate::handlers::common::find_identifier_token_at_offset;
+use crate::handlers::common::location_from_span;
+use crate::handlers::common::position_to_offset;
+use crate::handlers::find_all_references;
+use crate::types::CallKind;
+
+/// A callable item.
+enum Callable<'a> {
+    /// A workflow.
+    Workflow(&'a Workflow),
+    /// A task.
+    Task(&'a Task),
+}
+
+impl Callable<'_> {
+    /// Get the name of this callable.
+    fn name(&self) -> &str {
+        match self {
+            Callable::Workflow(w) => w.name(),
+            Callable::Task(t) => t.name(),
+        }
+    }
+
+    /// Get the [`Span`] of the callable's name.
+    fn name_span(&self) -> Span {
+        match self {
+            Callable::Workflow(w) => w.name_span(),
+            Callable::Task(t) => t.name_span(),
+        }
+    }
+
+    /// Get the [`SymbolKind`] for this callable.
+    fn symbol_kind(&self) -> SymbolKind {
+        match self {
+            Callable::Workflow(_) => SymbolKind::FUNCTION,
+            Callable::Task(_) => SymbolKind::METHOD,
+        }
+    }
+
+    /// Attempt to convert this callable to a [`CallHierarchyItem`].
+    fn as_call_hierarchy_item(
+        &self,
+        analysis_doc: &Document,
+        lines: &LineIndex,
+    ) -> Result<CallHierarchyItem> {
+        let name_range = location_from_span(analysis_doc.uri(), self.name_span(), lines)?.range;
+
+        Ok(CallHierarchyItem {
+            name: self.name().to_string(),
+            kind: self.symbol_kind(),
+            tags: None,
+            detail: None,
+            uri: (**analysis_doc.uri()).clone(),
+            range: name_range,
+            selection_range: name_range,
+            data: None,
+        })
+    }
+}
+
+/// Attempt to get a [`Callable`] from the specified position.
+fn find_callable_at_position<'a>(
+    graph: &'a DocumentGraph,
+    document_uri: &Url,
+    position: SourcePosition,
+    encoding: SourcePositionEncoding,
+) -> Result<Option<(Document, Arc<LineIndex>, Callable<'a>)>> {
+    let index = graph
+        .get_index(document_uri)
+        .ok_or_else(|| anyhow!("document `{uri}` not found in graph", uri = document_uri))?;
+
+    let node = graph.get(index);
+    let (root, lines) = match node.parse_state() {
+        ParseState::Parsed { lines, root, .. } => {
+            (SyntaxNode::new_root(root.clone()), lines.clone())
+        }
+        _ => bail!("document `{uri}` has not been parsed", uri = document_uri),
+    };
+
+    let Some(analysis_doc) = node.document() else {
+        bail!("document analysis data not available for {}", document_uri);
+    };
+
+    let offset = position_to_offset(&lines, position, encoding)?;
+    let Some(token) = find_identifier_token_at_offset(&root, offset) else {
+        return Ok(None);
+    };
+
+    if let Some(workflow) = analysis_doc.workflow()
+        && workflow.name() == token.text()
+    {
+        return Ok(Some((
+            analysis_doc.clone(),
+            lines,
+            Callable::Workflow(workflow),
+        )));
+    }
+
+    if let Some(task) = analysis_doc.task_by_name(token.text()) {
+        return Ok(Some((analysis_doc.clone(), lines, Callable::Task(task))));
+    }
+
+    Ok(None)
+}
+
+/// Creates a [`CallHierarchyItem`] for the given symbol, if applicable.
+///
+/// Implementation of [`textDocument/prepareCallHierarchy`]
+///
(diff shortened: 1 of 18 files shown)


crates/wdl-analysis/src/handlers/call_hierarchy.rs
```diff
@@ -0,0 +1,341 @@
...
+        let name_range = location_from_span(analysis_doc.uri(), self.name_span(), lines)?.range;
+
+        Ok(CallHierarchyItem {
+            name: self.name().to_string(),
+            kind: self.symbol_kind(),
+            tags: None,
+            detail: None,
+            uri: (**analysis_doc.uri()).clone(),
+            range: name_range,
+            selection_range: name_range,
```
claymcleod: Per LSP 3.17, `range` should be the full enclosing-symbol range (body and comments) while `selectionRange` is the name; clients render `range` as the symbol's clickable region, so collapsing both to the name span breaks highlight and reveal behavior. The same bug appears in the vacant arms of `incoming_calls` and `outgoing_calls`.

One option: take the `SyntaxNode` of the `TaskDefinition`/`WorkflowDefinition` and use its `text_range()` for `range`; keep `name_span()` for `selection_range`. For `incoming_calls`, extend `EnclosingScope` with a `full_range` field populated by `resolve_enclosing_scope` from the enclosing definition's syntax span. For `outgoing_calls`, the `TaskDefinition`/`WorkflowDefinition` is already in hand at lines 295/303 — use its `inner().text_range()` for `range`.

crates/wdl-analysis/src/handlers/call_hierarchy.rs
```diff
@@ -0,0 +1,341 @@
...
+    let mut calls: HashMap<(Url, String), (CallHierarchyItem, Vec<Range>)> = HashMap::new();
+    let scope = workflow.scope();
+
+    for (ident, call) in workflow.calls() {
+        let source_doc = call
+            .namespace()
+            .map(|ns| {
+                analysis_doc
+                    .namespace(ns)
+                    .expect("namespace should be present")
```
claymcleod: `analysis_doc.namespace(ns)` returns `None` when the import name was never registered—e.g. when the `import` statement has a diagnostic. The same risk applies to the four other `.expect(...)` calls in this function: `scope.lookup(ident).expect("should be in scope")` at line 288, and the two `.find(...).expect("should exist")` lookups at lines 297 and 305; each can fire when the imported document is partially analyzed or the call references a renamed/deleted target. LSP handlers must keep serving on broken code, but today any of these aborts the analyzer worker thread.

I think each site should `continue` rather than panic. Outgoing-call results are advisory, and dropping one entry is much better than crashing the worker.

crates/wdl-analysis/src/handlers/call_hierarchy.rs
```diff
@@ -0,0 +1,341 @@
...
+        return Ok(Some((
+            analysis_doc.clone(),
+            lines,
+            Callable::Workflow(workflow),
+        )));
+    }
+
+    if let Some(task) = analysis_doc.task_by_name(token.text()) {
+        return Ok(Some((analysis_doc.clone(), lines, Callable::Task(task))));
+    }
```
claymcleod: `find_callable_at_position` matches by token text alone—any identifier whose text matches a top-level workflow or task name returns that callable, even when the cursor is on an unrelated declaration. With a workflow `first` in the file, a local `String first = "x"` inside any other task or workflow makes `prepareCallHierarchy` claim the cursor is on the top-level `first`, and the false seed cascades into `incoming_calls` and `outgoing_calls`.

You could resolve the token via `goto_definition` and only accept the callable when the resolved span equals `workflow.name_span()` or `task.name_span()`. That also generalizes to cross-document calls, a cursor on a call site referencing an imported task.

crates/wdl-analysis/src/handlers/call_hierarchy.rs
```diff
@@ -0,0 +1,341 @@
...
+            }
+            CallKind::Workflow => {
+                let workflow = source_doc
+                    .root()
+                    .children::<WorkflowDefinition>()
+                    .find(|workflow| workflow.name().text() == call.name())
+                    .expect("should exist");
+                (SymbolKind::FUNCTION, workflow.name().span())
+            }
+        };
```
claymcleod: `outgoing_calls` walks the CST to find the called task or workflow, but the analysis model already exposes `Document::task_by_name` and `Document::workflow()`—the same lookups `find_callable_at_position` uses at line 135 to identify the seed callable. Swap the CST traversal for those accessors: it removes two of the `.expect("should exist")` panic sites, drops an O(tasks) scan per call, and keeps both ends of the call hierarchy symmetric.

I suggest promoting the private `Callable` enum to `document.rs` so both ends share the same accessor.

crates/wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -32,6 +34,59 @@ struct TargetDefinition {
...
+    pub location: Location,
+    /// The enclosing task or workflow scope that contains this reference.
+    ///
+    /// This is `None` for references at the document top-level (e.g., struct
+    /// definitions or import statements).
+    pub enclosing_scope: Option<EnclosingScope>,
+}
+
+/// The result of a find-all-references query.
+#[derive(Clone, Debug)]
```
claymcleod: `ScopedReferences` carries an `Option<EnclosingScope>` per reference, but only `incoming_calls` reads it—`rename` and the queue's `FindAllReferences` arm both call `.locations()` and throw the scope away. Move the enclosing-scope resolution inside `incoming_calls`: walk each returned `Location`, look up its document in the graph (the same pattern `outgoing_calls` already uses at lines 272–286), and compute the enclosing scope there. `find_all_references` then stays a single-purpose primitive returning `Vec<Location>`, `EnclosingScope`, `EnclosingScopeKind`, `ReferenceWithScope`, and `ScopedReferences` collapse into private helpers in `call_hierarchy.rs`, and `rename` stops paying for unused work.

This also drops the breaking change: `Analyzer::find_all_references` already returns `Vec<Location>` to external callers, so reverting the internal signature leaves the public API untouched and the `!` in the commit unnecessary.

crates/wdl-lsp/tests/call_hierarchy.rs
```diff
@@ -0,0 +1,331 @@
...
+    else {
+        panic!("expected call hierarchy")
+    };
+
+    verify_call_hierarchy(vec![ExpectedCallHierarchyItem::first()], &hierarchy);
+
+    let item = hierarchy.remove(0);
+    let incoming_calls = incoming_calls_request(&mut ctx, item)
+        .await
+        .expect("should return outgoing calls");
```
claymcleod:
```suggestion
        .expect("should return incoming calls");
```

crates/wdl-analysis/src/handlers/call_hierarchy.rs
```diff
@@ -0,0 +1,341 @@
...
+use crate::graph::ParseState;
+use crate::handlers::EnclosingScopeKind;
+use crate::handlers::common::find_identifier_token_at_offset;
+use crate::handlers::common::location_from_span;
+use crate::handlers::common::position_to_offset;
+use crate::handlers::find_all_references;
+use crate::types::CallKind;
+
+/// A callable item.
+enum Callable<'a> {
```
claymcleod:
```suggestion
#[derive(Debug)]
enum Callable<'a> {
```

crates/wdl-analysis/src/handlers/call_hierarchy.rs
```diff
@@ -0,0 +1,341 @@
...
+    let node = graph.get(index);
+    let (root, lines) = match node.parse_state() {
+        ParseState::Parsed { lines, root, .. } => {
+            (SyntaxNode::new_root(root.clone()), lines.clone())
+        }
+        _ => bail!("document `{uri}` has not been parsed", uri = document_uri),
+    };
+
+    let Some(analysis_doc) = node.document() else {
+        bail!("document analysis data not available for {}", document_uri);
```
claymcleod:
```suggestion
        bail!("document analysis data not available for `{document_uri}`");
```

crates/wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -32,6 +34,59 @@ struct TargetDefinition {
...
+    /// The kind of the enclosing scope.
+    pub kind: EnclosingScopeKind,
+    /// The name of the enclosing task or workflow.
+    pub name: String,
+    /// The location of the enclosing scope's name declaration.
+    pub location: Location,
+}
+
+/// Represents the kind of an enclosing scope.
+#[derive(Clone, Debug, PartialEq, Eq)]
```
claymcleod:
```suggestion
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```

crates/wdl-analysis/src/handlers/call_hierarchy.rs
```diff
@@ -0,0 +1,341 @@
...
+    let Some((analysis_doc, _, callable)) =
+        find_callable_at_position(graph, &document_uri, position, encoding)?
+    else {
+        return Ok(None);
+    };
+
+    let target_name = match &callable {
+        Callable::Workflow(workflow) => workflow.name().to_string(),
+        Callable::Task(task) => task.name().to_string(),
+    };
```
claymcleod:
```suggestion
    let target_name = callable.name().to_string();
```

crates/wdl-lsp/tests/call_hierarchy.rs
```diff
@@ -0,0 +1,331 @@
...
+    mut expected: Vec<ExpectedCallHierarchyItem>,
+    received: &[CallHierarchyItem],
+) {
+    for item in received {
+        let matched = expected.iter().position(|expected| expected == item);
+
+        if let Some(index) = matched {
+            expected.remove(index);
+        } else {
+            panic!("Unexpected call hierarchy item returned: {item:?}");
```
claymcleod:
```suggestion
            panic!("unexpected call hierarchy item returned: {item:?}");
```

crates/wdl-lsp/tests/call_hierarchy.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    assert!(
+        expected.is_empty(),
+        "Some expected items were not returned: {expected:?}"
```
claymcleod:
```suggestion
        "some expected items were not returned: {expected:?}"
```

crates/wdl-lsp/tests/call_hierarchy.rs
```diff
@@ -0,0 +1,331 @@
...
+                && expected
+                    .from_ranges
+                    .iter()
+                    .all(|expected_range| call.from_ranges.contains(expected_range))
+        });
+
+        if let Some(index) = matched {
+            expected.remove(index);
+        } else {
+            panic!("Unexpected outgoing call returned: {call:?}");
```
claymcleod:
```suggestion
            panic!("unexpected outgoing call returned: {call:?}");
```

crates/wdl-lsp/tests/call_hierarchy.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    assert!(
+        expected.is_empty(),
+        "Some expected items were not returned: {expected:?}"
```
claymcleod:
```suggestion
        "some expected items were not returned: {expected:?}"
```

crates/wdl-lsp/tests/call_hierarchy.rs
```diff
@@ -0,0 +1,331 @@
...
+) {
+    for call in received {
+        let matched = expected.iter().position(|expected| {
+            expected.from == call.from && expected.from_ranges == call.from_ranges
+        });
+
+        if let Some(index) = matched {
+            expected.remove(index);
+        } else {
+            panic!("Unexpected incoming call returned: {call:?}");
```
claymcleod:
```suggestion
            panic!("unexpected incoming call returned: {call:?}");
```

crates/wdl-lsp/tests/call_hierarchy.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    assert!(
+        expected.is_empty(),
+        "Some expected items were not returned: {expected:?}"
```
claymcleod:
```suggestion
        "some expected items were not returned: {expected:?}"
```

claymcleod (changes requested):
Looking good!

claymcleod:
Thanks for the quick turnaround. Upon review, I think three items from the original review are still open.

Comment 2 (panic sites in `outgoing_calls`) is only partly addressed. The `analysis_doc.namespace(ns)` site correctly `continue`s now, but three `.expect(...)` calls remain in `crates/wdl-analysis/src/handlers/call_hierarchy.rs`. Each of these still aborts the analyzer worker on partially-analyzed imports, renamed targets, or calls to a definition that no longer exists. They should `continue` for the same reason the namespace site does.

Comment 4's second half wasn't taken—the CST traversal was swapped for `task_by_name`/`workflow()` (good), but `Callable` is still a private enum in `call_hierarchy.rs`. Promoting it to `document.rs` next to the accessors would let `find_callable_at_position` and the outgoing-call lookup share one type rather than rediscovering it on each end. If you don't think we should do this, will you let me know why?

On comment 5, now that `find_all_references` is back to returning `Vec<Location>`, is anything else here actually a breaking change? If not, the `!` on the commit and PR title can come off.

claymcleod (approved):

PR #881, opened 2026-05-28
feat: add Renovate configuration files for dependency management

crh225:
Adds self-managed [Renovate](https://docs.renovatebot.com) via a GitHub Actions workflow and config file to automate dependency updates for `Cargo.toml`, `Cargo.lock`, and GitHub Actions versions. It keeps things current without manual chores while leaving humans in the loop for anything risky.

**How it runs:** A scheduled GitHub Action (weekly, Monday 8 AM CT) runs Renovate using the existing GitHub App token (`APP_ID`/`APP_PRIVATE_KEY`). Supports manual dispatch and dry-run mode. Changes to the config or workflow trigger a dry-run on PR to validate before merge.

**Key configuration choices:**

- **14-day `minimumReleaseAge`** — freshly published crates and actions get a buffer before adoption, catching yanked releases and supply-chain issues.
- **Minor/patch automerge** — auto-merges via GitHub's native `platformAutomerge` with a squash strategy once CI passes. Major updates open a labeled PR for human review.
- **Vulnerability alerts** (Dependabot + OSV) bypass the release-age window and run on any schedule so CVE fixes are not delayed.
- **`rangeStrategy: "update-lockfile"`** — `Cargo.toml` constraints are only rewritten when the new version falls outside the existing range, keeping PRs to lockfile-only churn most of the time.
- **Weekly `lockFileMaintenance`** for `Cargo.lock` refresh.
- **Semantic commits** — PRs use `build(deps):` commit convention.

For all contributors:

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -11,7 +11,7 @@
   ],
   "timezone": "America/Chicago",
   "schedule": [
-    "before 6am"
+    "at any time"
   ],
   "prHourlyLimit": 4,
   "prConcurrentLimit": 10,
@@ -28,7 +28,7 @@
   "lockFileMaintenance": {
     "enabled": true,
     "schedule": [
-      "before 6am on monday"
+      "at any time"
     ],
     "automerge": true
   },
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,81 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":dependencyDashboard",
+    ":semanticCommits",
+    ":semanticCommitTypeAll(build)"
+  ],
+  "labels": [
+    "dependencies"
+  ],
+  "timezone": "America/Chicago",
+  "schedule": [
+    "before 6am"
+  ],
+  "prHourlyLimit": 4,
+  "prConcurrentLimit": 10,
+  "minimumReleaseAge": "14 days",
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -0,0 +1,81 @@
...
+    ":semanticCommits",
+    ":semanticCommitTypeAll(build)"
+  ],
+  "labels": [
+    "dependencies"
+  ],
+  "timezone": "America/Chicago",
+  "schedule": [
+    "before 6am"
+  ],
```
claymcleod, line 15: The workflow's only scheduled trigger is the Monday 13:00 UTC cron, which is 8 AM in `America/Chicago`, but this `schedule` only allows Renovate to create branches `before 6am`. Per the Renovate scheduling docs, branches are never created outside the schedule window, so every weekly run lands at 8 AM, falls outside the window, and Renovate opens no PRs. The same applies to `lockFileMaintenance.schedule` (`before 6am on monday`); only `vulnerabilityAlerts` (`at any time`) would ever fire. Since the cron already controls cadence, I'd let the in-config schedules always match.

```suggestion
  "schedule": [
    "at any time"
  ],
```

The same change is needed for `lockFileMaintenance.schedule`, or drop both `schedule` keys entirely and let the cron drive timing.

claymcleod (changes requested):

renovate.json
crh225: code changed to "at any time"

PR #888, opened 2026-05-29
feat(wdl-lsp)!: switch to `async-lsp`

### LINKED ISSUE #586: LSP shutdown request returns error instead of `null`

Serial-ATA:
Lot of noise here from updating the tests, which are actually much simpler now (see the diff for `common.rs`). All of the manual request/response handling was removed, since `async-lsp` *also* gives us the ability to create clients

closes #586

supersedes #826 

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lsp/src/server.rs
```diff
@@ -1229,151 +1159,345 @@ impl<S: 'static> LanguageServer for Server<S> {
...
-            )
-            .await
-            .map_err(|e| RpcError {
-                code: ErrorCode::InternalError,
-                message: e.to_string().into(),
-                data: None,
-            })?;
+        // Notify the analyzer that the document has changed
+        let state = self.state.clone();
+        tokio::task::spawn(async move {
```
claymcleod: Each of `did_open`, `did_change`, and `did_close` now spawns an independent `tokio` task, so the order in which they acquire `state.read()` and call `notify_incremental_change` is no longer deterministic.

`async-lsp` hands notifications to `notify` in receipt order, but spawning drops that guarantee, so two rapid `did_change` notifications (or a `did_open` followed immediately by `did_change`) can reach the analyzer out of order and corrupt incremental edit state. 

My recommendation would be to drain events through an mpsc actor rather than spawning per message so receipt order is preserved.

crates/wdl-lsp/src/server.rs
```diff
@@ -1229,151 +1159,345 @@ impl<S: 'static> LanguageServer for Server<S> {
...
+            if !params.event.added.is_empty() {
+                for folder in &params.event.added {
+                    if let Err(e) = state
+                        .config
+                        .analyzer
+                        .add_directory(folder.uri.to_file_path().expect("should be a file path"))
+                        .await
+                    {
+                        error!("failed to add documents from directory to analyzer: {e}");
+                    }
```
claymcleod: `did_change_workspace_folders` panics here on any non-`file` URI, where `initialize` handles the same case with a guard. Because this runs inside a spawned task, the panic isn't caught by `CatchUnwindLayer` (which only wraps the request service), so it silently aborts the folder add. Let's mirror the guarded pattern already used in `initialize`.

```suggestion
                for folder in &params.event.added {
                    if let Ok(path) = folder.uri.to_file_path()
                        && let Err(e) = state.config.analyzer.add_directory(path).await
                    {
                        error!("failed to add documents from directory to analyzer: {e}");
                    }
```

crates/wdl-lsp/src/server.rs
```diff
@@ -431,675 +483,525 @@ impl<S: 'static> Server<S> {
...
+                .analyzer
+                .analyze_document(ProgressToken::default(), params.text_document.uri.clone())
+                .await
+                .map_err(|e| ResponseError::new(ErrorCode::INTERNAL_ERROR, e))?;
+
+            let mut matcher = options.baseline.as_ref().map(|b| b.matcher());
+            proto::document_diagnostic_report(params, results, &name, matcher.as_mut()).ok_or_else(
+                || {
+                    ResponseError::new(
+                        ErrorCode::UNKNOWN_ERROR_CODE,
```
claymcleod: This changes the no-report case from `REQUEST_CANCELLED` to `UNKNOWN_ERROR_CODE`: why'd we go that direction? `RequestCancelled` is handled silently, whereas a generic error code can surface to the user as a failed diagnostic request.

claymcleod (changes requested):

crates/wdl-lsp/src/server.rs
Serial-ATA: `REQUEST_CANCELLED` isn't an appropriate error code, https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#errorCodes:

> The client has canceled a request and a server has detected the cancel.

I can change it to `REQUEST_FAILED`, but the error should surface regardless. The only time `document_diagnostic_report()` can fail is if the diagnostic spans don't actually map to positions in the document, which is a bug.

crates/wdl-lsp/src/server.rs
```diff
@@ -895,102 +798,329 @@ impl<S: 'static> LanguageServer for Server<S> {
...
+                    folding_range_provider: Some(FoldingRangeProviderCapability::Simple(true)),
+                    ..Default::default()
+                },
+                server_info: Some(info),
+            })
+        })
+    }
 
-        Ok(result)
+    fn shutdown(&mut self, _: ()) -> BoxFuture<'static, Result<(), Self::Error>> {
```
claymcleod, line 1674: `shutdown` returns right away while `notification_loop` may still be draining queued edits, and `_notification_loop_handle` is stored on the struct but never awaited or aborted. This works since `exit` follows, but keeping the handle only to drop it reads as if the task is managed when it isn't. Consider either dropping the field (a plain detached `spawn` says the same thing) or using it here to abort or join the loop so in-flight analyzer work settles before responding (I'd recommend the latter).

crates/wdl-lsp/src/server.rs
```diff
@@ -426,242 +493,152 @@ impl ServerOptions {
...
     /// Runs the server until a request is received to shut down.
     ///
     /// See also: [`Self::new()`]
     pub async fn run(
         options: ServerOptions,
+        user_options: UserOptions,
         log_handle: Option<FilterReloadHandle<S>>,
-    ) -> Result<()> {
+    ) -> anyhow::Result<()> {
         debug!("running LSP server: {options:#?}");
```
claymcleod: Now that `user_options` is a separate argument (log level and lint config), this line logs only `options` and drops it. One option is to include it:

```suggestion
        debug!("running LSP server: {options:#?}; user options: {user_options:#?}");
```

crates/wdl-lsp/src/server.rs
```diff
@@ -1229,151 +1445,152 @@ impl<S: 'static> LanguageServer for Server<S> {
...
-    async fn symbol(
-        &self,
-        params: WorkspaceSymbolParams,
-    ) -> RpcResult<Option<Vec<SymbolInformation>>> {
-        let config = self.config.read().await;
+    fn did_change_workspace_folders(
+        &mut self,
+        params: DidChangeWorkspaceFoldersParams,
+    ) -> Self::NotifyResult {
+        debug!("received `workspace/didChangeWorkspaceFolders` request: {params:#?}");
```
claymcleod: These `did*` handlers are LSP notifications, not requests, but several of the entry logs call them a `request`: `didChangeWorkspaceFolders` here, plus `didOpen`, `didChange`, `didClose`, and `didChangeWatchedFiles`. `didChangeConfiguration` already says `notification`. The wording is carried over from the previous implementation, but since these lines are being rewritten anyway it would be nice to make the five consistent. Relatedly, `shutdown` is the only handler with no entry log at all, and a `debug!` there would help since it is the path this PR fixes for #586.

crates/wdl-lsp/src/server.rs
```diff
@@ -840,23 +745,24 @@ impl<S: 'static> LanguageServer for Server<S> {
         // Progress the added folders
         if !params.event.added.is_empty() {
             for folder in &params.event.added {
-                if let Err(e) = config
-                    .analyzer
-                    .add_directory(folder.uri.to_file_path().expect("should be a file path"))
-                    .await
+                if let Ok(path) = folder.uri.to_file_path()
```
claymcleod: Workspace folders whose URI does not convert to a local path are silently dropped here, and the same pattern repeats in `initialize`. That is a reasonable thing to skip (a remote or virtual workspace this server cannot analyze), but a `warn!` noting the skipped URI would make it diagnosable instead of invisible. `error!` would be too loud, since non-`file` folders are a legitimate case.

claymcleod (changes requested):
A few things beyond the inline notes. 

1. The `S` type parameter threaded through `Server` and `ServerState` exists only to carry the `tracing-subscriber` reload handle; boxing that handle would let the whole file drop the generic, worth considering as a follow-up. I know this isn't related to your changes, but it might be a good time to clean that up if you agree.
2. The notification actor also preserves order among notifications but not against requests, since a request handler can take `state.read()` and run ahead of a `didChange` still queued in the channel; if that ordering matters for incremental edits, routing requests through the same serialized path, or documenting why it is safe, would close the gap.
3. Separately, the `shutdown` fix for #586 has no test, which would be cheap to add now that the harness can issue requests directly.

crates/wdl-lsp/src/server.rs
Serial-ATA: no need to keep those logs around, switched to the `TracingLayer` from `async-lsp` in [`8a52985` (this PR)](https://github.com/stjude-rust-labs/sprocket/pull/888/commits/8a529852f630bc52b3358f1718cad5b1f599a998)

PR #891, opened 2026-06-01
fix(sprocket): separate test cancellation contexts

Serial-ATA:
Since all tests shared the same `CancellationContext`, any failing test would take down any other running tests. Added `CancellationContext::child()` to create individual contexts for each test, while still allowing global cancellation

For external contributors:

For all contributors:

      is implemented.


claymcleod (approved):

claymcleod:
Actually, after approving I noticed that this is not quite right—the `state()` and `user_cancelled()` haven't been updated to reflect this relationship, so it gives incorrect hierarchical information. This is further confounded by the fact that cancellations in the parent (even user cancellations) always show up as error cancellations in the child. Let me do a bit of hacking to see what I can come up with.

claymcleod:
@Serial-ATA @peterhuene what do we think of this solution?

crates/wdl-engine/src/eval.rs
```diff
@@ -161,14 +164,61 @@ impl CancellationContext {
...
+                let level = level.max(parent_level);
+                let user = if parent_level != CANCELLATION_STATE_NOT_CANCELED {
+                    parent_user
+                } else {
+                    user
+                };
+                (level, user)
+            }
+            None => (level, user),
+        }
```
Serial-ATA, line 216:
```suggestion
        let user_initiated =
            level != CANCELLATION_STATE_NOT_CANCELED && (raw & CANCELLATION_STATE_ERROR == 0);

        match &self.parent {
            Some(parent) => {
                let (parent_level, parent_user_initiated) = parent.effective();
                let level = level.max(parent_level);
                let user_initiated = if parent_level != CANCELLATION_STATE_NOT_CANCELED {
                    parent_user_initiated
                } else {
                    user_initiated
                };
                (level, user_initiated)
            }
            None => (level, user_initiated),
        }
```

For consistency

crates/wdl-engine/src/eval.rs
```diff
@@ -161,14 +164,61 @@ impl CancellationContext {
...
+        match &self.parent {
+            Some(parent) => {
+                let (parent_level, parent_user) = parent.effective();
+                let level = level.max(parent_level);
+                let user = if parent_level != CANCELLATION_STATE_NOT_CANCELED {
+                    parent_user
+                } else {
+                    user
+                };
+                (level, user)
```
Serial-ATA: If a child errors and a parent is cancelled _after_, the `user`value is wrong for the child:

```rust
let parent = CancellationContext::new(FailureMode::Fast);
let child = parent.child(FailureMode::Fast);

// This is fine
child.error(&EvaluationError::Canceled);
assert_eq!(child.state(), CancellationContextState::Canceling);
assert!(!child.user_canceled());

assert_eq!(parent.cancel(), CancellationContextState::Canceling);

// Now the parent trampled the user cancellation, this fails
assert!(!child.user_canceled());
```

Serial-ATA (commented):
@claymcleod

peterhuene (approved):

tests/cli/test/cancellation/inputs/source.wdl
```diff
@@ -0,0 +1,17 @@
+version 1.3
```
peterhuene, line 6: Nit: could we add some comments here explaining the intention of the test?

Serial-ATA: Ah, yeah good call

PR #908, opened 2026-06-03
feat(wdl-grammar): consolidate unknown tokens

Serial-ATA:
For the following:

```wdl
version 1.3

task foo {
    ;;;;
}
```

The parser used to emit a diagnostic for each unknown character:

```
error: an unknown token was encountered
  ┌─ source.wdl:4:5
  │
4 │     ;;;;
  │     ^ this is not a supported WDL token

error: an unknown token was encountered
  ┌─ source.wdl:4:6
  │
4 │     ;;;;
  │      ^ this is not a supported WDL token

error: an unknown token was encountered
  ┌─ source.wdl:4:7
  │
4 │     ;;;;
  │       ^ this is not a supported WDL token

error: an unknown token was encountered
  ┌─ source.wdl:4:8
  │
4 │     ;;;;
  │        ^ this is not a supported WDL token
```

Now it groups them together:

```
error: unknown start of token: ;
  ┌─ source.wdl:4:5
  │
4 │     ;;;;
  │     ^^^^ this is not a supported WDL token
```

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-grammar/src/parser.rs
+++ b/crates/wdl-grammar/src/parser.rs
@@ -1068,12 +1068,6 @@ where
                 }
             }
             Err(_) => {
-                let unexpected_char = self
-                    .source(span)
-                    .chars()
-                    .next()
-                    .expect("should exist, we just parsed this character");
-
                 let mut unknown_span = span;
                 let lexer = self.lexer.as_mut().expect("should have a lexer");
 
--- a/crates/wdl-grammar/src/parser.rs
+++ b/crates/wdl-grammar/src/parser.rs
@@ -1092,11 +1092,10 @@ where
                 }
 
                 self.diagnostic(
-                    Diagnostic::error(format!("unknown start of token: {unexpected_char}"))
-                        .with_label(
-                            Self::unsupported_token_text(self.source(span)),
-                            unknown_span,
-                        ),
+                    Diagnostic::error("an unknown token was encountered").with_label(
+                        Self::unsupported_token_text(self.source(span)),
+                        unknown_span,
+                    ),
                 );
 
                 Event::Token {
--- a/crates/wdl-grammar/src/parser.rs
+++ b/crates/wdl-grammar/src/parser.rs
@@ -1058,29 +1058,65 @@ where
                     return Some((token, span));
                 }
 
+                if peeked {
+                    self.lexer.as_mut().expect("should have a lexer").next();
+                }
+
                 Event::Token {
                     kind: token.into_syntax(),
                     span,
                 }
             }
             Err(_) => {
+                let lexer = self.lexer.as_ref().expect("should have a lexer");
+                let remaining_source_len = lexer.source_len() - span.start();
+
+                // Consecutive unknown tokens of the same type get condensed into a single
+                // diagnostic
+                let mut source_chars = self
+                    .source(Span::new(span.start(), remaining_source_len))
+                    .chars();
+                let unexpected_char = source_chars
+                    .next()
+                    .expect("should exist, we just parsed this character");
+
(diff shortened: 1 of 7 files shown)


crates/wdl-grammar/src/parser.rs
```diff
@@ -1058,29 +1058,65 @@ where
...
                     span,
                 }
             }
             Err(_) => {
+                let lexer = self.lexer.as_ref().expect("should have a lexer");
+                let remaining_source_len = lexer.source_len() - span.start();
+
+                // Consecutive unknown tokens of the same type get condensed into a single
+                // diagnostic
+                let mut source_chars = self
```
peterhuene: I wonder if we should just merge any unknown tokens together regardless of whether or not they are the same character.

Do something like:

* If we're peeked, move next (i.e. consume the current unknown token).
* Peek in a loop:
  * Break out if end-of-input or if the peeked token's kind is _not_ `SyntaxKind::Unknown`.
  * Adjust the span's end to the peeked token's end.
  * Move the lexer next.

What do you think about that approach?

Serial-ATA: @peterhuene By that, do you mean produce a single event but still emit two different diagnostics like:

```wdl
version 1.3

task foo {
    ;;;;````
}
```

```
error: unknown start of token: ;
  ┌─ source.wdl:4:5
  │
4 │     ;;;;
  │     ^^^^ this is not a supported WDL token

error: unknown start of token: `
  ┌─ source.wdl:4:9
  │
4 │     ````
  │     ^^^^ this is not a supported WDL token
```

or group everything together?

```
error: unknown start of token: ;
  ┌─ source.wdl:4:5
  │
4 │     ;;;;````
  │     ^^^^^^^^ this is not a supported WDL token
```

peterhuene: I was thinking the grouped like:

```
error: an unknown token was encountered
  ┌─ source.wdl:4:5
  │
4 │     ;;;;````
  │     ^^^^^^^^ this is not supported by WDL
```

And there would being a single `Event:Token` for that span.

peterhuene:
For the purpose of fuzzing, we should probably have some configuration around the maximum number of diagnostics that the parser will push given how it's currently unbounded, which isn't good.

crates/wdl-grammar/src/parser.rs
Serial-ATA: @peterhuene done in [`705f495` (this PR)](https://github.com/stjude-rust-labs/sprocket/pull/908/commits/705f4954cb6313f8d3f866fd22e44be25e894891)

crates/wdl-grammar/src/parser.rs
```diff
@@ -1058,29 +1058,60 @@ where
...
+                        unknown_span.start(),
+                        peeked_span.end() - unknown_span.start(),
+                    );
+                    lexer.next();
+                }
+
                 self.diagnostic(
-                    Diagnostic::error("an unknown token was encountered")
-                        .with_label(Self::unsupported_token_text(self.source(span)), span),
+                    Diagnostic::error(format!("unknown start of token: {unexpected_char}"))
```
peterhuene: Nit: I think we can keep the old error message; the span makes it clear what the "token" was comprised of, in my opinion.

peterhuene (commented):
Looks great, just a nit on the diagnostic's error message itself.

crates/wdl-grammar/src/parser.rs
Serial-ATA: sure, reverted

peterhuene (approved):

PR #914, opened 2026-06-04
fix: treat local paths as urls

adthrasher:
Fixes #912. Accepted local paths can also be specified as URLs. This is useful for the future `retry` functionality because local source paths get converted to `file://` URLs internally. Allowing specified paths to also be specified as URLs, the `retry` function can reuse the prior submission cleanly.

Before submitting this PR, please make sure:

For external contributors:

- [ ] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [ ] You have not used AI on any parts of this pull request.
- [ ] You have added a few sentences describing the PR here.
- [ ] Your code builds clean without any errors or warnings.

For all contributors:

- [ ] You have added tests (when appropriate).
- [x] You have added an entry in the CHANGELOG (when appropriate).
- [ ] You have updated the README or other documentation to account for these changes (when appropriate).
- [ ] You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).

For PRs containing lint rule changes:

- [ ] You have updated any and all effected entries within `RULES.md`.
- [ ] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

PR #915, opened 2026-06-04
feat: workflow management

### LINKED ISSUE #910: `dev server`: workflow management
Currently, a user creates a server with `dev server` and they can submit workflows/tasks with `dev submit`. Any other interaction with the run must be done via the API.

It would be nice for users to have a more fully featured CLI for interacting with the server. In [Oliver](https://github.com/stjudecloud/oliver), we had `inspect <workflow-id>`, `status [-d]`, `kill`, and `retry` subcommands, in addition to `submit`. `retry` also accepted command-line arguments that would override those in the original submission to make updating resources, etc., a painless experience.

Much of this can be done via API calls.

**Additional context**

adthrasher:
Adds utility functions for interacting with a `sprocket dev server`.

Adds `retry`, `cancel`, `inspect`, and `status` subcommands. Also adds an API endpoint for listing `tasks` associated with a `run`.

Right now, all commands are under `sprocket dev`. We may want to consider a new subcommand `sprocket server` with the existing `server` command becoming `server start` and the other new commands similarly nested under `server`.

Depends on #914.

For external contributors:

For all contributors:

      is implemented.


src/commands/retry.rs
```diff
@@ -0,0 +1,203 @@
...
+    // `validate_inputs` with only the override slice (a partial set) would
+    // incorrectly fail with "missing required input" for every untouched input.
+    //
+    // WDL source re-analysis above still catches structural changes to the
+    // WDL. Value-level validation for the complete merged set is performed by
+    // the server when the run executes.
+    //
+    // Note: `@file` overrides and multi-value array appends are not currently
+    // supported in the retry path.
+    for item in &args.overrides {
```
claymcleod: I see the comment here, but we should support `@file` overrides and multi-value array appends before we ship this feature IMO.

src/commands/status.rs
```diff
@@ -0,0 +1,246 @@
...
+    ///
+    /// Valid values: `queued`, `running`, `completed`, `failed`, `canceling`,
+    /// `canceled`. Only used when no `RUN_ID` is provided.
+    #[clap(long, value_name = "STATUS")]
+    status: Option<String>,
+
+    /// Maximum number of runs to return per page.
+    ///
+    /// Only used when no `RUN_ID` is provided.
+    #[clap(long, value_name = "N", default_value = "100")]
```
claymcleod: `--limit 0` can make the list path loop forever: the server returns `next_token=0` when runs exist, and `status_list` keeps requesting the same page. Let's validate this as `>= 1` in the CLI, and ideally have the API reject non-positive limits as well.

src/server/api/v1/tasks.rs
```diff
@@ -211,6 +277,89 @@ pub async fn list_tasks(
...
+            .parse::<i64>()
+            .map_err(|_| Error::BadRequest(format!("invalid `next_token`: `{}`", t)))?,
+        None => 0,
+    };
+    let limit = query.limit.unwrap_or(100);
+
+    let response = send_command(&state.run_manager_tx, |rx| RunManagerCmd::ListTasks {
+        run_id: Some(id),
+        status: query.status,
+        limit: query.limit,
```
claymcleod: This endpoint should reject non-positive `limit` values and negative `next_token` values before forwarding them to the database. SQLite treats `LIMIT -1` as unbounded, and `limit=0` can produce a repeated pagination token.

src/commands/cancel.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Arguments for the `cancel` subcommand.
+#[derive(Parser, Debug)]
+#[command(author, version, about)]
+pub struct Args {
+    /// The run to cancel.
+    ///
+    /// May be a UUID or the human-readable generated name of the run (e.g.
+    /// `happy-dolphin-42`).
+    #[clap(value_name = "RUN_ID")]
```
claymcleod: Since this accepts either a UUID or the generated run name, `RUN` would be a clearer value name than `RUN_ID`. Could we use that across the new `cancel`, `inspect`, `retry`, and `status` commands so the signatures read as `<RUN>` / `[RUN]` while keeping the commands under `dev` for now?

src/commands.rs
```diff
@@ -132,12 +137,20 @@ pub enum Commands {
 #[derive(Subcommand, Debug)]
 #[allow(clippy::large_enum_variant)]
 pub enum DevCommands {
+    /// Cancel a running or queued run.
+    Cancel(cancel::Args),
```
claymcleod: Let's nest these run-management commands under `dev server` instead of adding more flat `dev` subcommands. The shape would be `sprocket dev server start`, `sprocket dev server submit`, `sprocket dev server status [RUN]`, `sprocket dev server inspect <RUN>`, `sprocket dev server cancel <RUN>`, and `sprocket dev server retry <RUN> [OVERRIDES]`, which keeps the namespace under `dev` for now while grouping the server lifecycle and server-client actions together.

src/commands/client.rs
```diff
@@ -0,0 +1,191 @@
...
+use crate::config::Config;
+use crate::server::ErrorResponse;
+use crate::server::ListRunsResponse;
+use crate::server::ListTasksResponse;
+use crate::server::RunTaskCountsResponse;
+use crate::server::Task;
+
+/// CLI arguments for connecting to a Sprocket server instance.
+#[derive(ClapArgs, Debug)]
+pub struct SprocketClientConnectionArgs {
```
claymcleod: If these commands are moving under `dev server`, `ServerConnectionArgs` would be a clearer and shorter name for this shared argument group.

tests/api/tasks.rs
```diff
@@ -0,0 +1,162 @@
...
+/// Seed a run with the given name and return its UUID.
+async fn seed_run(db: &Arc<dyn Database>, session_id: Uuid, name: &str) -> Uuid {
+    let run_id = Uuid::new_v4();
+    db.create_run(run_id, session_id, name, "test.wdl", Some("wf"), "{}")
+        .await
+        .unwrap();
+    run_id
+}
+
+#[sqlx::test]
```
claymcleod, line 80: These tests cover `/runs/{id}/tasks/counts`, but the new `/runs/{id}/tasks` route is still untested. Let's add coverage for run filtering, status filtering, pagination, and the invalid `limit` / `next_token` cases called out above.

src/commands/client.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    Err(CommandError::Single(anyhow::anyhow!(msg)))
+}
+
+/// Fetches the per-status task counts for a run.
+///
+/// Queries `GET /api/v1/runs/{uuid}/tasks/counts`. The endpoint reports
+/// all-zero counts (rather than an error) for unknown runs.
+pub async fn fetch_task_counts(base_url: &str, uuid: Uuid) -> CommandResult<RunTaskCountsResponse> {
+    let url = format!("{base_url}/api/v1/runs/{uuid}/tasks/counts");
```
claymcleod: It might be worth having the server module expose route templates/builders and using those from the CLI client instead of duplicating API path strings here. For example, the server could define `RUN_TASK_COUNTS: &str = "/runs/{id}/tasks/counts"` for router/OpenAPI registration plus `run_task_counts_path(id: Uuid) -> String` for clients; then route-shape changes become harder to miss because the router, docs, and client all refer to the same named route.

src/commands/cancel.rs
```diff
@@ -0,0 +1,60 @@
...
+    let body: CancelRunResponse = resp
+        .json()
+        .await
+        .context("failed to deserialize cancel response")?;
+
+    println!(
+        "Run {uuid} has been signaled for cancellation.",
+        uuid = body.uuid,
+    );
+    println!(
```
claymcleod, line 41: This note should only print when the server is actually running in slow-failure mode. Could we add a small server metadata/config route that reports the active failure mode, then have `cancel` query it and conditionally show this message? That keeps the CLI from warning about slow cancellation behavior when the server is configured for fast failure.

src/commands/cancel.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let body: CancelRunResponse = resp
+        .json()
+        .await
+        .context("failed to deserialize cancel response")?;
+
+    println!(
+        "Run {uuid} has been signaled for cancellation.",
+        uuid = body.uuid,
+    );
```
claymcleod, line 44:
```suggestion
    println!(
        "Run `{uuid}` has been signaled for cancellation.",
        uuid = body.uuid,
    );
```

src/commands/client.rs
```diff
@@ -0,0 +1,191 @@
...
+        0 => Err(CommandError::Single(anyhow::anyhow!(
+            "no run found with name or UUID `{input}`"
+        ))),
+        1 => Ok(matches[0]),
+        _ => {
+            let ids = matches
+                .iter()
+                .map(|u| u.to_string())
+                .collect::<Vec<_>>()
+                .join(", ");
```
claymcleod, line 205:
```suggestion
            let ids = matches
                .iter()
                .map(|u| format!("`{u}`"))
                .collect::<Vec<_>>()
                .join(", ");
```

src/commands/inspect.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    if let Some(summary) = task_counts_summary(&counts, colorize) {
+        field!("Tasks:", summary);
+    }
+
+    if let Some(target) = &run.target {
+        field!("Target:", target);
+    }
+
+    field!("Source:", run.source);
```
claymcleod:
```suggestion
    field!("Name:", format!("`{}`", run.name));
    field!("UUID:", format!("`{}`", run.uuid));
    field!("Status:", status_display);

    if let Some(summary) = task_counts_summary(&counts, colorize) {
        field!("Tasks:", summary);
    }

    if let Some(target) = &run.target {
        field!("Target:", format!("`{target}`"));
    }

    field!("Source:", format!("`{}`", run.source));
```

src/commands/inspect.rs
```diff
@@ -0,0 +1,488 @@
...
+    if let Some(directory) = &run.directory {
+        field!("Directory:", directory);
+
+        // Note the outputs file location if outputs are available.
+        if run.outputs.is_some() {
+            let outputs_path = std::path::Path::new(directory).join("outputs.json");
+            field!(
+                "Outputs:",
+                format!("available at {}", outputs_path.display())
+            );
```
claymcleod, line 295:
```suggestion
        field!("Directory:", format!("`{directory}`"));

        // Note the outputs file location if outputs are available.
        if run.outputs.is_some() {
            let outputs_path = std::path::Path::new(directory).join("outputs.json");
            field!(
                "Outputs:",
                format!("available at `{}`", outputs_path.display())
            );
```

src/commands/status.rs
```diff
@@ -0,0 +1,246 @@
...
+    println!(
+        "{short_uuid}  {name:<30}  {status}{elapsed}",
+        short_uuid = &run.uuid.to_string()[..8],
+        name = run.name,
+        status = status_display,
+        elapsed = elapsed,
+    );
+
+    if let Some(target) = &run.target {
+        println!("{:>14}  {target}", "Target:");
```
claymcleod, line 145:
```suggestion
    let name_display = format!("`{}`", run.name);
    println!(
        "`{short_uuid}`  {name:<32}  {status}{elapsed}",
        short_uuid = &run.uuid.to_string()[..8],
        name = name_display,
        status = status_display,
        elapsed = elapsed,
    );

    if let Some(target) = &run.target {
        println!("{:>14}  `{target}`", "Target:");
    }
```

src/commands/status.rs
```diff
@@ -0,0 +1,246 @@
...
+            .unwrap_or(run.created_at)
+            .format("%Y-%m-%d %H:%M:%S UTC");
+
+        println!(
+            "{short_uuid}  {name:<30}  {status:<12}  {target:<20}  {timestamp}",
+            short_uuid = &run.uuid.to_string()[..8],
+            name = run.name,
+            status = status_display,
+            target = target,
+            timestamp = timestamp,
```
claymcleod:
```suggestion
        let target = run
            .target
            .as_deref()
            .map(|target| format!("`{target}`"))
            .unwrap_or_else(|| "-".to_string());
        let name_display = format!("`{}`", run.name);
        let timestamp = run
            .completed_at
            .or(run.started_at)
            .unwrap_or(run.created_at)
            .format("%Y-%m-%d %H:%M:%S UTC");

        println!(
            "`{short_uuid}`  {name:<32}  {status:<12}  {target:<22}  {timestamp}",
            short_uuid = &run.uuid.to_string()[..8],
            name = name_display,
            status = status_display,
            target = target,
            timestamp = timestamp,
        );
```

src/commands/client.rs
```diff
@@ -0,0 +1,191 @@
...
+/// Queries `GET /api/v1/runs/{uuid}/tasks`. The returned tasks are sorted by
+/// creation time ascending (oldest first), so callers see them in execution
+/// order regardless of the server's page ordering.
+pub async fn fetch_run_tasks(base_url: &str, uuid: Uuid) -> CommandResult<Vec<Task>> {
+    let client = reqwest::Client::new();
+    let mut next_token: Option<String> = None;
+    let mut tasks = Vec::new();
+
+    loop {
+        let mut url = format!("{base_url}/api/v1/runs/{uuid}/tasks?limit=100");
```
claymcleod: nit: Could we make this `100` a named client page-size constant and reuse it for both paginated helpers in this file?

```suggestion
        let mut url = format!("{base_url}/api/v1/runs/{uuid}/tasks?limit={DEFAULT_PAGE_SIZE}");
```

claymcleod (changes requested):

src/commands/status.rs
adthrasher: Screenshot 2026-06-29 at 10 09 55 AM
As a user experience, I think the backticks are worse.

src/commands/inspect.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    if let Some(directory) = &run.directory {
+        field!("Directory:", format!("`{directory}`"));
+
+        // Note the outputs file location if outputs are available.
+        if run.outputs.is_some() {
+            let outputs_path = std::path::Path::new(directory).join("outputs.json");
+            field!(
+                "Outputs:",
+                format!("available at `{}`", outputs_path.display())
```
adthrasher, line 294: This is insufficient for users to find the outputs easily. The path to the `out` directory needs to be reported somewhere, possibly in the server metadata endpoint.

```
  Directory:  `runs/variant_calling/2026-07-07_133441831049615`
    Outputs:  available at `runs/variant_calling/2026-07-07_133441831049615/outputs.json`
```

src/commands/status.rs
claymcleod: I agree with your assessment. Let's change it back if that's okay.

src/commands/status.rs
```diff
@@ -0,0 +1,253 @@
...
+            .unwrap_or_else(|| "-".to_string());
+        let name_display = format!("`{}`", run.name);
+        let timestamp = run
+            .completed_at
+            .or(run.started_at)
+            .unwrap_or(run.created_at)
+            .format("%Y-%m-%d %H:%M:%S UTC");
+
+        println!(
+            "`{short_uuid}`  {name:<32}  {status:<12}  {target:<22}  {timestamp}",
```
claymcleod: These columns are padded to fixed widths (`name:<32`, `target:<22`) but never truncated, so any run name or target longer than its column overflows and pushes the rest of the row out of alignment. Worth truncating with an ellipsis (e.g. `variant_calling_wi…`) so the table stays aligned regardless of input length. The single-run summary above doesn't have this problem since it's not columnar—this is just for the list view.

adthrasher: For subsequent calls, the user needs either the full UUID or the full name. I'm currently printing just the short UUID. So we'll need to ensure we print at least one in full.

adthrasher: Practically speaking, the longest word in the adjective [list](https://github.com/allenap/rust-petname/blob/master/words/large/adjectives.txt) is 25 characters. The longest noun in the [list](https://github.com/allenap/rust-petname/blob/master/words/large/nouns.txt) is 11 characters. So with the 6 hex and two `-` characters, the upper limit is 44 characters.

claymcleod (approved):
Needs `CHANGELOG` entries, but otherwise good.

PR #917, opened 2026-06-05
chore(wdl-lsp): client notification tests

Serial-ATA:
With #888 we can easily create our own clients for tests, making it possible to test the notification handling.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lsp/tests/notifications/did_change_configuration.rs
```diff
@@ -0,0 +1,197 @@
...
+        .expect("failed to get diagnostics");
+
+    match diagnostic_report {
+        DocumentDiagnosticReportResult::Report(DocumentDiagnosticReport::Full(report)) => {
+            let items = report.full_document_diagnostic_report.items;
+            assert_eq!(items.len(), 1);
+            assert_eq!(
+                items[0].code,
+                Some(NumberOrString::String(String::from("SnakeCase")))
+            );
```
claymcleod, line 149: This assertion is too strict for the fixture: the test currently fails because linting produces another valid diagnostic alongside `SnakeCase`. We should assert that `SnakeCase` is present instead of assuming it is the only result.

```suggestion
            assert!(
                items.iter().any(|item| {
                    matches!(
                        item.code.as_ref(),
                        Some(NumberOrString::String(code)) if code == "SnakeCase"
                    )
                }),
                "expected a `SnakeCase` diagnostic, got: {items:?}"
            );
```

crates/wdl-analysis/src/graph.rs
```diff
@@ -574,11 +574,74 @@ impl DocumentGraph {
...
+                if dependent == index {
+                    return;
+                }
+
+                let dep_node = graph.get_mut(dependent);
+                trace!(
+                    "document `{uri}` needs to be reanalyzed",
+                    uri = dep_node.uri
+                );
+                dep_node.reanalyze();
```
claymcleod: We need to rebuild more than the analyzed document here. `reanalyze()` only clears `document` and `analysis_error`, but the dependent still keeps its old parsed imports while `delete_node()` removes the imported graph node. The next analysis can then hit `resolve_import()`'s `expect("missing import node in graph")` instead of reporting a missing-import diagnostic, so deletion should either keep a placeholder import node or force dependents through the parse/dependency-edge update path.

crates/wdl-lsp/tests/requests/diagnostics.rs
```diff
@@ -93,27 +98,18 @@ async fn no_baseline_reports_all_diagnostics() {
...
+            let hash = blake3::hash(b"x").to_hex();
+            let baseline = Baseline::new(vec![
+                BaselineEntry::new("InputName", "source.wdl", hash),
+                BaselineEntry::new("UnusedInput", "source.wdl", hash),
+            ])
+            .with_base_dir(workspace.to_path_buf());
+
+            server_options.baseline = Some(baseline);
+        },
+    );
```
claymcleod: This test lost the lint-enabled setup during the builder migration, so it now passes without proving that the baseline suppresses anything. Let's keep linting enabled here like the other baseline tests.

```suggestion
    let mut ctx = TestContextBuilder::new("baseline")
        .user_options(UserOptions {
            lint: LintOptions {
                enabled: true,
                ..Default::default()
            },
            ..Default::default()
        })
        .build_with_options_fn(|workspace, server_options, _| {
            let hash = blake3::hash(b"x").to_hex();
            let baseline = Baseline::new(vec![
                BaselineEntry::new("InputName", "source.wdl", hash),
                BaselineEntry::new("UnusedInput", "source.wdl", hash),
            ])
            .with_base_dir(workspace.to_path_buf());

            server_options.baseline = Some(baseline);
        });
```

crates/wdl-lsp/tests/notifications/did_change_configuration.rs
```diff
@@ -0,0 +1,197 @@
...
+    );
+
+    current_level = LevelFilter::TRACE;
+    config.lock().await.log_level = current_level.into();
+    ctx.notify::<DidChangeConfiguration>(DidChangeConfigurationParams {
+        settings: Value::Null,
+    })
+    .expect("failed to send didChangeConfiguration notification");
+
+    // Give the server some time to update...
```
claymcleod: This makes the test timing-dependent: a loaded runner can take longer than `100ms` to process the notification even when the server is correct. It would be more reliable to wait with a deterministic barrier, or poll `log_handle.clone_current()` under `tokio::time::timeout()` until it matches the expected level.

crates/wdl-analysis/src/analyzer.rs
```diff
@@ -360,15 +365,14 @@ where
     /// initialize a thread-local validator.
     ///
     /// The analyzer must be constructed from the context of a Tokio runtime.
-    pub fn new_with_validator<Progress, Return, Validator>(
+    pub fn new_with_validator<Progress, Return>(
```
claymcleod: This leaks the queue's sharing strategy into the public `wdl-analysis` API and makes existing callers wrap ordinary closures in `Arc`. We can keep this constructor generic over `Fn() -> Validator`, then erase it to `ValidatorFn` internally before storing it in the queue.

crates/wdl-analysis/src/queue.rs
```diff
@@ -471,6 +492,25 @@ where
 
                     completed.send(()).ok();
                 }
+                Request::Delete(DeleteRequest {
+                    documents,
+                    completed,
+                }) => {
+                    let start = Instant::now();
+                    debug!(
+                        "received request to delete {count} documents(s)",
```
claymcleod:
```suggestion
                        "received request to delete {count} document(s)",
```

claymcleod (changes requested):
I left the blocking issues inline. The notification coverage is valuable, but the failing test and graph-deletion invariant need fixes before this is ready to merge.

crates/wdl-analysis/src/analyzer.rs
```diff
@@ -688,6 +719,29 @@ where
         Ok(())
     }
 
+    /// Deletes the specified documents from the analyzer.
+    ///
+    /// This differs from [`Self::remove_documents()`], as a deletion will occur
```
claymcleod: Now that we have both behaviors, `remove_documents()` is a little ambiguous: it really means “unroot these documents, but keep them if imports still reference them,” while `delete_documents()` means “forcefully remove them from the graph.” Could we rename the existing API to something like `unroot_documents()` (or `remove_roots()`) so callers can tell which behavior they are choosing?

crates/wdl-lsp/tests/notifications/did_change.rs
```diff
@@ -0,0 +1 @@
+//! Integration tests for the `workspace/didChange` notification.
```
claymcleod: This module only contains the doc comment right now, so it adds a `mod did_change;` entry without any `textDocument/didChange` coverage. Could we either add the intended `didChange` test here or drop this file until there’s a test to include?

crates/wdl-lsp/src/server.rs
```diff
@@ -1529,30 +1540,39 @@ impl<S: 'static> Server<S> {
...
 
             match event.typ {
                 FileChangeType::CREATED => {
-                    let Some(path) = to_wdl_file_path(&event.uri) else {
+                    let Some(path) = to_existing_wdl_file_path(&event.uri) else {
                         continue;
                     };
 
                     debug!("document `{uri}` has been created", uri = event.uri);
                     added.push(path_to_uri(&path).expect("should convert to uri"));
```
claymcleod, line 1565: This filter drops deleted-directory notifications before they can reach `delete_documents()`. `DocumentGraph::delete()` already supports prefix deletion by converting the URI to a path and deleting every indexed document below it, but a directory URI usually won’t pass `is_wdl_file_uri()`. Could we route deleted directory events through the delete path too, so deleting a watched folder invalidates/removes the WDL files under it?

crates/wdl-analysis/src/graph.rs
```diff
@@ -625,11 +625,76 @@ impl DocumentGraph {
...
+                }
+
+                let dep_node = graph.get_mut(dependent);
+                trace!(
+                    "document `{uri}` needs to be reanalyzed",
+                    uri = dep_node.uri
+                );
+
+                // Drop any dependents down to an unparsed state
+                dep_node.notify_change(true);
```
claymcleod: Passing `true` here clears the dependent node’s pending in-memory change as well as its parse state. If that dependent document is open in the editor with unsaved edits, deleting one of its imports can discard the queued text change and re-analyze from disk instead of the editor buffer. Could this invalidation keep pending changes intact, or otherwise requeue the open document content before forcing reanalysis?

crates/wdl-analysis/src/queue.rs
```diff
@@ -1329,6 +1387,18 @@ where
...
+
+        for uri in uris {
+            graph.delete(&uri);
+        }
+
+        graph.gc();
+    }
+
     /// Awaits the given set of futures while providing progress to the given
     /// callback.
```
claymcleod, line 1403: Starting from `graph.roots()` means the validator swap only invalidates roots and whatever is reachable from them in this traversal, but imported documents can be non-root graph nodes that still produce diagnostics under the new validator. Since this method’s contract says it marks all documents for re-analysis, could we invalidate every graph node instead of walking only from roots?

crates/wdl-lsp/tests/notifications/did_change_watched_files.rs
```diff
@@ -0,0 +1,104 @@
...
+    assert_document_symbol_request(&mut ctx, enum_wdl.clone(), false).await;
+
+    ctx.notify::<DidChangeWatchedFiles>(DidChangeWatchedFilesParams {
+        changes: vec![
+            FileEvent {
+                uri: enum_wdl.clone(),
+                typ: FileChangeType::CREATED,
+            },
+            FileEvent {
+                uri: third_wdl.clone(),
```
claymcleod, line 59: This sends a `DELETED` notification while `third.wdl` still exists on disk. That covers the analyzer removal path, but it doesn’t exercise the real watched-file scenario where the file is gone by the time the server handles the event. Could the test remove `third.wdl` before sending the notification so we also cover the filesystem-deleted behavior?

claymcleod (dismissed):
The notification-test direction looks good, but I left a few comments around document removal semantics, configuration reanalysis, and a couple of test-coverage gaps.

peterhuene (approved):

PR #864, opened 2026-05-20
chore(ci): improve ci speeds by caching dependencies and downloading pre-built binaries

BD103:
Currently it takes about 21 minutes for CI to run on a PR. This due to the duration of the `test` and `engine-test` jobs on Windows, which average 21 and 17 minutes respectively ([you can verify this here](https://github.com/stjude-rust-labs/sprocket/actions/metrics/performance?tab=jobs)). In this PR I cut the average CI time down to about 14 minutes through dependency caching and installing pre-built binaries.

Note that the speedups due to caching will only be visible once this PR is merged, as the cache is created on the `main` branch!

## Improving `engine-test`

In [my baseline run](https://github.com/BD103/sprocket/actions/runs/26170587122/job/76986808401), the `engine-test` job takes 21 minutes to run without a cache. The largest contributors are:

- Compiling and installing `cargo-llvm-cov`: 2.5 minutes
- Compiling and running tests: 11 minutes
- Compressing and uploading a new cache: 6.5 minutes

Instead of compiling `cargo-llvm-cov` from scratch every time, we can install a pre-built copy using [`taiki-e/install-action`](https://github.com/taiki-e/install-action) (which we already use to install `nextest`). Installing the pre-built copy takes only 5 seconds, saving us about 2.5 minutes.

Additionally, I ended up removing caching from this job entirely. The [cache quota is 10 GB](https://github.com/stjude-rust-labs/sprocket/actions/caches), and each cached binary takes about 3 ± 0.5 GB in my experimentation. Because we are caching Linux, Windows, and Mac, it results in `engine-test` alone taking 9 GB of the 10 GB quota. I found it was faster to cache the `test` job and not cache this `engine-test` job, since we cannot cache both.

In the end, I decreased the `engine-test` runtime from 17 minutes on average to 11.5 minutes on average. [Here is the final result.](https://github.com/BD103/sprocket/actions/runs/26188567731/job/77050612176)

## Improving `test`

In [my baseline run](https://github.com/BD103/sprocket/actions/runs/26170587122/job/76986808342), the `test` job takes 23 minutes without caching. The largest contributors are:

- Compiling and installing `cargo-llvm-cov`: 2 minutes
- Compiling and running tests: 20 minutes

Here I also used [`taiki-e/install-action`](https://github.com/taiki-e/install-action) to install a pre-built copy of `cargo-llvm-cov`, saving 2 minutes.

[...]

PR #919, opened 2026-06-07
chore(py): add unit testing, type checking, and code formatting

BD103:
This PR sets up various quality assurance checks for the Python bindings.

- Unit testing using [`pytest`](https://docs.pytest.org)
- Type and stub type checking using [`mypy`](https://mypy.readthedocs.io)
- Code formatting using [`black`](https://black.readthedocs.io)

These tools are automatically installed when you run `maturin develop`, I've included instructions on running them in `python/README.md`. Neither `pytest` nor `black` supports Python 3.9, so I increased the minimum Python version to 3.10 in 2148efb6a265fdb984913890f9bbc36bfa17446f. I also fixed a few typos as a drive-by in b0996e52bbf3bda21b148c10bb060c7a744d73ab.

All of these tools how been added to CI and are run on every pull request (see `python-ci.yml`). If these checks take too long, I may adjust them to only run when Python-specific files are modified.

For external contributors:

For all contributors:


BD103:
Rebased onto main. I was worried #937 would cause issues, but it passed all of my tests locally! @claymcleod could I have a review when you have the time? :)

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,126 @@
...
+        if: ${{ matrix.os == 'windows-latest' }}
+        run: |
+          Add-Content -Value "VIRTUAL_ENV=$PWD\.venv" -Path $Env:GITHUB_ENV
+          Add-Content -Value "$PWD\.venv\Scripts" -Path $Env:GITHUB_PATH
+
+      - name: Activate virtual environment (non-Windows)
+        if: ${{ matrix.os != 'windows-latest' }}
+        run: |
+          echo "VIRTUAL_ENV=$(pwd)/.venv" >> "${GITHUB_ENV}"
+          echo "$(pwd)/.venv/bin" >> "${GITHUB_PATH}"
```
BD103, line 62: This does the same thing as `source .venv/bin/activate` and `.venv\Scripts\activate.bat` in a way that persists environmental variables across steps in Github Actions.

- <https://docs.python.org/3/library/venv.html#how-venvs-work>
- <https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands>

pyproject.toml
```diff
@@ -20,6 +20,23 @@ dynamic = ["version"]
 [project.urls]
 repository = "https://github.com/stjude-rust-labs/sprocket"
 
+# Dependencies for local development, not visible to downstream consumers.
+[dependency-groups]
+# Installed by default with `maturin develop`.
+dev = ["pytest~=9.0", "mypy~=2.1", "black~=26.5"]
+
 [tool.maturin]
+module-name = "sprocket_bio._sprocket_bio"
```
BD103, line 29: This renames the Python extension from `python/sprocket_bio/sprocket_bio.so` to `python/sprocket_bio/_sprocket_bio.so`. The `_` prefix marks the extension as private and not intended to be imported directly. It also makes writing type stubs that pass MyPy's stubtest easier.

python/tests/test_init.py
```diff
@@ -0,0 +1,5 @@
+import sprocket_bio
+
+
+def test_has_package_docs():
```
BD103, line 4: This is a silly test, but it provides an example of what tests look like. Take a look at [Pytest's getting started guide](https://docs.pytest.org/en/stable/getting-started.html) for more examples.

python/sprocket_bio/py.typed
BD103, line 1: This tells static analyzers that this Python package includes type hints (in our case through out type stub `*.pyi` files).

python/tests/test_init.py
```diff
@@ -0,0 +1,5 @@
...
+    assert len(sprocket_bio.__doc__) > 0
```
claymcleod: `Module.__doc__` is `str | None`, so this crashes with `TypeError` rather than a clean assertion failure if the Rust extension ever loses its docstring.

```suggestion
    assert sprocket_bio.__doc__, "sprocket_bio must have a module docstring"
```

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,126 @@
...
+
+      - name: Build package
+        uses: PyO3/maturin-action@v1
+        with:
+          command: develop
+          args: --locked
+
+      - name: Check formatting
+        run: black --check python
```
claymcleod: `black` reads `.py` files directly—it never imports the package, so the Rust toolchain and `maturin develop` are not needed here. This saves several minutes of compile time on every PR.

```suggestion
  format:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v6

      - name: Setup Python
        uses: actions/setup-python@v6
        with:
          python-version: "3.10"

      - name: Install black
        run: pip install black~=26.5

      - name: Check formatting
        run: black --check python
```

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,126 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, macos-latest, ubuntu-latest]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+      
+      - name: Update Rust
+        run: rustup update stable && rustup default stable && rustup component add llvm-tools
```
claymcleod: `llvm-tools` is only needed in `CI.yml` for `cargo-llvm-cov` coverage. None of the Python CI jobs use it.

```suggestion
        run: rustup update stable && rustup default stable
```

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,126 @@
...
+        run: pytest
+
+  type-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+      
+      - name: Update Rust
+        run: rustup update stable && rustup default stable && rustup component add llvm-tools
```
claymcleod: Same as above, `llvm-tools` is not needed here.

```suggestion
        run: rustup update stable && rustup default stable
```

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,126 @@
...
+        if: ${{ matrix.os != 'windows-latest' }}
+        run: |
+          echo "VIRTUAL_ENV=$(pwd)/.venv" >> "${GITHUB_ENV}"
+          echo "$(pwd)/.venv/bin" >> "${GITHUB_PATH}"
+
+      - name: Build package
+        uses: PyO3/maturin-action@v1
+        with:
+          command: develop
+          args: --locked
```
claymcleod, line 68: `CI.yml` uses `Swatinem/rust-cache@v2` with `save-if: ${{ github.ref == 'refs/heads/main' }}` before each maturin build—the same pattern here would avoid a full cold compile on every PR push. The cache step should go after `Update Rust` and before this step.

```suggestion
      - name: Cache Rust build artifacts
        uses: Swatinem/rust-cache@v2
        with:
          save-if: ${{ github.ref == 'refs/heads/main' }}

      - name: Build package
        uses: PyO3/maturin-action@v1
        with:
          command: develop
          args: --locked
```

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,126 @@
+name: Python CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
claymcleod, line 10: Every Rust-only PR currently triggers full Python CI across three platforms—including a Rust compile each time. A `paths:` filter restricts the trigger to PRs that actually touch Python-related files.

```suggestion
  pull_request:
    paths:
      - "python/**"
      - "pyproject.toml"
      - "crates/sprocket-py/**"
      - ".github/workflows/python-ci.yml"
```

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,126 @@
...
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, macos-latest, ubuntu-latest]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
```
claymcleod: Trailing whitespace on this blank line.

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,126 @@
...
+      - name: Run tests
+        run: pytest
+
+  type-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
```
claymcleod: Trailing whitespace on this blank line.

python/sprocket_bio/__init__.py
```diff
@@ -1,5 +1,8 @@
+# Import the Python extension (`sprocket_py`).
```
claymcleod: `sprocket_py` is neither the old module name nor the new one—looks like a copy-paste artifact.

```suggestion
# Import the Python extension (`_sprocket_bio`).
```

claymcleod (changes requested):

.github/workflows/python-ci.yml
BD103: I don't think this would benefit us. All of the Python workflows take less than 7 minutes, while the Rust CI takes an average of 15 minutes to complete. Speeding up the Python workflow would have little visible effect on overall CI speeds.

Additionally, part of the reason why the Rust CI only takes 15 minutes is due its use of caching (#864), but [the cache for a single run takes up our entire 10 GB quota](https://github.com/stjude-rust-labs/sprocket/actions/caches). If we cached the Python CI, it would be taking up valuable space and cause the Rust CI cache to get evicted, which would result in CI getting _slower_ overall.

claymcleod (approved):

PR #922, opened 2026-06-08
feat(wdl-grammar): diagnostic brace tracking

Serial-ATA:
This adds more context to unclosed brace diagnostics, pretty much doing the exact same thing as `rustc`.

For the following:

```wdl
version 1.3

task foo {
    input {
        Map[String, Array[Int] foo
}
```

Before we produced:

```wdl
error: expected `]`, but found identifier
  ┌─ tmp.wdl:5:32
  │
5 │         Map[String, Array[Int] foo
  │            -                   ^^^ unexpected identifier
  │            │                    
  │            this `[` is not matched

error: expected `}`, but found end of input
  ┌─ tmp.wdl:7:1
  │
3 │ task foo {
  │          - this `{` is not matched
  ·
7 │ 
  │ ^ unexpected end of input

error: failing due to 2 errors
```

With this, it now produces:

```wdl
error: expected `{`, but found end of input
  ┌─ tmp.wdl:3:10
  │
3 │ task foo {
  │          ^ unclosed delimiter
4 │     input {
  │           - this delimiter might not be properly closed...
5 │         Map[String, Array[Int] foo
6 │ }
  │ - ...as it matches this but it has different indentation

error: expected `]`, but found identifier
  ┌─ tmp.wdl:5:12
  │
5 │         Map[String, Array[Int] foo
  │            ^             -   - --- unexpected identifier
  │            │             │   │  
  │            │             │   ...matches this closing brace
  │            │             this opening brace...
  │            this `[` is not matched

error: failing due to 2 errors
```

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-grammar/src/grammar/v1.rs
+++ b/crates/wdl-grammar/src/grammar/v1.rs
@@ -2209,6 +2209,7 @@ fn call_statement(
         // Given the optional `input:` that we need to parse after the open brace, we
         // unfortunately can't use `Parser::matching_delimited` here
         let open_span = parser.require(Token::OpenBrace);
+        parser.push_open_delimiter(open_span, Token::OpenBrace);
 
         if parser.next_if(Token::InputKeyword).is_some() {
             expected!(parser, marker, Token::Colon);
@@ -2223,6 +2224,7 @@ fn call_statement(
         );
 
         parser.consume_close_token(Token::OpenBrace, open_span, Token::CloseBrace);
+        parser.pop_open_delimiter();
     }
 
     marker.complete(parser, SyntaxKind::CallStatementNode);
--- a/crates/wdl-grammar/src/grammar/v1.rs
+++ b/crates/wdl-grammar/src/grammar/v1.rs
@@ -710,7 +710,7 @@ fn import_statement(
     // Form 1 suffix: optional `as <alias>` and zero-or-more `alias` clauses.
     // Forms 2 and 3 do not accept either.
     if !has_selection {
-        if parser.next_if(Token::AsKeyword) {
+        if parser.next_if(Token::AsKeyword).is_some() {
             ident!(parser, marker, "import namespace");
         }
 
@@ -749,7 +749,7 @@ fn import_member(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marker,
     expected_in!(parser, marker, ANY_IDENT, "selected member");
     parser.update_last_token_kind(SyntaxKind::Ident);
 
-    if parser.next_if(Token::AsKeyword) {
+    if parser.next_if(Token::AsKeyword).is_some() {
         expected_in!(parser, marker, ANY_IDENT, "member alias");
         parser.update_last_token_kind(SyntaxKind::Ident);
     }
@@ -769,7 +769,7 @@ fn symbolic_module_path(
     expected_in!(parser, marker, ANY_IDENT, "symbolic module path");
     parser.update_last_token_kind(SyntaxKind::Ident);
 
-    while parser.next_if(Token::Slash) {
+    while parser.next_if(Token::Slash).is_some() {
         expected_in!(parser, marker, ANY_IDENT, "symbolic module path component");
         parser.update_last_token_kind(SyntaxKind::Ident);
     }
@@ -1109,7 +1109,7 @@ fn input_item(
     expected_fn!(parser, marker, ty);
     ident!(parser, marker, "input name");
 
-    let kind = if parser.next_if(Token::Assignment) {
+    let kind = if parser.next_if(Token::Assignment).is_some() {
         expected_fn!(parser, marker, expr);
         SyntaxKind::BoundDeclNode
     } else {
@@ -2136,7 +2136,7 @@ fn conditional_else_if_clause(
     marker: Marker,
 ) -> Result<(), (Marker, ParseDiagnostic)> {
     parser.require(Token::ElseKeyword);
-    if parser.next_if(Token::IfKeyword) {
+    if parser.next_if(Token::IfKeyword).is_some() {
         paren!(parser, marker, |parser, _| {
             expected_fn!(parser, expr);
             Ok(())
@@ -2210,7 +2210,7 @@ fn call_statement(
         // unfortunately can't use `Parser::matching_delimited` here
         let open_span = parser.require(Token::OpenBrace);
 
-        if parser.next_if(Token::InputKeyword) {
+        if parser.next_if(Token::InputKeyword).is_some() {
             expected!(parser, marker, Token::Colon);
         }
 
@@ -2233,7 +2233,7 @@ fn call_statement(
 fn call_target(parser: &mut Parser<'_>, marker: Marker) -> Result<(), (Marker, ParseDiagnostic)> {
     ident!(parser, marker, "call target name");
 
-    while parser.next_if(Token::Dot) {
+    while parser.next_if(Token::Dot).is_some() {
         ident!(parser, marker, "call target name");
     }
 
@@ -2268,7 +2268,7 @@ fn call_input_item(
     ident!(parser, marker, "call input key");
     parser.update_last_token_kind(SyntaxKind::Ident);
 
-    if parser.next_if(Token::Assignment) {
+    if parser.next_if(Token::Assignment).is_some() {
         expected_fn!(parser, marker, expr);
     }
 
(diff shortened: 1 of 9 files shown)


crates/wdl-grammar/src/parser.rs
```diff
@@ -679,23 +701,173 @@ where
...
+                    Span::new(open.start(), close.start() - open.start()),
+                    same_indent,
+                )
+            })
+            .collect();
+
+        // Sort by `start` position, working outside-in
+        matched_spans.sort_by_key(|(open, ..)| open.start());
+
+        for i in 0..matched_spans.len() {
```
claymcleod, line 794: We should avoid quadratic work here. The parser runs on untrusted WDL input, so a large file with many matched delimiters followed by one mismatch can make this diagnostic path compare every matched span against every later span. Could we keep enough stack/indent metadata while consuming delimiters to select the suspicious candidate in one pass instead of post-processing all prior pairs?

crates/wdl-grammar/src/parser.rs
```diff
@@ -679,23 +701,173 @@ where
...
+    }
+
+    /// Create a diagnostic reporting all unclosed delimiters up to EOF.
+    fn eof_err(&mut self, open: T) {
+        if self.diagnostic_context.eof {
+            return;
+        }
+
+        const UNCLOSED_DELIMITER_SHOW_LIMIT: usize = 3;
+        let mut diagnostic = self.unexpected_inner(open.describe(), None, false);
```
claymcleod: I think this should use the expected closing delimiter, or a dedicated unclosed-delimiter headline. Passing `open.describe()` makes an unclosed `{` report `expected `{`, but found end of input`, even though the parser is missing `}`; that makes the primary diagnostic point users at the wrong token.

crates/wdl-grammar/src/parser.rs
```diff
@@ -648,10 +660,20 @@ where
 
         cb(self, open_span)?;
```
claymcleod: It might be worth guarding the delimiter stack so callback errors cannot leave stale state behind. `matching` pushes `open_span` before this call, but `?` returns before the pop below; a parse error inside the callback can then contaminate later EOF or mismatch diagnostics with an already-abandoned delimiter.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -2210,7 +2210,7 @@ fn call_statement(
         // unfortunately can't use `Parser::matching_delimited` here
         let open_span = parser.require(Token::OpenBrace);
 
-        if parser.next_if(Token::InputKeyword) {
+        if parser.next_if(Token::InputKeyword).is_some() {
```
claymcleod, line 2214: Please make this custom brace path participate in the same delimiter tracking as `matching_delimited`. This branch records `open_span` locally, but it never pushes that span into `open_delimiters`, so EOF inside a `call { ... }` block cannot get the new unclosed-delimiter context that the rest of the parser now emits.

crates/wdl-grammar/src/parser.rs
```diff
@@ -679,23 +701,173 @@ where
...
+                    "...as it matches this but it has different indentation",
+                    Span::new(block_span.end(), 1),
+                );
+        } else if let Some(&(parent_open, parent_close)) =
+            self.diagnostic_context.matching_block_spans.last()
+        {
+            // Optional fallback if we don't have a specific candidate
+            diagnostic.inner = diagnostic
+                .inner
+                .with_label("this opening brace...", parent_open)
```
claymcleod: Consider using generic delimiter wording here, or storing the delimiter kind alongside the spans. `matching_block_spans` records brackets and parentheses too, and the updated `mismatched-bracket` fixture now says `opening brace` / `closing brace` for an `Array[String]` bracket pair.

crates/wdl-grammar/src/parser.rs
```diff
@@ -679,23 +701,173 @@ where
...
+        for (span, _) in &self.diagnostic_context.open_delimiters[..delimiters_shown] {
+            diagnostic.inner = diagnostic.inner.with_label("unclosed delimiter", *span);
+        }
+
+        // Summarize the rest
+        if let Some((span, _)) = self
+            .diagnostic_context
+            .open_delimiters
+            .get(UNCLOSED_DELIMITER_SHOW_LIMIT)
+            && self.diagnostic_context.open_delimiters.len() >= UNCLOSED_DELIMITER_SHOW_LIMIT + 2
```
claymcleod: One small boundary case: exactly four unclosed delimiters labels the first three and then does not summarize the fourth. Since `UNCLOSED_DELIMITER_SHOW_LIMIT` is `3`, this should probably summarize whenever `open_delimiters.len() > UNCLOSED_DELIMITER_SHOW_LIMIT`.

claymcleod (changes requested):
One additional issue does not attach cleanly to the diff: the recovery rewind path should snapshot and restore the new delimiter-tracking state too. That path truncates events and resets the lexer after an abandoned parse attempt, but `open_delimiters` and `matching_block_spans` keep any state the abandoned attempt recorded, so later diagnostics can describe delimiters from code the parser rolled back.

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -2209,8 +2209,9 @@ fn call_statement(
         // Given the optional `input:` that we need to parse after the open brace, we
         // unfortunately can't use `Parser::matching_delimited` here
         let open_span = parser.require(Token::OpenBrace);
+        parser.push_open_delimiter(open_span, Token::OpenBrace);
 
-        if parser.next_if(Token::InputKeyword) {
+        if parser.next_if(Token::InputKeyword).is_some() {
             expected!(parser, marker, Token::Colon);
```
claymcleod, line 2215: This early return can leave the delimiter stack dirty. After `push_open_delimiter()`, `expected!` returns from `call_statement()` immediately if `input` is not followed by `:`, so the `pop_open_delimiter()` below never runs and later EOF or mismatch diagnostics can include a stale `call {` delimiter. Could we put this custom branch behind a small cleanup guard or closure so the delimiter is popped on every exit path?

crates/wdl-grammar/src/parser.rs
```diff
@@ -633,25 +637,61 @@ where
...
+            // Check to see if the close token is immediately following the opening
+            if allow_empty {
+                match parser.peek() {
+                    Some((t, span)) if t == close => {
+                        parser.next();
+                        parser
+                            .diagnostic_context
+                            .matching_block_spans
+                            .push((open_span, span));
+                        parser.diagnostic_context.open_delimiters.pop();
```
claymcleod, line 661: This empty-pair fast path pops the delimiter stack, and then `matching()` pops again after `inner()` returns. I don’t see a current caller passing `allow_empty = true`, so this is not a live bug, but it would drop the parent delimiter if that mode is used later. Could we keep cleanup centralized in the outer `self.pop_open_delimiter()` and remove this inner pop?

crates/wdl-grammar/src/parser.rs
```diff
@@ -679,23 +719,189 @@ where
...
+                .inner
+                .with_label(
+                    "this delimiter might not be properly closed...",
+                    Span::new(block_span.start(), 1),
+                )
+                .with_label(
+                    "...as it matches this but it has different indentation",
+                    Span::new(block_span.end(), 1),
+                );
+        } else if let Some(&(parent_open, parent_close)) =
```
claymcleod, line 825: This fallback can point at an unrelated earlier delimiter pair. `matching_block_spans.last()` is global parser state, so if no suspicious candidate is found for the current error, the diagnostic can borrow context from a previous sibling block; the updated `missing-brace` fixture shows the `task bar` error pointing at `task foo { ... }`. Could we only use matched pairs that overlap the current error/open span, or skip the fallback when there is no relevant pair?

crates/wdl-grammar/src/grammar/v1.rs
```diff
@@ -2479,7 +2481,7 @@ fn pair_or_paren_expr(
 
     expected_fn!(parser, marker, expr);
 
-    if parser.next_if(Token::CloseParen) {
+    if parser.next_if(Token::CloseParen).is_some() {
```
claymcleod, line 2484: This manual `(` / `)` parsing path still bypasses the new delimiter tracking, so pair literals and parenthesized expressions do not contribute to `open_delimiters` or `matching_block_spans`. That means an unclosed parenthesized expression can miss the new EOF context, and a successfully matched pair cannot help nearby mismatch diagnostics. Could we add a small parser helper or guard for manually parsed delimiters and use it here too?

claymcleod (changes requested):
This is close after the earlier review pass, but I left a few more comments around delimiter-state cleanup and making sure the new diagnostic context stays relevant to the current parse error. We also need a determination on whether the existing unresolved performance thread should be addressed before merge.

PR #932, opened 2026-06-12
feat: add fuzz targets

Serial-ATA:
Adds fuzz targets for `wdl-ast` and `wdl-format`

For external contributors:

For all contributors:

      is implemented.


fuzz/fuzz_targets/wdl_lint.rs
```diff
@@ -0,0 +1,72 @@
...
+            return;
+        }
+
+        let mut file = NamedTempFile::new().expect("failed to create temp file");
+        file.write_all(data.as_bytes()).expect("failed to write to temp file");
+
+        let ctx = CONTEXT.get().expect("context should be initialized");
+        let _ = ctx.runtime.block_on(async {
+            ctx.analyzer
+                .analyze_document((), Url::from_file_path(file.path()).unwrap())
```
claymcleod: Let's register the temp file before analyzing it. `analyze_document` returns an empty result for unknown documents, so this target currently stops after parse validation and never exercises the `Linter` visitor. After adding the document, please also make the analyzer config use the same fallback version as the pre-parse so the target filters and analyzes under the same version policy.

fuzz/common.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// Generate an initial corpus directory for the target.
+///
+/// This fills `fuzz/corpus/<target>` with all of the WDL test assets in the
+/// repo.
+pub fn init_corpus_dir(target: &str) -> std::io::Result<()> {
+    let corpus_dir = Path::new(env!("CARGO_MANIFEST_DIR"))
+        .join("corpus")
+        .join(target);
+    if let Err(e) = std::fs::create_dir_all(&corpus_dir) {
```
claymcleod: We should make this seeding idempotent. `create_dir_all` returns `Ok(())` when the directory already exists, so the `AlreadyExists` branch never handles normal repeat runs; each fuzz invocation copies the same `.wdl` files again under fresh UUID names. One option is to skip seeding when `corpus_dir` is already non-empty, or use deterministic filenames derived from the source path.

fuzz/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+These are intended to be run with [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz).
+
+## Usage
+
+```shell
+cargo fuzz run <TARGET>
+```
+
+Where `<TARGET>` is one of the targets in `./fuzz_targets`.
```
claymcleod: Consider listing the concrete `cargo-fuzz` target names here. `cargo fuzz run` takes the binary names (`wdl-ast`, `wdl-format`, and `wdl-lint`), while `./fuzz_targets` contains underscore-named source files.

fuzz/README.md
```diff
@@ -0,0 +1,13 @@
+Fuzz targets for the `*-wdl` crates.
```
claymcleod:
```suggestion
Fuzz targets for the `wdl-*` crates.
```

fuzz/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Usage
+
+```shell
+cargo fuzz run <TARGET>
+```
+
+Where `<TARGET>` is one of the targets in `./fuzz_targets`.
+
+See the [cargo-fuzz book](https://rust-fuzz.github.io/book/introduction.html) for more information.
```
claymcleod:
```suggestion
See the [`cargo-fuzz` book](https://rust-fuzz.github.io/book/introduction.html) for more information.
```

claymcleod (changes requested):

fuzz/common.rs
Serial-ATA: huh, thought it behaved like `create_dir`

fuzz/README.md
Serial-ATA: just gonna point to the CLI, no need to explicitly list them

claymcleod:
Thanks for the quick updates. I rechecked the branch locally, and `cargo check -p fuzz --bins --locked` currently fails: `common.rs` moves `corpus_dir` into `read_dir` and then uses it again, and `wdl_lint.rs` returns different types from the async block after `add_document`. I think this points to a missing CI gate for the new fuzz crate: normal CI does not need to run fuzzing, but it should at least compile the fuzz targets on every PR, e.g., `cargo check -p fuzz --bins --locked`. That would catch this class of issue without making PR CI run a fuzzing campaign.

Separately, we should decide what interval makes sense for actually running the fuzzers. That could be a scheduled job or some other non-PR workflow, but it should be explicit so the targets do more than compile over time.

claymcleod (approved):
This addresses the review feedback well enough for me. Thank you!

PR #933, opened 2026-06-12
fix: allow subpaths of input directories for task outputs.

### LINKED ISSUE #931: eval: output mapping restriction on known guest paths is too strict
**Describe the bug**

1. Create `repro.wdl` with the following contents:

```wdl
version 1.3

task repro {
    input {
        Directory dir
    }

    command <<<>>>

    output {
        Directory sub = join_paths(dir, "sub")
    }
} 
```

2. Run `mkdir -p foo/sub`.
3. Run `sprocket run repro.wdl repro.dir=foo`.

**Expected behavior**

The evaluation should succeed and the output `sub` should be to the host directory `foo/sub`.

**Terminal Output**

Instead evaluation fails:

```bash
error: failed to evaluate output `sub` for task `repro`: guest path `/home/peterhuene/src/peterhuene/sprocket/foo/sub` is not an input or within the task's working directory
   ┌─ repro.wdl:11:19
   │
11 │         Directory sub = join_paths(dir, "sub")
   │                   ^^^
```

**Additional context**

This is due to a having a strict mapping between guest and host paths when evaluating outputs.

Instead, upon failing to find a direct mapping between the guest and host path, we should fall back to a linear scan over the known guest paths and see if the evaluated output is prefixed by any of them. If so, strip the prefix from the guest path and join it with the associated host path.

peterhuene:
This PR fixes guest <-> host path conversion such that it properly handles paths prefixed with guest/host paths.

Additionally, it ensures that the conversion to host paths for output values accepts existing host paths and does the correct guest -> host path translation.

Fixes #931.

For external contributors:

For all contributors:

      is implemented.


peterhuene:
Note: the relevant code changes are in the first commit and fifth commits; the second commit just shuffles what's in `task.rs` around so that there aren't multiple `impl State` in `task.rs`.

As a result, the PR diff looks super messy, so probably best to review the two commits separately.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -556,1490 +569,1583 @@ impl<'a> State<'a> {
...
-        }
+    /// Returns `None` if there is no host path representation of the guest
+    /// path.
+    fn host_path(&self, path: &GuestPath) -> Option<HostPath> {
+        self.path_map.get_by_right(path).cloned().or_else(|| {
+            // A direct mapping between the guest and host wasn't found, so scan for a
+            // matching guest prefix
+            for (host, guest) in self.path_map.iter() {
+                // Check to see if the provided guest path is prefixed by this entry
+                if let Some(remainder) = path
```
claymcleod: Let's normalize the candidate subpath before accepting this prefix match. As written, a guest path like ``${dir}/../sibling`` still matches the mapped input prefix, then `clean(...)` maps it to a host sibling outside the input tree. We should reject remainders that escape the mapped guest/host root, and apply the same containment check after `Url::join`, before returning a host path.

crates/wdl-engine/tests/tasks/subdir-output/source.wdl
```diff
@@ -0,0 +1,23 @@
...
+version 1.3
+
+task test {
+    input {
+        Directory dir
+    }
+
+    # Perform some unnecessary conversions which internally map guest <-> host
+    String s = dir + "/sub"
+    Directory d = s + "/../sub/../sub"
```
claymcleod: It might be worth adding a negative regression for the boundary case here. This fixture verifies that `..` segments which normalize back inside the input are accepted, but it does not cover `..` segments that escape above the mapped input and should still be rejected.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -556,1490 +569,1583 @@ impl<'a> State<'a> {
...
+            .coerce(Some(evaluator.context()), &ty)
             .map_err(|e| runtime_type_mismatch(e, &ty, name.span(), &value.ty(), expr.span()))?;
+        value = value
+            .resolve_paths(
+                ty.is_optional(),
+                self.base_dir.as_local(),
+                Some(self.transferer().as_ref()),
+                &|path| {
+                    // If the path is already a host path, or prefixed with one, return it as-is.
+                    if self.path_map.contains_left(path)
```
claymcleod: We should apply the same containment check to this host-path fast path. A value like `/mapped/input/../sibling` can pass this raw `strip_prefix` test and return as-is, even though the normalized path is outside the mapped input tree.

crates/wdl-engine/tests/tasks/subdir-output/source.wdl
```diff
@@ -0,0 +1,23 @@
+## This is a test to ensure that task outputs may refer to existing sub paths of a directory input.
```
claymcleod:
```suggestion
## This is a test to ensure that task outputs may refer to existing sub-paths of a directory input.
```

crates/wdl-engine/tests/tasks/url-subpath/source.wdl
```diff
@@ -0,0 +1,23 @@
+## This is a test to ensure that task outputs may refer to existing sub paths of URL inputs
```
claymcleod:
```suggestion
## This is a test to ensure that task outputs may refer to existing sub-paths of URL inputs
```

claymcleod (changes requested):

claymcleod (approved):

PR #940, opened 2026-06-15
feat: implement dynamic arguments for Slurm and LSF backends.

peterhuene:
This PR implements dynamically passing command line arguments to `sbatch` and `bsub` to the Slurm and LSF backends, respectively.

To accomplish this, the backend configurations now have "conditional" arguments that each contain a "condition" that is a WDL expression. The first conditional arguments that has its condition evaluate to `true` will be added to the command line arguments when spawning `sbatch` or `bsub`.

The WDL expressions are statically analyzed when deserializing the Sprocket configuration files. This helps to catch errors before any tasks are executed.

Several names are made available to the expressions representing a task's requirements and hints:

* `cpu` - the task's `cpu` requirement (`Float`)
* `memory` - the task's `memory` requirement, in bytes (`Int`)
* `gpu` - the task's `gpu` requirement (`Boolean`)
* `fpga` - the task's `fpga` requirement (`Boolean`)
* `disks` - the task's `disks` requirement, in bytes (`Int`)
* `hint` - the task's `hints` section as an `Object`

As a result of this change, there was a breaking change to how existing unconditional arguments to `sbatch` and `bsub` are specified in `sprocket.toml` so that they are more consistent with the way conditional arguments are specified.

As the Slurm and LSF backends have matured, I'm removing the requirement of enabling experimental features for their use.

For external contributors:

For all contributors:

      is implemented.


test-configs/slurm-apptainer-engine.toml
```diff
@@ -6,4 +6,3 @@ backends.default.default_slurm_partition.max_cpu_per_task = 4
 backends.default.default_slurm_partition.max_memory_per_task = "16 GiB"
 backends.default.extra_sbatch_args = ["--time=1"]
```
claymcleod: This sample still uses the removed `extra_sbatch_args` key. Since `SlurmApptainerBackendConfig` denies unknown fields now, this config no longer loads when used via `SPROCKET_TEST_ENGINE_CONFIG`.

```suggestion
backends.default.sbatch.args = ["--time=1"]
```

crates/wdl-engine/src/config.rs
```diff
@@ -1589,6 +1689,376 @@ impl ApptainerConfig {
...
+                let ty = ExprTypeEvaluator::new(&mut context)
+                    .evaluate_expr(&expr)
+                    .unwrap_or(Type::Union);
+
+                if !context.0.is_empty() {
+                    return Err(context.0);
+                }
+
+                match ty {
+                    Type::Primitive(PrimitiveType::Boolean, false) | Type::Union => {}
```
claymcleod, line 1791: This accepts `Union` as a valid top-level condition type, which lets expressions like `hint.foo` pass config deserialization even though they can evaluate to a non-Boolean later during job submission. We should require the final expression type to be statically `Boolean` so invalid backend conditions fail at config load time; `Union` should only be acceptable inside expressions whose overall type is Boolean, e.g., comparisons or `defined(...)`.

crates/wdl-engine/src/config.rs
```diff
@@ -1589,6 +1689,376 @@ impl ApptainerConfig {
     }
 }
 
+/// Represents a condition in a conditional argument.
+///
+/// The expression is evaluated in a context where a task's evaluated
+/// `requirements` and `hints` sections are made available as `Object`
+/// `requirement` and `hint` variables, respectively.
```
claymcleod: The documented condition scope does not match the implementation: `requirement` is not exposed, and individual computed values like `cpu` and `memory` are exposed instead.

```suggestion
/// The expression is evaluated in a context where a task's computed `cpu`,
/// `memory`, `gpu`, `fpga`, and `disks` values and evaluated `hint` object are
/// available as variables.
```

crates/wdl-engine/src/config.rs
```diff
@@ -1589,6 +1689,376 @@ impl ApptainerConfig {
...
+                    "fpga" => Ok((!self.request.constraints.fpga.is_empty()).into()),
+                    "disks" => Ok(self
+                        .request
+                        .constraints
+                        .disks
+                        .iter()
+                        .map(|(_, s)| *s)
+                        .sum::<i64>()
+                        .into()),
+                    "hint" => Ok(self.request.hints.clone().into()),
```
claymcleod, line 1851: `hint` only exposes the evaluated section object here, so backend conditions can miss input-provided hint overrides that the rest of task constraint evaluation honors through `inputs.hint(...).or_else(|| hints.get(...))`. We should evaluate conditions against the same effective hint/requirement view used by the backend helpers; otherwise a config condition like `defined(hint.foo)` can be false even when the caller supplied `foo` as a task hint override.

claymcleod (changes requested):
It may be worth narrowing the evaluator used for backend conditions. The current implementation passes a full `EvaluationContext` with `base_dir`, `temp_dir`, and `transferer` into the regular WDL expression evaluator, but submit-argument selection only needs a small task metadata context. A restricted `TaskSubmissionContext` shared by queue/partition selection and conditional argument selection would make this path more predictable and avoid giving config predicates access to unrelated WDL runtime capabilities.

crates/wdl-engine/src/config.rs
peterhuene: We have to accept `Union` as, for example, the `hints` section is represented as an `Object` type and access into an `Object` type (e.g. `hint.foo`) is always `Union` via static analysis since the members of the object are not known until runtime.

Adding a `defined` call doesn't help as `defined(hint.foo) && hint.foo` is still `Union` as `hint.foo` is `Union`.

That is to say the static analysis is "best effort" but there will always be expressions that should be supported which can only to validated at runtime.

crates/wdl-engine/src/config.rs
peterhuene: Yep, this comment was out of date after some design changes.

peterhuene:
> It may be worth narrowing the evaluator used for backend conditions. The current implementation passes a full EvaluationContext with base_dir, temp_dir, and transferer into the regular WDL expression evaluator, but submit-argument selection only needs a small task metadata context. A restricted TaskSubmissionContext shared by queue/partition selection and conditional argument selection would make this path more predictable and avoid giving config predicates access to unrelated WDL runtime capabilities.

I think we can probably do a follow-up to further restrict what these expressions have access to from the WDL stdlib.

PR #911, opened 2026-06-04
feat(py): add bindings to `wdl::diagnostics` and friends

BD103:
This PR adds Python bindings for the `wdl::diagnostics` module and other diagnostics-related APIs. This makes it possible to create and emit diagnostics with pure Python:

```python
from sprocket_bio.diagnostics import emit_diagnostics, Mode
from sprocket_bio.grammar import Diagnostic


diagnostics = [
    Diagnostic.error("oh no!")
        .with_help("something went wrong")
        .with_fix("try this instead")
        .with_rule("MyRule"),
]

emit_diagnostics(
    "file.txt",
    "This is the file contents.",
    diagnostics,
    Mode.default(),
    True,
)
```

image

In this PR's current state, I've written everything except for the unit tests. I plan on deferring those to a separate PR, as I need to create Python-specific CI checks first.

- [x] The bindings themselves
- [x] Documentation
- [x] Type stubs
- [ ] Unit tests

## APIs with Bindings

- `wdl::diagnostics::Mode`
- `wdl::diagnostics::emit_diagnostics`
- `wdl::grammar::Span`
- `wdl::grammar::Severity`
- `wdl::grammar::Diagnostic`
- `wdl::grammar::Label`

## APIs Intentionally Skipped

- `wdl::diagnostics::DiagnosticCount` (low impact)
- `wdl::diagnostics::emit_diagnostics_with_backtrace` (may be added with `wdl::engine` bindings)
- `wdl::diagnostics::get_diagnostics_display_config` (low impact)
- Everything else in `wdl::grammar` (will be added later)

Before submitting this PR, please make sure:

For external contributors:

- [x] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [x] You have not used AI on any parts of this pull request.
- [x] You have added a few sentences describing the PR here.
- [x] Your code builds clean without any errors or warnings.

For all contributors:

- [ ] You have added tests (when appropriate).
- [ ] You have added an entry in the CHANGELOG (when appropriate).
- [ ] You have updated the README or other documentation to account for these changes (when appropriate).
- [ ] You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).

PR #941, opened 2026-06-15
feat(py): add python bindings for `wdl`'s diagnostics apis

BD103:
> [!NOTE]
> ~~This PR is stacked on top of #919, so that will need to be merged before this one is. For a diff of changes made exclusively in this PR, see [`python-ci...diagnostics-bindings-new-approach`](https://github.com/BD103/sprocket/compare/python-ci...diagnostics-bindings-new-approach).~~

## Overview

This PR adds Python bindings for a subset of `wdl::diagnostics` and `wdl::grammar`, enabling emitting diagnostics exclusively using Python. This PR completes part 2 of being able to write WDL lints using Python.

```mermaid
graph LR;
  ci[ci ✅] --> diagnostics[diagnostics ✅] --> lexing --> ast --> linting;
```

Here is an example creating and emitting diagnostics using Python:

```python
from sprocket_bio.diagnostics import emit_diagnostics, Mode
from sprocket_bio.grammar import Diagnostic


diagnostics = [
    Diagnostic.error("oh no!")
        .with_help("something went wrong")
        .with_fix("try this instead")
        .with_rule("MyRule"),
]

emit_diagnostics(
    "file.txt",
    "This is the file contents.",
    diagnostics,
    Mode.default(),
    True,
)
```

image

I specifically adding bindings for 6 APIs that I felt were most useful in creating diagnostics:

- `wdl::diagnostics::Mode`
- `wdl::diagnostics::emit_diagnostics()`
- `wdl::grammar::Span`
- `wdl::grammar::Severity`
- `wdl::grammar::Diagnostic`
- `wdl::grammar::Label`

There were a few other APIs that I intentionally skipped adding bindings for:

- `wdl::diagnostics::DiagnosticCount` (low impact)
- `wdl::diagnostics::emit_diagnostics_with_backtrace()` (may be added with wdl::engine bindings)
- `wdl::diagnostics::get_diagnostics_display_config()` (low impact)
- Everything else in `wdl::grammar` (will be added in part 3)

## Testing

To test the bindings, you must first create a virtual environment and compile the Python extension. Make sure to use Python 3.10 or greater.

```sh
# Create the Python virtual environment, installing the latest version of pip and setuptools.
python -m venv --upgrade-deps .venv

# Activate the virtual environment.
source .venv/bin/activate

# Install the build system, Maturin.
pip install maturin

# Compile and install `sprocket_bio` in the virtual environment.
maturin develop
```

When testing, I recommend opening a Python REPL by running the `python` command in the terminal:

```python
>>> from sprocket_bio.grammar import Diagnostic
>>> d = Diagnostic.error("oh no!")
>>> print(d.message)
oh no!
```

There are also several automated tests and checks you can run:

```sh
# Run unit tests.
pytest

# Check types and type stubs.
mypy --package sprocket_bio
python -m mypy.stubtest sprocket_bio

# Format code.
black python/
```

## Reviewing Notes

I recommend reviewing this PR commit-by-commit! I'll also leave comments in the Github UI for specific places where I have questions or point out certain design decisions. Additionally, I recommend comparing this PR with #911 and seeing which approach you like better in terms of the location of the Python bindings.

cc @claymcleod :)

---

For external contributors:

For all contributors:

      is implemented.


crates/wdl-diagnostics/src/lib.rs
```diff
@@ -233,3 +243,40 @@ pub fn emit_diagnostics_with_backtrace<'a>(
 
     Ok(())
 }
+
+/// Python-specific APIs.
+#[cfg(feature = "python")]
+pub mod python {
```
BD103: My original approach in #911 was to add all of this code to the `sprocket-py` crate, however that led to a lot of duplication. This approach where each crate gets its own `python` module + feature reduces that duplication a decent amount.

crates/wdl-diagnostics/Cargo.toml
```diff
@@ -12,11 +12,13 @@ rust-version.workspace = true
 
 [features]
 backtrace = ["dep:wdl-engine"]
+python = ["dep:pyo3"]
```
BD103: How should I document this feature? Are there any API guarantees? Should this be marked as perma-unstable and not subject to breaking change guarantees?

crates/sprocket-py/src/lib.rs
```diff
@@ -10,4 +10,79 @@ use pyo3::prelude::*;
...
+    /// directly. For example, running `from sprocket_bio.diagnostics import
+    /// Mode` will make the Python interpreter look for
+    /// `sprocket_bio/diagnostics.py` or `sprocket_bio/diagnostics/__init__.py`.
+    /// Neither of these files exist, however, and importing will result in
+    /// a `ModuleNotFoundError`. To fix this, we patch [`sys.modules`](https://docs.python.org/3.10/library/sys.html#sys.modules)
+    /// in this initializer so that the Python interpreter can import these
+    /// submodules even though they aren't represented on the file system. For
+    /// more information, see [pyo3#759](https://github.com/PyO3/pyo3/issues/759).
+    #[pymodule_init]
+    fn init(module: &Bound<'_, PyModule>) -> PyResult<()> {
```
BD103, line 52: This initializing is a little tricky to understand. Let me know if you have any questions!

python/sprocket_bio/diagnostics.pyi
BD103, line 1: These are type stubs, and add [type hints](https://docs.python.org/3/library/typing.html) to our Python package. [MyPy has some good documentation on type stubs](https://mypy.readthedocs.io/en/stable/stubs.html), for reference. While these files look like a lot of duplication, [MyPy's `stubtest` tool](https://mypy.readthedocs.io/en/stable/stubtest.html) helps us stay synchronized.

python/tests/test_diagnostics.py
BD103, line 1: In a future PR I want to add code coverage for these unit tests. That's a bit too complicated for this PR, though, so I'm deferring it.

Edit: tracked in #960

python/sprocket_bio/examples/emit_diagnostics.py
BD103, line 1: I wasn't sure whether to put the example within the Python package or outside of it. When it's in the package, anyone who installed `sprocket_bio` can run this example with `python -m sprocket_bio.examples.emit_diagnostics`. The negative is that it does increase the size of the package, however.

crates/wdl-grammar/src/diagnostic.rs
```diff
@@ -446,3 +492,180 @@ impl Label {
...
+        #[pyo3(name = "with_rule")]
+        fn py_with_rule(&self, rule: &str) -> Self {
+            self.clone().with_rule(rule)
+        }
+
+        /// Sets the help message for the diagnostic.
+        ///
+        /// This is different from the `fix` message, as it only serves to
+        /// provide more context to the issue, rather than a solution.
+        #[pyo3(name = "with_help")]
```
claymcleod, line 617: `with_help()` is exposed as a builder, but there's no corresponding `help` getter—every other settable field (`rule`, `severity`, `message`, `fix`, `labels`) is readable as a Python property. A lint author who calls `.with_help("hint text")` has no way to verify the value was attached without emitting the diagnostic.

Add a getter alongside the others:
```rust
/// Gets the optional help message of the diagnostic.
#[getter(help)]
fn py_help(&self) -> Option<&str> {
    self.help.as_deref()
}
```
And add `help: str | None` to `Diagnostic` in `grammar.pyi`.

crates/wdl-grammar/src/diagnostic.rs
```diff
@@ -168,6 +192,17 @@ impl FromStr for Severity {
...
 /// Represents a diagnostic to display to the user.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[cfg_attr(
+    feature = "python",
+    pyo3::pyclass(
+        module = "sprocket_bio.grammar",
+        frozen,
+        skip_from_py_object,
+        eq,
+        ord,
```
claymcleod, line 203: Not directly related to your change, but I noticed this while reviewing: the `ord` attribute wires `__lt__`/`__gt__` to `impl Ord for Diagnostic`, but that impl omits the `help` field while `Eq` (derived) includes it. Two diagnostics differing only in `help` will have `d1 != d2` yet `d1 < d2` returns `False` and `d1 > d2` returns `False`—breaking the expected total-order contract.

Add `help` to the comparison chain in `impl Ord`:
```rust
match self.fix.cmp(&other.fix) {
    Ordering::Equal => {}
    ord => return ord,
}
self.help.cmp(&other.help)
```

crates/wdl-diagnostics/src/lib.rs
claymcleod: IMO, the module should be private to match the `wdl-grammar` pattern—`pub` here only to satisfy the `use wdl_diagnostics::python::py_emit_diagnostics` import in `sprocket-py`. We can keep it private and re-export just the function:

```rust
#[cfg(feature = "python")]
mod python { ... }

#[cfg(feature = "python")]
pub use python::py_emit_diagnostics;
```

crates/wdl-grammar/src/diagnostic.rs
```diff
@@ -446,3 +492,180 @@ impl Label {
...
+        /// Calculates an intersection of two spans, if one exists.
+        ///
+        /// If spans are adjacent, a zero-length span is returned.
+        ///
+        /// Returns `None` if the two spans are disjoint.
+        ///
+        /// # Examples
+        ///
+        /// ```python
+        /// >>> Span.intersect(Span(0, 10), Span(5, 10))
```
claymcleod: The doctest uses class-method syntax, but `intersect` is an instance method. `Span.intersect(Span(0, 10), Span(5, 10))` would raise `TypeError` at runtime.

```suggestion
        /// >>> Span(0, 10).intersect(Span(5, 10))
```

python/sprocket_bio/examples/emit_diagnostics.py
```diff
@@ -0,0 +1,32 @@
+from sprocket_bio.diagnostics import Mode, emit_diagnostics
+from sprocket_bio.grammar import Diagnostic, Span
+
+import os
+
+workflow_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "example.wdl")
+
+with open(workflow_path, "rt") as f:
```
claymcleod: The file open and `diagnostics` list run at module level, so any import of this file—by mypy, pytest's collection pass, or `from sprocket_bio.examples import *`—tries to open `example.wdl` immediately. Move both into `main()` so the I/O only happens when the script is actually run.

crates/wdl-grammar/Cargo.toml
```diff
@@ -10,10 +10,14 @@ homepage.workspace = true
 repository = "https://github.com/stjude-rust-labs/sprocket/tree/main/crates/wdl-grammar"
 documentation = "https://docs.rs/wdl-grammar"
 
+[features]
+python = ["dep:pyo3"]
```
claymcleod: Both `wdl-grammar` and `wdl-diagnostics` are published crates, so `python` becomes a publicly discoverable feature. A downstream Rust user who runs `cargo add wdl-grammar --features python` would pull in pyo3 unexpectedly. We recommend renaming it to `unstable_python` to signal that it is not part of the stable public API. Same applies to `wdl-diagnostics/Cargo.toml`.

python/tests/test_grammar.py
```diff
@@ -0,0 +1,81 @@
+from sprocket_bio.grammar import Diagnostic, Label, Span
```
claymcleod: A few round-trips are missing from the test suite. Once `help` has a getter (see other comment), it should be covered here. Also missing: `with_rule` + `d.rule` round-trip, `Diagnostic.note()` constructor, `Severity` ordering (`Severity.ERROR < Severity.WARNING`), and `Diagnostic` equality (`d1 == d2`).

crates/wdl-grammar/src/diagnostic.rs
```diff
@@ -446,3 +492,180 @@ impl Label {
...
+            self.intersect(*other.get())
+        }
+
+        /// Gets the length of the span.
+        fn __len__(&self) -> usize {
+            self.len()
+        }
+
+        /// Returns a printable representation of this object.
+        fn __repr__(&self) -> String {
```
claymcleod, line 579: `__repr__` emits `Span(start..end)` but the constructor takes `(start, len)`, so `eval(repr(x)) != x`. One option is to emit `Span({start}, {len})` to match the constructor, or keep the range form and add a note that it's display-only.

crates/wdl-diagnostics/src/lib.rs
```diff
@@ -90,6 +90,16 @@ impl DiagnosticCounts {
...
 /// The diagnostic mode to use for reporting diagnostics.
 #[derive(Clone, Copy, Debug, Default, ValueEnum, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "python",
+    pyo3::pyclass(
+        module = "sprocket_bio.diagnostics",
+        frozen,
+        rename_all = "SCREAMING_SNAKE_CASE",
+        skip_from_py_object,
+        eq
```
claymcleod, line 102: `Mode` is `frozen` + `eq` but missing `hash`—all the other frozen types in this PR have it, and without it `Mode.FULL` cannot be used as a `dict` key or in a `set`. Consider adding `hash` to the pyclass annotation.

claymcleod (changes requested):

crates/wdl-grammar/src/diagnostic.rs
BD103: I tested it and was able to run both forms without any exceptions.

```python
from sprocket_bio.grammar import Span
x = Span(0, 10)
y = Span(5, 10)
print(x.intersect(y)) # 5..10
print(Span.intersect(x, y)) # 5..10
```

Even still, I do like your version better aesthetically! I'll switch it over :)

crates/wdl-grammar/src/diagnostic.rs
BD103: Is the `__repr__` form meant to be round-tripped? I assumed it was more akin to Rust's `Debug` representation. Either way, I updated `__repr__` to return `format!("Span({}, {})", self.start, self.len())`.

crates/wdl-diagnostics/src/lib.rs
BD103: Although `hash` does require `frozen`, I don't annotate all `frozen` classes with `hash`. Instead, my rationale is:

1. I try to annotate as many classes as possible with `frozen`, if they can reasonably be made immutable, for performance reasons.
2. I only annotate a class with `eq` if it implements `PartialEq`, `ord` for `PartialOrd`, and `hash` for `Hash`.

In this case, `Mode` doesn't derive `Hash`, so I didn't make the class hashable either. If you'd like me to make the enum hashable I can, but I figured if it didn't derive `Hash` already, it probably isn't necessary.

crates/wdl-grammar/src/diagnostic.rs
BD103: @riyacore404 handled this already in #943, so I'll address this when #919 is merged and I rebase this PR.

python/tests/test_grammar.py
BD103: I'll address this when I rebase the PR :)

BD103:
Looks like we're having issues downloading from <https://crates.io>; the CI failures so far do not appear to be related to this PR:

```
Caused by:
  failed to load source for dependency `reqwest-middleware`

Caused by:
  unable to update registry `crates-io`

Caused by:
  download of re/qw/reqwest-middleware failed

Caused by:
  curl failed

Caused by:
  [16] Error in the HTTP2 framing layer (Send failure: Connection reset by peer)
```

crates/wdl-diagnostics/Cargo.toml
```diff
@@ -12,11 +12,17 @@ rust-version.workspace = true
 
 [features]
 backtrace = ["dep:wdl-engine"]
+# Enables APIs related to Sprocket's Python bindings.
+#
+# This feature is not subject to `wdl-diagnostics`'s semver guarantees. There may be breaking
+# changes across patch versions. Use at your own risk.
+unstable-python = ["dep:pyo3"]
```
claymcleod: `unstable-python` currently enables `pyo3`, but it does not enable the Python bindings for the concrete `Diagnostic` type used by `py_emit_diagnostics`. That means `cargo test -p wdl-diagnostics --features unstable-python --quiet` fails when someone builds this crate feature directly, even though `sprocket-py` happens to enable the missing feature through its separate `wdl-grammar` dependency. Let's make this feature self-contained by forwarding the Python feature through the dependency chain, e.g., add an `unstable-python` feature to `wdl-ast` that enables `wdl-grammar/unstable-python`, then include that from `wdl-diagnostics/unstable-python`.

crates/wdl-grammar/src/diagnostic.rs
```diff
@@ -451,3 +504,147 @@ impl Label {
...
+mod python {
+    use pyo3::prelude::*;
+
+    use super::*;
+
+    #[pymethods]
+    impl Span {
+        /// Creates a new span from the given start and length.
+        #[new]
+        fn __new__(start: usize, len: usize) -> Self {
```
claymcleod: This exposes the unchecked `start + len` in `Span::new` to Python callers, so `Span(2**64 - 1, 1)` raises a Rust panic instead of a Python exception. We should make this constructor return `PyResult<Self>` and reject overflow with `checked_add`, mapped to a `ValueError` or `OverflowError`, so invalid Python input cannot unwind through the extension.

python/sprocket_bio/grammar.pyi
```diff
@@ -0,0 +1,67 @@
+import typing
+
+@typing.final
+class Diagnostic:
+    rule: str | None
+    severity: Severity
+    message: str
+    help: str | None
+    fix: str | None
+    labels: list[Label]
```
claymcleod: The stubs describe these `get_all` fields as mutable attributes, but the runtime classes are frozen and `labels` returns a copied list. `mypy` accepts `Span(1, 2).start = 3` and `d.labels.append(...)`, while runtime either raises or leaves the diagnostic unchanged. It would be better to model the fields as read-only `@property` values and type `labels` as an immutable view, e.g., `Sequence[Label]`, so type checking matches runtime behavior.

crates/sprocket-py/src/lib.rs
```diff
@@ -10,4 +10,79 @@ use pyo3::prelude::*;
...
+                let item_name: &Bound<'_, PyString> = item_name.cast()?;
+                // Get the actual item from its name.
+                let item = module.getattr(item_name)?;
+
+                // If the item is a submodule...
+                if let Ok(submodule) = item.cast::<PyModule>() {
+                    let submodule_name = format!("{parent_name}.{item_name}");
+
+                    // ...add the submodule to `sys.modules`.
+                    sys_modules.set_item(&submodule_name, submodule)?;
```
claymcleod, line 72: This makes `from sprocket_bio.diagnostics import Mode` work, but the aliased module still has extension-internal metadata: `__name__` is `_sprocket_bio.diagnostics`, and both `__package__` and `__spec__` are `None`. Consider either setting the standard module metadata while registering the alias, or adding real `diagnostics.py` and `grammar.py` shim modules that re-export from `._sprocket_bio`; that would make runtime introspection line up with the `.pyi` module layout.

python/sprocket_bio/examples/emit_diagnostics.py
```diff
@@ -0,0 +1,35 @@
...
+]
+
+
+def main() -> None:
+    # Get the full path of `example.wdl` from this script's location.
+    workflow_path = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "example.wdl"
+    )
+
+    with open(workflow_path, "rt") as f:
```
claymcleod: It might be worth specifying `encoding="utf-8"` here so the example behaves the same way across platforms if the WDL source ever includes non-ASCII text.

python/sprocket_bio/diagnostics.pyi
```diff
@@ -0,0 +1,21 @@
+from .grammar import Diagnostic
```
claymcleod: Nit: please put the stdlib `typing` import before the relative `.grammar` import, with a blank line between the import groups.

python/sprocket_bio/examples/emit_diagnostics.py
```diff
@@ -0,0 +1,35 @@
+from sprocket_bio.diagnostics import Mode, emit_diagnostics
+from sprocket_bio.grammar import Diagnostic, Span
+
+import os
```
claymcleod: Nit: let's put the stdlib `os` import first, then a blank line, then the `sprocket_bio` imports.

python/tests/test_diagnostics.py
```diff
@@ -0,0 +1,58 @@
+from pytest import CaptureFixture
+from sprocket_bio.grammar import Diagnostic, Span
+from sprocket_bio.diagnostics import Mode, emit_diagnostics
```
claymcleod, line 3: Nit: please separate the third-party `pytest` import from the first-party `sprocket_bio` imports with a blank line.

python/tests/test_grammar.py
```diff
@@ -0,0 +1,106 @@
+from sprocket_bio.grammar import Diagnostic, Label, Span, Severity
+
+import pytest
```
claymcleod: Nit: please put the third-party `pytest` import before the first-party `sprocket_bio` import, with a blank line between the import groups.

crates/wdl-grammar/CHANGELOG.md
```diff
@@ -17,6 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   diagnostic ([#927](https://github.com/stjude-rust-labs/sprocket/pull/927)).
 * `Parser` now has a hard recursion limit of `128` on nested expressions ([#930](https://github.com/stjude-rust-labs/sprocket/pull/930)).
 
+#### Fixed
+
+* `Diagnostic` now considers its `help` field in its `Ord` implementation
```
claymcleod: Nit: please include the PR reference and terminal punctuation for consistency with the neighboring entries.

```suggestion
* `Diagnostic` now considers its `help` field in its `Ord` implementation ([#941](https://github.com/stjude-rust-labs/sprocket/pull/941)).
```

crates/wdl-diagnostics/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+* Added the `unstable-python` feature flag, which enables APIs related to Sprocket's Python bindings ([#941](https://github.com/stjude-rust-labs/sprocket/pull/941))
```
claymcleod:
```suggestion
* Added the `unstable-python` feature flag, which enables APIs related to Sprocket's Python bindings ([#941](https://github.com/stjude-rust-labs/sprocket/pull/941)).
```

crates/wdl-grammar/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+* Added the `unstable-python` feature flag, which enables APIs related to Sprocket's Python bindings ([#941](https://github.com/stjude-rust-labs/sprocket/pull/941))
```
claymcleod:
```suggestion
* Added the `unstable-python` feature flag, which enables APIs related to Sprocket's Python bindings ([#941](https://github.com/stjude-rust-labs/sprocket/pull/941)).
```

claymcleod (changes requested):
`python/README.md` still tells contributors to run `mypy --package sprocket_bio`, but this PR changes CI to run `mypy python`. Please update the README so local validation covers the same files as CI; it would also be helpful to add a short example of the new `sprocket_bio.diagnostics` and `sprocket_bio.grammar` APIs while touching this section.

BD103:
> it would also be helpful to add a short example of the new `sprocket_bio.diagnostics` and `sprocket_bio.grammar` APIs while touching this section.

I'm going to hold off for now. The diagnostics APIs are not the most glamorous, and I'd rather showcase actual WDL analysis in the `README.md`.

crates/wdl-grammar/src/diagnostic.rs
BD103: In general we don't need to worry about panics, as PyO3 catches them at the FFI boundary and converts them into [`PanicException`](https://docs.rs/pyo3/latest/pyo3/panic/struct.PanicException.html)s. Even still, `OverflowError` in this scenario is much nicer on the Python side.

python/sprocket_bio/diagnostics.pyi
BD103: Got it! In the future, I'll setup [`isort`](https://isort.readthedocs.io) to handle this automatically.

crates/sprocket-py/src/lib.rs
BD103: I'd like to defer this to a separate PR. There are multiple solutions to this, and I would like to do more research in order to find which one best suites us.

Edit: I'm tracking this in #957!

claymcleod (approved):

PR #942, opened 2026-06-17
fix(test): make should_fail functional for task tests

### LINKED ISSUE #897: `dev test`: support `should_fail` on tasks
In the test YAML, there's a special assertion `should_fail`, which is only available for workflows. When used with tasks, it does nothing: https://github.com/stjude-rust-labs/sprocket/blob/88603edc6dce549c2784488ded8d6bd1a8a85a7e/src/test/assertions.rs#L64-L66

For consistency, the `should_fail` assertion could be used on tasks to imply any nonzero exit code is expected. Having both `should_fail` and `exit_code` in an assertions list should return an error.

riyacore404:
### What this PR does

Previously, using `should_fail: true` in a task test silently emitted a warning and did nothing — the assertion was ignored entirely. This PR makes `should_fail` functional for tasks:

#### Changes

- A task test with `should_fail: true` now **passes** when the task exits with a nonzero exit code, and **fails** when the task exits with code `0`.
- The `should_fail` check is evaluated before `exit_code`, so `exit_code` is effectively bypassed when `should_fail` is set (consistent with the existing workflow behavior).
- The silent `warn!("ignoring \`should_fail\` assertion for task")` is removed.
- Updated doc comments on `Assertions` and `ParsedAssertions` to describe the new per-target semantics.
- Added two new unit tests in `src/test/assertions.rs` covering the parse path.
- Added `--fixtures-dir` and `--run-dir` CLI flags to `sprocket test` and extracted a `resolve_test_paths()` helper with six unit tests.
- Added an integration fixture at `tests/cli/test/task_should_fail/` demonstrating the new behavior end-to-end.

Closes #897.

### Checklist

For external contributors:

For all contributors:

- [ ] You have updated any and all affected entries within `RULES.md`.
- [ ] You have added a test case in `crates/wdl-lint/tests/lints` that covers every possible diagnostic emitted for the rule within the file where the rule is implemented.


src/test/assertions.rs
```diff
@@ -21,7 +21,11 @@ pub(crate) struct Assertions {
     /// The expected exit code of the task (ignored when testing workflows).
     #[serde(default)]
     pub exit_code: i32,
-    /// Whether a workflow should fail or not (ignored when testing tasks).
+    /// Whether the test is expected to fail.
+    ///
+    /// For workflows, any failure is expected.
+    /// For tasks, the engine determines failure based on the exit code and
+    /// any `return_codes` specified in the task's `runtime` block.
```
Serial-ATA:
```suggestion
    /// For tasks, the engine determines failure based on the exit code.
```

This isn't true ([yet](https://github.com/stjude-rust-labs/sprocket/issues/897#issuecomment-4616038516))

src/commands/test.rs
```diff
@@ -285,11 +285,24 @@ impl TestIteration {
...
+
+                        if assertions.should_fail {
+                            if actual_exit_code == 0 {
+                                return Ok(IterationResult::Fail(anyhow!(
+                                    "{id} exited with code `0` but `should_fail` expected a \
+                                     nonzero exit code: see `{dir}`",
+                                    dir = run_dir.display(),
+                                )));
+                            }
+                            return Ok(IterationResult::Success);
```
Serial-ATA: This shouldn't return early, the other cases below still need to be tested

src/test/assertions.rs
```diff
@@ -1046,4 +1048,28 @@ mod tests {
...
+            ..Default::default()
+        };
+        let result = assertions.parse(false, &IndexMap::new());
+        assert!(result.is_ok());
+        assert!(result.unwrap().should_fail);
+    }
+
+    #[test]
+    fn parse_should_fail_with_exit_code_ok_for_task() {
+        // should_fail + exit_code together is NOT an error
```
Serial-ATA: It *should* be an error. `should_fail` is, by default, an alias for `exit_code: 0`, and in the future will determine the expected exit codes from the task definition

Serial-ATA (changes requested):
Thanks for picking this up! Outside of these comments, this will also need a changelog entry

src/test/assertions.rs
riyacore404: Fixed: updated the doc comment to accurately describe the semantics

src/test/assertions.rs
riyacore404: I misread the intended semantics here. Removed the incorrect test that expected this to be Ok, added parse_should_fail_with_nonzero_exit_code_errors_for_task to assert the correct error behavior, and restored the original bail! that I had mistakenly removed.

src/commands/test.rs
riyacore404: Fixed: removed the early return Ok(IterationResult::Success) and converted to else if for the exit_code check

src/test/assertions.rs
```diff
@@ -61,8 +65,11 @@ impl Assertions {
                 warn!("ignoring `stderr` assertion for workflow");
             }
         } else {
-            if self.should_fail {
-                warn!("ignoring `should_fail` assertion for task");
+            if self.should_fail && self.exit_code != 0 {
```
Serial-ATA: That means this is valid:

```yml
foo:
  - name: foo
    assertions:
      exit_code: 0
      should_fail: true
```

Instead, `exit_code` should be `Option<i32>` that defaults to `0` if not set. *Any* combination of `should_fail` + `exit_code` should error for consistency.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* Fixed `sprocket dev test` where `should_fail` assertion was silently
```
Serial-ATA: This entry should go under `Added`, since we're adding `should_fail` support to tasks

tests/cli/test/task_should_fail/inputs/source.wdl
Serial-ATA, line 1: Please add a newline at the end

tests/cli/test/task_should_fail/inputs/test/source.yml
Serial-ATA, line 1: Same here

Serial-ATA (changes requested):
Thanks! Just a few more things.

Also, outside of the comments below, you need to fix your test in `tests/cli/test/task_should_fail`. There should be an `args` file, see https://github.com/stjude-rust-labs/sprocket/blob/main/tests/cli/test/cancellation/args for an example.

tests/cli/test/task_should_fail/inputs/source.wdl
Serial-ATA: Also would be nice to test the error case in this file:

```yml
  - name: always_succeeds
    assertions:
      should_fail: true
```

```wdl
task always_succeeds {
    command <<<
        exit 0
    >>>
}
```

src/test/assertions.rs
riyacore404: `exit_code` is now `Option<i32>` (defaults to `0`), and `should_fail` + `exit_code` together is now a parse-time error for both tasks and workflows.

riyacore404:
Hi @Serial-ATA ! I tried adding this, but ran into something I wasn't sure how to handle.

When I added the `always_succeeds` task and test case, the CLI test kept failing even though the behavior itself looked correct. After digging into it, it seems the error message includes the full path to the test's temporary run directory, e.g.:

```text
exited with code `0` but `should_fail` expected a nonzero exit code:
see `/private/var/.../tmpXXXXXX/test/runs/...`
```

That path appears to be generated dynamically each run (via `TempDir::new()` in `tests/cli.rs`), so the blessed stderr file never matches. I reran it several times locally and got a different path each time.

I noticed `tests/cli.rs` already normalizes things like timestamps/UUIDs, but not this kind of path. I wasn't sure whether I should add path normalization there, since that feels a bit outside the scope of this PR.

I haven't pushed anything addressing this yet and wanted to check first. Would you prefer that I:

- (a) add path normalization to `tests/cli.rs` as part of this PR,
- (b) open a separate issue/PR for the harness fix and leave the `always_succeeds` case out of this one for now, or
- (c) do something else I'm missing?

For reference, here's the regex that's used for normalization in `tests/cli.rs`:

```rust
static TIMESTAMP_PATTERN: LazyLock<Regex> =
    LazyLock::new(|| Regex::new(r"\d{4}-\d{2}-\d{2}_\d{12,}").unwrap());
```

It only matches date-shaped timestamps, not the random alphanumeric suffix `TempDir::new()` produces (e.g. `tmpGMxwHo`), so there's currently no normalization path that would make this fixture's `stderr` stable.

Serial-ATA:
@riyacore404 Yeah, looks like we didn't have any tests for failing test runs. I went ahead and added the path normalization

riyacore404:
@Serial-ATA 

Found one edge case running locally on macOS: /var is a symlink to /private/var there, so TempDir::new() gives a path like /var/folders/..., but the subprocess actually reports back /private/var/folders/.... The literal replace in normalize_string() doesn't catch that, leaving a stray /private prefix:

  expected: see `_TEMP_DIR_/test/runs/...`
  actual:   see `/private_TEMP_DIR_/test/runs/...`

Pushed a fix in 294862c — canonicalizing the working test directory once in run_test() before it's passed downstream. Didn't touch normalize_string or any of the comparison functions, so it should be a no-op on Linux.

Verified locally: task_should_fail passes consistently across repeated runs, full CLI suite (93/93) and unit suite (105/106, the one failure is the pre-existing can_submit_and_complete Docker flake) both pass, fmt/clippy clean.

Let me know if you'd prefer a different approach.

Serial-ATA (approved):

src/test/assertions.rs
```diff
@@ -19,9 +19,16 @@ use wdl::analysis::types::Type;
 #[derive(Default, serde::Deserialize, Debug)]
 pub(crate) struct Assertions {
     /// The expected exit code of the task (ignored when testing workflows).
+    ///
+    /// Defaults to `0` if not specified. Cannot be combined with
+    /// `should_fail`.
     #[serde(default)]
-    pub exit_code: i32,
-    /// Whether a workflow should fail or not (ignored when testing tasks).
+    pub exit_code: Option<i32>,
```
claymcleod, line 76: We should distinguish an omitted `exit_code` from an explicitly null one here. With `Option<i32>`, `exit_code: null` deserializes as `None`, so it defaults to `0` and bypasses the `should_fail`/`exit_code` conflict check. A custom deserializer or presence-aware raw field would let the parser reject malformed input instead of silently accepting it.

src/test/assertions.rs
```diff
@@ -19,9 +19,16 @@ use wdl::analysis::types::Type;
...
+    /// Whether the test is expected to fail.
+    ///
+    /// For workflows, any failure is expected.
+    /// For tasks, any nonzero exit code is expected.
+    /// Cannot be combined with `exit_code`.
     #[serde(default)]
     pub should_fail: bool,
```
claymcleod: Let's make `should_fail` presence-aware too if the intended rule is that any `should_fail`/`exit_code` combination is invalid. As written, `should_fail: false` deserializes the same as an omitted field, so `should_fail: false` plus `exit_code` is accepted even though the YAML still specifies both knobs.

src/test/assertions.rs
```diff
@@ -378,8 +385,13 @@ impl OutputAssertion {
 #[derive(Debug)]
 pub(crate) struct ParsedAssertions {
     /// The expected exit code of the task (ignored when testing workflows).
+    ///
+    /// Defaults to `0`.
     pub exit_code: i32,
-    /// Whether a workflow should fail or not (ignored when testing tasks).
+    /// For workflows: whether the workflow is expected to fail.
+    /// For tasks: whether any nonzero exit code is acceptable (i.e. the task
+    /// is expected to fail). Combining this with a nonzero `exit_code`
```
claymcleod: This should say that combining `should_fail` with any specified `exit_code` is rejected, not just a nonzero one. The parser rejects `Some(0)` too, and the earlier review thread explicitly asked for any combination to error.

tests/cli.rs
```diff
@@ -548,16 +562,12 @@ fn compare_test_results(
     if env::var_os("BLESS").is_some() {
         fs::write(
             &expected_stderr_file,
-            TIMESTAMP_PATTERN
-                .replace_all(&command_output.stderr, "_TIMESTAMP_")
-                .as_ref(),
+            normalize_string(&command_output.stderr, working_test_directory),
```
claymcleod, line 594: It might be worth normalizing text files copied into `outputs/` during `BLESS` as well. `stderr` and `stdout` now replace the temp directory with `_TEMP_DIR_`, but `recursive_copy` preserves output file contents verbatim; if a test output file contains the temp path, blessing writes a one-run absolute path that later comparisons will not match.

tests/cli/test/bad_test_definitions/inputs/bar.yaml
```diff
@@ -16,4 +16,8 @@ bar:
       two:
         - correct
       three:
-       - also correct
\ No newline at end of file
+        - also correct
+  - name: conflicting options
+    assertions:
+      exit_code: 0
+      should_fail: true
```
claymcleod: We should add a case for explicit `should_fail: false` with `exit_code` if the intended rule is that any specified combination is invalid. This fixture covers the truthy path, but it would not catch the current parse behavior where explicit `false` is treated the same as omitting `should_fail`.

CHANGELOG.md
```diff
@@ -19,7 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `--show-task-stderr` option to the `run` subcommand to show task stderr during execution ([#743](https://github.com/stjude-rust-labs/sprocket/pull/743)).
 * Added `--fixtures-dir` and `--run-dir` options to the `sprocket dev test`
   command ([#747](https://github.com/stjude-rust-labs/sprocket/pull/747)).
-  
+* Added support for the `should_fail` assertion on tasks in `sprocket dev test` ([#942](https://github.com/stjude-rust-labs/sprocket/issues/942)).
```
claymcleod: I think this reads more like a bugfix than an added feature, since task `should_fail` was already accepted but silently ignored. Could we move this under `Fixed` and link to the PR instead of `/issues/942`?

tests/cli/test/task_should_fail/inputs/source.wdl
```diff
@@ -0,0 +1,14 @@
...
+        exit 1
+    >>>
+}
+
+task always_succeeds {
+    command <<<
+        exit 0
+    >>>
+}
+
```
claymcleod: IMO we should remove this extra blank line so `git diff --check` passes cleanly.

tests/cli/test/task_should_fail/inputs/test/source.yml
```diff
@@ -0,0 +1,10 @@
+always_fails:
+  - name: should_fail_passes
+    assertions:
+      should_fail: true
+
+always_succeeds:
+  - name: always_succeeds
+    assertions:
+      should_fail: true
+
```
claymcleod: Please remove this trailing blank line so `git diff --check` passes cleanly.

claymcleod (changes requested):

CHANGELOG.md
Serial-ATA: This is new functionality, not a bugfix https://github.com/stjude-rust-labs/sprocket/pull/942#discussion_r3443058864

tests/cli/test/task_should_fail/inputs/source.wdl
```diff
@@ -0,0 +1,14 @@
+version 1.2
```
claymcleod: This should be WDL 1.3.

CHANGELOG.md
claymcleod: Alright I suppose it falls somewhere in the middle in my mind. I'm happy to call it a feature.

src/test/assertions.rs
riyacore404: Already addressed in the previous iteration: `exit_code` is `Option<i32>` and `parse()` uses` is_some()` for the conflict check, so `exit_code: null` deserializes as `None` and is treated as unspecified.

src/test/assertions.rs
riyacore404: Fixed: changed `should_fail` to `Option<bool>`. The conflict check now fires when either field is `Some(_)`, so `should_fail: false` combined with any `exit_code` is correctly rejected. Added unit tests covering both `should_fail: false` + `exit_code` (errors) and `should_fail: false` alone (ok).

tests/cli.rs
riyacore404: added `normalize_output_files()` helper that walks all text files copied into the expected `outputs/` directory during BLESS and replaces any occurrences of the temp path with `_TEMP_DIR_`, matching the normalization already applied to `stderr` and `stdout`.

tests/cli/test/bad_test_definitions/inputs/bar.yaml
riyacore404: added a `conflicting options with should_fail false` entry with `exit_code: 1` and `should_fail: false` to catch the case where explicit `false` was previously treated the same as omitting the field.

src/test/assertions.rs
claymcleod: I think the earlier thread is about omission, not explicit `null`. Making `exit_code` optional is right so an omitted field can default to `0`, but `exit_code: null` is still the user specifying the `exit_code` key with a nonsensical value. We should reject that as malformed input rather than treating it the same as omission; otherwise the “any specified `should_fail` + `exit_code` combination is invalid” rule still has a hole.

claymcleod:
Thank you!

PR #953, opened 2026-06-25
feat(sprocket): add JSON schema for `dev test` YAMLs

Serial-ATA:
Adds a (mostly) auto-generated JSON schema provided by [schemars](https://crates.io/crates/schemars), and a new `sprocket dev test schema` subcommand to print it.

I also have a copy of the schema in the `jsonschemas` directory for tools to point to. It'll need to stay in place and be kept up to date. There's a test verifying it matches the current types in `src/test.rs`

Before submitting this PR, please make sure:

For external contributors:

- [ ] You have read the [contributing guide](https://github.com/stjude-rust-labs/sprocket/blob/main/CONTRIBUTING.md) in its entirety.
- [ ] You have not used AI on any parts of this pull request.
- [ ] You have added a few sentences describing the PR here.
- [ ] Your code builds clean without any errors or warnings.

For all contributors:

- [x] You have added tests (when appropriate).
- [x] You have added an entry in the CHANGELOG (when appropriate).
- [ ] You have updated the README or other documentation to account for these changes (when appropriate).
- [ ] You have made a PR to the `next` branch in the [sprocket.bio repository](https://github.com/stjude-rust-labs/sprocket.bio) (when appropriate).

For PRs containing lint rule changes:

- [ ] You have updated any and all effected entries within `RULES.md`.
- [ ] You have added a test case in `crates/wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.

PR #958, opened 2026-06-26
feat(sprocket): add JSON schema for `sprocket.toml`

Serial-ATA:
Adds an auto-generated JSON schema for `sprocket.toml` with [schemars](https://crates.io/crates/schemars), and a new `sprocket config schema` subcommand to print it.

I also updated the output of `sprocket config init` to include a schema directive for `taplo` to automatically pick it up.

depends on #953

For external contributors:

For all contributors:

      is implemented.


jsonschemas/sprocket.toml.json
```diff
@@ -0,0 +1,1560 @@
...
+      "format": "uint",
+      "minimum": 0
+    },
+    "NewlineStyle": {
+      "description": "The newline style to use when formatting.",
+      "oneOf": [
+        {
+          "description": "Use the native newline style of the platform.",
+          "type": "string",
+          "const": "Auto"
```
claymcleod: The generated schema currently uses Rust variant names for several string-backed TOML values, but the config parser/serializer uses the lowercase/display forms. For example, `sprocket config init` emits `newline_style = "auto"`, `parallelism = "available"`, `max_concurrent_runs = "unlimited"`, `common.color = "auto"`, `common.report_mode = "full"`, `large_file_warning = "1.0 MiB"`, and `trust_mode = "auto"`, while the schema allows values like `Auto`, `Available`, `Unlimited`, and `Full`. That means the schema directive points Taplo at a schema that rejects the file Sprocket just generated. Can we add custom `JsonSchema` implementations or `#[schemars(with = ...)]` wrappers for these display-parsed values, plus a test that validates `config init` output against `config schema`?

src/config.rs
```diff
@@ -242,77 +271,95 @@ impl fmt::Display for FallbackVersion {
...
 
 /// Represents the configuration for the Sprocket `run` command.
-#[derive(Debug, Clone, PartialEq, Eq, Toml)]
+#[derive(Debug, Clone, PartialEq, Eq, Toml, JsonSchema)]
 #[toml(Toml, rename_all = "snake_case", deny_unknown_fields)]
+#[schemars(rename_all = "snake_case", deny_unknown_fields)]
 pub struct RunConfig {
     /// The engine configuration.
     #[toml(default, flatten, with = flatten_any)]
+    #[schemars(default)]
```
claymcleod: This field is flattened in `sprocket.toml`, but the generated schema models it as a nested `engine` property under `[run]`. `sprocket config init` writes keys like `backend`, `fail`, and `[run.http]` directly under `[run]`, while the schema only allows `engine`, `output_dir`, and `events_capacity` there with `additionalProperties: false`. That makes the generated `[run]` section fail schema validation even before the enum casing issue. Can the schema representation flatten `EngineConfig` the same way `toml_spanner` does, and can the init-output-vs-schema test cover this shape?

claymcleod (changes requested):

jsonschemas/sprocket.toml.json
Serial-ATA: Yeah, good call on the `config init` test. Super easy to catch errors now

claymcleod:
The schema still rejects two valid serialized `sprocket.toml` values. `format.indent = "tabs"` is accepted and emitted by the config code, but the schema currently allows only `"Tabs"`; `format.max_line_length = "none"` is accepted and emitted for no limit, but the schema allows only integer/null. Please update the schema generation or annotations so those TOML forms validate, and add non-default schema coverage so the test exercises values beyond `Config::default()`.

claymcleod (approved):

PR #977, opened 2026-07-03
chore(ci): use python-version-file when installing python

### LINKED ISSUE #972: ci(py): use `python-version-file` when installing python

We currently manually specify the Python version in CI:

https://github.com/stjude-rust-labs/sprocket/blob/854f13febe71482313ae473ae166f3da9d861fc7/.github/workflows/python-ci.yml#L44-L47

This means we duplicate the Python version (3.10) for every job that installs Python.

`actions/setup-python` supports the [`python-version-file`](https://github.com/actions/setup-python/blob/v6/docs/advanced-usage.md#using-the-python-version-file-input) input, where it can read the Python version from `pyproject.toml`. We should switch to this, so `pyproject.toml` becomes the single source of truth for what version of Python we use.

riyacore404:
## Description

`actions/setup-python` supports the `python-version-file` input, which reads the Python version from `pyproject.toml` (via `requires-python`). This replaces the hardcoded `python-version: "3.10"` in the `test` and `type-check` jobs, so `pyproject.toml` becomes the single source of truth instead of duplicating the version in every job.

## Changes

- `.github/workflows/python-ci.yml`: replaced hardcoded `python-version` with `python-version-file: "pyproject.toml"` in the `test` and `type-check` jobs.

## Testing

- Confirmed `pyproject.toml` declares `requires-python = ">=3.10"`.

- Ran `actionlint` against the workflow file — no new issues introduced.

- CI on this PR itself is the functional test.

#### Fixes #972

For external contributors:

For all contributors:

      is implemented.


BD103 (changes requested):
I took a look at the CI logs, and it appears that this change makes the action install Python 3.14, not 3.10. I think it's because `pyproject.toml` has `requires-python = ">=3.10"`, so the action opportunistically uses the latest possible version.

When I created the action I purposefully installed the minimum supported Python version (3.10) so that I didn't accidentally use unstable features. I want to keep that, but I don't know if there's a way to use `python-version-file` _and_ install Python 3.10 with our current `pyproject.toml`. Do you know of a way to do so?

riyacore404:
You're right — `actions/setup-python` resolves `pyproject.toml`'s `requires-python` to the **latest** Python version satisfying the specified range, not the minimum. That's the documented and intended behavior, and the maintainers have confirmed there isn't an option to instead select the lower bound of the range (see `actions/setup-python#542`).

Since `requires-python = ">=3.10"` will resolve to the newest supported version, pointing `python-version-file` at `pyproject.toml` can't guarantee that CI runs on Python 3.10.

A reasonable alternative would be to use a dedicated `.python-version` file:

```yaml
- name: Setup Python
  uses: actions/setup-python@v6
  with:
    python-version-file: .python-version
```

with `.python-version` containing:

```text
3.10
```

This still avoids duplicating the version across multiple workflow jobs while keeping CI pinned to the intended minimum supported version.

I also think this is a cleaner separation of responsibilities:

- `requires-python` declares the supported Python **version range** for the package.
- `.python-version` specifies the exact interpreter version that a particular development or CI environment should use.

That avoids conflating package metadata with CI configuration and prevents the CI version from changing automatically as the supported range evolves.

If you'd prefer, we could also leave the existing `python-version: "3.10"` as-is and consider this a non-change, since `python-version-file` can't achieve the intended behavior when pointed at `pyproject.toml`.

BD103:
Oooo `.python-version` is a good idea! I definitely want to get the maintainers to sign-off on it, but that's the approach that I like the best. :)

riyacore404:
Sounds good — this does touch on CI conventions the maintainers set, so their sign-off makes sense before locking it in.
I've pushed the `.python-version` + workflow update in the meantime so it's ready to go once there's agreement (see the latest commit). Happy to adjust based on whatever direction the maintainers land on — including reverting to the original hardcoded `python-version: "3.10"` if they'd rather not introduce a second file for this.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+* CI now reads the Python version from a `.python-version` file via `python-version-file` instead of hardcoding it in each job ([#972](https://github.com/stjude-rust-labs/sprocket/pull/977)).
```
BD103: I'm not sure if this change needs to be documented in the changelog, as it isn't user-facing. Then again, I'm not certain what the exact policy on what gets included in the changelog is, so I may be wrong :)

BD103 (approved):
Looks good on my end :)

CHANGELOG.md
Serial-ATA: Yeah, the changelog only needs to show user-facing changes

riyacore404: thanks for confirming! I'll leave this out of the changelog then.

Serial-ATA (approved):

PR #978, opened 2026-07-03
Add `strongish` content digest mode for call caching

### LINKED ISSUE #974: _strongish_ content digests for call caching
For content digests, `weak` is great if you're very rigorous and have complete control over your environment, but isn't as trustworthy if someone else can write your files.  `strong` is great, but slow for large files.  It would be nice to have an intermediate strategy that makes use of the file contents, but runs faster.

Add an intermediate _strongish_ content digest strategy, between `weak` and `strong`, similar to Cromwell's [`fingerprint`](https://cromwell.readthedocs.io/en/latest/Configuring/#call-cache-strategy-options-for-local-filesystem) strategy that combines mtime, size, and a hash of the first 10 MB.

riyacore404:
Adds a `strongish` content digest mode, an intermediate strategy between `weak` and `strong`, similar to Cromwell's `fingerprint` call cache strategy: hashes file size, last modified time, and only the first 10 MiB of file content.

## Changes
- `ContentDigestMode::Strongish` variant in `config.rs`
- Implementation in `digest.rs` (`calculate_file_digest`), sharing the size/mtime hashing logic with `Weak` via a new `hash_file_metadata` helper
- Unit tests covering digest stability/change behavior and confirming content beyond the 10 MiB prefix is ignored (unlike `strong`)

## Testing
- `cargo test -p wdl-engine`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo fmt --check`

#### Closes #974

For external contributors:

For all contributors:

      is implemented.


crates/wdl-engine/src/digest.rs
```diff
@@ -146,38 +154,72 @@ async fn calculate_file_digest(path: &Path, mode: ContentDigestMode) -> Result<D
...
+                let mut file = fs::File::open(&path)
+                    .with_context(|| format!("failed to open `{path}`", path = path.display()))?;
+
+                let mut buf = Vec::new();
+                file.by_ref()
+                    .take(STRONGISH_DIGEST_PREFIX_LEN)
+                    .read_to_end(&mut buf)
+                    .with_context(|| {
+                        format!("failed to read contents of `{path}`", path = path.display())
+                    })?;
```
peterhuene: Rather than reading the 10 MiB into a buffer, it might be better to use `std::io::copy` to copy from the `take`n file into the hasher.

peterhuene (changes requested):

crates/wdl-engine/src/digest.rs
riyacore404: Makes sense, `Hasher` already implements `Write`, so `io::copy` is a cleaner fit than reading into a `Vec` first. Updated.

peterhuene (approved):
Looks great! Thanks very much for implementing this @riyacore404!

PR #979, opened 2026-07-04
fix(wdl-analysis): make runtime key type checking version-aware

### LINKED ISSUE #811: Runtime key type checking should be version-aware
I came across this while running `sprocket lint` on the [broadinstitute/warp](https://github.com/broadinstitute/warp) codebase. Several tasks there declare `cpu` as a `String` (e.g., `cpu: "1"`), and `sprocket lint` flags these as errors. I put up a fix in [broadinstitute/warp#1811](https://github.com/broadinstitute/warp/pull/1811), but in doing so I realized the lint diagnostic isn't quite right for WDL 1.0 documents.

The WDL 1.0 spec does not formally define the `cpu` key. In 1.0, runtime keys are essentially arbitrary:

> The runtime section defines key/value pairs for runtime information needed for this task. Individual backends will define which keys they will inspect so a key/value pair may or may not actually be honored depending on how the task is run.
>
> ([WDL 1.0 spec, Runtime Section](https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#runtime-section))

Only `docker` and `memory` are given "recommended conventions" in 1.0. The formal type definitions for `cpu`, `gpu`, `disks`, `maxRetries`, and `returnCodes` were introduced in WDL 1.1.

Flagging a `String`-typed `cpu` value as an error in a `version 1.0` document is technically incorrect, since the spec imposes no type constraint there. The type checking for these keys should only apply to documents declaring `version 1.1` or later.

riyacore404:
## Summary

WDL 1.0's runtime section only gives formal type conventions for the `docker`/`container` and `memory` keys. The `cpu`, `gpu`, `disks`, `maxRetries`, and `returnCodes` keys weren't formally typed until WDL 1.1, but `wdl-analysis` was type-checking them unconditionally regardless of declared document version, producing false-positive type-mismatch diagnostics on valid WDL 1.0 documents.

## Changes

- `task_requirement_types()`: gate `cpu`, `disks`, `gpu`, and the `maxRetries`/`returnCodes` aliases to `version >= 1.1`. `docker`/`memory` remain checked at all versions.
- `evaluate_runtime_item()`: for those same keys on WDL 1.0 documents, skip type checking entirely rather than letting the lookup miss and fall through to `hints`-key type checking — `gpu` and `disks` are also valid hint key names with different expected types, so the fallthrough would otherwise produce incorrect diagnostics of its own.
- Added a regression test:
  `crates/wdl-analysis/tests/analysis/runtime-section-v1-0/`.

Closes #811.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Fixed
-
-* `runtime` section key type checking (e.g. for `cpu`, `gpu`, `disks`,
-  `maxRetries`, and `returnCodes`) is now version-aware: these keys are no
-  longer type checked in WDL 1.0 documents, since they were not formally
-  typed until WDL 1.1 ([#811](https://github.com/stjude-rust-labs/sprocket/pull/979)).
-
 ## 0.27.0 - 2026-06-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* `runtime` section key type checking (e.g. for `cpu`, `gpu`, `disks`,
+  `maxRetries`, and `returnCodes`) is now version-aware: these keys are no
+  longer type checked in WDL 1.0 documents, since they were not formally
+  typed until WDL 1.1 ([#811](https://github.com/stjude-rust-labs/sprocket/pull/979)).
+
 ## 0.27.0 - 2026-06-26
 
 ### Added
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* `runtime` section key type checking (e.g. for `cpu`, `gpu`, `disks`,
```
peterhuene: Could we move this addition to the `CHANGELOG.md` in the `wdl-analysis` crate?

peterhuene (commented):
Thanks for fixing this @riyacore404. Looks good, just the one comment relating to the CHANGELOG entry.

peterhuene:
Oh it also looks like the gauntlet baselines need updating as a result of these changes.

peterhuene (approved):
Thanks again for fixing this!

PR #980, opened 2026-07-05
feat(py): `wdl-grammar` bindings

BD103:
## Overview

This PR adds Python bindings for a subset of `wdl::grammar`, making it possible to parser WDL documents into an `Event` stream. `Event` streams can be treated as a flat list of tokens, useful for writing syntax highlighters, or can be treated as a list of nested nodes and tokens, useful for writing rudimentary analyzers.

This PR completes the third milestone in being able to write WDL lints in Python:

```mermaid
graph LR;
  ci[ci ✅] --> diagnostics[diagnostics ✅] --> lexing[lexing ✅] --> ast --> linting;
```

## What's In The Box?

- [`wdl_grammar::grammar::document()`](https://docs.rs/wdl-grammar/0.24.0/wdl_grammar/grammar/fn.document.html)
- [`wdl_grammar::version::SupportedVersion`](https://docs.rs/wdl-grammar/0.24.0/wdl_grammar/version/enum.SupportedVersion.html)
- [`wdl_grammar::version::V1`](https://docs.rs/wdl-grammar/0.24.0/wdl_grammar/version/enum.V1.html)
- [`wdl_grammar::parser::Event`](https://docs.rs/wdl-grammar/0.24.0/wdl_grammar/parser/enum.Event.html)
- [`wdl_grammar::SyntaxKind`](https://docs.rs/wdl-grammar/0.24.0/wdl_grammar/enum.SyntaxKind.html)

The core of this PR is the `document()` function, which parses a WDL document into an event stream. An event stream is composed of `Event`s, which can be either `Token`s, `NodeStarted`s, or `NodeFinished`s. `Token`s and `NodeStarted`s expose a `SyntaxKind`, which is used to label the token / node as a specific type. (E.g. a comment token is labeled with `SyntaxKind.COMMENT`.)

## Examples

In this PR I've included two examples using the event stream. The first, titled `event_stream.py`, parses a WDL document and prints its events to the console in a similar format to [`SyntaxTree`](https://docs.rs/wdl-grammar/0.24.0/wdl_grammar/struct.SyntaxTree.html)'s `Debug` implementation:

```sh
(.venv) $ python -m sprocket_bio.examples.event_stream --help
usage: event_stream.py [-h] <source_file>

Parses a WDL document and prints the event stream

positional arguments:
  <source_file>  path to WDL document

options:
  -h, --help     show this help message and exit
```

```sh
(.venv) $ python -m sprocket_bio.examples.event_stream python/sprocket_bio/examples/example.wdl
SyntaxKind.ROOT_NODE
  SyntaxKind.VERSION_STATEMENT_NODE
    SyntaxKind.VERSION_KEYWORD@0..7 'version'
    SyntaxKind.WHITESPACE@7..8 ' '
    SyntaxKind.VERSION@8..11 '1.3'
  SyntaxKind.WHITESPACE@11..13 '\n\n'
  SyntaxKind.TASK_DEFINITION_NODE
    SyntaxKind.TASK_KEYWORD@13..17 'task'
    SyntaxKind.WHITESPACE@17..18 ' '
    SyntaxKind.IDENT@18..27 'say_hello'
    # ...
```

The second example, my personal favorite, is a syntax highlighter that parses a WDL document and emits a styled HTML document. This example is an improved version of my original proof-of-concept that I wrote for my GSoC proposal.

```sh
(.venv) $ python -m sprocket_bio.examples.syntax_highlighter --help
usage: syntax_highlighter.py [-h] [-o <OUTPUT_FILE>] [--open] <source_file>

Syntax highlights a WDL document, outputting HTML that can be viewed in your browser

positional arguments:
  <source_file>         path to WDL document

options:
  -h, --help            show this help message and exit
  -o <OUTPUT_FILE>, --output <OUTPUT_FILE>
                        the path to write to, defaults to <source_file>.html
  --open                opens the output in a web browser
```

```sh
(.venv) $ python -m sprocket_bio.examples.syntax_highlighter --open python/sprocket_bio/examples/example.wdl
```

a screenshot of a web browser displaying the syntax highlighted WDL

## Testing

To test the bindings, you must first create a virtual environment and compile the Python extension. Make sure to use Python 3.10 or greater.

```sh
# Create the Python virtual environment, installing the latest version of pip and setuptools.
python -m venv --upgrade-deps .venv

# Activate the virtual environment.
source .venv/bin/activate

# Install the build system, Maturin.
pip install maturin

# Compile and install `sprocket_bio` in the virtual environment.
maturin develop
```

This will enable you to run the examples from above, spawn a REPL to test the bindings yourself, and to run other automated checks like `pytest` and `mypy`.

## For Reviewers

Like last time, I think this PR is best reviewed commit-by-commit. I rebased, amended, and squashed the commits several times over my development in order for them to be consumable this way.

cc @claymcleod

### What is `PyEvent`? Why isn't `#[pyclass]` derived directly on `Event`?

In short, `#[pyclass]` doesn't work on `Event` because `Event::NodeFinished` is a unit variant, not a tuple or structure variant. Trying to derive `#[pyclass]` directly on `Event` leads to PyO3 emitting the following error:

```
error: Unit variant `NodeFinished` is not yet supported in a complex enum
       = help: change to an empty tuple variant instead: `NodeFinished()`
       = note: the enum is complex because of non-unit variant `NodeStarted`
  --> crates/wdl-grammar/src/parser.rs:42:5
   |
42 |     /// A node has finished.
   |     ^^^^^^^^^^^^^^^^^^^^^^^^
```

As stated in the error message, this could be fixed by making `Event::NodeFinished` a tuple variant, but that is a breaking change that I wasn't willing to make. `PyEvent` is my solution: a re-implementation of `Event` that is annotated with `#[pyclass]`. It's possible to create a `PyEvent` from an `Event` using `PyEvent::from_event()`, and if it becomes necessary to convert a `PyEvent` back to an `Event`, I will implement that as well.

### Why does the Python equivalent of `document()` take a string and not a a `Parser`?

For the same reason I'm purposefully not implementing the [`wdl_grammar::lexer`](https://docs.rs/wdl-grammar/0.24.0/wdl_grammar/lexer/index.html) module: generics. `Parser`, and the `Lexer` that is needed to construct it, are both generic over the token type. Generics are _really hard_ to represent in Python bindings. While possible (using [`__class_getitem__()`](https://docs.python.org/3/reference/datamodel.html#emulating-generic-types), see [PyO3's docs on generics](https://pyo3.rs/v0.29.0/python-typing-hints.html#supporting-generics)), I decided it wasn't worth the complexity in this case. This can absolutely be changed in the future if the need for these types arises.

---

For external contributors:

For all contributors:


python/sprocket_bio/examples/event_stream.py
```diff
@@ -0,0 +1,65 @@
...
+        match event:
+            case Event.NodeStarted(kind, _):
+                print("  " * indent + str(kind))
+                indent += 1
+            case Event.NodeFinished():
+                indent -= 1
+            case Event.Token(kind, span):
+                print(
+                    "  " * indent + f"{kind}@{span} {source[span.start : span.end]!r}"
+                )
```
claymcleod: `Span` comes from Rust byte offsets, but Python slices `str` by code point. For WDL containing non-ASCII text before or inside a token, this prints the wrong substring. We should slice the UTF-8 bytes and decode the token text back to `str`.

```suggestion
                token_text = source.encode("utf-8")[span.start : span.end].decode("utf-8")
                print("  " * indent + f"{kind}@{span} {token_text!r}")
```

python/sprocket_bio/examples/syntax_highlighter.py
```diff
@@ -0,0 +1,141 @@
...
+            # Filter the parser events for tokens, which represent individual pieces of syntax.
+            if isinstance(event, Event.Token):
+                # If this token should be colored then write it inside of a styled `<span>`, else
+                # just write the plain text.
+                if color := syntax_kind_color(event.kind):
+                    f.write(
+                        f'<span style="color: {color}">{source[event.span.start : event.span.end]}</span>'
+                    )
+                else:
+                    f.write(source[event.span.start : event.span.end])
```
claymcleod: Same byte-offset issue here: `event.span` indexes UTF-8 bytes from Rust, while `source[...]` indexes Python code points. Highlighting a file with non-ASCII text before a token can color the wrong text; separately, writing token text directly into HTML lets comments or strings become browser-interpreted markup. Let’s encode once, decode each token slice, and escape it before writing. It would also be worth adding a non-ASCII regression test for both examples so this does not regress.

```suggestion
        from html import escape

        source_bytes = source.encode("utf-8")

        for event in events:
            # Filter the parser events for tokens, which represent individual pieces of syntax.
            if isinstance(event, Event.Token):
                # If this token should be colored then write it inside of a styled `<span>`, else
                # just write the plain text.
                token_text = escape(
                    source_bytes[event.span.start : event.span.end].decode("utf-8")
                )
                if color := syntax_kind_color(event.kind):
                    f.write(f'<span style="color: {color}">{token_text}</span>')
                else:
                    f.write(token_text)
```

python/sprocket_bio/examples/syntax_highlighter.py
```diff
@@ -0,0 +1,141 @@
...
+    # Emit diagnostics, if there are any.
+    emit_diagnostics(
+        str(args.source_file), source, diagnostics, report_mode=Mode.FULL, colorize=True
+    )
+
+    # If any of the diagnostics are errors, exit.
+    if any([d.severity is Severity.ERROR for d in diagnostics]):
+        sys.exit(1)
+
+    with open(args.output, mode="wt", encoding="utf-8") as f:
```
claymcleod, line 123: If `--output` points at the input WDL file, this opens it with `"wt"` and truncates the source before writing the HTML. We should reject that case before opening the output path.

```suggestion
    if args.output.resolve() == args.source_file.resolve():
        sys.exit(f"refusing to overwrite source file: {args.output}")

    with open(args.output, mode="wt", encoding="utf-8") as f:
```

python/tests/grammar/test_grammar.py
```diff
@@ -0,0 +1,84 @@
+from sprocket_bio.diagnostics import emit_diagnostics, Mode
```
claymcleod: `mypy python` currently fails before checking the rest of the package because this new file and `python/tests/test_grammar.py` both map to the module name `test_grammar` without a package marker:

```text
python/tests/test_grammar.py: error: Duplicate module named "test_grammar" (also at "python/tests/grammar/test_grammar.py")
```

Could you add `python/tests/grammar/__init__.py` or rename this file so the grammar tests have a distinct module name?

crates/wdl-grammar/src/parser.rs
```diff
@@ -1287,6 +1289,100 @@ where
...
+                } => Ok(format!(
+                    "Event.NodeStarted({}, {})",
+                    // Equivalent to `repr(SyntaxKind)`.
+                    kind.into_bound_py_any(py)?.repr()?.to_str()?,
+                    // If `forward_parent` is `Some` write the plain integer, else write "None".
+                    std::fmt::from_fn(|f| match forward_parent {
+                        Some(forward_parent) => write!(f, "{forward_parent}"),
+                        None => write!(f, "None"),
+                    }),
+                )),
```
claymcleod, line 1356: This is why the `lint` job fails: `std::fmt::from_fn` is newer than the workspace MSRV, and Clippy rejects it under `-D warnings`. We can format the optional value with an MSRV-compatible `match` first.

```suggestion
                } => {
                    let forward_parent = match forward_parent {
                        Some(forward_parent) => forward_parent.to_string(),
                        None => "None".to_owned(),
                    };

                    Ok(format!(
                        "Event.NodeStarted({}, {})",
                        // Equivalent to `repr(SyntaxKind)`.
                        kind.into_bound_py_any(py)?.repr()?.to_str()?,
                        forward_parent,
                    ))
                }
```

crates/wdl-grammar/src/tree.rs
```diff
@@ -902,6 +914,59 @@ impl SyntaxTokenExt for SyntaxToken {
...
+        #[pyo3(name = "is_symbolic")]
+        fn py_is_symbolic(&self) -> bool {
+            self.is_symbolic()
+        }
+
+        /// Describes the syntax kind.
+        #[pyo3(name = "describe")]
+        fn py_describe(&self) -> &'static str {
+            self.describe()
+        }
```
claymcleod, line 953: `SyntaxKind.describe()` is now public Python API, but `describe()` panics for exposed symbolic values like `UNKNOWN`, `UNPARSED`, `ABANDONED`, and `MAX`. A Python caller can hit this directly with `SyntaxKind.UNKNOWN.describe()`, so let’s turn that into a normal Python exception before calling the Rust method.

```suggestion
mod python {
    use pyo3::exceptions::PyValueError;
    use pyo3::prelude::*;

    use super::*;

    #[pymethods]
    impl SyntaxKind {
        /// Returns whether the token is a symbolic `SyntaxKind`.
        ///
        /// Generally speaking, symbolic `SyntaxKind`s have special meanings
        /// during parsing—they are not real elements of the grammar but rather
        /// an implementation detail.
        #[pyo3(name = "is_symbolic")]
        fn py_is_symbolic(&self) -> bool {
            self.is_symbolic()
        }

        /// Describes the syntax kind.
        #[pyo3(name = "describe")]
        fn py_describe(&self) -> PyResult<&'static str> {
            if self.is_symbolic() {
                return Err(PyValueError::new_err(format!(
                    "cannot describe symbolic syntax kind `{self:?}`"
                )));
            }

            Ok(self.describe())
        }
```

crates/wdl-grammar/src/grammar.rs
```diff
@@ -159,3 +162,35 @@ fn version_statement(
...
+        // is why it is substituted out here.
+        source: &str,
+        fallback_version: Option<SupportedVersion>,
+    ) -> (Vec<PyEvent>, Vec<Diagnostic>) {
+        let (events, diagnostics) = document(Parser::new(Lexer::new(source)), fallback_version);
+
+        // Converts `Event`s into `PyEvent`s.
+        let events = events
+            .into_iter()
+            .map(|event| PyEvent::from_event(event))
```
claymcleod: This is one of the current `lint` failures: Clippy flags this as `redundant_closure` under `-D warnings`.

```suggestion
            .map(PyEvent::from_event)
```

crates/wdl-grammar/src/parser.rs
```diff
@@ -1287,6 +1289,100 @@ where
     }
 }
 
+#[cfg(feature = "unstable-python")]
+mod python {
```
claymcleod, line 1294: This is the other `lint` failure in this file: the crate denies `clippy::missing_docs_in_private_items`, so the private Python module needs a doc comment.

```suggestion
/// Python-specific APIs.
#[cfg(feature = "unstable-python")]
mod python {
```

python/sprocket_bio/grammar/parser.pyi
```diff
@@ -0,0 +1,36 @@
...
+    @typing.final
+    class Token(Event):
+        kind: SyntaxKind
+        span: Span
+        __match_args__ = ("kind", "span")
+
+        def __new__(cls, kind: SyntaxKind, span: Span) -> Event.Token: ...
+
+    # `Event` cannot be constructed directly, you must construct one of its subclasses.
+    def __new__(cls) -> typing.NoReturn: ...
```
claymcleod: After the duplicate test module is fixed, `mypy` still rejects this stub because `@disjoint_base` sees `Event` as `Callable[[], Never]`, which also makes `Event.abandoned()` look missing. Removing the base `__new__` stub keeps the variants typed correctly and matches the rest of the file.

```suggestion
```

python/sprocket_bio/grammar/version.pyi
```diff
@@ -0,0 +1,33 @@
...
+    def __lt__(self, other: typing.Any, /) -> bool: ...
+    def __le__(self, other: typing.Any, /) -> bool: ...
+    def __gt__(self, other: typing.Any, /) -> bool: ...
+    def __ge__(self, other: typing.Any, /) -> bool: ...
+
+@disjoint_base
+class SupportedVersion:
+    @typing.final
+    class V1(SupportedVersion):
+        _0: V1
```
claymcleod: The Rust class is frozen, so `_0` is not writable at runtime. The stub should mark it as final the same way the other frozen attributes do.

```suggestion
        _0: typing.Final[V1]
```

python/tests/grammar/test_grammar.py
```diff
@@ -0,0 +1,84 @@
...
+def test_document_error() -> None:
+    events, [diagnostic] = document("", SupportedVersion.V1(V1.ZERO))
+
+    assert events == [
+        Event.NodeStarted(SyntaxKind.ROOT_NODE, None),
+        Event.NodeFinished(),
+    ]
+
+    assert diagnostic.severity is Severity.ERROR
+    assert "must start with a version statement" in diagnostic.message
```
claymcleod: The Python `document()` wrapper takes `fallback_version`, but the tests only cover the missing-version error path. A small unsupported-version test would lock in the fallback behavior that callers need when parsing development or future WDL versions.

```suggestion
    assert "must start with a version statement" in diagnostic.message


def test_document_uses_fallback_version_for_unsupported_version() -> None:
    events, diagnostics = document("version 2.0\n\nworkflow test {}", SupportedVersion.V1(V1.ZERO))

    assert diagnostics == []
    assert Event.NodeStarted(SyntaxKind.WORKFLOW_DEFINITION_NODE, None) in events
```

claymcleod (changes requested):

claymcleod:
As this Python API surface grows, it might be worth considering `pyo3-stub-gen` or another stub-generation workflow so the `.pyi` files do not drift from the PyO3 bindings. It looks possible here because this is a PyO3 + maturin project, but probably not free: the generator needs Rust-side annotations and manual overrides for complex cases like nested modules, enum variants, and public re-exports.

crates/wdl-grammar/src/parser.rs
BD103: Darn, I was really hoping to avoid the extra `"None".to_owned()` allocation! I supported I could use `Cow` here, but that's beginning to feel like premature optimization to me.

python/sprocket_bio/grammar/parser.pyi
BD103: Thank you, I was very confused on why this was happening! The error message `Value of type variable "_TC" of "disjoint_base" cannot be "Callable[[], Never]"` isn't very clear to me 😅

python/sprocket_bio/examples/event_stream.py
BD103: Oooo good catch, I haven't considered that!

BD103:
> As this Python API surface grows, it might be worth considering `pyo3-stub-gen` or another stub-generation workflow so the `.pyi` files do not drift from the PyO3 bindings. It looks possible here because this is a PyO3 + maturin project, but probably not free: the generator needs Rust-side annotations and manual overrides for complex cases like nested modules, enum variants, and public re-exports.

Agreed. This is especially true as I've found some cases that `mypy.stubtest` does not catch. For example, it's unable to catch incorrect function return types, as verifying that requires running the function. Switching to a generator means we don't need to completely rely on `mypy.stubtest` for correctness.

claymcleod:
Thanks for addressing the review comments. One small issue with the overwrite guard: `os.path.samefile(args.source_file, args.output)` raises `FileNotFoundError` when `--output` points at a file that does not exist yet, including the default `source.wdl.html` path. Could you switch this back to comparing resolved paths, or guard `samefile` behind `args.output.exists()`?

BD103:
> One small issue with the overwrite guard: `os.path.samefile(args.source_file, args.output)` raises `FileNotFoundError` when `--output` points at a file that does not exist yet, including the default `source.wdl.html` path. Could you switch this back to comparing resolved paths, or guard `samefile` behind `args.output.exists()`?

Done! I like `samefile()` because it handles symlinks, so I chose to check that the output exists first.

PR #982, opened 2026-07-06
feat(sprocket): test YAML parse diagnostics

Serial-ATA:
Replaced the `serde-yaml-ng` parsing with `serde-saphyr`. Now we're able to produce spanned diagnostics and report multiple errors at once.

This also moves all test types to a new `sprocket-test-types` crate for future use in the LSP.

For external contributors:

For all contributors:

      is implemented.


crates/sprocket-test-types/src/assertions.rs
```diff
@@ -1,188 +1,332 @@
...
+            "`{}` assertions have no effect for workflows",
+            field.0.value
+        ))
+        .with_highlight(convert_yaml_span(field.0.defined.span()))
+}
+
+/// A regex failed to compile.
+fn invalid_regex(re: &str, field: &Spanned<Value>) -> Diagnostic {
+    Diagnostic::error(format!("invalid regex: `{re}`"))
+        .with_help("regexes must follow the syntax defined here: <https://docs.rs/regex/latest/regex/#syntax>")
```
a-frantz, line 60: this is a good help note. I didn't think about how we have to use rust flavored REs.

crates/sprocket-test-types/src/assertions.rs
```diff
@@ -1,188 +1,332 @@
...
+    /// The expected exit code of the task.
+    ///
+    /// If undefined, this defaults to `0`.
+    pub fn exit_code(&self) -> i32 {
+        self.exit_code
+            .as_ref()
+            .map_or(0, |exit_code| exit_code.value)
+    }
+
     /// Parse the assertions from the serde definitions.
```
a-frantz:
```suggestion
    /// Parse the assertions from the user supplied values.
```

crates/sprocket-test-types/src/inputs.rs
```diff
@@ -0,0 +1,308 @@
...
+            format!("`{}` has {first_group_length} entries", first_group.0.value),
+            convert_yaml_span(first_group.0.defined.span()),
+        )
+        .with_label(
+            format!("`{}` has {group_length} entries", group.0.value),
+            convert_yaml_span(group.0.defined.span()),
+        )
+        .with_help(
+            "all sequences in a group must have the same length, with the first group setting the \
+             precedent",
```
a-frantz:
```suggestion
            "all sequences in a group must have the same length",
```
nit: as they must be equal none of them set a precedent, and the second part of this is unnecessary

crates/sprocket-test-types/src/inputs.rs
```diff
@@ -0,0 +1,308 @@
...
+use wdl_ast::Diagnostic;
+
+use crate::convert_yaml_span;
+use crate::expected_mapping;
+use crate::yaml::MaybeMap;
+use crate::yaml::Spanned;
+use crate::yaml::SpannedField;
+
+/// A key is defined multiple times.
+fn duplicate_key(first: &Spanned<String>, duplicate: &Spanned<String>) -> Diagnostic {
```
a-frantz, line 20: maybe this will be answered further in my review, but isn't this a violation of the YAML spec? I wouldn't expect to need to write this diagnostic out manually

crates/sprocket-test-types/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "sprocket-test-types"
+version = "0.1.0"
+license.workspace = true
+edition.workspace = true
+authors.workspace = true
+homepage.workspace = true
+rust-version.workspace = true
+repository = "https://github.com/stjude-rust-labs/sprocket/tree/main/crates/wdl-analysis"
```
a-frantz:
```suggestion
repository = "https://github.com/stjude-rust-labs/sprocket/tree/main/crates/sprocket-test-types"
```

tests/cli/dev/test/invalid-yaml/unsupported-fields/stderr
```diff
@@ -0,0 +1,24 @@
+warning: useless `exit_code` assertion
```
a-frantz, line 1:
```suggestion
warning: ignored `exit_code` assertion
```
nit: "ignored" reads better than "useless" to me. Same for all the following diagnostics as well.

crates/sprocket-test-types/src/inputs.rs
Serial-ATA: usually, but not when groups are involved. see here: https://github.com/Serial-ATA/sprocket/blob/3b3523fb4ad369557456dcb180036ec1bf0a8fe1/tests/cli/dev/test/invalid-yaml/bad-input-groupings/stderr#L12-L19

from this definition: https://github.com/Serial-ATA/sprocket/blob/3b3523fb4ad369557456dcb180036ec1bf0a8fe1/tests/cli/dev/test/invalid-yaml/bad-input-groupings/inputs/foo/foo.yml#L12-L22

a-frantz: got it, thanks!

tests/cli/dev/test/invalid-yaml/unsupported-fields/stderr
Serial-ATA: IMO "useless" makes it immediately clear to me that the assertion isn't necessary and I can delete it. "ignored" makes me think "why is it being ignored?"

a-frantz: something about "useless" feels incorrect to me, but I can't put my finger on why 🤔 but, not worth holding up the PR over a bad vibe, so this can be ignored and resolved

a-frantz (approved):

Serial-ATA:
oops, I somehow overwrite a completely unrelated test file

PR #992, opened 2026-07-09
perf: cache compiled regexes in string stdlib functions

### LINKED ISSUE #988: stdlib regex functions recompile their pattern on every call
The `sub`, `find`, `matches`, and `split` stdlib functions call `Regex::new` on each invocation ([`sub.rs`](https://github.com/stjude-rust-labs/sprocket/blob/f557310de46e8402dc14dd9352853f94f2910158/crates/wdl-engine/src/stdlib/sub.rs#L96), [`find.rs`](https://github.com/stjude-rust-labs/sprocket/blob/f557310de46e8402dc14dd9352853f94f2910158/crates/wdl-engine/src/stdlib/find.rs#L36), [`matches.rs`](https://github.com/stjude-rust-labs/sprocket/blob/f557310de46e8402dc14dd9352853f94f2910158/crates/wdl-engine/src/stdlib/matches.rs#L32), [`split.rs`](https://github.com/stjude-rust-labs/sprocket/blob/f557310de46e8402dc14dd9352853f94f2910158/crates/wdl-engine/src/stdlib/split.rs#L36)). These functions typically run per element inside scatters or array operations, so the same pattern gets recompiled thousands of times in a workflow that uses, say, `sub()` on every row of an input table. Compiling a regex is orders of magnitude slower than matching with one.

A small global cache keyed by pattern string would fix this—something like a `LazyLock<Mutex<HashMap<String, Regex>>>` shared by the four functions, following the same shape as the [digest memoization](https://github.com/stjude-rust-labs/sprocket/blob/f557310de46e8402dc14dd9352853f94f2910158/crates/wdl-engine/src/digest.rs#L67). A plain `HashMap` that grows with the number of distinct patterns is probably fine (workflows use a handful), but an LRU would bound memory if we want to be careful. Invalid patterns should keep returning the existing diagnostic rather than being cached.

Each of the four files has unit tests at the bottom that should pass unchanged, and a `cache hit returns same behavior` style test would be a welcome addition.

Rehan30g:
Closes #988.

The `sub`, `find`, `matches`, and `split` stdlib functions each called `Regex::new` on every invocation. These run per element inside scatters and array operations, so the same pattern ends up recompiled thousands of times in a workflow that uses one of them on every row of an input. Compiling a regex is a lot more expensive than matching with one that already exists.

### What changed

I added a small shared cache in `stdlib.rs`, a `LazyLock<Mutex<HashMap<String, Regex>>>` with a `cached_regex` helper, following the same shape as the digest memoization the issue pointed at. The four functions now call `super::cached_regex(...)` instead of `Regex::new(...)`. A `Regex` is cheap to clone since it is reference counted internally, so callers get their own handle without recompiling.

Invalid patterns are not inserted into the cache and their compilation error is returned to the caller exactly as before, so the error path is unchanged.

### Tests

Added two unit tests covering the cache: valid patterns compile and get reused, invalid patterns keep returning an error and are not cached. Existing `sub`/`find`/`matches`/`split` tests still pass.

```
cargo test -p wdl-engine --lib stdlib::
test result: ok. 68 passed; 0 failed
```

`cargo clippy -p wdl-engine` is clean and the code is formatted with the repo's nightly rustfmt config. I also added a CHANGELOG entry under Unreleased.

### DIFF
--- a/crates/wdl-engine/CHANGELOG.md
+++ b/crates/wdl-engine/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+* The `sub`, `find`, `matches`, and `split` standard library functions now reuse compiled regular expressions through a shared cache instead of recompiling the pattern on every call ([#988](https://github.com/stjude-rust-labs/sprocket/issues/988)).
+
 ### Added
 
 * Added a `strongish` content digest mode (`run.task.digests = "strongish"`) that hashes file size, last modified time, and the first 10 MiB of a file's contents; this is an intermediate strategy between `weak` and `strong`, similar to Cromwell's `fingerprint` call caching strategy ([#978](https://github.com/stjude-rust-labs/sprocket/pull/978)).
--- a/crates/wdl-engine/src/stdlib.rs
+++ b/crates/wdl-engine/src/stdlib.rs
@@ -4,11 +4,13 @@ use std::collections::HashMap;
 use std::path::Path;
 use std::path::PathBuf;
 use std::sync::LazyLock;
+use std::sync::Mutex;
 
 use anyhow::Result;
 use anyhow::anyhow;
 use anyhow::bail;
 use futures::future::BoxFuture;
+use regex::Regex;
 use tempfile::TempPath;
 use wdl_analysis::stdlib::Binding;
 use wdl_analysis::types::Type;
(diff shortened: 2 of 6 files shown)


claymcleod:
Hi @Rehan30g! Thanks for this submission. Would you mind filling out the contributor checklist in the PR template? Thank you!

Rehan30g:
Hey, sorry about this, I need to be upfront. I used AI assistance to help put this PR together, which goes against what your contributor checklist asks for. I should have caught that before submitting. Closing this one out, appreciate you taking a look regardless.

PR #999, opened 2026-07-10
feat: add `sprocket module` porcelain

claymcleod:
Added the experimental `sprocket dev module` porcelain for creating, resolving, updating, inspecting, signing, verifying, and trusting WDL modules. The command group now manages `module.json`, reproducible `module-lock.json` files, Git shorthand and monorepo paths, default-branch fallback when no matching version tags exist, scoped cache operations, and signer-change policy; `sprocket run` and `sprocket submit` regenerate missing or stale lockfiles before execution.

Expanded `wdl-modules` with reusable project discovery, lossless manifest editing, crash-safe manifest/lockfile transactions, relocking, cache verification, and trust operations, while Sprocket retained CLI configuration, prompting, and presentation. The module formats were finalized around Git tag versions, `sha`-pinned lock entries, signer-bound `module.sig` payloads, and stricter tree and import validation; analysis also gained version-compatible task and workflow re-exports for module imports. Existing pre-release module manifests, lockfiles, and signatures may need regeneration.

For all contributors:


adthrasher:
Two configuration suggestions for defaults to the `init` command. I'd want to put a configuration key in sprocket.toml like:

```
authors = ["Ari Frantz <Ari.Frantz@STJUDE.ORG>","Andrew Thrasher <Andrew.Thrasher@STJUDE.ORG>"]
```

I think people will also want to override `license`, so we should make that configurable as well.

crates/wdl-modules/src/dependency/source.rs
```diff
@@ -259,8 +282,32 @@ pub enum GitSelector {
     Tag(String),
     /// A Git branch name.
     Branch(String),
-    /// A full Git commit SHA.
-    Commit(GitCommit),
+    /// A Git commit SHA, or any unique prefix of one (7–40 hex chars).
```
peterhuene: Double checking that this comment is accurate. The lower bound seems like it's 4 hex characters in other places.

a-frantz:
A git repo configured with SSH instead of HTTP breaks `module init`:
```
$  sprocket dev module init
error: validating generated manifest

Caused by:
    0: invalid `module.json` JSON
    1: relative URL without a base: "git@github.com:stjudecloud/workflows.git"
```

crates/wdl-modules/src/project/mutation/transaction.rs
```diff
@@ -0,0 +1,969 @@
...
+) -> Result<(), ProjectMutationError> {
+    let active = journal_root.join(ACTIVE_DIRECTORY);
+    match std::fs::symlink_metadata(&active) {
+        Ok(metadata) if metadata.is_dir() && !metadata.file_type().is_symlink() => {}
+        Ok(_) => {
+            return Err(ProjectMutationError::InvalidPath {
+                path: active,
+                expected: "directory",
+            });
+        }
```
peterhuene: Could this be simplified to:


```suggestion
        Ok(metadata) => {
            validate_directory_metadata(&active, &metadata)?;
        }
```

crates/wdl-modules/src/project/mutation/transaction.rs
```diff
@@ -0,0 +1,969 @@
...
+        Err(source) => Err(ProjectMutationError::Io {
+            operation: "inspecting",
+            path: path.to_path_buf(),
+            source,
+        }),
+    }
+}
+
+/// Attempts to remove the journal directory when it is empty; silently
+/// ignores errors so that unrelated journal contents are not affected.
```
peterhuene: Should this ignore _all_ unrelated errors? Seems like it should scope to perhaps ignoring `ErrorKind::DirectoryNotEmpty` at a minimum.

crates/wdl-modules/src/project/mutation/transaction.rs
```diff
@@ -0,0 +1,969 @@
...
+    sync_project_files(project.manifest_path(), project.lockfile_path())?;
+    remove_directory_if_present(&active)?;
+    sync_directory(journal_root)?;
+    cleanup_journal_if_empty(journal_root);
+    Ok(())
+}
+
+#[cfg(test)]
+thread_local! {
+    static FAIL_MANIFEST_RESTORE: std::cell::RefCell<Option<PathBuf>> =
```
peterhuene: Despite being test code, I think this warrants a comment describing its purpose for injecting a simulated I/O error for testing rollback behavior.

Couldn't hurt to have a comment on the related functions either.

crates/wdl-modules/src/project/mutation/transaction.rs
```diff
@@ -0,0 +1,969 @@
...
+impl Drop for ManifestRestoreFailureGuard {
+    fn drop(&mut self) {
+        FAIL_MANIFEST_RESTORE.with(|path| {
+            path.borrow_mut().take();
+        });
+    }
+}
+
+#[cfg(test)]
+fn fail_manifest_restore_for_test(path: &Path) -> ManifestRestoreFailureGuard {
```
peterhuene: I think we can probably move this function and `ManifestRestoreFailureGuard` to the `test` submodule.

In fact, I'd be inclined to move all of this `#[cfg(test)]` code to the `test` submodule and make it and `maybe_fail_manifest_restore` use `pub(super)` visibility.

crates/wdl-modules/src/project/mutation/transaction.rs
```diff
@@ -0,0 +1,969 @@
...
+    if metadata.file_type().is_symlink() || !metadata.is_dir() {
+        return Err(ProjectMutationError::InvalidPath {
+            path: path.to_path_buf(),
+            expected: "directory",
+        });
+    }
+    Ok(())
+}
+
+/// Ensures `path` is a non-symlink directory owned by this process, creating
```
peterhuene: > Ensures `path` is a non-symlink directory owned by this process...

How does it ensure the directory is "owned" by the calling process?

crates/wdl-modules/src/project/mutation/transaction.rs
```diff
@@ -0,0 +1,969 @@
...
+                operation: "serializing `module-lock.json`",
+                path: PathBuf::from("module-lock.json"),
+                source,
+            })?;
+    }
+    Ok(())
+}
+
+/// Writes the manifest document to `path` atomically.
+fn write_manifest_atomically(
```
peterhuene: On the surface, this feels like a lot of duplicated functionality with `write_lockfile_atomically` (and possibly `write_bytes_atomically`).

Perhaps instead a single `write_atomically` function that accepts a closure which takes the (temp) `&mut File` and returns `io::Result<()>`?

crates/wdl-modules/src/project/mutation/transaction.rs
```diff
@@ -0,0 +1,969 @@
...
+fn snapshot_file(journal: &Path, label: &str, path: &Path) -> Result<(), ProjectMutationError> {
+    match std::fs::symlink_metadata(path) {
+        Ok(metadata) if !metadata.is_file() || metadata.file_type().is_symlink() => {
+            return Err(ProjectMutationError::InvalidPath {
+                path: path.to_path_buf(),
+                expected: "file",
+            });
+        }
+        Ok(_) => {
+            let bytes = std::fs::read(path).map_err(|source| ProjectMutationError::Io {
```
peterhuene: Could we utilize `std::io::copy` instead of buffering the entire file contents?

crates/wdl-modules/src/project/mutation/transaction.rs
```diff
@@ -0,0 +1,969 @@
...
+
+/// Restores one file from the journal snapshot at `<label>.before` or removes
+/// it when the journal records `<label>.absent`.
+fn restore_snapshot(journal: &Path, label: &str, path: &Path) -> Result<(), ProjectMutationError> {
+    let snapshot = journal.join(format!("{label}.before"));
+    let absent = journal.join(format!("{label}.absent"));
+
+    match std::fs::symlink_metadata(&snapshot) {
+        Ok(metadata) if metadata.is_file() && !metadata.file_type().is_symlink() => {
+            let bytes = std::fs::read(&snapshot).map_err(|source| ProjectMutationError::Io {
```
peterhuene: This also seems like a place where `std::io::copy` could be utilized instead of buffering the file.

crates/wdl-modules/src/project/document.rs
```diff
@@ -0,0 +1,302 @@
...
+use crate::dependency::DependencyName;
+use crate::dependency::DependencySource;
+use crate::manifest::ManifestError;
+
+/// An error parsing or editing a `module.json` document.
+#[derive(Debug, Error)]
+pub enum ManifestDocumentError {
+    /// The `module.json` bytes failed strict manifest validation.
+    #[error("invalid module manifest")]
+    Manifest(#[source] ManifestError),
```
peterhuene: Should this and the `Json` variant use `#[from]` so that we can omit the `map_err` conversions below?

crates/wdl-modules/src/project/document.rs
```diff
@@ -0,0 +1,302 @@
...
+        let mut value = self.value.clone();
+        let root = value
+            .as_object_mut()
+            .ok_or(ManifestDocumentError::RootNotObject)?;
+        let dependencies = root
+            .entry("dependencies")
+            .or_insert_with(|| serde_json::Value::Object(serde_json::Map::new()))
+            .as_object_mut()
+            .ok_or(ManifestDocumentError::DependenciesNotObject)?;
+        if let Some(existing) = self.matching_dependency_key(&name)
```
peterhuene: This could use a comment explaining why we need to remove existing entries prior to insertion rather than relying on `insert` to replace the entry (i.e. dependency name normalization).

crates/wdl-modules/src/project/document.rs
```diff
@@ -0,0 +1,302 @@
...
+        &self.manifest
+    }
+
+    /// Sets or replaces one dependency entry in `module.json`.
+    ///
+    /// Unrelated manifest fields and dependency extension fields stay intact.
+    /// The in-memory document changes only after the edited result reparses as
+    /// a valid manifest, so failed edits leave both the document bytes and the
+    /// validated view unchanged.
+    pub fn set_dependency(
```
peterhuene: A better name might be `insert_dependency` which follows the common "insert or update" pattern of an associative container in Rust.

crates/wdl-modules/src/project/document.rs
```diff
@@ -0,0 +1,302 @@
...
+        let mut bytes =
+            serde_json::to_vec_pretty(&self.value).map_err(ManifestDocumentError::Json)?;
+        bytes.push(b'\n');
+        Ok(bytes)
+    }
+
+    /// Replaces the stored JSON value after revalidating the full document.
+    fn replace_validated(&mut self, value: serde_json::Value) -> Result<(), ManifestDocumentError> {
+        let mut bytes = serde_json::to_vec_pretty(&value).map_err(ManifestDocumentError::Json)?;
+        bytes.push(b'\n');
```
peterhuene: Do we really require the file to end with a new line to parse it?

crates/wdl-modules/src/project/document.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// with a trailing newline.
+    pub fn to_bytes(&self) -> Result<Vec<u8>, ManifestDocumentError> {
+        let mut bytes =
+            serde_json::to_vec_pretty(&self.value).map_err(ManifestDocumentError::Json)?;
+        bytes.push(b'\n');
+        Ok(bytes)
+    }
+
+    /// Replaces the stored JSON value after revalidating the full document.
+    fn replace_validated(&mut self, value: serde_json::Value) -> Result<(), ManifestDocumentError> {
```
peterhuene: I'm curious as to why this validation needs to occur at insertion/removal time rather than being ok with delaying it until we attempt to serialize the modified in-memory representation.

It seems like an expensive operation to do for the unlikely event that a change can't round-trip.

That said, what kind of edits do we envision that this would guard against? The corresponding test case does it for an invalid dependency name, which is caught _prior_ to actually modifying anything internally.

If it's for the checks for the root or dependencies being an object, that feels like something that `Manifest::parse` should be ensuring and an invariant maintained for `ManifestDocument`. 

If it's to catch errors for _serializing_ a dependency being inserted, that can also be done prior to modifying the internal fields in any way.

crates/wdl-modules/src/project/document.rs
```diff
@@ -0,0 +1,302 @@
...
+        bytes.push(b'\n');
+        let manifest = Manifest::parse(&bytes).map_err(ManifestDocumentError::Manifest)?;
+        self.value = value;
+        self.manifest = manifest;
+        Ok(())
+    }
+
+    /// Parses one dependency key with manifest validation rules.
+    fn parse_dependency_name(name: &str) -> Result<DependencyName, ManifestDocumentError> {
+        name.parse().map_err(|_| {
```
peterhuene: I wonder if we could leverage the returned `DependencyNameError` in `ManifestError::InvalidDependencyName` rather than discarding it here.

crates/wdl-modules/src/project/mutation.rs
```diff
@@ -0,0 +1,352 @@
...
+            Self::Manifest(_) => None,
+        }
+    }
+}
+
+/// An error applying a mutation to a module project.
+#[derive(Debug, Error)]
+pub enum ProjectMutationError {
+    /// Loading, reloading, or recovering a project path failed.
+    #[error("project error")]
```
peterhuene: Feels like this should be `#[error(transparent)]`. This message isn't adding any context.

crates/wdl-modules/src/project/mutation.rs
```diff
@@ -0,0 +1,352 @@
...
+    Project(#[from] ProjectError),
+    /// The proposed `module.json` update could not be serialized.
+    #[error("invalid manifest update")]
```
peterhuene: Likewise, I'd consider this one for also being `#[error(transparent)]`.

crates/wdl-modules/src/project/mutation/transaction.rs
peterhuene, line 1: I'm not entirely convinced that implementing "transactional" mutations of `module.json` and the corresponding lock file are worth the complexity, to be honest.

I don't believe `cargo` and `cargo-edit` go to these lengths to handle synchronizing updates to `Cargo.toml` and `Cargo.lock`.

Tools that edit the dependencies in `Cargo.toml` generally do so by deserializing the file, performing in-memory edits, serializing the in-memory changes back to `Cargo.toml` (one could still utilize an atomic rename of a temporary file to prevent partial writes from a failed serialization, if desired) and then, upon success, attempt to synchronize the lock file.

Note that the edit to `Cargo.toml` isn't done under an advisory lock; it's always last one in wins because the edit could come from `cargo-edit` or a text editor, it isn't "owned" by those programs the way `Cargo.lock` is fully maintained by `cargo` so there's no way to coordinate changes to `Cargo.toml`.

`Cargo.lock` synchronization typically starts by reading it under a shared advisory lock into an in-memory representation (a default one if `Cargo.lock` doesn't exist) considered to be the "baseline", then resolving dependencies utilizing that lock file representation, then generating a new in-memory representation of the lock file from what was resolved, performing a diff of that new lock file representation against the baseline, and then finally, only if they differ, does it acquire an exclusive lock and write the new contents of `Cargo.lock`. It's absolutely possible for `Cargo.toml` and `Cargo.lock` to be out-of-sync and that's okay; keeping them in-sync is really just a reduction of work optimization for future invocations of `cargo`.

I much prefer that model which doesn't require backing up or restoring files, manual calls to `fsync`, etc.

crates/wdl-modules/src/project/mutation.rs
```diff
@@ -0,0 +1,352 @@
...
+    /// Caller-rooted journal directory for this project key.
+    journal_root: PathBuf,
+    /// Open lock file handle that keeps the operating-system lock alive.
+    _lock: File,
+}
+
+impl LockedModuleProject {
+    /// Acquires the project mutation lock under `state_root`.
+    ///
+    /// This blocks on the per-project global lock in `<state_root>/locks`,
```
peterhuene: Given the number of directories here and in `transaction.rs`, I think a detailed comment somewhere that gives the overall filesystem layout of a module "project" would be helpful for people to keep track of what goes where.

_Edit from progressing further in the review: it looks like this might have been in another file, so ignore if that's the case._

crates/wdl-modules/src/resolver/git_resolver.rs
peterhuene, line 1: It feels like this should probably be in the `git` module as `git_resolver` being a submodule of `resolver` seems redundant to me.

Perhaps we can merge this file with what's currently in `git.rs` but move the functions currently in `git` into an `ops` submodule of `git`?

crates/wdl-modules/src/resolver/git_resolver/locked.rs
```diff
@@ -0,0 +1,254 @@
...
+use crate::dependency::DependencyName;
+use crate::lockfile::DependencyMap;
+use crate::lockfile::ResolvedSource;
+use crate::module::Module;
+use crate::resolver::cache::CacheKey;
+use crate::resolver::error::ResolverError;
+use crate::resolver::scope::DependencyScope;
+use crate::resolver::scope::ResolutionMode;
+
+impl GitResolver {
```
peterhuene: I'm not the biggest of fans of splitting the `impl` of a type across several modules, to be honest.

I get the intent, as the resolver is quite complex, but I personally wouldn't mind if all of the `GitResolver` code was in `git.rs` (we could still split out the tests into a `tests` submodule file, though; I think one file would be fine, but I'm fine with splitting test code into separate files).

crates/wdl-modules/src/resolver/git_resolver/materialize.rs
```diff
@@ -0,0 +1,556 @@
...
+    pub(super) async fn plan_git_materialization(
+        &self,
+        name: &DependencyName,
+        url: &url::Url,
+        selector: &GitSelector,
+        path: &Option<GitModulePath>,
+        scope: DependencyScope,
+        mode: ResolutionMode<'_>,
+    ) -> Result<GitMaterializationPlan, ResolverError> {
+        let path_prefix = path.as_ref().map(GitModulePath::as_str).map(str::to_string);
```
peterhuene: Seems like we could skip the `to_string` call here in favor of the following below:

```rust
let sparse_path = path_prefix.unwrap_or(".").to_string();
```

crates/wdl-modules/src/project.rs
```diff
@@ -0,0 +1,276 @@
+//! Module project loading, upward discovery, and paired persistence.
+//!
+//! [`ModuleProject`] loads an exact `module.json` path and remembers the
+//! sibling `module-lock.json` path beside it. [`ModuleProject::discover`]
+//! walks upward from a caller-supplied start path until it finds `module.json`
+//! or reaches a `.git` boundary, so project discovery returns `Ok(None)` when
+//! no ancestor project exists. [`ManifestDocument`] preserves unknown manifest
+//! extension fields while revalidating each edit immediately. When callers
+//! need to write the manifest and lockfile together, [`LockedModuleProject`]
```
peterhuene: Could you explain why it would be necessary to get a global lock to write the lock file rather than just taking a lock on the lock file itself to coordinate updates?

tests/module/fixtures.rs
```diff
@@ -0,0 +1,31 @@
...
+mod signer;
+
+pub(crate) use assertions::*;
+pub(crate) use command::*;
+pub(crate) use git::*;
+pub(crate) use module::*;
+pub(crate) use signer::*;
+
+#[test]
+fn focused_fixture_modules_preserve_the_test_api() {
```
peterhuene: What is the point of this test?

src/analysis.rs
```diff
@@ -446,21 +423,15 @@ pub(crate) fn resolution_context_for_manifest(
...
-/// discovered manifest is malformed or when `starts` spans more than one
-/// manifest.
+/// LSP server. Each path in `starts` is walked upward (stopping at a repository
+/// root) for a `module.json`. The default null-resolver context is returned
+/// when no manifest is found; an error is returned when a discovered manifest
+/// is malformed or when `starts` spans more than one manifest.
 pub(crate) fn resolution_context_from_paths(
     modules_config: &wdl_modules::resolver::ModulesConfig,
-    feature_flags: &FeatureFlags,
+    _feature_flags: &FeatureFlags,
```
peterhuene: Is this being kept because we anticipate using it again in the future?

src/commands/printer.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /// Prints a confirmation prompt to stderr and reads one line from stdin.
+    ///
+    /// Appends ` [y/N] ` to the prompt. Any case variant of `y`/`yes`
+    /// accepts; anything else — including EOF and an empty line — declines.
+    pub fn confirm(&self, prompt: impl fmt::Display) -> anyhow::Result<bool> {
+        eprint!("{prompt} [y/N] ");
+        io::stderr().flush().context("flushing prompt")?;
+
```
peterhuene: It might not be worth the dependency, but `dialoguer::Confirm` is pretty nice in that it doesn't have to read in an entire line for the prompt.

src/commands/printer.rs
```diff
@@ -0,0 +1,141 @@
...
+use anyhow::Context as _;
+use colored::Colorize as _;
+
+/// A console printer for command status lines.
+///
+/// Owns the colorization decision so subcommands don't thread a bare
+/// `bool` through every call. Cheap to copy; construct it once from the
+/// resolved color mode and pass it down by value.
+#[derive(Clone, Copy, Debug)]
+pub struct Printer {
```
peterhuene: Honestly it feels like this type should be merged with `CommandOutput`.

src/commands/module/verify.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    if problems.is_empty() {
+        Ok(())
+    } else {
+        anyhow::bail!(
+            "verification with `--require-signatures` requires signatures for every package:\n  {}",
+            problems.join("\n  ")
+        );
+    }
```
peterhuene: Nit:


```suggestion
    if !problems.is_empty() {
        anyhow::bail!(
            "verification with `--require-signatures` requires signatures for every package:\n  {}",
            problems.join("\n  ")
        );
    }
    
    Ok(())
```

src/commands/module/project.rs
```diff
@@ -0,0 +1,81 @@
...
+        "discovered module project"
+    );
+}
+
+/// Loads `module-lock.json` when present.
+pub(super) fn load_lockfile(project: &Project) -> anyhow::Result<Option<Lockfile>> {
+    let lockfile = project.load_lockfile()?;
+    match &lockfile {
+        Some(lockfile) => {
+            tracing::trace!(
```
peterhuene: This `tracing::trace!` call seems misplaced to me.

src/commands/module/lock.rs
peterhuene, line 1: Could you explain at a high level how command this differs from `module update`?

src/commands/module/lock.rs
```diff
@@ -0,0 +1,160 @@
...
+use crate::config::Config;
+
+const LOCK: Action = Action::new("Locked", "lock");
+
+/// Arguments to `sprocket dev module lock`.
+#[derive(Parser, Debug)]
+pub struct Args {
+    /// Fail if `module-lock.json` is missing or out of date.
+    #[arg(long)]
+    pub locked: bool,
```
peterhuene, line 28: It feels like pretty much any command that reads the lock file should have this option, right?

src/commands/module/display.rs
```diff
@@ -0,0 +1,207 @@
+//! Module-specific value formatting.
+
+use wdl_modules::dependency::GitSelector;
+use wdl_modules::lockfile::ResolvedSource;
+use wdl_modules::resolver::lock::DependencyUpdate;
+
+/// Formats a dependency update for an action line.
+#[cfg(test)]
+fn update_message(
```
peterhuene: Can this move to the `tests` submodule?

src/analysis.rs
```diff
@@ -429,6 +395,17 @@ pub(crate) fn resolution_context_for_manifest(
     manifest_path: &Path,
     manifest: wdl_modules::Manifest,
 ) -> anyhow::Result<wdl::analysis::ResolutionContext> {
+    let lockfile_path = manifest_path.with_file_name(wdl_modules::LOCKFILE_FILENAME);
+    if lockfile_path.exists()
+        && let Ok(lock_bytes) = std::fs::read(&lockfile_path)
+        && let Ok(lock) = wdl_modules::Lockfile::parse(&lock_bytes)
+        && !lock.satisfies_manifest(&manifest)
+    {
+        tracing::warn!(
```
peterhuene, line 402: Question: why aren't we automatically updating the lock file like we do for `sprocket run`?

src/commands/run.rs
```diff
@@ -638,6 +645,14 @@ pub async fn run(
     let report_mode = args.report_mode.unwrap_or(config.common.report_mode);
     args.apply_engine_config(&mut config.run.engine);
 
+    // Regenerate a stale or missing module lockfile before executing so
+    // the run proceeds against a consistent, reproducible tree.
+    if let Some(dir) = source.local_start_dir() {
```
peterhuene, line 751: Shouldn't this command have a `--locked` option to prevent evaluation if the lock file is missing or out of date?

src/commands/module/auto_lock.rs
```diff
@@ -0,0 +1,126 @@
...
+    let Some(project) = wdl_modules::project::ModuleProject::discover(start)? else {
+        return Ok(());
+    };
+    if project.manifest().dependencies.is_empty() {
+        return Ok(());
+    }
+
+    // Avoid acquiring the mutation lock when the current lockfile already
+    // satisfies the manifest.
+    let existing = load_lockfile(&project)?;
```
peterhuene: This redundant loading of the lock file seems like it could be solved via taking a shared advisory lock on the lock file itself?

src/commands/module/init.rs
```diff
@@ -0,0 +1,567 @@
...
+    let repository = repository.trim();
+    (!repository.is_empty())
+        .then(|| sanitize_repository(repository))
+        .flatten()
+}
+
+/// Removes credentials and request metadata from an inferred repository URL.
+fn sanitize_repository(repository: &str) -> Option<String> {
+    let Ok(mut url) = url::Url::parse(repository) else {
+        return Some(repository.to_string());
```
peterhuene: We probably need to convert SCP and local file path remotes into URLs here to ensure we can validate the serialized manifest, per @a-frantz's comment relating to being unable to `module init` in the workflows repo with an SCP remote.

src/commands/module/add/source.rs
```diff
@@ -0,0 +1,646 @@
...
+                "expanded hosted Git shorthand"
+            );
+            Ok(Some(url))
+        }
+        SourceKind::ScpLike => {
+            let (user_host, path) =
+                scp_like_parts(source).expect("source kind was inferred from scp-like syntax");
+            anyhow::bail!(
+                "`{source}` looks like an scp-style Git URL, which is not supported; use \
+                 `ssh://{user_host}/{path}` instead"
```
peterhuene: Is there a reason _not_ to support the syntax for the command and simply auto-convert it to a URL?

peterhuene (commented):
My initial review comments.

Overall, it looks good and there's a lot of needed functionality here.

I think it seems overly complex in places and some of the review comments I have can probably be addressed in follow-up PRs to improve the developer experience.

crates/wdl-modules/src/project/document.rs
claymcleod: No—the parser does not require a trailing newline. The newline was appended while serializing the edited document, before the old full-document round-trip validation. That reserialize-and-reparse step is gone; edits now validate their inputs before changing either representation, while `to_bytes` still emits newline-terminated JSON as an output-formatting choice.

src/analysis.rs
claymcleod: This path also constructs resolution contexts for the analyzer/LSP. Keeping it read-only matters because analysis runs implicitly when a workspace opens and repeatedly as files change; writing here could dirty the checkout merely by opening it, race with the editor or another tool updating `module.json`, and retrigger filesystem watchers and analysis. `run` and `submit` are explicit execution commands, so they auto-lock before execution. Analysis instead warns when the lockfile is stale, while commands that consume the lockfile expose `--locked` when the caller wants stale or missing state to be an error.

crates/wdl-modules/src/resolver/git.rs
```diff
@@ -396,19 +396,7 @@ impl GitResolver {
...
-                {
-                    tracing::warn!(
-                        path = %plan.leaf.display(),
-                        error = %io_err,
-                        "failed to clean up cache leaf after materialization failure",
-                    );
-                }
-                return Err(err);
-            }
+            Err(err) => return Err(err),
```
peterhuene: Nit: could we do a `let fetched = ...?;` above rather than manually doing it on `result` now that the error branch is just a return?

peterhuene (approved):
Looks good. Just the one nit comment, so otherwise approving.

src/analysis.rs
peterhuene: To be fair, `rust-analyzer` will write a new `Cargo.lock` when it starts or when `Cargo.toml` changes. Generally that's the experience I would expect with an LSP because I frequently manually add dependencies to `Cargo.toml` and expect the LSP to just automatically resolve them into the lock file.

I wouldn't want the lock file updated for _every_ change in the workspace, just when the workspace initially opens and the `module.json` changes.

PR #1017, opened 2026-07-16
refactor(sprocket)!: unify lint rules and tags

Serial-ATA:
This makes lint tags simpler to work with:

| Before | After |
|--------|-------|
| `check --all-lint-rules` | `check --tag all` |
| `check --filter-lint-tag <TAG>` | `check --except <TAG>` |
| `sprocket check --only-lint-tag <TAG>` | `check --except all --tag <TAG>` |

* `--tag` makes it possible to extend the default tag list (where `--only-lint-tag` would require you to specify every tag)
* `--except` now accepts both rule IDs and tag names
* `--except` and `--tag` can be used together with `--tag` taking precedence, so `--only-lint-tag` can be replaced with `--except all --tag <TAG>`

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lint/src/tags.rs
```diff
@@ -53,63 +85,54 @@ impl std::fmt::Display for UnknownTagError {
...
 impl TagSet {
+    /// A tag set containing all tags.
+    pub const ALL: Self = Self({
+        let mut bits = 0;
+        let mut i = 0;
+        while i < Tag::All as u8 {
+            bits |= 1 << i;
+            i += 1;
+        }
+        bits
```
peterhuene, line 107: Do we ever envision using bits in the set that won't represent a tag?

If not, then perhaps simply `u32::MAX` rather than relying on the value of `Tag::All` (which at a minimum would need a comment on it to say it should come last)?

crates/wdl-lint/src/tags.rs
```diff
@@ -53,63 +85,54 @@ impl std::fmt::Display for UnknownTagError {
...
 
-        let mut bits = 0u32;
+        let mut ret = Self::EMPTY;
         let mut i = 0;
         while i < tags.len() {
-            bits |= Self::mask(tags[i]);
+            let tag = tags[i];
+
+            // `Tag::All` is just a marker, it shouldn't show up in the set.
+            if tag as u8 == Tag::All as u8 {
```
peterhuene, line 124: Nit: The cast seems unnecessary.


```suggestion
            if tag == Tag::All {
```

tests/cli/check/except/can_ignore_tag/inputs/input-keyword.wdl
```diff
@@ -0,0 +1,22 @@
+# Test for lint tag exceptions. This file should only produce rules under the
+# `Deprecated` tag, which is excepted.
```
peterhuene: Unless I'm misunderstanding this test:

```suggestion
# Test for lint tag exceptions. This file should only produce a `RequirementsSection`
# diagnostic as the `CallInputKeyword` diagnostic is excepted via the `deprecated` tag.
```

peterhuene (commented):
Discussion of the "pedantic" tag aside, I have just a few comments.

crates/wdl-lint/src/tags.rs
Serial-ATA: I wish, but:

```rust
error[E0015]: cannot call non-const operator in constant functions
   --> crates/wdl-lint/src/tags.rs:123:16
    |
123 |             if tag == Tag::All {
    |                ^^^^^^^^^^^^^^^
```

peterhuene: Oh right, we're in a const context. Never mind.

crates/wdl-lint/src/tags.rs
peterhuene: Although, looking at how we implemented other functions to rely on the behavior that only one bits in the set correspond to actual valid tags, I'm fine with this.

I'll add a comment for more documentation on the variant, though.

crates/wdl-lint/src/tags.rs
```diff
@@ -39,8 +56,23 @@ pub enum Tag {
     /// Rules associated with keeping WDL compatible with other Sprocket
     /// commands (e.g. `doc`).
     SprocketCompatibility,
+
+    /// Rules that may be overly strict or produce false positives.
+    Pedantic,
+
+    /// Rules from all tags.
```
peterhuene, line 64:
```suggestion
    /// Rules from all tags.
    ///
    /// This variant is handled specially by [`TagSet`] and must be defined last.
```

Serial-ATA: Sure, though I already have an assertion that verifies that at compile time:

```rs
error[E0080]: evaluation panicked: `Tag::All` must be the last variant
  --> crates/wdl-lint/src/tags.rs:68:5
   |
68 | /     assert!(
69 | |         Tag::All as usize == Tag::COUNT - 1,
70 | |         "`Tag::All` must be the last variant"
71 | |     );
   | |_____^ evaluation of `tags::_` failed here
```

peterhuene: Indeed (and great to have), I just like to see a comment closer to where someone might add a new value.

tests/cli/check/except/can_ignore_tag/inputs/input-keyword.wdl
Serial-ATA: Ah, nope that was actually a bug. `RequirementsSection` and `CallInputKeyword` are both `deprecated` so it should've produced nothing

peterhuene: Ah weird, I forgot that we roll both a check for a missing `requirements` section and a check for using a deprecated `runtime` section into the same lint, which is why on the surface I thought that `RequirementsSection` was expected (i.e. it was checking solely for a missing `requirements` section).

That lint should probably be broken apart (in the future) since the lint about a missing `requirements` section isn't really about deprecation.

Serial-ATA: Yeah I was thinking the same. I'll make an issue

src/commands/check.rs
```diff
@@ -51,52 +50,32 @@ pub struct Common {
     #[clap(value_name = "SOURCE")]
     pub sources: Vec<Source>,
 
-    /// Excepts (ignores) an analysis or lint rule.
+    /// Excepts (ignores) a lint rule or tag from running if it would have been
+    /// included otherwise.
     ///
     /// Repeat the flag multiple times to except multiple rules. This is
```
```suggestion
    /// Repeat the flag multiple times to except multiple rules or tags. This is
```

src/analysis.rs
```diff
@@ -560,15 +541,17 @@ fn get_diagnostics_config(exceptions: &HashSet<String>) -> DiagnosticsConfig {
...
-            .iter()
-            .any(|exception| exception.eq_ignore_ascii_case(rule.id()))
+    if exceptions
+        .iter()
+        .any(|exception| exception.eq_ignore_ascii_case(rule.id()))
+    {
+        return false;
+    }
+
+    enabled_lint_tags.intersect(rule.tags()) == rule.tags()
```
a-frantz, line 524: is this right? Looks like a user would need to be opted into every tag of a multi-tag rule for it to run. I would expect a rule to run so long as at least one of its tags is enabled.

tests/cli/check/except/tag_precedence/inputs/source.wdl
a-frantz, line 1: I was wondering about this case. I think `--except all` should be called out somewhere in the CLI `--help` as being possible. I don't think the help text for either `--tag` or `--except` mentions `all` is an option.

crates/wdl-lint/src/tags.rs
```diff
@@ -42,8 +59,24 @@ pub enum Tag {
 
     /// Rules associated with the performance of a document.
     Performance,
+
+    /// Rules that may be overly strict or produce false positives.
+    Pedantic,
```
a-frantz:
```suggestion
```
can we remove this from this PR? I think this needs some more tire kicking before we include it.

src/analysis.rs
Serial-ATA: yeah I wasn't 100% on that behavior change (I even had to change up some tests), but I think it makes sense:

`EmptyDocComment` is `Clarity+Documentation` and both of those tags are enabled by default. if I disabled `Clarity` (`sprocket check --lint --except clarity`) it wouldn't make sense for `EmptyDocComment` to still apply, despite `Documentation` still being enabled.

tests/cli/check/except/tag_precedence/inputs/source.wdl
Serial-ATA: it shows up in the help text for `check --tag` since it's one of the `Tag` variants:

```
Checks a document or a directory containing documents                                                                                                                                                                                                                                                               

Usage: sprocket check [OPTIONS] [SOURCE]...

Arguments:
  [SOURCE]...
          A set of source documents as files, directories, or URLs

Options:
      ...

      --tag <TAG>
          Includes a lint tag for running.
          
          Repeat the flag multiple times to include multiple tags. `--except <RULE|TAG>` can be used in conjunction with this argument. This is additive with tags selected via config files.
          
          [possible values: All, Clarity, Completeness, Correctness, Deprecated, Documentation, Naming, Pedantic, Performance, Portability, Sorting, Spacing, SprocketCompatibility, Style]

      ...
```

src/analysis.rs
a-frantz: This API makes the most sense with the current set up where we have a default set of tags enabled by default, but I think that deserves revisiting.

At this moment I'm not ready to throw my weight behind any particular alternative, or to throw out the current design, but I don't think we've nailed the "which lint rules to run by default" question.

tests/cli/config/init/run/stdout
```diff
@@ -14,9 +14,7 @@ deny_warnings = false
 deny_notes = false
 hide_notes = false
 hide_warnings = false
-all_lint_rules = false
-only_lint_tags = []
-filter_lint_tags = []
+tags = []
```
a-frantz, line 17: now that "all_lint_rules" has been removed, this should populate with the default set. That would deserve a big call out in the top level changelog, as it could cause silent regressions (unlike the other config changes which will blow up)

a-frantz (changes requested):
Sorry for the long delay on this review! I initially paused to think over the question about `--tag` and `--except` interactions, and then forgot to come back

src/analysis.rs
a-frantz: To state explicitly, this thread can be resolved now 👍

tests/cli/config/init/run/stdout
Serial-ATA: I don't think that needs to change (and it might even be a bit confusing if it does). The defaults are always enabled unless _explicitly_ disabled, so if we default to:

```toml
[check]
tags = ["completeness", "naming", "clarity", "portability", "correctness", "deprecated", "documentation"]
```

Dropping `documentation` doesn't disable it, you'd instead need:

```toml
[check]
tags = ["completeness", "naming", "clarity", "portability", "correctness", "deprecated"]
except = ["documentation"]
```

https://github.com/stjude-rust-labs/sprocket/pull/1017/changes#diff-1f3674ae3616d00554dec466b2555f119070b7bc1eefb1c236f108626fc3b96cR243-R249

Do you think that the default tag set should only be used if no tags are explicitly specified then?

a-frantz: o right. 🤔 ok, per my other comment, I think we want to hit the drawing board again, but I'll approve for now. This is an improvement and the right direction 🎉 , but I'm wondering if we want to more drastically redesign this part of the linting interface.

a-frantz (approved):
once all the tests have been fixed, this LGTM

Issue #666, opened 2026-02-25
Configuration for call input formatting in `wdl-format`


Serial-ATA:
**Is your feature request related to a problem? Please describe.**
The `CallInputSpacing` lint and current `wdl-format` behavior force users into a single style of call statement.

Currently, the following:

```wdl
call lorem { foo, bar, baz = 1 }

call ipsum { foo, bar, baz = 1, qux = "hello", quux = 15 }
```

Would be formatted as:

```wdl
call lorem {
    foo,
    bar,
    baz = 1,
}

call ipsum {
    foo,
    bar,
    baz = 1,
    qux = "hello",
    quux = 15,
}
```

**Describe the solution you'd like**
We could probably just add a configurable `max_call_width` that determines whether or not to format the arguments vertically. So by default, the above example could be formatted as:

```wdl
call lorem { foo, bar, baz = 1 }

call ipsum {
    foo,
    bar,
    baz = 1,
    qux = "hello",
    quux = 15,
}
```

a-frantz:
due to the way we have `PreToken` and `PostToken` division, and only `PostToken`s have a notion of line width, this would be tricky, unless `max_call_width` refers to _number of inputs_ to put on a single line. e.g. your example could be `max_call_width = 3`. I'm not really sure this would ever be desirable though. We used to put calls with a single input all on the same line, but it led to some awkwardness which led to always formatting them across multiple lines.

mcovarr:
I was curious about this issue since call inlining was a formatting style I didn't recall working with before. 

Looking at data across 307 WDL files from [broadinstitute/warp](https://github.com/broadinstitute/warp) (`master`) and [broadinstitute/gatk](https://github.com/broadinstitute/gatk) (`ah_var_store`), there are currently 1,631 call statements. Thirteen are written on a single line, and eleven of those have no inputs at all (`call foo`). Among the 1,620 calls that do have inputs, exactly two are inline.

Only 100 of the 1,631 calls would fit in 90 columns inline, and 13 of those 100 are the one-liners; i.e. even when an inline formatting would fit, the authors in these repos use non-inline formatting the vast majority of the time.

tl;dr if inline call formatting is implemented, the authors in these repos would probably want the ability to opt out.

a-frantz:
[...]

PR #1034, opened 2026-07-20
feat(wdl-format): extraneous token dropping

### LINKED ISSUE #799: wdl-format: comments are reordered

### LINKED ISSUE #855: wdl-format: drop empty call statement bodies

In the following:

```wdl
workflow foo {
    call bar {}
}
```

The `{}` is not necessary in the call statement

Maybe we could add a config option to drop empty call bodies? So the above example would be formatted as:

```wdl
workflow foo {
    call bar
}
```

### LINKED ISSUE #1035: bad formatting of symbolic imports with comments
**Describe the bug**

these are some contrived cases I came up with to demonstrate issues with the approach taken to formatting in #831 . Sharing here for posterity and to prevent regression.

```shell
$  cat ~/tmp/symbolic_imports_long_path_comments.wdl
version 1.4

import
openwdl/very/long/nested/module/path/that/definitely/exceeds/the/max/line/length/csvkit # interrupted
/and/continued/on/new/line
as csv
import { Foo } from openwdl/very/long/nested/module/path/that/definitely/exceeds/the/max/line/length/csvkit/                    # interrupted by very weird whitespace
          and/continued/on/new/line # and then some more weirdo space
workflow test {}
$  sprocket format view ~/tmp/symbolic_imports_long_path_comments.wdl
version 1.4

import openwdl/very/long/nested/module/path/that/definitely/exceeds/the/max/line/length/csvkit # interrupted
/and/continued/on/new/line
    as csv
import {
    Foo,
} from openwdl/very/long/nested/module/path/that/definitely/exceeds/the/max/line/length/csvkit/                    # interrupted by very weird whitespace
          and/continued/on/new/line

workflow test {
}
```

**Additional context**

Fixed by #1034

a-frantz:

closes #855
closes #1035
closes #799 

the most important change in this PR is not user-facing; it involves a refactor/rewrite of the "`flush()`" method during PreToken to PostToken conversion. That method had been expanded on piecemeal over a variety of new features, and the resulting code got fairly spaghettified during all those distinct additions. This axed the meat of that method and rewrote it from scratch so it will be easier to follow and maintain.

the "fit or split" logic has been removed and will be added in a follow up PR (this one got thorny enough without that feature confusing the diff), so #721, #666, and #278 will remain open till the follow up lands

For external contributors:

For all contributors:

      is implemented.


crates/wdl-format/src/token/post.rs
```diff
@@ -527,125 +522,136 @@ impl Postprocessor {
...
-                let break_kind = potential_line_breaks.get(&i).unwrap();
-                if let Some(also_break_on) = tandem_line_break(*break_kind) {
+                if let Some(also_break_on) = tandem_line_break(cached_on) {
                     let tandem_break = TandemBreak {
-                        open: *break_kind,
+                        open: cached_on,
                         close: also_break_on,
                         depth: 0,
                     };
                     break_stack.push(tandem_break);
```
claymcleod, line 617: `Postprocessor` still needs to account for the indentation depth of each active tandem break. This push no longer increments `indent_level`, and the matching pop no longer decrements it, so nested implicit breaks all receive one continuation indent. For example, nested calls place inner arguments at the outer call indentation. Please restore the increment/decrement bookkeeping when tandem breaks are opened and closed.

crates/wdl-format/src/token/post.rs
```diff
@@ -527,125 +522,136 @@ impl Postprocessor {
...
+                    // the line is too long after the next step.
+                    cache = Some(post_buffer.clone());
+                    cached_self = Some(self.clone());
+                    cached_on = Some(*kind);
+                }
+                prev_kind = Some(*kind);
+            } else if let Some(k) = prev_kind.take()
+                && matches!(can_be_line_broken(k), Some(LineBreak::After))
+            {
                 cache = Some(post_buffer.clone());
```
claymcleod: This checkpoint path also runs when `max_line_length = "none"`, even though no rollback can occur. Every token after a breakable `After` token clones the entire `post_buffer` and `Postprocessor`, so long comma-separated expressions perform repeated full-buffer copies and can become O(n²) in line length. Please gate checkpoint creation on an active width limit, or restore the cheap first pass and speculate only after detecting overflow.

crates/wdl-format/src/element.rs
```diff
@@ -74,6 +75,23 @@ impl FormatElement {
...
+    pub fn has_comment(&self) -> bool {
+        if let Some(node) = self.element().as_node() {
+            return node
+                .inner()
+                .children_with_tokens()
+                .any(|c| c.kind() == wdl_ast::SyntaxKind::Comment);
+        };
+        let token = self.element().as_token().expect("must be node or token");
+        token.inner().inline_comment().is_some()
+            || token.inner().preceding_trivia().next().is_some()
```
claymcleod: `preceding_trivia()` also yields whitespace containing blank lines, so this treats a comma with only blank-line trivia as comment-protected. With `trailing_commas = false`, that comma remains in the output even though it has no comment. Please restrict this check to `SyntaxKind::Comment`; the normal token emission path can preserve blank-line trivia separately.

claymcleod (changes requested):

crates/wdl-format/src/token/post.rs
a-frantz: Good catch! I had intended for `prev_kind` to only be populated when the width limit was active, but I see that the later block is also missing that if guard. I'll fix and make the cases more explicit

crates/wdl-format/src/token/post.rs
```diff
@@ -527,125 +522,144 @@ impl Postprocessor {
...
-        // Reset self.
-        self.interrupted = false;
-        self.position = LinePosition::StartOfLine;
-        self.temp_indent = starting_temp_indent;
-        self.indent_level = starting_indent;
+                // line break
+                self.interrupted = true;
+                self.end_line(&mut post_buffer);
+                if let Some(also_break_on) = tandem_line_break(cached_on) {
+                    self.indent_level += 1;
```
claymcleod, line 610: Because `prev_kind` remains populated after the `After` break is inserted, the next token can enter this rollback path for the same opening delimiter. If that token also exceeds the width, this increments `indent_level` and pushes a second `TandemBreak` for the same `OpenParen`; the single closing parenthesis pops only one entry, leaving the enclosing block over-indented. You should avoid reopening an already-active tandem break, or clear `prev_kind` once the earlier break has been applied.

claymcleod (changes requested):

crates/wdl-format/tests/format/symbolic_imports_short/source.formatted.wdl
```diff
@@ -3,10 +3,20 @@ version 1.4
 import openwdl/csvkit
 import openwdl/csvkit as csv
 import * from openwdl/csvkit
-import { sort } from openwdl/csvkit
-import { sort as sorter } from openwdl/csvkit
-import { CsvSort, CsvSortStable } from openwdl/csvkit
-import { sort, cut } from openwdl/csvkit
+import {
+    sort,
+} from openwdl/csvkit
```
adthrasher, line 8: This follows our "break everything" approach to inserting line breaks, but I think it is suboptimal here. Fine to keep it for now, but we should revisit it in the future.

a-frantz: Yes, that's going to be addressed in the follow up PR I have planned. This should return to a single line style for the release following today's 🚀

crates/wdl-format/CHANGELOG.md
```diff
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+* Extraneous tokens (e.g. trailing commas and empty call input blocks) are dropped from the output ([#1034](https://github.com/stjude-rust-labs/sprocket/pull/1034))).
```
adthrasher:
```suggestion
* Extraneous tokens (e.g. trailing commas and empty call input blocks) are dropped from the output ([#1034](https://github.com/stjude-rust-labs/sprocket/pull/1034)). 
```

crates/wdl-format/CHANGELOG.md
```diff
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+* Extraneous tokens (e.g. trailing commas and empty call input blocks) are dropped from the output ([#1034](https://github.com/stjude-rust-labs/sprocket/pull/1034))). 
+
 #### Fixed
 
+* Comments within import statements are preserved ([#1034](https://github.com/stjude-rust-labs/sprocket/pull/1034))).
```
adthrasher:
```suggestion
* Comments within import statements are preserved ([#1034](https://github.com/stjude-rust-labs/sprocket/pull/1034)).
```

crates/wdl-format/CHANGELOG.md
```diff
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+* Non-directive comment ordering and blank lines are preserved ([#1034](https://github.com/stjude-rust-labs/sprocket/pull/1034))).
```
adthrasher:
```suggestion
* Non-directive comment ordering and blank lines are preserved ([#1034](https://github.com/stjude-rust-labs/sprocket/pull/1034)).
```

crates/wdl-format/src/v1/import.rs
```diff
@@ -49,193 +43,60 @@ pub fn format_import_statement(
...
+            && (items.peek().is_some() || comma.has_comment())
+        {
+            (comma).write(stream, config);
+            if items.peek().is_some() {
+                stream.end_line();
             }
+        } else if config.trailing_commas {
+            stream.push_literal(",".into(), SyntaxKind::Comma);
         }
     }
```
adthrasher, line 93: Should this get abstracted into a helper? It's reimplemented in multiple places.

crates/wdl-format/src/token/post.rs
```diff
@@ -527,125 +522,148 @@ impl Postprocessor {
...
+                && let Some(cached_self) = cached_self.take()
+                && let Some(cached_on) = cached_on.take()
+            {
+                // revert
+                post_buffer = cache;
+                // reset self
+                self.indent_level = cached_self.indent_level;
+                self.temp_indent = cached_self.temp_indent;
+                self.interrupted = cached_self.interrupted;
+                self.position = cached_self.position;
```
adthrasher: Can this just be `*self = cached_self;` or do we need to enumerate only a subset of fields here?

crates/wdl-format/CHANGELOG.md
a-frantz: One extra paren is a typo, three is a misconfigured nvim plugin 🤣

crates/wdl-format/src/token/post.rs
a-frantz: no, you're right that is much better! TBH I forgot the right syntax for doing the self assignment and neglected to look it up and come back around.

crates/wdl-format/src/v1/import.rs
a-frantz: I considered making a macro for it, a helper function would be suitable as well.

FWIW, there are a lot of repetitive patterns in the `v1` module due to the nature of formatting. It's likely possible to abstract a few more of these patterns into helpers.

Would it be suitable to log it as an issue for follow up? I think one PR that combs over the entirety of `wdl-format::v1` would make sense (and could even be a `good first issue`)

Happy to knock out this pattern by itself before merging if we want

adthrasher: I'm happy to leave it for a follow-up.

crates/wdl-format/src/token/post.rs
```diff
@@ -527,125 +522,145 @@ impl Postprocessor {
...
-                    Some(LineBreak::Before) => {
-                        potential_line_breaks.insert(i, *kind);
+        while let Some(token) = pre_buffer.next() {
+            let mut cache = None;
+            let mut cached_self = None;
+            let mut cached_on = None;
+
+            if let Some(max) = max_length {
+                if let PreToken::Literal(_, kind) = token
+                    && let Some(LineBreak::Before) = can_be_line_broken(*kind)
```
claymcleod: The pop is gated on `can_be_line_broken`, so it never fires for placeholders. `tandem_line_break` pairs `PlaceholderOpen` with `CloseBrace`, and `CloseBrace` isn't in either `LineBreak` arm, so the break we push for `~{` can't ever come back off the stack. The `indent_level += 1` sticks around and everything after the placeholder shifts over a level.

No flags needed, this is the default width:

```wdl
version 1.2

task t {
    command <<<
        some_program --xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx ~{aaaaaaaaaaaaaaaaaaaa + bbbbbbbbbbbbbbbbbbbbbbbb}
        echo done
    >>>

    output {
        String s = "x"
    }
}
```

becomes

```wdl
    command <<<
        some_program --xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx ~{
            aaaaaaaaaaaaaaaaaaaa + bbbbbbbbbbbbbbbbbbbbbbbb}
            echo done
        >>>

        output {
            String s = "x"
        }
    }
```

The `>>>`, the `output` section, and the task's closing brace all moved. Run it again and `echo done` goes from 12 spaces to 16, then 20, then 24 — it never settles, so `format check` stays dirty. `main` is correct for this one.

I tried adding `CloseBrace` to the `Before` arm. That fixes the leak and all 38 fixtures still pass, but it also starts breaking before `}` inside placeholders, and there's still a non-idempotent case at width 60. So I think the pop needs to come out from under `can_be_line_broken` rather than just adding the kind. Either way, worth a fixture that breaks at `~{`.

claymcleod (changes requested):

crates/wdl-format/src/token/post.rs
```diff
@@ -527,125 +522,147 @@ impl Postprocessor {
     ) {
         assert!(!self.interrupted);
         assert!(self.position == LinePosition::StartOfLine);
```
adthrasher:
```suggestion
        assert!(self.position == LinePosition::StartOfLine, "flush must be called from start of line");
```

crates/wdl-format/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
```
adthrasher, line 10: This appears to be internally consistent, but I think this should be `###` instead which (I think) would be consistent with the other crates.

adthrasher (approved):

crates/wdl-format/CHANGELOG.md
a-frantz: I'm not sure what the release process exactly entails now, but this was done a while back to facilitate copy-pasting from crate-changelogs into github releases.

@claymcleod is this extra header padding still useful for our release process?

claymcleod: I think it makes the release notes look better, yes. That's reason we're doing it for now, even if we skip that third heading level.

crates/wdl-format/tests/format/clays_complex_script/source.formatted.wdl
```diff
@@ -1,6 +1,7 @@
 ## # Header
-## part of preamble
 # regular comment
+
+## part of preamble
```
claymcleod, line 4: This doesn't seem correct, right?

crates/wdl-format/tests/format/comment_normalization/source.formatted.wdl
```diff
@@ -1,9 +1,11 @@
 ## I am a test of weird trivia
-##
-## ok this should be sufficient for testing
+
 # whoa
+##
 
 # this is an odd jumble
+## ok this should be sufficient for testing
```
claymcleod, line 7: Same here, this isn't right.

crates/wdl-format/tests/format/symbolic_imports_long_path_with_comments/source.wdl
```diff
@@ -0,0 +1,9 @@
+version 1.4
+
+import
+openwdl/very/long/nested/module/path/that/definitely/exceeds/the/max/line/length/csvkit # interrupted
```
claymcleod, line 4: Comments aren't really allowed here, so I'm not sure this is useful to test.

crates/wdl-format/tests/format/symbolic_imports_short/source.formatted.wdl
claymcleod: Agree that this is suboptimal.

claymcleod:
I added c9630aa to fix a problem with non-idempotent output, just FYI.

crates/wdl-format/tests/format/clays_complex_script/source.formatted.wdl
a-frantz: this closes/"fixes" #799 . For context, re-ordering comments (doc comments lead, followed by "regular" comments, tailed by directives) was originally a feature and this test was intended to validate the behavior. But in hindsight, I agree with Alex that this appears as a bug. It's too disruptive and opinionated for a formatter. Reverting the feature and keeping doc comment trivia in input order is intentional.

I could rewrite this test as the text here makes less sense now.

crates/wdl-format/tests/format/symbolic_imports_long_path_with_comments/source.wdl
a-frantz: I'm not aware of anything that prohibits comments here. In WDL, wherever whitespace trivia (as simple as a space) is allowed, so are comments and newlines. Does the import RFC include specific language preventing trivia within imports? Even if it does, there's nothing about trivia here that causes ambiguity or difficulty parsing, and there is no validation in sprocket that trivia doesn't appear within imports.

It's also a matter of ensuring correctness of the algorithm, even if this is a contrived test that is unlikely to appear in "real" code. `format` should never drop user written text/comments, and #1035 demonstrates that the latest release of `sprocket` drops these comments from the output.

claymcleod (approved):

PR #1042, opened 2026-07-25
feat(py): `wdl-ast` bindings

BD103:
## Overview

This PR adds Python bindings for a subset of `wdl::ast`. My current approach is to create procedural attribute macros to generate the AST bindings, as manually implementing them would take a lot of up-front effort and be difficult to maintain.

## Progress

- [x] `#[ast]` macro
  - [x] Creates Python version of annotated struct
  - [x] Supports `SyntaxNode`
  - [x] Supports `SyntaxTree`
  - [x] Validates fields and generics
  - [x] `From` impl for converting between original and Python struct
  - [x] `IntoPyObject` for original struct
  - [x] Supports enums
  - [x] Supports custom `module`, `str`, and `eq` arguments
  - [x] Documented
  - [x] Tested
- [x] `#[ast_methods]` macro
  - [x] Creates `#[pymethods]` version of annotated impl
  - [x] Method body, handling `&self`, `&mut self`, `self`, and associated function
    - `&self` and `&mut self`: `Ast::from(self.clone()).foo()`
    - `self`: `Ast::from(self).foo()`
    - Associated function: `Ast::foo()`
  - [x] Strip `<N: SyntaxNode>` generics method return type
    -  (Optional) And see if necessary for method arguments as well
  - [x] Special case convert `impl Iterator` into `PyList` for return types
  - [x] Documented
  - [x] Tested
  - [x] (Optional) Supports `#[skip]` avoid processing methods
- [x] Apply macros
  - [x] Annotate AST nodes and tokens with `#[ast]`
  - [x] Annotate `impl` blocks with `#[ast_methods]`
- [x] Write bindings for other types that `#[ast]` and friends don't cover
- [x] Investigate items that I didn't add `#[ast]` or `#[ast_methods]` to
- [x] Type stubs ~~(most likely we'll switch to automatically generating these)~~
- [ ] Override `ASTNode` / `ASTToken` methods for each AST element
- [ ] (Optional) Python example
- [ ] (Optional) Python unit tests
- [ ] (Optional) Make `ASTNode` / `ASTToken` store the `ThreadSafeSyntaxNode` / `ThreadSafeSyntaxToken`

---

For external contributors:

For all contributors:


BD103:
@claymcleod this is ready for an initial review! The main machinery is in `crates/sprocket-py-macros` and `crates/wdl-ast/src/python.rs`. Hopefully you have an easier time reviewing it than I had writing it 😅

Please hold off merging this yet. Off the top of my head, here are the remaining things I want to finish:

- There are about 10 items I didn't add `#[ast]` or `#[ast_methods]` to due to small issues. None of them are super critical in my eyes, but I'd like to investigate a bit more before fully skipping them
- Type stubs (most likely we'll switch to automatically generating these)
- A basic example using the API
- Add some form of Python unit tests
- Update `#[ast]`'s docs to mention how it handles enums
- Support `str` ~~and `rename_all`~~ as arguments to `#[ast]`

crates/sprocket-py-macros/src/ast_methods.rs
```diff
@@ -0,0 +1,1686 @@
...
+        } else {
+            replace_self_with_original_type_path(type_, original_type_path)?;
+
+            if let Some(ast_generic_ident) = ast_generic_ident {
+                strip_path_generic(type_, ast_generic_ident)?;
+            }
+        }
+    }
+
+    // Add "py_" prefix.
```
claymcleod: Could we restore the original Python name here with `#[pyo3(name = "...")]`? `#[pymethods]` otherwise exposes the renamed Rust identifier, so methods become `py_items`, `py_version`, etc., rather than the AST API’s `items`, `version`, and so on.

crates/sprocket-py-macros/src/ast/struct_.rs
```diff
@@ -0,0 +1,394 @@
...
+                    AstKind::Token => "PyAstToken",
+                },
+                Span::call_site(),
+            );
+
+            parse_quote!(#[::pyo3::pyclass(module = #module, name = #class_name, extends = crate::#extends, frozen, from_py_object, eq)])
+        },
+        // `#[pymethods]` relies on the Python struct being cloneable. We additionally derive
+        // `PartialEq` for parity with the original struct.
+        parse_quote!(#[derive(Clone, PartialEq)]),
```
claymcleod: The derived equality is not necessarily parity with the wrapped type. For example, `TokenText` explicitly compares and hashes token text rather than token position, while `PyTokenText` derives equality over its thread-safe syntax-token wrapper. Could the generated `PartialEq` delegate to the original AST type instead, or could `#[ast]` avoid `eq` for wrappers with custom equality?

crates/sprocket-py/src/lib.rs
```diff
@@ -60,6 +60,401 @@ mod _sprocket_bio {
...
+            #[pymodule_export]
+            use wdl_ast::v1::PyPlaceholder;
+            #[pymodule_export]
+            use wdl_ast::v1::PyPlaceholderOpen;
+            #[pymodule_export]
+            use wdl_ast::v1::PyPlaceholderOption;
+            #[pymodule_export]
+            use wdl_ast::v1::PyPlus;
+            #[pymodule_export]
+            use wdl_ast::v1::PyPrimitiveType;
```
claymcleod, line 386: `PrimitiveType.kind()` is exposed and returns `PrimitiveTypeKind`, but this module does not export the generated `PyPrimitiveTypeKind`. Python callers therefore receive a value they cannot name as `sprocket_bio.ast.v1.PrimitiveTypeKind` for comparison or type checks. Please export `PyPrimitiveTypeKind` alongside `PyPrimitiveType`.

crates/wdl-ast/src/v1/task.rs
```diff
@@ -1017,6 +1056,7 @@ pub enum StrippedCommandPart<N: TreeNode = SyntaxNode> {
 
 /// Represents a command section in a task definition.
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "unstable-python", sprocket_py_macros::ast)]
```
claymcleod: Could we add `#[cfg_attr(feature = "unstable-python", sprocket_py_macros::ast_methods)]` to this implementation? `TaskDefinition.command()` is exported and returns `CommandSection`, but without generated methods Python cannot read its heredoc state or command parts. `strip_whitespace()` should remain skipped because `StrippedCommandPart` has no Python wrapper.

crates/wdl-ast/src/v1/expr.rs
```diff
@@ -1970,6 +1981,7 @@ impl<T: TreeToken> LiteralStringText<T> {
 
 /// Represents a literal string.
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "unstable-python", sprocket_py_macros::ast)]
```
claymcleod: `PyLiteralString` is exported, but this implementation is not annotated with `ast_methods`, so Python cannot inspect a literal string at all. Please expose an appropriate Python-facing accessor—at minimum its text—while skipping methods whose `LiteralStringKind` or `StringPart` results are not bound.

crates/sprocket-py/src/lib.rs
```diff
@@ -60,6 +60,401 @@ mod _sprocket_bio {
...
+    mod ast {
+        use pyo3::prelude::*;
+        #[pymodule_export]
+        use wdl_ast::PyAst;
+        #[pymodule_export]
+        use wdl_ast::PyAstNode;
+        #[pymodule_export]
+        use wdl_ast::PyAstToken;
+        #[pymodule_export]
+        use wdl_ast::PyComment;
```
claymcleod, line 77: This module exports `PyComment`, but not the `PyCommentKind` and `PyDirectiveKind` classes generated for `CommentKind` and `DirectiveKind`; it also omits the `Directive` value returned by `Comment.directive()`. Please register the generated result types with the `ast` module so callers can import and inspect values returned by the exposed `Comment` API.

crates/sprocket-py-macros/src/ast/struct_.rs
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        // Let the Python struct be converted directly into a Python object. This lets the Python
+        // struct be stored in enums.
+        impl<'py> ::pyo3::conversion::IntoPyObject<'py> for #py_ident {
+            type Target = ::pyo3::types::PyAny;
+            type Output = ::pyo3::Bound<'py, Self::Target>;
+            type Error = ::pyo3::PyErr;
+
+            fn into_pyobject(self, py: ::pyo3::marker::Python<'py>) -> Result<Self::Output, Self::Error> {
```
claymcleod: Could we qualify this as `::std::result::Result` in both struct and enum expansions? Macro output resolves this bare `Result` at the invocation site, where `wdl-ast` already has one-parameter `Result<T>` aliases. Annotating a type in such a module produces a generic-arity error in generated code rather than a usable binding.

crates/sprocket-py-macros/src/ast/enum_.rs
```diff
@@ -0,0 +1,571 @@
...
+        // Make generated enum a pyclass.
+        {
+            let module = &args.module;
+            let class_name = LitStr::new(&original.ident.to_string(), original.ident.span());
+
+            parse_quote!(#[::pyo3::pyclass(module = #module, name = #class_name, frozen, from_py_object, eq)])
+        },
+        // `Clone` is required by `from_py_object`, and `PartialEq` is for parity with the original
+        // enum.
+        parse_quote!(#[derive(Clone, PartialEq)]),
```
claymcleod: This has the same parity problem for enums. `Type` implements semantic equality by delegating to its variants, but `PyType` derives equality over the generated wrapper payloads. Please forward the generated Python enum equality to the original enum rather than deriving it, or omit `eq` when parity cannot be preserved.

crates/wdl-ast/src/v1/task.rs
```diff
@@ -1327,6 +1369,7 @@ impl<N: TreeNode> CommandPart<N> {
 
 /// Represents a requirements section in a task definition.
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "unstable-python", sprocket_py_macros::ast)]
```
claymcleod: The same omission affects `RequirementsSection`, `RequirementsItem`, `RuntimeSection`, and `RuntimeItem`: each gets a Python wrapper but none gets `ast_methods`. As a result, `TaskDefinition.requirements()` and `.runtime()` return opaque values. Please add generated methods to these implementations, skipping the unbound `Container` conversion methods.

crates/sprocket-py-macros/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+///   implementation of the underlying Rust datatype. This is forwarded to
+///   `#[pyclass]`, and by default is omitted. Note that format strings (like
+///   `str = "{format_str:?}"`) are not supported because this macro internally
+///   uses `#[pyclass(name = ...)]`.
+///
+/// # Requirements
+///
+/// - The type generics must be either `<N: TreeNode = SyntaxNode>` or `<T:
+///   TreeToken = SyntaxToken>`.
+/// - The type must be a tuple struct with a single field for the node or token
```
claymcleod: These requirements describe structs only, but `#[ast]` also accepts enums. Please document the enum form and its constraints separately so callers do not infer that the macro rejects every enum.

crates/sprocket-py-macros/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+#### Added
+
+* Created `sprocket-py-macros` ([#TODO](TODO)).
```
claymcleod:
```suggestion
* Created `sprocket-py-macros` ([#1042](https://github.com/stjude-rust-labs/sprocket/pull/1042)).
```

claymcleod (changes requested):
Additionally, the macOS and Ubuntu lint jobs are failing.

crates/sprocket-py-macros/src/ast/enum_.rs
```diff
@@ -0,0 +1,571 @@
+//! TODO
```
claymcleod: Also just spotted this.

crates/sprocket-py-macros/src/ast/mod.rs
```diff
@@ -0,0 +1,155 @@
+//! TODO
```
claymcleod: Same here.

crates/sprocket-py-macros/src/ast/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+        args_parser.parse2(args_stream)?;
+
+        Ok(args)
+    }
+}
+
+impl Default for Args {
+    fn default() -> Self {
+        Self {
+            module: LitStr::new("sprocket_bio.ast.v1", Span::call_site()),
```
claymcleod, line 70: const this IMO

crates/sprocket-py-macros/src/ast_methods.rs
```diff
@@ -0,0 +1,1686 @@
+//! TODO
```
claymcleod: and here

crates/sprocket-py-macros/src/ast_methods.rs
```diff
@@ -0,0 +1,1686 @@
...
+    /// Succeeds if `stream` parses as exactly `feature = "unstable-python",
+    /// skip`, fails otherwise.
+    ///
+    /// This is used to determine if the inner token stream of
+    /// `#[cfg_attr(...)]` contains `skip`.
+    fn parse_cfg_attr_tokens(stream: ParseStream<'_>) -> Result<()> {
+        let feature = stream.parse::<Ident>()?;
+
+        if feature != "feature" {
+            return Err(stream.error(""));
```
claymcleod: These probably need to be filled in.

claymcleod:
Two things about the thread-safe representation, separate from the equality threads.

`#[ast]` only accepts `module` and `str`, so there is no way to opt a generated class into `hash` or `ord`. By the convention you laid out in #941, `eq` for `PartialEq` and `hash` for `Hash`, three annotated types should be hashable and currently are not: `TokenText`, which has a hand-written `Hash` and exists specifically to hash by text, and which `Ident.hashable()` hands to Python; `PrimitiveTypeKind`, which is also `Ord` and already carries a `// TODO: add ord` next to the attribute; and `ConditionalStatementClauseKind`. `ExceptRule` gets `hash` only because it is annotated by hand rather than through the macro. Adding `hash` and `ord` arguments would close the gap.

`ThreadSafeSyntaxToken` can drop its `parent` field. Resolution currently descends twice, once to rebuild the parent through `SyntaxNodePtr` and again through `token_at_offset`. Since `token_at_offset` already descends from whatever node it is called on, and a token's start offset uniquely identifies it in an immutable tree, `{ root, offset }` resolved from the root yields the same token in one descent. Storing the token's `SyntaxKind` as well would let the mismatch arm return a real error instead of `unreachable!()`.

crates/sprocket-py-macros/src/ast/mod.rs
BD103: `LitStr` has private fields, and `LitStr::new()` is not `const`, so I'm not sure this is possible.

crates/sprocket-py-macros/src/ast_methods.rs
BD103: Fair enough! I skipped it initially since the error messages would never be exposed to the user.

crates/sprocket-py-macros/src/ast_methods.rs
BD103: Oh, I totally forgot about that! Thank you for catching this

BD103:
While I added `#[ast(eq)]`, I'm going to skip adding support for `ord` and `hash` until a separate PR.

BD103:
While creating the type stubs, I used [PyO3 to generate the initial stub file](https://pyo3.rs/v0.29.2/type-stub.html) and have been editing it to match out current conventions. I used [this patch](https://gist.github.com/BD103/519d2774a72d528b01629b35500dfec2) in order for the generator to properly fill out type information for `#[ast]` types. I'm not committing this, however, because I don't think PyO3's stub generator is complete enough to enable in the codebase.

claymcleod (approved):

PR #1049, opened 2026-07-29
Refine `sprocket dev doc` output and gate accessibility in CI

claymcleod:
Rounded out the `sprocket dev doc` work on this branch and added an accessibility test.

The generated documentation now bundles the WDL grammar so highlighting works offline, themes code highlighting to the site's brand tokens in both light and dark, and highlights task command sections as bash with WDL `~{...}` placeholders preserved. The pages gained richer source cards, tightened navigation and "on this page" spacing, mono type-listing links, and an optional Slack project link; generation also no longer fails with `Not a directory` when a module holds documents beside its collapsed entrypoint.

A new `a11y` CI job builds the docs, serves them, and runs axe-core (WCAG 2.1 A/AA) in both themes, failing on confirmed violations. Reaching a clean bar drove keyboard-operable sidebar folders, accessible names on the width controls, `<li>`-wrapped navigation, an `inert` hidden sidebar, and darker light-mode brand colors so text meets AA contrast.

Two things for review: the light-mode accent palette is now noticeably darker, and axe "incomplete" results (such as the ⌘K keyboard-shortcut badges) are reported as warnings rather than failing the build.

For all contributors:


crates/wdl-doc/src/docs_tree.rs
```diff
@@ -962,9 +996,16 @@ impl DocsTree {
             .collect::<Vec<String>>()
             .join(", ");
 
+        let is_module = self.workspace_metadata.is_some();
+        let show_workflows_field = if is_module {
```
claymcleod, line 1026: Note that this is a design decision to consider: when rendering a module at the root, the workflows tab is not shown.

crates/wdl-doc/a11y/run.mjs
```diff
@@ -0,0 +1,146 @@
...
+        violations.push(`[${theme}] ${rel}\n    ${v.id} (${v.impact})\n    ${v.help} <${v.helpUrl}>\n    ${n.target.join(", ")}`);
+      }
+    }
+    await page.close();
+  }
+}
+await browser.close();
+
+if (incompleteCount > 0) {
+  console.log(`\n${incompleteCount} axe "incomplete" result(s) reported as warnings (manual review; not failing).`);
```
a-frantz, line 140: If this is being run in CI, will we notice (and care) if these warnings accumulate? We can say "we'll manual review as these come up" but if there's no reminder mechanism, will we actually review them?

crates/wdl-doc/src/docs_tree.rs
a-frantz: That's fine. An earlier design for `doc` placed more emphasis on the concept of workflows being the "true first class citizens" of WDL, but that has been de-emphasized. When we go through another round of design, I could see us dropping the workflow tab entirely.

crates/wdl-doc/src/docs_tree.rs
```diff
@@ -1524,3 +1770,431 @@ fn sort_workflow_categories(categories: HashSet<String>) -> Vec<String> {
...
+            }
+        }
+    }
+}
+
+/// Converts a kebab-case or snake_case module manifest name into a
+/// human-friendly title by replacing separators with spaces and
+/// capitalizing each word (e.g. `spellcraft-showcase` becomes `Spellcraft
+/// Showcase`).
+fn humanize_module_name(name: &str) -> String {
```
a-frantz, line 1798: Don't modules already have an optional `name` field in the JSON for this purpose? Can we use that field if present, and if not wrap the module identifier in `<code>` to match the logic used elsewhere?

a-frantz (commented):
comments from reviewing the code. I have not yet actually built and explored these changes in a browser, which I plan to do this afternoon. Might have more thoughts after that

src/commands/doc.rs
```diff
@@ -50,6 +51,9 @@ pub struct Args {
     /// An optional link to the project's GitHub repository.
     #[arg(long, value_name = "LINK TO GITHUB")]
     pub github_url: Option<Url>,
+    /// An optional link to the project's Slack workspace.
+    #[arg(long, value_name = "LINK TO SLACK")]
+    pub slack_url: Option<Url>,
```
a-frantz, line 56: There are a lot of args to the `doc` command that I'm sure really fit as CL args. Many of the existing args make a lot more sense to specify only via `sprocket.toml` as they are both specific to a workspace and unlikely to change between invocations. I think `slack_url` is one such arg that we can probably trim off, but this addition is consistent with the current CL API so I'm ok leaving it in this PR.

I might put up a PR after this release cycle to sharpen this API surface (and remove this arg you're adding now, but that can be debated/decided later)

a-frantz (approved):

PR #1054, opened 2026-08-03
fix(tests): fix Windows path normalization

Serial-ATA:
`normalize_string` was breaking URLs and multiline JSON strings when blessing tests

For external contributors:

For all contributors:

      is implemented.


a-frantz (commented):
I tested this locally and seems there's still a bad replacement:

Image

tests/cli.rs
```diff
@@ -259,21 +259,26 @@ fn resolve_env_config(test_path: &Path) -> Result<Option<NamedTempFile>> {
...
-    static DRIVE_PREFIX: std::sync::LazyLock<regex::Regex> =
-        std::sync::LazyLock::new(|| regex::Regex::new(r"[A-Za-z]:(/\S)").unwrap());
-    let s = DRIVE_PREFIX.replace_all(&s, "$1");
+        .replace("sprocket.exe", "sprocket");
+
+    // HACK: Assuming all Windows paths are absolute and have at least 2 segments.
+    // Lots of tests have multiline JSON strings with literal "\n" that
+    // would otherwise be normalized improperly.
+    static WINDOWS_PATH: LazyLock<Regex> =
+        LazyLock::new(|| Regex::new(r#"(?<drive>[A-Za-z]:)\\[^\r\n"'\\]*\\[^\r\n"']*"#).unwrap());
```
claymcleod: This pattern is not anchored to a path boundary, so it can match the `s:` in a valid escaped URL such as `https:\/\/example.com`. The replacement then produces `http////example.com`, which reintroduces the corruption this change is meant to remove. Please require a path boundary before the drive prefix and cover this case with a regression test.

tests/cli.rs
```diff
@@ -259,21 +259,26 @@ fn resolve_env_config(test_path: &Path) -> Result<Option<NamedTempFile>> {
...
+    let s = WINDOWS_PATH.replace_all(&s, |caps: &regex::Captures<'_>| {
+        // Replace backslashes and strip Windows drive prefixes (e.g., `C:`) from
+        // absolute paths.
+        let s = caps.get(0).unwrap().as_str();
+        let drive = caps.name("drive").unwrap().as_str();
+
+        let stripped_drive = &s[drive.len()..];
+        stripped_drive.replace('\\', "/")
```
claymcleod, line 280: The final character class permits arbitrary backslashes, so a path followed by a literal `\n` (e.g., `C:\foo\bar\nnext`) is consumed as part of the path and normalized to `/foo/bar/nnext`. We should stop matching at path components without consuming JSON escape sequences.

claymcleod, line 280: JSON-encoded Windows paths contain doubled textual backslashes, but `replace('\\', '/')` converts each one independently: `C:\\foo\\bar` becomes `//foo//bar`, which cannot match the Unix form. We should treat escaped separator pairs as one path separator while leaving unrelated escapes intact.

tests/cli.rs
claymcleod: The matcher only accepts backslash-separated paths with at least two components, so valid forms such as `C:\foo`, `C:/foo/bar`, and UNC paths remain platform-specific. We should support the path forms this normalizer is expected to compare, or document and test the narrower contract.

tests/cli.rs
```diff
@@ -259,21 +259,26 @@ fn resolve_env_config(test_path: &Path) -> Result<Option<NamedTempFile>> {
...
-    // Strip Windows drive prefixes (e.g., `C:`) from absolute paths.
-    static DRIVE_PREFIX: std::sync::LazyLock<regex::Regex> =
-        std::sync::LazyLock::new(|| regex::Regex::new(r"[A-Za-z]:(/\S)").unwrap());
-    let s = DRIVE_PREFIX.replace_all(&s, "$1");
+        .replace("sprocket.exe", "sprocket");
+
+    // HACK: Assuming all Windows paths are absolute and have at least 2 segments.
+    // Lots of tests have multiline JSON strings with literal "\n" that
+    // would otherwise be normalized improperly.
+    static WINDOWS_PATH: LazyLock<Regex> =
```
claymcleod: We should add focused unit coverage for this branch. The current CLI suite passes but does not exercise escaped URLs, raw or JSON-escaped Windows paths, paths followed by `\n`, single-component paths, forward-slash paths, or UNC paths, so regressions in this matcher can still pass CI.

tests/cli.rs
```diff
@@ -259,21 +259,26 @@ fn resolve_env_config(test_path: &Path) -> Result<Option<NamedTempFile>> {
...
-        .replace("\\", "/")
-        .replace("//", "/");
-
-    // Strip Windows drive prefixes (e.g., `C:`) from absolute paths.
-    static DRIVE_PREFIX: std::sync::LazyLock<regex::Regex> =
-        std::sync::LazyLock::new(|| regex::Regex::new(r"[A-Za-z]:(/\S)").unwrap());
-    let s = DRIVE_PREFIX.replace_all(&s, "$1");
+        .replace("sprocket.exe", "sprocket");
+
+    // HACK: Assuming all Windows paths are absolute and have at least 2 segments.
```
claymcleod, line 268:
```suggestion
    // NOTE: Assuming all Windows paths are absolute and have at least 2 segments.
```

tests/cli.rs
```diff
@@ -259,21 +259,26 @@ fn resolve_env_config(test_path: &Path) -> Result<Option<NamedTempFile>> {
...
+    // Lots of tests have multiline JSON strings with literal "\n" that
```
claymcleod:
```suggestion
    // Lots of tests have multiline JSON strings with literal `\\n` that
```

tests/cli.rs
```diff
@@ -259,21 +259,26 @@ fn resolve_env_config(test_path: &Path) -> Result<Option<NamedTempFile>> {
...
+    // would otherwise be normalized improperly.
+    static WINDOWS_PATH: LazyLock<Regex> =
+        LazyLock::new(|| Regex::new(r#"(?<drive>[A-Za-z]:)\\[^\r\n"'\\]*\\[^\r\n"']*"#).unwrap());
+    let s = WINDOWS_PATH.replace_all(&s, |caps: &regex::Captures<'_>| {
```
claymcleod:
```suggestion
    static WINDOWS_PATH_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
        Regex::new(r#"(?<drive>[A-Za-z]:)\\[^\r\n"'\\]*\\[^\r\n"']*"#).unwrap()
    });
    let s = WINDOWS_PATH_PATTERN.replace_all(&s, |caps: &regex::Captures<'_>| {
```

claymcleod (changes requested):

tests/cli.rs
Serial-ATA: Why? No matter what the pattern matching is gonna be hacky

tests/cli.rs
Serial-ATA: The comment already points out what the pattern covers. This isn't expected to match every Windows path format

a-frantz (approved):
ran a `BLESS=1` locally and it all looks good 👍

tests/cli.rs
claymcleod: This wasn't to say that it wasn't hacky, just suggesting the convention that we use throughout the code base for these kinds of things (including hacks). It could have been labelled as a nit I suppose.

claymcleod (approved):

Issue #812, opened 2026-04-11
Backport optional-to-default call input coercion to WDL 1.0 and 1.1


claymcleod:
Sprocket currently allows passing `T?` to a non-optional call input with a default (e.g., `Boolean x = false`) in WDL 1.2+, but flags it as a type mismatch in WDL 1.0 and 1.1. This was formalized in the 1.2 spec via openwdl/wdl#462 and openwdl/wdl#634, but Cromwell and miniwdl already implement this behavior for all WDL versions.

I've opened openwdl/wdl#761 to backport the clarification to the 1.0 and 1.1 specs. Once that lands, Sprocket should implement the same coercion for earlier versions.

PR #1058, opened 2026-08-04
fix: treat `None` as an identifier in WDL 1.0.

### LINKED ISSUE #1038: sprocket check/lint accept the WDL 1.1 `None` literal in `version 1.0` documents
**Describe the bug**

A document declared as `version 1.0` that uses the `None` literal passes
`sprocket check` and `sprocket lint` with no diagnostic. `None` is not part of
the WDL 1.0 specification -- it was introduced in [WDL 1.1](https://github.com/openwdl/wdl/blob/legacy/versions/1.1/SPEC.md?plain=1#L459).
The WDL 1.0 spec has no occurrences of `None`.

Because sprocket accepts it silently, code that passes `sprocket check`/`lint`
can still fail to parse on other WDL 1.0 engines. In our case a `version 1.0`
workflow using `None` inside a `scatter` block was merged after passing
sprocket, and only later did we discover it fails on miniwdl, Cromwell's
womtool, and wdlTools:

- miniwdl: `unexpected keyword None`
- womtool: `Cannot lookup value 'None', it is never declared`
- wdlTools: `Identifier None is not defined`

**Expected behavior**

`sprocket check` (or `sprocket lint`) should flag use of the `None` literal in
a document declared as `version 1.0`, the same way it would flag any other
WDL 1.1-only syntax used in a 1.0 document.

**Terminal Output**

Minimal repro (`none_literal.wdl`):

```wdl
version 1.0

workflow test_none_literal {
  input {
    String value
  }

  String? maybe_value = if (value == "match") then value else None

  output {
    String? result = maybe_value
  }
}
```

sprocket accepts it with no diagnostic:

```bash
$ sprocket check none_literal.wdl
$ echo "exit: $?"
exit: 0

$ sprocket lint none_literal.wdl
note[MetaSections]: workflow `test_none_literal` is missing both `meta` and `parameter_meta` sections
  ┌─ none_literal.wdl:3:10
  │
3 │ workflow test_none_literal {
  │          ^^^^^^^^^^^^^^^^^ this workflow is missing both `meta` and `parameter_meta` sections
  │
  = fix: add both the `meta` and `parameter_meta` sections

$ echo "exit: $?"
exit: 0
```

Other WDL 1.0 engines reject the same file:

```bash
$ miniwdl check --strict none_literal.wdl
(none_literal.wdl Ln 8 Col 63) unexpected keyword None
      String? maybe_value = if (value == "match") then value else None
                                                                  ^^^^
$ echo "exit: $?"
exit: 2

$ java -jar womtool-92.jar validate none_literal.wdl
Failed to process workflow definition 'test_none_literal' (reason 1 of 1): Cannot lookup value 'None', it is never declared. Available values are: ['maybe_value', 'value', 'result']
$ echo "exit: $?"
exit: 1

$ java -jar wdlTools-0.17.17.jar check none_literal.wdl
=====================================
Type-check errors in none_literal.wdl
=====================================
Line:Col | Description
-------------------------------------
8:62-8:66| Identifier None is not defined
```

**Screenshots**

N/A

- OS: Linux (RHEL 9, kernel `5.14.0-611.55.1.el9_7.x86_64`), x86_64
- Version: `sprocket unknown (a44828a55 2026-07-19) dirty 1 modification` (local build from `main`) (from https://github.com/stjude-rust-labs/sprocket/pull/1027#event-28301908481)
- Shell: zsh (`/usr/bin/zsh`)
- Toolchain version: `rustc 1.97.1 (8bab26f4f 2026-07-14)`, `cargo 1.97.1 (c980f4866 2026-06-30)`

peterhuene:
The `None` keyword was introduced in WDL 1.1.

As a result, WDL 1.0 source should not treat it as a literal none expression; instead it should be treated as a name reference expression.

Updated existing tests that incorrectly used `None` in WDL 1.0 source.

Fixes #1038.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/tests/analysis/inputs-with-none-v1_0/source.wdl
```diff
@@ -33,9 +33,6 @@ workflow test {
     }
 
     call t as t2 { input:
```
claymcleod, line 39: Removing these assignments makes `t2` equivalent to `t3` (and `w2` equivalent to `w3`), so this fixture no longer tests the optional-to-default coercion from #812. We should preserve the explicit-absence case with a valid WDL 1.0 optional expression.

crates/wdl-analysis/tests/analysis/inputs-with-none-v1_0-errors/source.wdl
```diff
@@ -27,19 +27,11 @@ task t {
 
 workflow test {
     call t as t1 { input:
```
claymcleod, line 33: These empty input blocks now duplicate the omitted-input calls below and replace the original optional-to-required type errors with missing-input warnings. Let's keep the #812 regression by passing a valid WDL 1.0 `Int?` expression to all four inputs and expecting the type mismatch only for required `input4`.

crates/wdl-engine/tests/workflows/inputs-with-none-v1_0/source.wdl
```diff
@@ -35,9 +35,6 @@ workflow test {
 
     call t as t2 {
```
claymcleod, line 40: Could we feed `t2`/`w2` from a valid WDL 1.0 optional value and keep `t3`/`w3` as omission controls? The `t2_output2` and `w2_output2` expectations should then remain `null`.

crates/wdl-grammar/tests/parsing/none-1.0/source.wdl
```diff
@@ -0,0 +1,14 @@
+version 1.0
```
claymcleod, line 1: The dotted fixture directory names are truncated by the harness's `Path::file_stem()`, so both `none-1.0` and `none-1.1` register as `none-1` and can't be filtered or identified independently. It might be worth renaming the parsing and analysis directories to `none-1-0` and `none-1-1`, matching `reserved-identifier-1-0`/`1-1`.

crates/wdl-grammar/tests/parsing/none-1.0/source.wdl
```diff
@@ -0,0 +1,14 @@
...
+
+workflow test_none_literal {
+  input {
+    String value
+  }
+
+  # In version 1.0, `None` should parse as an name reference expression
```
claymcleod:
```suggestion
  # In version 1.0, `None` should parse as a name reference expression.
```

crates/wdl-analysis/tests/analysis/none-1.0/source.wdl
```diff
@@ -0,0 +1,15 @@
+version 1.0
+
+workflow test_none_literal {
+  input {
+    String value
+  }
+
+  # In version 1.0, `None` should parse as an name reference expression
```
claymcleod:
```suggestion
  # In version 1.0, `None` should parse as a name reference expression.
```

claymcleod (changes requested):

crates/wdl-analysis/tests/analysis/inputs-with-none-v1_0-errors/source.wdl
```diff
@@ -1,4 +1,4 @@
-## This is a test that passing `None` to a required input without a
+## This is a test that failing to pass a required input without a
```
claymcleod:
```suggestion
## This is a test that fails to pass a required input without a
```

crates/wdl-analysis/tests/analysis/inputs-with-none-v1_0/source.wdl
peterhuene: I'm not sure what you mean by that, could you demonstrate?

peterhuene: As far as I'm aware, only a name reference expression to an optional variable would work.

claymcleod: Correct—I'm talking about creating an optional a different way, assigning that to a variable, and passing the name reference to the call rather than passing `None` directly.

E.g.,

```wdl
workflow test {
    if (false) {
        Int absent = 0
    }

    call t as t2 { input:
        input1 = absent,
        input2 = absent,
        input3 = absent,
        input4 = 42,
    }
```

claymcleod: This is only for the v1.0 test by the way.

crates/wdl-grammar/tests/parsing/none-1.0/source.wdl
peterhuene: Looking at the code, there doesn't seem to be an obvious reason why `file_stem` was being used vs `file_name` in the tests. There were already a few directories in there than shared the same file stem.

I think we should normalize the tests to use the full directory name and not the stem.

crates/wdl-analysis/tests/analysis/inputs-with-none-v1_0-errors/source.wdl
```diff
@@ -26,20 +26,22 @@ task t {
...
+    }
+
     call t as t1 { input:
-        input1 = None,
-        input2 = None,
-        input3 = None,
-        input4 = None,
+        input1 = absent,
+        input2 = absent,
+        input3 = absent,
```
claymcleod, line 36: `input4` is dropped rather than assigned, so `t1` now produces the same "missing required call input" warning as the bare `t2` below it, and the fixture emits four warnings and zero errors despite being named `-errors`. Assigning it restores the #812 guard:

```suggestion
        input3 = absent,
        input4 = absent,
```

I think `w1` wants the same treatment.

crates/wdl-analysis/tests/analysis/inputs-with-none-v1_0-errors/source.wdl
```diff
@@ -1,4 +1,4 @@
-## This is a test that passing `None` to a required input without a
+## This is a test that fails to pass a required input without a
 ## default is still an error in WDL 1.0.
```
claymcleod: My suggestion here was wrong and left the sentence ungrammatical—sorry about that. Now that the calls pass an absent optional, this reads better as:

```suggestion
## This is a test that passing an absent optional value to a required
## input without a default is still an error in WDL 1.0.
```

claymcleod (approved):

PR #1069, opened 2026-08-06
feat: add `sprocket dev flowchart mermaid` subcommand

claymcleod:
Added a new `sprocket dev flowchart mermaid` subcommand that takes a WDL workflow file and emits a Mermaid flowchart diagram to stdout.

The renderer walks the workflow's AST and recursively expands called workflows into labelled subgraphs (with a cycle guard to handle mutual recursion). A `--depth` flag caps the expansion depth, defaulting to unlimited. Scatter blocks render as a `procs` shape followed by a subgraph for their body. Plain `if` conditionals show the condition in the diamond with `yes`/`no` edges; `if`/`else if`/`else` chains collapse into a single `"if"` diamond with per-clause labelled fan-out edges.

For all contributors:


src/commands/flowchart/mermaid.rs
```diff
@@ -0,0 +1,812 @@
...
+pub struct Args {
+    /// The path or URL to a `WDL` document containing the workflow.
+    #[clap(value_name = "SOURCE")]
+    pub source: Source,
+
+    /// The name of the workflow to render.
+    ///
+    /// If not specified and the document contains exactly one workflow, that
+    /// workflow is used. If the document contains multiple workflows, this
+    /// argument is required.
```
adthrasher: The current spec specifies a maximum of one workflow per file.

src/commands/flowchart/mermaid.rs
```diff
@@ -0,0 +1,812 @@
...
+            .ast()
+            .into_v1()
+            // SAFETY: each test source declares a supported WDL version.
+            .unwrap()
+            .workflows()
+            .next()
+            // SAFETY: each test source declares one workflow.
+            .unwrap();
+
+        render_mermaid(&workflow, &document, Some(0))
```
adthrasher: I think we probably need at least one test at a depth greater than `0`.

Also test that depth cap 0 is effective when there are subworkflows. Ideally, also `scatter` and workflow expansion.

src/commands/flowchart/mermaid.rs
```diff
@@ -0,0 +1,812 @@
...
+}
+
+/// Gets the syntax node that identifies a visible workflow statement.
+fn statement_key(statement: &WorkflowStatement) -> Option<SyntaxNode> {
+    (!matches!(statement, WorkflowStatement::Declaration(_))).then(|| statement.inner().clone())
+}
+
+/// Escapes characters that have special meaning in `Mermaid` label strings.
+fn escape(s: &str) -> String {
+    s.replace('"', "&quot;")
```
adthrasher:
```suggestion
    s.replace('&', "&amp;")
          .replace('"', "&quot;")
```

I had trouble finding documentation, but stackoverflow things & also needs an escape.

src/commands/flowchart/mermaid.rs
```diff
@@ -0,0 +1,812 @@
...
+    dependencies
+}
+
+/// Gets the syntax node that identifies a visible workflow statement.
+fn statement_key(statement: &WorkflowStatement) -> Option<SyntaxNode> {
+    (!matches!(statement, WorkflowStatement::Declaration(_))).then(|| statement.inner().clone())
+}
+
+/// Escapes characters that have special meaning in `Mermaid` label strings.
+fn escape(s: &str) -> String {
```
adthrasher, line 610: Also, what happens if the label contains a WDL placeholder like `~{bam}`. Are the braces problematic?

src/commands/flowchart/mermaid.rs
```diff
@@ -0,0 +1,812 @@
...
+    exits: Vec<Exit>,
+}
+
+/// Counter-based context for generating unique `Mermaid` node IDs.
+#[derive(Default)]
+struct MermaidCtx {
+    counter: usize,
+    /// Tracks which workflow URIs have already been expanded to prevent
+    /// infinite recursion through mutually-recursive workflows.
+    expanded: HashSet<String>,
```
adthrasher: This comment is narrower than the implementation and comment below.

src/commands/flowchart/mermaid.rs
```diff
@@ -0,0 +1,812 @@
...
+
+            WorkflowStatement::Conditional(cond) => {
+                let clauses: Vec<_> = cond.clauses().collect();
+                let has_else = clauses
+                    .iter()
+                    .any(|c| c.kind() == ConditionalStatementClauseKind::Else);
+                let is_multi_clause = clauses.len() > 1
+                    || clauses
+                        .first()
+                        .is_some_and(|c| c.kind() != ConditionalStatementClauseKind::If);
```
adthrasher: Can you clarify this check? The first clause in a conditional should be an `if` right?

src/commands.rs
```diff
@@ -139,6 +140,9 @@ pub enum Commands {
 pub enum DevCommands {
     /// Document a workspace.
     Doc(doc::Args),
+    /// Renders a `WDL` workflow as a diagram.
```
adthrasher: The backticks are inconsistent with the `test` help text. Personally, I prefer it without the backticks when rendered.

adthrasher (commented):
Also needs an updated to CHANGELOG.

PR #1101, opened 2026-08-10
feat(wdl-analysis)!: incremental analysis

Serial-ATA:
Previously, any analysis request on a document would:

1. Start from scratch
2. Unconditionally invalidate all dependents

That made working with the LSP in large workspaces heavier than necessary.

This adds a new per-document `AnalysisCache` that gets populated with all items and a graph of their dependencies on the first pass, then incrementally edited on any later analysis requests.

The cache works off two layers of hashes: `SignatureHash` (name+inputs+outputs) and `BodyHash` (any body statements)

In the local document, a difference in either hash will mark the item as dirty. For imports, only `SignatureHash` differences are significant.

For example:

```wdl
# greet.wdl
version 1.3

task say_hello {
    input {
        String name
    }

    command <<<
        echo "Hello, ~{name}!"
    >>>

    output {
        String greeting = read_string(stdout())
    }
}

# main.wdl
version 1.3

import "greet.wdl"

# `do_say_hello` depends on the "greet.wdl" import
workflow do_say_hello {
    call greet.say_hello { name = "John" }
}
```

Changing the signature of `greet.say_hello` would invalidate:

* `greet.say_hello`
* The `"greet.wdl"` import
* `main.do_say_hello`

However, changing the _body_ of `greet.say_hello` will only ever invalidate `greet.say_hello`.

Note that this is only incremental **analysis**. Each analysis request will still do a full _validation_ pass.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/src/handlers/common/docs.rs
```diff
@@ -297,12 +301,35 @@ fn render_enum_doc(n: &EnumDefinition, computed_type: Option<&str>) -> String {
     s
 }
 
+/// Get the CST root for an item that may be imported.
+fn get_root(
```
claymcleod, line 305: An imported item's `name_span()` is a span in the importing document, but `get_root` hands back the defining document's tree, so this resolves an offset against the wrong file — it lands on whatever node happens to occupy that byte, and panics outright when the offset runs past the end of the defining document. `TaskRef::name_span()` returns `ImportedTask::span`, documented as the span of the import statement that introduced the task, while `token_at_offset` asserts the offset lies inside the root it is given. `provide_struct_documentation` and `provide_enum_documentation` have the same shape, and all three are now reachable from plain completion requests, since `add_callable_completions` iterates imported tasks.

One option is to drop the offset lookup for imported items entirely and resolve the definition in its own document instead: `ImportedTask` carries both `document()` and `name()`, and `StructRef`/`EnumRef` already expose `definition()`. That would let `get_root` go away.

Worth knowing that the three new fixtures pass only because the offsets happen to coincide. In `hover_selected` the member span starts at byte 22 and the single task in `lib.wdl` spans bytes 13 to 48, so any offset in that window finds the right task; adding a second task ahead of `add`, or shortening `lib.wdl`, turns the test red.

crates/wdl-analysis/src/graph.rs
```diff
@@ -270,20 +278,34 @@ impl DocumentGraphNode {
...
     }
 
+    /// Gets the analysis cache for this node, if one exists.
+    pub(crate) fn cache(&self) -> Option<&AnalysisCache> {
+        self.cache.as_deref()
+    }
+
+    /// Takes the analysis cache out of the node.
+    pub(crate) fn take_cache(&mut self) -> Option<AnalysisCache> {
+        self.cache.take().and_then(Arc::into_inner)
```
claymcleod, line 288: `take_cache` recovers the cache only when the node holds the last `Arc`, so any surviving `Document` clone silently discards it and the next analysis runs cold, with no log line to say so. That is the normal case rather than a rare race: `analysis_completed` deliberately keeps two clones, and a `Namespace` holds a full `Document` of every imported document, which the importer then stores in its own retained cache. An imported document's cache therefore never reaches a strong count of one, so imported documents never reuse their cache at all — and in a workspace those are the shared libraries, exactly where the reuse was meant to pay off.

The tests in this PR already show it. `should_isolate_local_body_changes` drops the previous `AnalysisResult` before its second pass and logs `re-analyzing 2 items` then `1 items`. `should_track_local_dependencies` is the same shape but keeps the result alive, and logs `3 items` then `3 items`. In `should_invalidate_dependents_of_imported_structs`, `foo.wdl` re-analyzes both of its items on every pass.

Rather than moving an owned value out and back, one option is to hold the cache as `Arc<AnalysisCache>` end to end and pass `Arc::clone(&cache)` into `Document::from_graph_node`, so recovery never depends on unique ownership. Whichever way it goes, it seems worth logging when the cache cannot be recovered, because a silent fall back to cold analysis is very hard to notice.

This one also hides other problems: several defects elsewhere in this PR only become reachable once the cache genuinely engages.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -2601,17 +2868,28 @@ fn populate_types(document: &mut DocumentData) {
...
+                    document,
+                    offset,
+                    dependent,
+                });
                 match converter.convert_struct_type(&definition) {
                     Ok(ty) => {
-                        let s = &mut document.structs[index];
-                        assert!(s.ty.is_none(), "type should not already be present");
+                        let s = document.cache.struct_by_index_mut(index).unwrap();
+                        assert!(s.ty().is_none(), "type should not already be present");
```
claymcleod, line 2892: This assert panics on any document with two or more structs where one signature changes. The gate above only asks whether *some* local type is missing, but this loop then re-converts *every* local struct, and a struct that survived the edit still carries the type it was given on the previous pass. The enum arm takes the opposite approach for the same situation and silently overwrites an already-populated type, so the two disagree.

Skipping nodes that are already typed keeps the retained type, which is still correct, because an invalidated struct is removed and re-inserted with no type:

```suggestion
                        if s.ty().is_some() {
                            continue;
                        }
```

Testing the condition before `convert_struct_type` runs would avoid the wasted conversion as well, and it would be worth making the enum arm agree either way.

This is currently masked: because the cache is discarded whenever a `Document` clone survives, `populate_types` usually sees a cold cache and the assert never trips.

crates/wdl-analysis/src/document/cache.rs
```diff
@@ -0,0 +1,1381 @@
...
+                continue;
+            }
+
+            let new_body_hash = match cache_item {
+                CachedItemRef::Import(_) => {
+                    let import_stmt = current_ast
+                        .imports()
+                        .find(|(h, _)| **h == signature_hash)
+                        .map(|(_, i)| i)
+                        .expect("should exist because current_ast contains hash");
```
claymcleod, line 1234: The invariant this `expect` relies on does not hold, because the guard and the lookup disagree about kind: `contains` matches a hash against items of every kind, while `imports()` yields only imports. A cached import whose signature hash equals that of a struct or enum in the new AST therefore passes the guard and finds nothing here, aborting the analyzer rather than returning a diagnostic.

Nothing rules that out today, since struct, enum, and import hashes carry no per-kind tag — tasks and workflows are separated only incidentally, by `keyword()` landing in the digest first. I have not built an input that reaches it, so treat this as an unsound invariant rather than a demonstrated crash.

Two independent fixes: look the statement up by kind instead of leaning on `contains`, which makes the code correct regardless of hashing, and give each hasher a kind tag, which removes the collision class underneath it.

crates/wdl-analysis/src/document/cache/hash.rs
```diff
@@ -0,0 +1,480 @@
...
+        let mut signature_hasher = Sha256::default();
+
+        Digest::update(&mut signature_hasher, self.name().text());
+        if let Some(ty) = self.type_parameter() {
+            HashableElement::hash(&ty.ty(), &mut signature_hasher);
+        }
+
+        for choice in self.choices() {
+            Digest::update(&mut signature_hasher, choice.name().text());
+        }
```
claymcleod, line 178: Two separate changes to an enum leave its signature hash untouched, so the enum and everything depending on it keep serving stale analysis. Choice names are concatenated with no separator, so `enum Foo { Bar, Baz }` and `enum FooBar { Baz }` digest the identical bytes `FooBarBaz`, which makes any rename that moves a character boundary invisible. And `choice.value()` never reaches the hasher at all, so `enum E[String] { A = "a" }` and `enum E[String] { A = 111 }` are indistinguishable, even though choice values determine the inferred type and raise their own diagnostics. Enums carry no body hash, so nothing downstream catches either one.

```suggestion
        for choice in self.choices() {
            Digest::update(
                &mut signature_hasher,
                (choice.name().text().len() as u64).to_le_bytes(),
            );
            Digest::update(&mut signature_hasher, choice.name().text());
            if let Some(value) = choice.value() {
                HashableElement::hash(&value, &mut signature_hasher);
            }
        }
```

The missing separator is not local to enums — every place a string is concatenated into a digest has it, and import aliases collide the same way, with `alias A as B, alias C as D` matching `alias AB as CD`. It is worth applying length prefixes uniformly rather than patching this one site.

crates/wdl-analysis/src/document/cache/hash.rs
```diff
@@ -0,0 +1,480 @@
...
+            Type::Object(_) => {
+                Digest::update(hasher, "Object");
+            }
+            Type::Ref(type_ref) => {
+                Digest::update(hasher, type_ref.name().text());
+            }
+            Type::Primitive(p) => {
+                HashableElement::hash(p, hasher);
+            }
+        }
```
claymcleod, line 218: Changing a declared type between several genuinely different forms leaves the enclosing signature hash unchanged, so no caller gets re-analyzed. Because no variant tag is written, `Array` digests exactly like its own element type, which makes `Int`, `Array[Int]`, and `Array[Array[Int]]` one and the same, and `Map[String, Int]` matches `Pair[String, Int]` since both emit two sub-hashes with nothing to distinguish them. `ArrayType::is_non_empty` never reaches the hasher, so `Array[Int]+` matches `Array[Int]`. `Type::Object` writes the literal text `Object`, which a type reference named `Object` also produces. Since declared types feed input and output hashing, a task's signature can change in any of these ways while every call site keeps its stale cached analysis.

Optionality has the same weakness: a byte is written only in the optional case, so `Int?` and `Array[Int]?` also collide.

```suggestion
        Digest::update(hasher, [self.is_optional() as u8]);
        match self {
            Type::Map(map) => {
                Digest::update(hasher, [2]);
                let (k, v) = map.types();
                HashableElement::hash(&k, hasher);
                HashableElement::hash(&v, hasher);
            }
            Type::Array(array) => {
                Digest::update(hasher, [3, array.is_non_empty() as u8]);
                HashableElement::hash(&array.element_type(), hasher);
            }
            Type::Pair(pair) => {
                Digest::update(hasher, [4]);
                let (l, r) = pair.types();
                HashableElement::hash(&l, hasher);
                HashableElement::hash(&r, hasher);
            }
            Type::Object(_) => {
                Digest::update(hasher, [5]);
            }
            Type::Ref(type_ref) => {
                Digest::update(hasher, [6]);
                Digest::update(
                    hasher,
                    (type_ref.name().text().len() as u64).to_le_bytes(),
                );
                Digest::update(hasher, type_ref.name().text());
            }
            Type::Primitive(p) => {
                Digest::update(hasher, [7]);
                HashableElement::hash(p, hasher);
            }
        }
```

crates/wdl-analysis/src/document/cache/hash.rs
```diff
@@ -0,0 +1,480 @@
...
+        let mut signature_hasher = Sha256::default();
+        let mut body_hasher = Sha256::default();
+
+        Digest::update(&mut signature_hasher, self.keyword().text());
+        Digest::update(&mut signature_hasher, self.name().text());
+
+        for item in self.items() {
+            match item {
+                TaskItem::Input(i) => HashableElement::hash(&i, &mut signature_hasher),
+                TaskItem::Output(o) => HashableElement::hash(&o, &mut signature_hasher),
```
claymcleod, line 63: Moving a defaulted declaration between `input` and `output` produces the identical signature hash, so the task's callers are never re-analyzed even though the change alters whether they may bind that name and whether it resolves as an output. Both sections fold into the same digest with no marker saying which one they came from, and their element hashers emit the same byte shape — type, then name, then expression — so `input { String x = "a" }` and `output { String x = "a" }` are indistinguishable.

```suggestion
                TaskItem::Input(i) => {
                    Digest::update(&mut signature_hasher, [1]);
                    HashableElement::hash(&i, &mut signature_hasher);
                }
                TaskItem::Output(o) => {
                    Digest::update(&mut signature_hasher, [2]);
                    HashableElement::hash(&o, &mut signature_hasher);
                }
```

The workflow hasher has the same pair of arms and needs the same treatment.

crates/wdl-analysis/src/document/cache/hash.rs
```diff
@@ -0,0 +1,480 @@
...
+
+        match self.source() {
+            ImportSource::Uri(uri) => match uri.text() {
+                None => Digest::update(&mut signature_hasher, ""),
+                Some(text) => Digest::update(&mut signature_hasher, text.text()),
+            },
+            ImportSource::ModulePath(path) => {
+                Digest::update(&mut signature_hasher, path.text());
+            }
+        }
```
claymcleod, line 128: `import "foo.wdl"` and `import * from "foo.wdl"` produce the identical signature hash, so switching between them never invalidates the import or anything downstream, even though the two have completely different scope semantics — one reaches the imported tasks through a namespace, the other merges every name directly into document scope. The form never reaches the digest: for the wildcard case `members()` is `None`, `aliases()` is empty, and `explicit_namespace()` returns `None` because it bails out unless the form is `Namespace`, which leaves the source text as the only thing hashed, exactly as for the plain namespace form.

This one stings more than the other collisions, because imports invalidate with the signature strategy and so are the one item whose change is meant to reach every dependent. Instead it reaches none of them.

Feeding a discriminant for `self.form()` into the hasher before the source fixes it, which needs `ImportForm` brought into scope here.

Smaller thing in the same block: the `None` arm digests an empty string, so an import whose source text is missing collides with a module path whose text is empty. Hashing a distinct marker for the absent case would separate them.

crates/wdl-analysis/src/document/cache/hash.rs
```diff
@@ -0,0 +1,480 @@
...
+        Digest::update(&mut signature_hasher, self.name().text());
+
+        for item in self.items() {
+            match item {
+                TaskItem::Input(i) => HashableElement::hash(&i, &mut signature_hasher),
+                TaskItem::Output(o) => HashableElement::hash(&o, &mut signature_hasher),
+                TaskItem::Command(c) => HashableElement::hash(&c, &mut body_hasher),
+                TaskItem::Requirements(r) => HashableElement::hash(&r, &mut body_hasher),
+                TaskItem::Hints(h) => HashableElement::hash(&h, &mut body_hasher),
+                TaskItem::Runtime(r) => HashableElement::hash(&r, &mut body_hasher),
```
claymcleod, line 67: Moving an item between `runtime`, `requirements`, and `hints` leaves the body hash unchanged, so the task is never re-analyzed even though which section a key belongs in is version-dependent and raises its own diagnostics. All three element hashers emit name then expression, and all three arms feed the same digest with nothing to say which section produced them, so relocating `docker` between them is invisible.

```suggestion
                TaskItem::Requirements(r) => {
                    Digest::update(&mut body_hasher, [3]);
                    HashableElement::hash(&r, &mut body_hasher);
                }
                TaskItem::Hints(h) => {
                    Digest::update(&mut body_hasher, [4]);
                    HashableElement::hash(&h, &mut body_hasher);
                }
                TaskItem::Runtime(r) => {
                    Digest::update(&mut body_hasher, [5]);
                    HashableElement::hash(&r, &mut body_hasher);
                }
```

Narrower than the signature collisions, since a body change only invalidates the task itself, but the stale result is still wrong.

crates/wdl-analysis/src/document.rs
```diff
@@ -1240,29 +1515,25 @@ impl Document {
...
+                    Some(t)
+                } else {
+                    None
                 }
-
-                &self.data.tasks[index - 1]
-            }
-        };
+            })
+            .find_or_last(|t| t.scope().span().start() == position)?;
```
claymcleod, line 1536: This assumes the cache yields tasks in document order, which stops being true after the first invalidation, so scope lookups return the wrong task and then nothing. `find_or_last` falls back to the last candidate the iterator produced, which is only the positionally nearest task while insertion order matches document order — and `remove_item` uses `shift_remove`, so an invalidated task is re-inserted at the end of the map. After editing one task, that task is last, becomes the answer for every position past its own start, fails the `contains` check on the next line, and the function returns `None`. Every consumer of scope-at-position is affected: completion, hover, goto-definition, find-references, and inlay hints all stop resolving inside the other tasks.

Choosing the greatest start that is not past the position removes the ordering assumption:

```suggestion
            .max_by_key(|t| t.scope().span().start())?;
```

Worth noting that both the current code and this fix scan every local task on each lookup, where the code this replaced binary-searched a map kept sorted by scope start. Keeping a span-sorted index would restore that and remove the ordering assumption at the same time, so the suggestion above is the small fix rather than the best one.

Like several others here, this is masked while the cache is being silently discarded.

crates/wdl-analysis/src/document/cache.rs
```diff
@@ -0,0 +1,1381 @@
...
+    /// Change the item's CST offset.
+    fn swap_offset(&mut self, offset: usize) {
+        match self {
+            Self::Struct(s) => s.offset = offset,
+            Self::Enum(e) => e.offset = offset,
+            Self::Task(t) => t.offset = offset,
+            Self::Workflow(w) => w.offset = offset,
+            Self::Import(i) => i.offset = offset,
+        }
+    }
```
claymcleod, line 621: Retaining an item across an edit updates the item's recorded offset but leaves every span stored inside it pointing at pre-edit coordinates, so position-based lookups resolve against the old document. `intersect` repositions only the cached diagnostics and then calls this, which touches nothing but `CachedItem::offset` — while `Task::name_span`, `Task::span`, each `Scope::span`, each `Name::span`, and the struct and enum name spans were all built from absolute AST spans when the item was first analyzed, and nothing ever rewrites them.

`trivia_should_shift_diagnostics` sets up exactly the case: inserting a comment inside a task retains the task, so the diagnostics move but the task's own spans stay where they were. That makes hover, goto-definition, completion, scope lookup, and semantic tokens all point at the wrong offsets after precisely the edits this cache is designed to make cheap.

Note this also defeats the ordering fix for `find_scope_by_position`: even with the candidate chosen correctly, the comparison still runs against stale `Scope::span` values, so the two need fixing together.

Storing these spans relative to the item and adding `CachedItem::offset` when they are read would match the approach already taken for diagnostics; rewriting them alongside the offset here would also work, though it means walking every scope on each retained item.

crates/wdl-analysis/src/document/cache.rs
```diff
@@ -0,0 +1,1381 @@
...
+            for label in diagnostic.labels_mut() {
+                let start = original_item_offset + label.span().start();
+                let end = original_item_offset + label.span().end();
+
+                let mut start_diff = 0isize;
+                let mut end_diff = 0isize;
+
+                for edit in edits {
+                    let edit_start = edit.range.start;
+                    let edit_end = edit.range.start + edit.replacement_length;
```
claymcleod, line 582: `edit_end` mixes coordinate spaces: `range` is documented as the range in the old string that was replaced, so the old-text boundary is `range.end`, while this adds the *new* text's length to the old start. For a deletion the computed end collapses back to the start, so every label at or after it takes the first branch and receives the whole negative difference, including labels that fell inside the deleted range and should have clamped to the edit site. The clamping branch below is unreachable for deletions as a result.

```suggestion
                    let edit_end = edit.range.end;
```

That alone is not sufficient, though. `apply_changes` mutates the source as it walks the edit list, so each returned range is relative to the source *after* the preceding edits, whereas this loop accumulates them all against one position computed from the original text. For a batch of more than one edit the total is wrong regardless of how `edit_end` is derived. Folding the position through the edits in order — moving it past each edit, or clamping it into that edit's replaced range, before considering the next — handles both problems, since each step then works in the coordinate space its own range belongs to.

Only insertions are covered today: `trivia_should_shift_diagnostics` is the sole test reaching this code with `start: None`, and both of its edits are pure insertions, where `range.len()` is zero and the wrong formula happens to agree with the right one. Deletions, replacements, and multi-edit batches have no coverage.

crates/wdl-grammar/src/diagnostic.rs
```diff
@@ -352,6 +352,15 @@ impl Diagnostic {
         self
     }
 
+    /// Offsets the spans of all labels in the diagnostic by the given amount.
+    pub fn offset(&mut self, offset: isize) {
+        for label in &mut self.labels {
+            let start = (label.span.start() as isize + offset) as usize;
```
claymcleod, line 358: Casting a negative sum straight back to `usize` wraps instead of clamping, and `Span::new` then adds the length to that wrapped start, which overflows. `shift_diagnostic_offsets` is the caller that makes it reachable: it passes the item's offset negated, so any label positioned before the item it was attributed to produces a negative intermediate. In a debug build that aborts the analyzer on the addition; in release it wraps and the following `saturating_sub` collapses the label onto the item start, so the label silently points at the wrong code instead.

```suggestion
            let start = (label.span.start() as isize + offset).max(0) as usize;
```

That stops the abort, but it is worth asking why a label ends up before its own item at all. The insert paths take the whole shared diagnostics buffer with `std::mem::take`, so a diagnostic left behind by a construct that returned early gets stored on whichever item is cached next, and is then rebased against that item's larger offset. Clamping hides that; attributing each diagnostic to the construct that produced it would remove it.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -194,123 +198,201 @@ pub(crate) fn populate_document(
...
 
+    let ast_items = AstItems::new(ast);
+    document.cache.intersect(&ast_items, edits, |import| {
+        resolve_import(graph, import, index)
+            .ok()
+            .and_then(|(_, cache)| cache.map(|c| c.exports_hash()))
+    });
+
+    let dirty_items: Vec<_> = document.cache.dirty(&ast_items).collect();
+    if dirty_items.is_empty() {
```
claymcleod, line 209: Lint exceptions are now collected only when something is dirty, so a pass where nothing changed leaves the exception map empty and `MeaninglessLintDirective` stops firing. Validation still runs on every pass and seeds its map solely from `document.analysis_diagnostics().exceptions`, then reports the entries that were never applied — with an empty map there is nothing to report. Before this change the population ran unconditionally, ahead of item processing, so it is a regression rather than a pre-existing gap.

The user-visible effect is a diagnostic that depends on cache warmth: analyse the file cold and the directive is flagged, edit a comment and it quietly disappears. That is worth avoiding independently of the lint itself, since analysis output should not vary with whether the cache happened to be reused.

Moving the `add_exceptions` block above this check restores the old behaviour, and it needs to run before the early return rather than after it.

Suppression itself is unaffected, since that is read from the CST at the point each diagnostic is raised, so this is limited to the unapplied-directive report.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -194,123 +198,201 @@ pub(crate) fn populate_document(
...
 
     if let Some(severity) = document.config.diagnostics_config().unused_import {
-        for (name, ns) in document.namespaces.iter().filter(|(_, ns)| !ns.used) {
-            let Some(node) = import_nodes_by_namespace.get(name) else {
+        for ns in document.cache.namespaces().filter(|(_, ns)| !ns.used) {
+            let Some(node) = import_nodes_by_namespace.get(ns.1.name()) else {
                 continue;
             };
 
+            // TODO: Should go on the import item
```
claymcleod, line 310: The unused-import warning disappears on any pass where the import itself was not dirty, because `import_nodes_by_namespace` is filled in only while walking the dirty items. A clean import is absent from that map, the `else` arm skips it, and — as this TODO notes — the diagnostic is never attached to the cached import item either, so there is no cached copy to fold back in. The warning is reported the first time a document is analysed and then silently stops.

Same shape as the lint-exception population above: a diagnostic whose presence depends on whether the cache was reused. Doing what the TODO says and storing it on the import item fixes both halves at once, since the cached item would then carry it across clean passes.

claymcleod (changes requested):
The thing I would most like to see added is a test that analyses a fixture cold, applies an edit sequence, and then asserts the result matches a fresh analysis of the same final text, because every defect I have commented on would have been caught by one and without it the design stays one unhashed field away from silent staleness. This also needs a rebase before it can land, since `main` has rewritten `except_directive_valid.rs` in the meantime.

crates/wdl-grammar/src/diagnostic.rs
peterhuene: If we're really worried about this, then perhaps a checked add to keep the panic (documenting it) and the invariant that we don't overflow the operation, move the cast to `usize to the `Span::new` constructor, and assert the start isn't less than 0.

crates/wdl-analysis/src/document.rs
```diff
@@ -914,56 +1021,53 @@ impl Document {
     /// Creates a new analyzed document from a document graph node.
     pub(crate) fn from_graph_node(
         config: &Config,
-        graph: &DocumentGraph,
+        graph: &mut DocumentGraph,
```
peterhuene: I think instead we should take `graph` as `&DocumentGraph` and take the previous analysis cache for the node as an argument.

That would limit the scope of the write lock needed to take the node's analysis cache and thus the call to `analyze_node` could look like this:

```rust
let existing_cache = {
    graph.write().get_mut(index).take_cache()
};

let graph = graph.read();
Self::analyze_node(&config, &graph, existing_cache, index, &mut (validator)())
```

This will prevent the unintentional serializing of concurrent multi-file analysis, which, with these changes, is taking a write lock for the entire length of the analysis operation, which otherwise just needs to read graph state, not write to it.

crates/wdl-analysis/src/queue.rs
```diff
@@ -1315,11 +1315,12 @@ where
                     let config = self.config.clone();
                     let validator = validator.clone();
                     handles.push(RayonHandle::spawn(move || {
+                        let mut graph = graph.write();
```
peterhuene: As mentioned previously, the scope of this write lock is too broad and will force analysis to run serially when there are multiple files to analyze at the same time on the thread pool.

crates/wdl-analysis/src/document/v1.rs
```diff
@@ -179,13 +182,15 @@ fn sort_scopes(scopes: &mut Vec<Scope>) {
...
 }
 
-/// Creates a new document for a V1 AST.
+/// Creates a new document for a V1 AST with optional caching.
 pub(crate) fn populate_document(
     document: &mut DocumentData,
+    existing_cache: Option<Arc<AnalysisCache>>,
     config: &Config,
-    graph: &DocumentGraph,
+    graph: &mut DocumentGraph,
```
peterhuene: From a cursory check, it doesn't seem like this needs to be a `mut` at all.

crates/wdl-analysis/src/document.rs
```diff
@@ -52,17 +61,30 @@ pub struct Namespace {
...
     /// Gets the span of the import that introduced the namespace.
     pub fn span(&self) -> Span {
         self.span
     }
 
     /// Gets the URI of the imported document that introduced the namespace.
-    pub fn source(&self) -> &Arc<Url> {
-        &self.source
+    pub fn source(&self) -> Arc<Url> {
+        self.source.clone()
```
peterhuene, line 87: There are a few places where a reference to the `Url` is needed and not an owned `Arc`. Were there a lot of places where we ended up calling `source().clone()`?

crates/wdl-analysis/src/document.rs
```diff
@@ -653,21 +637,133 @@ impl Workflow {
...
+    }
+
+    /// Gets the offset of the struct in the source document's CST.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// Gets the URI of the document this struct was imported from.
+    pub fn source(&self) -> Arc<Url> {
+        self.document.uri()
```
peterhuene, line 677: Same comment here re: `&Arc<Url>` vs. `Arc<Url>` return.

crates/wdl-analysis/src/document.rs
```diff
@@ -653,21 +637,133 @@ impl Workflow {
...
+    }
+
+    /// Gets the offset of the enum in the source document's CST.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// Gets the URI of the document this enum was imported from.
+    pub fn source(&self) -> Arc<Url> {
+        self.document.uri()
```
peterhuene, line 735: Same comment here re: `&Arc<Url>` vs. `Arc<Url>` return.

crates/wdl-analysis/src/document.rs
```diff
@@ -995,8 +1099,8 @@ impl Document {
     }
 
     /// Gets the URI of the document.
-    pub fn uri(&self) -> &Arc<Url> {
-        &self.data.uri
+    pub fn uri(&self) -> Arc<Url> {
+        self.data.uri.clone()
```
peterhuene, line 1108: Same comment here re: `&Arc<Url>` vs. `Arc<Url>` return.

peterhuene (commented):
This looks good! Just a few comments to go over.

crates/wdl-analysis/src/document.rs
Serial-ATA: `&Arc<Url>` is just a strange return type, it should either be `&Url` or `Arc<Url>`, but I think the latter was needed since the `Arc<Url>` is used on other types

crates/wdl-analysis/src/document.rs
Serial-ATA: ah yep, good catch

crates/wdl-analysis/src/document.rs
peterhuene: I personally don't find it odd as it lets the caller make the decision if it needs to be cloned or not, which in a lot of cases the clone isn't necessary and we're just accessing a reference to the URL. Granted, the clone is basically just an atomic update to the reference count.

If it's too odd, perhaps a `fn uri(&self) -> &Url` and `uri_cloned(&self) -> Arc<Url>` given that there aren't too many of these called out in the review?

That said, I don't feel strongly about this so I'm also fine with just living with a single `fn uri(&self) -> Arc<Url>`.

peterhuene (approved):
The fix to the write lock scope for the graph looks good, so approving.

If you feel like addressing the comments on the `uri` methods, let me know.

crates/wdl-analysis/src/document.rs
Serial-ATA: Cloning an `Arc` is essentially free, so I wouldn't bother leaving it up to the caller. `&Arc<T>` is like `&String`, that extra indirection just isn't necessary.

At the end of the day though, we aren't calling this a million times in a loop, so neither performance concern would actually matter

claymcleod:
If Peter is good with it, no need for a second approval from me.

PR #1103, opened 2026-08-10
fix: improve call cache entry reuse.

### LINKED ISSUE #877: Call cache validation rejects repeated runs using `write_*()` generated files
**Describe the bug**

Call caching misses on repeated identical task runs when a task command uses a file generated by a `write_*()` function. While the task cache key is stable between runs, the existing cache entry is rejected during validation with:

```bash
ignoring call cache entry: the command of the task was modified
```

**Minimal runnable case**

`sprocket.toml`

```toml
[run.task]
cache = "on"
cache_dir = "call-cache"

[run.backends.default]
type = "local"
```

`input.wdl`

```wdl
version 1.2

task test {
    input {
        Array[String] xs = ["one", "two"]
    }

    command <<<
        cat ~{write_lines(xs)}
    >>>

    output {
        Array[String] lines = read_lines(stdout())
    }
}
```

Execute the same command twice, starting with a fresh cache and keeping `call-cache` between runs:

```sh
cargo run -- -vv run input.wdl -t test -c sprocket.toml -o runs/out
```

**Observed behavior**

Run 1 creates a cache entry:

```bash
using call cache directory `call-cache`
task cache key is `96ce43d9cb0144f191f9a530f969d9365c20d844f50053b3be93e6ebd86b37cd` task_id="test" task_name="test" document="file://.../input.wdl"
call cache miss for key `96ce43d9cb0144f191f9a530f969d9365c20d844f50053b3be93e6ebd86b37cd` task_id="test" task_name="test" document="file://.../input.wdl"
updated call cache entry for key `96ce43d9cb0144f191f9a530f969d9365c20d844f50053b3be93e6ebd86b37cd` task_id="test" task_name="test" document="file://.../input.wdl"
```

Run 2 computes the same task cache key, but rejects the entry:

```bash
using call cache directory `call-cache`
task cache key is `96ce43d9cb0144f191f9a530f969d9365c20d844f50053b3be93e6ebd86b37cd` task_id="test" task_name="test" document="file://.../input.wdl"
ignoring call cache entry: the command of the task was modified task_id="test" task_name="test" document="file://.../input.wdl"
updated call cache entry for key `96ce43d9cb0144f191f9a530f969d9365c20d844f50053b3be93e6ebd86b37cd` task_id="test" task_name="test" document="file://.../input.wdl"
```

**Expected behavior**

Repeated identical runs should hit the call cache. The files generated from `write_*()` functions should not invalidate cache reuse when their contents are unchanged. It looks like the command differs only because `write_lines()` produces a different generated file path on each run, even though the file contents are identical.

**Control case**

A task that writes the same content to a deterministic filename does hit the cache on the second run.

`input-stable.wdl`

```wdl
version 1.2

task test {
    input {
        Array[String] xs = ["one", "two"]
    }

    command <<<
        cat > lines.txt <<'EOF'
~{sep("\n", xs)}
EOF
        cat lines.txt
    >>>

    output {
        Array[String] lines = read_lines(stdout())
    }
}
```

Execute the same command twice, starting with a fresh cache and keeping `call-cache` between runs:

```sh
cargo run -- -vv run input-stable.wdl -t test -c sprocket.toml -o runs/out
```

Run 1 creates a cache entry:

```bash
using call cache directory `call-cache`
task cache key is `83647b98951f7e137e86c7541eb94ac5de73a6c7eac4339634b19c732abf1522` task_id="test" task_name="test" document="file://.../input-stable.wdl"
call cache miss for key `83647b98951f7e137e86c7541eb94ac5de73a6c7eac4339634b19c732abf1522` task_id="test" task_name="test" document="file://.../input-stable.wdl"
updated call cache entry for key `83647b98951f7e137e86c7541eb94ac5de73a6c7eac4339634b19c732abf1522` task_id="test" task_name="test" document="file://.../input-stable.wdl"
```

Run 2 hits the cache:

```bash
using call cache directory `call-cache`
task cache key is `83647b98951f7e137e86c7541eb94ac5de73a6c7eac4339634b19c732abf1522` task_id="test" task_name="test" document="file://.../input-stable.wdl"
task execution was skipped due to previous result being present in the call cache task_id="test" task_name="test" document="file://.../input-stable.wdl"
```

**Possible fix**

One possible fix would be for `write_*()` functions to return stable generated paths. For example, `write_lines()` could compute a digest of the generated file contents and write to a path such as `write_lines_<digest>.tmp`. Alternatively, cache validation could normalize paths generated by `write_*()` functions before comparing commands.

**Environment:**

- OS: Linux
- Version: 0.25.0
- Shell: bash

peterhuene:
This PR ensures that call cache entires will not be ignored if guest paths change between evaluation; this most commonly occurs with files created by the `write_*` stdlib functions as they are given random names.

There are two parts to this fix.

The first part is that command digest calculation should effectively treat a guest path reference as a reference to only the file name, as the directory containing the input is unstable and an implementation detail of the engine (e.g. `/mnt/task/inputs/<num>`). Additionally, if the input is a temporary file created by a `write_*` stdlib function call, the entire guest path is ignored for the purpose of the command's digest calculation.

The second part is that the backend input digests should be stored as a sorted list rather than a map that is keyed off of a path that might change.

Additionally, digest calculation of the temporary files created by the `write_*` stdlib functions is forced to be "strong", meaning file metadata is not part of the digest, only the content. As invocations of the `write_*` functions create new files, the file metadata should not be considered in the context of the call cache.

Note: this fix changes the calculation of call cache keys and therefore invalidate every existing cache entry.

Fixes #877.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-engine/src/cache.rs
```diff
@@ -51,6 +53,78 @@ mod lock;
...
+/// just the input's file name.
+///
+/// References to guest paths of temporary files are ignored entirely.
+fn hash_command(request: KeyRequest<'_>) -> ArrayString<64> {
+    let mut hasher = blake3::Hasher::new();
+    match request.guest_inputs_dir {
+        Some(prefix) => {
+            // Find the next prefix in the command string
+            let mut current = request.command;
+            while let Some(start) = current.find(prefix) {
```
claymcleod, line 67: nit: The normalizer rescans every backend input for every occurrence of the guest-input prefix. Consider exposing and reusing `InputTrie`’s longest-prefix lookup here, so each occurrence avoids scanning the full input slice.

crates/wdl-engine/src/cache.rs
```diff
@@ -51,6 +53,78 @@ mod lock;
...
+                current = &current[end..];
+            }
+
+            // Hash the remainder of the command
+            if !current.is_empty() {
+                current.hash(&mut hasher);
+            }
+        }
+        None => {
+            request.command.hash(&mut hasher);
```
claymcleod, line 129: This still misses the local backend. It has no `guest_inputs_dir`, so two `write_*()` calls with the same content still hash their different host paths and reject the cache entry. Normalize tracked `TempFile` host paths in this branch as well.

crates/wdl-engine/src/cache.rs
```diff
@@ -51,6 +53,78 @@ mod lock;
...
+                        let input = &request.backend_inputs[index];
+                        input.kind().hash(&mut hasher);
+                        match input.kind() {
+                            ContentKind::File | ContentKind::Directory => {
+                                // Hash just the file name
+                                // SAFETY: guest paths are always Unix style and have a slash
+                                let slash = start + current[start..end].rfind('/').unwrap();
+                                (&current[slash + 1..end]).hash(&mut hasher);
+                            }
+                            ContentKind::TempFile => {
```
claymcleod: This loses the relationship between a command position and its generated file. For example, after caching:

```wdl
version 1.3

task order_sensitive {
    command <<<
        printf '%s:%s\n' ~{write_lines(["left"])} ~{write_lines(["right"])}
    >>>
}
```

swapping the two `write_lines()` calls changes the command’s output but leaves the task inputs, sorted digest list, and normalized command digest unchanged. Compute each backend input digest before command normalization, then hash the matched temporary input’s strong digest at this position instead of dropping the path. That keeps regenerated paths reusable while preserving argument order.

crates/wdl-engine/src/cache.rs
```diff
@@ -277,6 +349,18 @@ impl Key {
             Ok(())
         }
 
+        if self.inputs.len() < entry.inputs.len() {
```
claymcleod, line 361: `cache.exclusions.inputs` no longer applies when validating backend inputs. An excluded WDL input keeps the same cache key, but changing its file still changes this digest vector and rejects the entry. Filter excluded inputs before building and comparing this vector, or retain enough identity to apply the exclusion here.

crates/wdl-engine/src/cache.rs
```diff
@@ -51,6 +53,78 @@ mod lock;
...
+
+                // Find the longest match from the backend input guest paths
+                let end = match request
+                    .backend_inputs
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(input_index, input)| {
+                        let p = input.guest_path()?.as_str();
+                        current[start..]
+                            .starts_with(p)
```
claymcleod: This also matches a guest path as a prefix of a different literal path. With an input at `/mnt/task/inputs/0/foo`, the command path `/mnt/task/inputs/0/foobar` is normalized as that input plus `bar`, which collides with a real `foo` reference followed by literal `bar`. Require a path boundary after the match before normalizing it.

crates/wdl-engine/src/cache.rs
```diff
@@ -836,6 +916,110 @@ mod test {
         );
     }
 
+    #[tokio::test]
+    async fn modified_guest_path() {
```
claymcleod, line 843: The regression lives in task evaluation, but these tests construct `KeyRequest` directly. Add a WDL-level test that runs a local task using `write_*()` twice and asserts the second run is cached; also cover a generated file that reaches the backend without appearing directly in the command.

claymcleod (changes requested):

crates/wdl-engine/src/cache.rs
peterhuene: This was intentional. I view the local backend as needed to get the engine off the ground during very early implementation and a candidate for being ripped out (especially for a v1); eventually resulting in removing this branch by always having a guest path prefix.

crates/wdl-engine/src/cache.rs
peterhuene: I actually had it the way you suggest originally, but just doing the content digest hash at the position of the input regardless of temp file or not.

I'll restore what I had earlier and also continue to hash the file name for non-temporary files.

crates/wdl-engine/src/cache.rs
peterhuene: This is an existing bug that was introduced when exclusions were added to the cache implementation.

The filtering that was done before would never have worked because the backend inputs were keyed by _path_ and not by input name.

There is actually no relationship between input and any backend inputs (it can be one to many) currently. I can see about fixing this.

crates/wdl-engine/src/cache.rs
peterhuene: In the context of a bash script, any character that isn't a null byte or a `/` might be a valid path character, though; it isn't an easy matter of detecting a "path boundary" apart from fully parsing the script like bash would. For example, the path might be quoted and thus whitespace could itself be part of the path.

This filter matches via longest matching input path. If `/mnt/task/inputs/0/foobar` was in the command text and there were two inputs `/mnt/task/inputs/0/foo` and `/mnt/task/inputs/0/foobar`, then the latter wins.

It would be very unlikely for an evaluated command to reference an input mount path directly such that `/mnt/task/inputs/0/foobar` appears in the command but there isn't a matching backend input with that guest path (i.e. just an input with guest path `/mnt/task/inputs/0/foo`).

crates/wdl-engine/src/cache.rs
peterhuene: The `InputTrie` is keyed off of the host paths (i.e. it is responsible for mapping host paths to guest paths such nodes with the same host parent directory get the same guest parent directory) and not the guest paths, which is what we're searching here.

crates/wdl-engine/src/cache.rs
peterhuene: The bug lived in the call cache implementation, not in task evaluation, so I think unit tests here are sufficient to test for future regressions.

As we don't capture messages logged by `wdl-engine` for the file evaluation tests as they're generally not easy to baseline, it's not easy for the tests to detect if a call cache hit or miss occurred (it's otherwise transparent to evaluation itself).

I'll add a test to cover a backend input that _isn't_ referenced by the command, though.

peterhuene: Never mind, you were right as we have unit tests in the `task` module that looks for logged messages for cache hits / misses, so I've added evaluation-level tests there.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -2565,4 +2589,367 @@ task test {
...
         );
     }
+
+    /// Tests that calls to `write_lines` is cacheable.
+    /// See: https://github.com/stjude-rust-labs/sprocket/issues/877
+    #[tokio::test]
+    #[traced_test]
+    async fn cache_write_lines() {
+        const SOURCE: &str = r#"
+version 1.2
```
claymcleod:
```suggestion
version 1.3
```

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -603,18 +606,23 @@ impl<'a> State<'a> {
...
         {
             // If the input has a guest path, map it
-            let input = &self.backend_inputs.as_slice()[index];
+            let input = &mut self.backend_inputs.as_slice_mut()[index];
             if let Some(guest_path) = input.guest_path() {
                 self.path_map.insert(path.clone(), guest_path.clone());
             }
 
+            if excluded_from_call_cache {
+                input.mark_excluded_from_call_cache();
```
claymcleod: This exclusion flag applies to the deduplicated backend input, so please keep the input cache-significant whenever any non-excluded WDL binding aliases the same path. For example, if `excluded` and `included` both resolve to `shared.txt`, evaluating `excluded` first marks their shared `Input` excluded, and evaluating `included` cannot clear it. `CallCache::key` then omits the file digest and `hash_command` skips the command reference, allowing changed contents to return stale cached output. I reproduced this with an evaluation-level regression where the second run incorrectly reported `cached() == true`; please add that alias case to the tests.

tests/cli/config/init/run/stdout
```diff
@@ -1,4 +1,4 @@
-#:schema https://raw.githubusercontent.com/stjude-rust-labs/sprocket/refs/heads/main/jsonschemas/sprocket.toml.json
+#:schema http/raw.githubusercontent.com/stjude-rust-labs/sprocket/refs/heads/main/jsonschemas/sprocket.toml.json
```
claymcleod: This isn't correct and needs to be fixed following my latest commit on `main`.

claymcleod (approved):
LGTM after these changes.

tests/cli/config/init/run/stdout
peterhuene: I pushed up one last commit ([2b5a673](https://github.com/stjude-rust-labs/sprocket/pull/1103/commits/2b5a6737e8e6381a04d5a36865e1590481e9ed7c)) after I rebased off of main that corrected the baselines. Looks like something changed in a `Cargo.lock` change that trimmed some whitespace in the baselines.

It did correct this commit, though.

PR #1107, opened 2026-08-11
fix(wdl-engine): WDL 1.2.1 `Directory` -> `String` coercion

### LINKED ISSUE #815: Ensure `Directory` to `String` coercion does not include a trailing slash
Relates to #637.

The [WDL v1.2.1 release](https://github.com/openwdl/wdl/releases/tag/v1.2.1) clarified that when a `Directory` is converted to a `String`, the resulting string does not include a trailing slash ([openwdl/wdl#745](https://github.com/openwdl/wdl/pull/745)).

The current `Directory` → `String` coercion in `crates/wdl-engine/src/value.rs` (around line 1617) converts the path as-is without stripping trailing slashes. `HostPath` is a raw `Arc<String>` wrapper that stores the path exactly as provided—no normalization occurs on construction. While `path_clean` is used elsewhere in the codebase and does strip trailing slashes, it's only applied within `EvaluationPath` (for base directory resolution), never to `HostPath` directly.

This means a trailing slash can survive into a String coercion when a Directory value is constructed from a WDL string literal (e.g., `Directory dir = "/some/path/"`), a String → Directory coercion, or a direct `HostPath::new()` call. There are currently no tests covering this behavior.

An explicit trailing-slash strip should be added—either in `HostPath::new()` (or a Directory-specific constructor) or at the Directory → String coercion boundary.

Serial-ATA:
closes #815

For external contributors:

For all contributors:

      is implemented.


crates/wdl-engine/src/value.rs
```diff
@@ -143,7 +144,7 @@ impl From<Arc<String>> for HostPath {
 
 impl From<HostPath> for Arc<String> {
     fn from(path: HostPath) -> Self {
-        path.0
+        path.to_string().into()
```
peterhuene: So I think it remains problematic that users can do a `guest_path.as_str().to_string()` to get a `String` that does not have a trailing slash.

It might be best to refactor `HostPath` and `GuestPath` in the following ways:

* Make the inner field not `pub` (ugh), forcing users to go the constructor and also use `as_str` for conversion to `&str`.
* Change the constructors to take `&str` instead of `Into<String>` and do the `trim_end_matches('/')` there.
* Add a `From` impl between `HostPath` and `GuestPath` that just clones the inner `Arc<String>`.

Thus we can convert to `Arc<String>` and do formatting without having to trim since it was already trimmed on construction. 

It'd be a bit more involved change, but probably for the best.

What are your thoughts?

Serial-ATA: I was thinking about moving it to the constructor, wasn't sure if we should just trim when needed and keep the original path around just in case. If we don't need the original path then yeah sure

peterhuene: Yeah I think it can be a lossy construction. We should probably document that it's guaranteed to not have a trailing slash as well; there's one or to places where we trim it for certain `format!` calls that could be updated to skip doing so too.

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -669,12 +669,11 @@ impl<'a> State<'a> {
...
                 match (&path_url, &host_url) {
                     (None, None) => {
-                        if let Some(remainder) = strip_path_prefix(path.0.as_str(), host.0.as_str())
-                        {
+                        if let Some(remainder) = strip_path_prefix(path, host) {
                             // Note: guest paths are always Unix-style paths
                             return Some(GuestPath::new(format!(
                                 "{base}/{remainder}",
-                                base = guest.0.trim_end_matches('/'),
+                                base = guest,
```
peterhuene: Nit: could we replace `base` in the format string with `guest`?

crates/wdl-engine/src/eval/v1/task.rs
```diff
@@ -684,7 +683,7 @@ impl<'a> State<'a> {
                             // Note: guest paths are always Unix-style paths
                             return Some(GuestPath::new(format!(
                                 "{base}/{remainder}",
-                                base = guest.0.trim_end_matches('/')
+                                base = guest
```
peterhuene: Nit: could we replace `base` in the format string with `guest`?

crates/wdl-engine/src/value.rs
```diff
@@ -72,12 +72,18 @@ use crate::path;
...
 impl HostPath {
     /// Constructs a new host path from a string.
+    ///
+    /// NOTE: Any trailing slashes will be stripped.
     pub fn new(path: impl Into<String>) -> Self {
-        Self(Arc::new(path.into()))
+        let path = path.into();
+
+        // https://github.com/openwdl/wdl/pull/745
+        let trimmed = path.trim_end_matches('/');
```
peterhuene: Nit: probably a micro-optimization, but could we perhaps check to see if the string ends with slash first and then trim it (thus reallocating) otherwise take the string as-is?

crates/wdl-engine/src/value.rs
```diff
@@ -137,7 +155,7 @@ fn write_escaped_wdl_string(f: &mut fmt::Formatter<'_>, s: &str) -> fmt::Result
 
 impl From<Arc<String>> for HostPath {
     fn from(path: Arc<String>) -> Self {
-        Self(path)
+        Self::new(path.as_str())
```
peterhuene: Nit:

```suggestion
        Self::new(path.unwrap_or_clone())
```

crates/wdl-engine/src/value.rs
peterhuene: I think we need to use the constructor here:

```suggestion
        Self::new(path.unwrap_or_clone())
```

peterhuene (commented):
Just a few nits.

peterhuene (approved):
Looks good!

PR #1109, opened 2026-08-12
fix: mark runs whose owning process stopped reporting as `orphaned`

### LINKED ISSUE #1091: Runs orphaned by a server restart are stuck `Running`/`Queued`/`Canceling` forever, and `/cancel` silently no-ops
**Describe the bug**

When the `sprocket server` process stops for any reason (crash, redeploy, manual restart, OOM-kill) while one or more runs are still `Queued`, `Running`, or `Canceling`, those runs are never revisited by the process that replaces it. A run's execution state — the in-memory cancellation/tracking handle and the `tokio` task actually driving its WDL evaluation — lives entirely inside the process that started it and is not persisted or reconstructed on startup.

As a result, after any restart:

- Every run that was in progress at the moment of the restart stays at its last non-terminal status (`Queued`/`Running`/`Canceling`) in the database **forever** — there is no executor left anywhere that could ever move it forward.
- `POST /api/v1/runs/{id}/cancel` against one of these runs returns `200 OK`, but does nothing: it neither halts anything (there's nothing left to halt) nor updates the run's status. Repeating the cancel request has the same effect indefinitely, giving no indication anything is wrong.

This is especially damaging for a long-lived server that gets restarted periodically (deploys, crashes, OOM), or that sees enough load to occasionally lose a worker: the number of permanently "stuck" runs accumulates over the server's lifetime, and operators have no way to clear them out through the API — cancel looks like it works but never actually resolves anything.

**Expected behavior**

At minimum:

- A restart must not silently orphan in-flight runs. Runs that can never make further progress should be transitioned to a terminal status (e.g. `Failed`) so the database only ever reflects runs the current process can actually act on.
- `/cancel` must never report success for a run it did not actually affect. If a run can't be canceled (for example, because it isn't tracked by the current process), the endpoint should return an error, not `200 OK`.

**Steps to Reproduce**

1. Start `sprocket server` and submit a workflow via `POST /api/v1/runs` that runs long enough to still be `Running` a few seconds later (or that will sit `Queued` if `max-concurrent-runs` is small enough).
2. Kill the server process (`kill -9 <pid>`, or a crash/redeploy) while the run is still `Running`.
3. Start a new `sprocket server` process pointed at the same database.
4. `GET /api/v1/runs/{id}` — the run is still reported as `Running` (or `Queued`/`Canceling`), even though nothing anywhere is executing it.
5. `POST /api/v1/runs/{id}/cancel` — returns `200 OK`.
6. `GET /api/v1/runs/{id}` again — the status is unchanged. The run is stuck permanently, and every future cancel attempt behaves the same way.

**Terminal Output**

```bash
$ curl -s -X POST localhost:4242/api/v1/runs/<id>/cancel
{"uuid":"<id>"}
# HTTP 200 — but the run's status never changes:
$ curl -s localhost:4242/api/v1/runs/<id> | jq .run.status
"running"
```

**Screenshots**

N/A

- OS: any
- Version: `sprocket` `main` (v0.29.0), commit `da931b8`
- Shell: n/a
- Toolchain version: n/a (server-side bug, not toolchain-dependent)

**Additional context**

Root cause (from `src/system/v1/exec/svc/run_manager.rs`):

- `RunManagerSvc` tracks active runs in an in-memory map (`runs: Arc<Mutex<HashMap<Uuid, CancellationContext>>>`). This map is populated only when a run is submitted in the *current* process, and starts empty on every `RunManagerSvc::new`. Nothing reconciles it against rows the database already has in `Running`/`Queued`/`Canceling` status when the process starts.
- `cancel_run` only takes action when the run's ID is present in that map, or when the database status is already `Canceling`. For a run whose status is `Running`/`Queued` but has no map entry (i.e. an orphan left behind by a previous process), neither branch is taken — the function falls through and returns success without ever touching the database, which the HTTP handler forwards to the client as `200 OK`.

I'm actively working on a fix for this (tracked on the `cancel-fix` branch): failing orphaned runs at startup — scoped so it only ever touches runs submitted through `sprocket server`, never `sprocket run` invocations sharing the same database — and making `cancel_run` return an explicit error instead of silently succeeding when it can't act on a run. Will link the PR here.

claymcleod:
Closes #1091. A run's execution state lived only in the process that started it, so runs left behind by a stopped `sprocket server` sat at `Queued`/`Running`/`Canceling` forever, and `cancel_run` returned `200 OK` for them without touching the database.

Processes that own a session now record a liveness heartbeat, and a running server sweeps runs whose owner has gone quiet into a new `orphaned` status. Cancelling an untracked run returns `409 Conflict`. Widening the `status` constraints meant the migration rebuilds `runs` and `tasks`.

For all contributors:


tests/cli/config/init/run/stdout
```diff
@@ -1,4 +1,4 @@
-#:schema https://raw.githubusercontent.com/stjude-rust-labs/sprocket/refs/heads/main/jsonschemas/sprocket.toml.json
+#:schema http/raw.githubusercontent.com/stjude-rust-labs/sprocket/refs/heads/main/jsonschemas/sprocket.toml.json
```
a-frantz: revert this

a-frantz (approved):
besides the bad test BLESS, LGTM. I ran it against killed server DB on the cluster and it correctly orphaned everything 🚀

Issue #595, opened 2026-02-04
sprocket check/lint runs on files that are in .gitignore


williamrowell:
**Describe the bug**
Running `sprocket check` or `sprocket lint` in a git repo will check all `.wdl` files in the repo, including those that that are excluded in .sprocketignore.

**Expected behavior**
I understand that there is a separate `.sprocketignore`, but it seems useful to ignore files that git ignores as well, e.g. temporary files in test output directories.

**Environment (please complete the following information):**
 - OS: Rocky 9, Linux
 - Version: sprocket 0.20.1 (2026-02-04)
 - Shell: bash, zsh
 - Toolchain version: rustc 1.93.0 (254b59607 2026-01-19)

**Additional context**
Wasn't sure whether to call this a bug or a feature request, but many tools that you expect to run in git repos tend to ignore files in .gitignore, so it felt like a borderline bug.

PR #1110, opened 2026-08-13
fix: add `--no-ignore` CLI option and fixes a11y test

claymcleod:
Added the global `--no-ignore` flag so commands can discover WDL documents despite `.sprocketignore` files. The `a11y` workflow uses the flag when generating its fixture documentation.

For all contributors:


a-frantz:
#595 
should we add a `git_ignore` config value that is carried through this same code path? I'm not sure it makes sense to exist as a CL flag. Conversely I think `--no-ignore` only makes sense as a CL flag and not a config value, which presents an odd asymmetry,  but 🤷‍♀️

peterhuene (approved):

peterhuene:
Shoot, I didn't see Ari's comment prior to approving (it seemed to auto-merge).

PR #1115, opened 2026-08-15
feat: harden Git fetch and cache operations for WDL modules

claymcleod:
Stacked on #999; the base branch is `feat/module-porcelain`, so that PR merges first.

Added the `[modules] max_transfer_bytes` cap so a runaway Git fetch aborts instead of filling the module cache. It defaults to 2 GiB and accepts `unlimited`.

Fixed `ssh-agent` authentication for Git dependencies reached over SSH. The credential callback returned credential types `libgit2` had not requested, which aborted the fetch; it now returns only a requested type, and a failing credential helper surfaces as an authentication error rather than a generic Git error.

Made cache reuse survive damaged state. A cache leaf whose sparse-checkout metadata cannot be parsed is evicted and re-cloned, that metadata is written through a temporary file and a rename, and materialized-tree limit checks read blob sizes from Git object headers instead of loading each blob's contents and silently skipping unreadable ones.

Split `resolver::git::ops` into `cache_store`, `checkout`, `creds`, `error`, and `remote` submodules, since the single file had grown past 1,600 lines. `wdl_modules::resolver` now also exports `TransferLimit` and `TransferLimitError`, which the public `ModulesConfig::max_transfer_bytes` field needs to be nameable downstream.

For all contributors:


claymcleod:
I'm going to go ahead and merge this into `feat/module-porcelain` so it can be reviewed in one sitting. Many of the changes are simply moves, which makes it difficult to tell which functionality changed in this PR.

PR #1117, opened 2026-08-17
feat(sprocket): show container image pull events

### LINKED ISSUE #363: Add containers being pulled to metrics
Currently, when containers are being pulled, the status menu looks like nothing is happening. We should add a metric there to show that work is being done via the container pulls.

Serial-ATA:
This shows the container image(s) being pulled during `sprocket run` and `sprocket dev test` using the new `crankshaft` events.

closes #363 

For external contributors:

For all contributors:

      is implemented.


crates/wdl-engine/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+<!-- TODO: Add PR number -->
```
claymcleod: This needs filling in.

src/commands/run.rs
```diff
@@ -481,6 +496,25 @@ async fn progress(
...
                     let removed = match event {
+                        CrankshaftEvent::ImagePullStarted { name, .. } => {
+                            // We can hit this multiple times for the same image
+                            if state.pulling_images.get_mut(&name).is_some() {
+                                continue;
+                            }
+
+                            let span = span!(parent: progress_bar.clone(), Level::WARN, "pull image");
+                            span.pb_set_style(&sub_task_style);
+                            span.pb_set_message(&format!("pulling image '{}'", name.green()));
```
claymcleod:
```suggestion
                            span.pb_set_message(&format!("pulling image `{}`", name.green()));
```

Cargo.toml
```diff
@@ -47,7 +47,8 @@ cloud-copy = { version = "0.10.0", features = ["cli"] }
 codespan-reporting = "0.13.1"
 colored = "3.1.1"
 convert_case = "0.11.0"
-crankshaft = { version = "0.10.0", features = ["docker"] }
+crankshaft = { git = "https://github.com/stjude-rust-labs/crankshaft.git", features = ["docker"] }
```
claymcleod, line 51: Perhaps we should just go ahead and release Crankshaft so that we can depend on the version here?

crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -276,15 +281,37 @@ impl ApptainerRuntime {
...
+        let display_name = container.uri();
+        let events = events_ctx.as_ref().map(|(sender, _)| sender.clone());
+        let task = events_ctx.as_ref().map(|(_, task_id)| *task_id);
+
         // For local SIF files, return the path directly.
         if let ContainerSource::SifFile(path) = container {
             return Ok(Some(path.clone()));
         }
 
+        send_event!(
```
claymcleod, line 295: `ImagePullStarted` is left unmatched on cache hits, pull failures, and cancellation; emit the start and exactly one terminal event at the same level around the `OnceCell` future. `ImagePullFinished` currently runs only inside the initializer after a network pull, so an initialized cell, an on-disk cache hit, an exhausted retry, or the canceled `select!` leaves `run` and `dev test` displaying this image for the rest of the command.

crates/wdl-engine/src/backend/docker.rs
```diff
@@ -302,18 +327,24 @@ impl<'a> DockerTask<'a> {
...
 
-        assert_eq!(statuses.len(), 1, "there should only be one exit status");
-        let status = statuses.first();
+        assert_eq!(results.len(), 1, "there should only be one result");
+        let result = results.first();
 
         Ok(Some(TaskExecutionResult {
-            container: Some(self.container),
-            exit_code: status.code().expect("should have exit code"),
+            container: Some(ContainerSource::Docker(
```
claymcleod, line 363: `TaskExecutionResult::container` now receives Docker’s inspected image ID; preserve the selected requested image reference instead. For WDL 1.2 tasks this value becomes `task.container` and is stored in the call cache, so `ubuntu:latest` now turns into `sha256:…`; `unwrap_or_default()` can also record an empty container when the backend supplies no image.

src/commands/test.rs
```diff
@@ -458,10 +473,97 @@ struct TestTask {
...
+            pulling_images: Arc::new(Mutex::new(HashSet::new())),
+            span: status_bar,
+        }
+    }
+
+    fn update_image_pull_status(&self, images: &HashSet<String>) {
+        const MAX_PULLING_IMAGES: usize = 3;
+
+        if images.is_empty() {
+            self.span.record("indicatif.pb_hide", true);
```
claymcleod: `record("indicatif.pb_hide", true)` cannot hide an already-active progress bar; close and drop the span when the image set becomes empty, then create a fresh span for the next pull. This span declares no such field, `IndicatifFilter` only checks field metadata when deciding whether to create a bar, and `pb_reset()` resets state without removing the bar, so `dev test` keeps the last pull spinner visible after completion.

src/commands/test.rs
```diff
@@ -421,13 +436,13 @@ impl TestIteration {
         if !quiet && self.cancellation.state() != CancellationContextState::Canceling {
             match &evaluation {
```
claymcleod, line 439: `dev test` result lines now move from stdout to stderr; write them through an Indicatif stdout writer instead. `initialize_logging` obtains this writer with `get_stderr_writer()`, so scripts that capture stdout lose every per-iteration result, including when `--no-status` disables the new display.

src/lib.rs
```diff
@@ -255,7 +256,12 @@ fn initialize_logging(
 
     // Set up an indicatif layer so that progress bars don't interfere with logging
     // output
-    let indicatif_layer = IndicatifLayer::new();
+    let pending_footer = ProgressStyle::with_template("    ...and {pending_progress_bars} more")?;
+    let indicatif_layer = IndicatifLayer::new()
+        .with_span_child_prefix_indent("    ")
+        .with_span_child_prefix_symbol("↳ ")
+        .with_max_progress_bars(commands::run::MAX_LINES, Some(pending_footer));
```
claymcleod: This process-wide cap also hides unrelated transfer progress bars; limit image-pull children locally instead. During concurrent localization, the run bar and one transfer consume both visible slots, so additional `cloud-copy` transfers that previously had their own bars collapse into the pending footer.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -980,90 +894,187 @@ impl TaskExecutionBackend for LsfApptainerBackend {
...
-                            self.events.crankshaft(),
-                            CrankshaftEvent::TaskFailed {
-                                id: crankshaft_id,
-                                message: format!("{e:#}"),
-                            },
-                        );
+                            exit_code
+                        },
+                        Err(e) => {
+                            send_event!(
```
claymcleod: This job-error path now emits `TaskFailed` twice; remove this inner send and let the new outer error handler report it once. Returning `Err(e)` reaches the handler at the end of `execute`, so `TaskMonitorSvc` processes the same terminal transition twice and the second, shorter message can overwrite the first.

crates/wdl-engine/src/backend/lsf_apptainer.rs
```diff
@@ -980,90 +894,187 @@ impl TaskExecutionBackend for LsfApptainerBackend {
...
+                    .generate_script(
+                        &backend_config.apptainer,
+                        &self.config.task.shell,
+                        &request,
+                        self.cancellation.first(),
+                        self.events.crankshaft().map(|e| (e.clone(), crankshaft_id)),
+                    )
+                    .await?
+                else {
                     return Ok(None);
```
claymcleod, line 960: Cancellation during the image pull leaves `TaskCreated` without a terminal event; send `TaskCanceled` before returning `Ok(None)` here and in the Slurm equivalent. Because `TaskCreated` now precedes `generate_script`, cancellation through `self.cancellation.first()` takes this branch before the later cancellation helper exists, violating the event contract and leaving the task in the progress and server state.

src/commands/test.rs
```diff
@@ -458,10 +473,97 @@ struct TestTask {
     inputs: EngineInputs,
 }
 
+#[derive(Clone)]
+struct StatusBar {
+    /// The images currently being pulled.
+    pulling_images: Arc<Mutex<HashSet<String>>>,
```
claymcleod: This shared `HashSet` undercounts concurrent pulls of the same image; track a reference count per image and remove the key only when its count reaches zero. Each test iteration creates its own backend and event channel, so two iterations can pull `ubuntu:latest` simultaneously; the first finish removes the only set entry while the second pull is still active.

crates/wdl-engine/src/backend/docker.rs
```diff
@@ -279,7 +304,7 @@ impl<'a> DockerTask<'a> {
             .name(&self.name)
             .executions(NonEmpty::new(
                 Execution::builder()
-                    .image(self.container.to_string())
+                    .images(images)?
```
claymcleod: When every candidate is unsupported, this now returns only `no image specified`; preserve the per-candidate rejection reasons in the returned error. The preceding `filter_map` drops `file://`, `oras://`, `library://`, and unknown candidates after logging warnings, whereas the removed `PullResults` path told the user which values failed and why.

claymcleod (changes requested):

crates/wdl-engine/src/backend/docker.rs
Serial-ATA: > ubuntu:latest now turns into sha256:…

That was a bug in crankshaft, that's been fixed on main

> unwrap_or_default() can also record an empty container when the backend supplies no image.

That isn't possible anyway. Crankshaft panics when if docker reports no image

crates/wdl-engine/src/backend/apptainer.rs
Serial-ATA: > an on-disk cache hit

Thats the intended behavior of start events: https://github.com/stjude-rust-labs/crankshaft/blob/81624d3977ddee898d1c20c20f1cc48c79d7dc37/crankshaft-events/src/lib.rs#L122-L123

The others are bugs, though

Cargo.toml
peterhuene: I may have other crankshaft changes before the next release as part of my refactoring.

I'm fine with depending on `main` out of Crankshaft for now.

crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -344,13 +373,23 @@ impl ApptainerRuntime {
             .await
             .with_context(|| format!("failed pulling Apptainer image `{container}`"))?;
 
+            send_event!(events_clone, Event::ImagePullFinished { id: task.unwrap(), name: container.uri() });
```
claymcleod: `ImagePullFinished` is still emitted twice for a fresh pull and not at all for cancellation; keep terminal emission only at the per-call `select!` boundary and emit one terminal event from its cancellation arm. The initializer sends a finish here, the success arm sends another, and `_ = token.cancelled()` returns without either terminal event, which still violates Crankshaft’s documented one-terminal-event contract and corrupts the UI reference counts.

Cargo.toml
claymcleod: Reminder that, before merging, we need to pin a released version of Crankshaft so that the Nix builds work.

src/commands/test.rs
```diff
@@ -458,10 +475,136 @@ struct TestTask {
...
+
+    fn update_image_pull_status(state: &mut StatusBarState) {
+        const MAX_PULLING_IMAGES: usize = 3;
+
+        if state.pulling_images.is_empty() {
+            state.span = tracing::Span::none();
+            return;
+        }
+
+        state.span = span!(Level::WARN, "status bar");
```
claymcleod, line 532: Apply the configured `ProgressStyle` to the span that is actually started, and avoid replacing that span on every event. `StatusBar::new` styles a span that is never started, then this line discards it and creates an unstyled span; tracing-indicatif’s default template has no `{msg}`, so the image name and color-aware template may not render, and elapsed time resets on every update.

claymcleod (changes requested):
Sorry, just a few more things.

Cargo.toml
Serial-ATA: @claymcleod Peter's waiting on his PR before the next crankshaft release. We can have a failing Nix build for now

src/commands/test.rs
claymcleod: On the span issue, I meant something like this:

```rust
Self {
    disabled: false,
    state: Arc::new(Mutex::new(StatusBarState {
        pulling_images: HashMap::new(),
        style: ProgressStyle::with_template(template).unwrap(),
        span: tracing::Span::none(),
    })),
}
```

Then create and start the span only when transitioning from no active pulls:

```rust
if state.pulling_images.is_empty() {
    state.span = tracing::Span::none();
    return;
}

if state.span.is_none() {
    state.span = span!(Level::WARN, "status bar");
    state.span.pb_set_style(&state.style);
    state.span.pb_start();
}
```

That keeps the same span while pulls remain active, so subsequent events only update its message instead of resetting the spinner and elapsed time.

claymcleod (approved):
This looks good with two caveats:

- It would be best to address the span issue this way IMO: https://github.com/stjude-rust-labs/sprocket/pull/1117#discussion_r3858535132
- I really want to wait until Peter releases Crankshaft, which he should do imminently. Not doing so will just mean I have to add another commit before I release tommorrow, so it's just worth waiting.

peterhuene:
@claymcleod it's not possible to update this PR to the latest `crankshaft` release as it needs my refactoring changes to fix the build.

I think it'd be to best to merge this with a git reference and the refactoring PR will be able to update to `crankshaft` 0.11.0.

claymcleod:
Alright, then we can proceed then.

PR #1125, opened 2026-08-18
refactor(wdl-lint)!: promote `ExceptDirectiveValid` to analysis

### LINKED ISSUE #997: wdl-lint: promote `ExceptDirectiveValid` to `wdl-analysis`
With #858, `wdl-analysis` rules are exceptable like `wdl-lint` rules, where each rule now specifies its exceptable nodes.

`ExceptDirectiveValid` currently lives in `wdl-lint`, however, so running `sprocket check` on the following:

```wdl
version 1.3

#@ except: UnusedImport
task say_hello {
    command <<<
        echo "Hello, world!"
    >>>
}
```

Emits `MeaninglessLintDirective` rather than `ExceptDirectiveValid`:

```wdl
note[MeaninglessLintDirective]: unnecessary `except` directive for lint rule `UnusedImport`
  ┌─ foo.wdl:3:12
  │
3 │ #@ except: UnusedImport
  │            ^^^^^^^^^^^^
```

`ExceptDirectiveValid` should be promoted to `wdl-analysis` so that *all* rules can be checked for validity

**Additional context**
Note that `MeaninglessLintDirective` implies that a lint exception *is* valid, but doesn't currently suppress anything. `ExceptDirectiveValid` means that the comment is misplaced and would *never* suppress that rule.

Serial-ATA:
closes #997

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/src/validation/exceptions.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    Diagnostic::note(format!(
+        "`except` directive `{id}` has no effect above {elem}",
+        elem = wrong_element.kind().describe()
+    ))
+    .with_rule(ExceptDirectiveValidRule::ID)
+    .with_label("cannot make an exception for this rule", span)
+    .with_label(
+        "invalid element for this `except` directive",
+        wrong_element.text_range(),
```
a-frantz: nit: we should only highlight the first token of `wrong_element`, otherwise a huge block of code my be highlighted for something that can be accurately conveyed with an individual token highlight. Something like:
```suggestion
        wrong_element.first_token().unwrap().span(),
```
would cause better rendering in an IDE.

Serial-ATA: TBF it'd also probably be fine to drop that label entirely. Highlighting the rule and pointing out the valid locations should be enough

Serial-ATA: went ahead and just removed the label

crates/wdl-lint/tests/lints/nonmatching-output/source.errors
```diff
@@ -68,16 +68,6 @@ warning[MatchingOutputMeta]: output `v` is missing from `meta.outputs` section i
    │
    = fix: add a description of output `v` to documentation in `meta.outputs`
```
claymcleod, line 70: Add a CLI-level regression test for a misplaced `wdl-lint` rule before removing this expectation, because no remaining test exercises the new cross-crate rule-metadata path. The analysis fixtures use only analysis rule IDs, while the lint harness constructs `DiagnosticsConfig::except_all()`, so regressions in `Linter::rules()` or `Validator::extend_rules()` would currently pass.

crates/wdl-analysis/src/validation/exceptions.rs
```diff
@@ -0,0 +1,198 @@
...
+                    && let Some(excepted_element) = &excepted_element
+                    && !exceptable_nodes.contains(&excepted_element.kind())
+                {
+                    let diagnostic = misplaced_except_directive(
+                        &rule.name,
+                        Span::new(
+                            start + comment.text().find(&rule.name).unwrap(),
+                            rule.name.len(),
+                        ),
+                        excepted_element,
```
claymcleod, line 162: Use `rule.span` for this label instead of searching the full comment, because a rule ID can be a substring of an earlier ID. With `#@ except: DeprecatedRuntimeSection, RuntimeSection`, this search labels the `RuntimeSection` suffix inside the first ID; that span does not match the second exception, so the validator emits both `ExceptDirectiveValid` and `MeaninglessLintDirective` for `RuntimeSection`.

crates/wdl-analysis/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `document()`, and source accessors, while `Document::imported_task_by_name`
   and `Document::imported_workflow_by_name` are now public
   ([#999](https://github.com/stjude-rust-labs/sprocket/pull/999)).
+* `ExceptDirectiveValid` rule, which ensures that `#@ except` comments are placed in valid locations ([#1125](https://github.com/stjude-rust-labs/sprocket/pull/1125)).
```
claymcleod, line 16: Document the breaking `Visitor::known_rules` → `Visitor::rules` and `Validator::extend_known_rules` → `Validator::extend_rules` API changes under `Changed`. Both signatures now carry exceptable-node metadata, so downstream `Visitor` implementations and validator embedders must migrate; this entry records only the promoted rule.

crates/wdl-analysis/src/visitor.rs
```diff
@@ -89,8 +89,8 @@ pub trait Visitor {
     /// Note that [`Validator`]s will expect this value to be static.
     ///
     /// [`Validator`]: crate::Validator
-    fn known_rules(&self) -> HashSet<String> {
-        HashSet::new()
+    fn rules(&self) -> HashMap<String, Option<&'static [SyntaxKind]>> {
```
claymcleod: Document this map’s contract: each key is a rule ID, `Some` lists the nodes where it can be excepted, and `None` means every node is valid. This is now a public extension point, and `Exceptions::comment` treats `None` as “skip placement validation,” so downstream `Visitor` implementations need the distinction to return correct metadata.

crates/wdl-lint/CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated `RequirementsSection` to only check for missing `requirements` sections ([#1040](https://github.com/stjude-rust-labs/sprocket/pull/1040)).
 
+#### Removed
+
+* Removed the `ExceptDirectiveValid` rule ([#1125](https://github.com/stjude-rust-labs/sprocket/pull/1125)).
```
claymcleod: Record the public `RULE_MAP` migration here instead of describing only the rule removal. `wdl_lint::rules::RULE_MAP` moved to `wdl_lint::RULE_MAP` and its key type changed from `&'static str` to `String`, so downstream users of that public static need a migration note; the wording should also say that `ExceptDirectiveValid` was promoted to `wdl-analysis`, not deleted outright.

crates/wdl-analysis/src/rules.rs
```diff
@@ -823,3 +836,79 @@ workflow example {
         self.0
     }
 }
+
+/// Detects unknown rules within lint directives.
```
claymcleod:
```suggestion
/// Detects improperly placed `except` directives.
```

claymcleod (changes requested):

crates/wdl-lint/CHANGELOG.md
Serial-ATA: IMO I don't think these things need points in the changelog. I doubt `RULE_MAP` being public was intentional in the first place. It wasn't being used anywhere outside of the rule's definition.

crates/wdl-analysis/src/rules.rs
```diff
@@ -823,3 +836,79 @@ workflow example {
...
+        "When writing WDL, `except` directives are used to suppress certain rules. If an `except` \
+         directive is misplaced, it will have no effect. This rule flags misplaced `except` \
+         directives to ensure they are in the correct location."
+    }
+
+    fn examples(&self) -> &'static [Example] {
+        &[Example {
+            negative: LabeledSnippet {
+                label: None,
+                snippet: r#"version 1.2
```
claymcleod: Should this be WDL 1.3

crates/wdl-lint/CHANGELOG.md
claymcleod: Okay, that's fine, mainly because this is a command line project primarily. Strictly speaking, we should add this kind of thing, but I get where you're coming from.

crates/wdl-analysis/CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * An import must now share the importing document's major version and have a
   minor version no greater than it; importing a newer minor version is rejected
   ([#999](https://github.com/stjude-rust-labs/sprocket/pull/999)).
+* `Vistor::known_rules()` and `Validator::extend_known_rules()` were renamed to `Visitor::rules()` and
```
claymcleod, line 23: Visitor.

claymcleod (approved):

PR #1130, opened 2026-08-19
feat(wdl-lsp): add `test` `textDocument/codeLens` support

Serial-ATA:
This adds `textDocument/codeLens` support for `dev test` YAMLs, making it possible to run individual tests or all tests under a target:

image

This has pending PRs for `sprocket-vscode` and `sprocket-intellij`. `sprocket-zed` unfortunately can't support them right now: https://github.com/stjude-rust-labs/sprocket-zed/issues/2

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lsp/src/test.rs
```diff
@@ -0,0 +1,215 @@
...
+
+impl SprocketTestCache {
+    /// Add a Sprocket test YAML file to the cache.
+    pub async fn open(&self, uri: Url) -> Result<Arc<SprocketTestYaml>> {
+        let Ok(path) = uri.to_file_path() else {
+            // `Analyzer` only supports `file://` URIs anyway.
+            bail!("unsupported uri: {uri}");
+        };
+
+        let content = tokio::fs::read_to_string(&path).await?;
```
claymcleod, line 56: Let’s seed this cache from `DidOpenTextDocumentParams.text_document.text` instead of rereading the URI; when the editor buffer differs from disk or the file has not been saved yet, the server computes stale lenses or caches nothing. Pass the document text into `open` and build `source` and `lines` from that authoritative buffer, matching the existing WDL `did_open` path.

crates/wdl-lsp/src/test.rs
```diff
@@ -0,0 +1,215 @@
...
+                    if start_offset <= end_offset && end_offset <= test_yaml.source.len() {
+                        test_yaml
+                            .source
+                            .replace_range(start_offset..end_offset, &change.text);
+                    }
+                }
+            }
+        }
+
+        test_yaml.lines = LineIndex::new(&test_yaml.source);
```
claymcleod, line 128: Please apply each content change against an up-to-date `LineIndex` and reject invalid or non-character-boundary offsets; when one `didChange` contains sequential edits, later ranges refer to the text produced by earlier edits, but this loop keeps using the original index. A valid Unicode edit can therefore make `replace_range` panic, while the `unwrap_or(0)` and `unwrap_or(source.len())` fallbacks can silently corrupt the cached document. Reuse the existing `SourceEdit` validation path or provide the same checks here.

crates/wdl-lsp/src/handlers/code_lens.rs
```diff
@@ -0,0 +1,124 @@
...
+
+/// Get the WDL file path associated with a test definition YAML file.
+///
+/// A test YAML file *must* have an associated WDL file, otherwise we don't
+/// consider it valid.
+///
+/// See [`is_sprocket_test_file()`](crate::test::is_sprocket_test_file)
+fn associated_wdl_file_path(path: &std::path::Path) -> Option<std::path::PathBuf> {
+    let base_name = path.file_stem()?;
+    let expected_wdl = std::path::Path::new(base_name).with_extension("wdl");
```
claymcleod, line 24: Consider deriving the WDL name from the complete YAML filename; when the basename contains an interior dot, `file_stem()` followed by `with_extension("wdl")` drops part of the name and disagrees with the CLI’s `find_yaml` mapping. For example, `align.bwa.yaml` currently maps to `align.wdl` instead of `align.bwa.wdl`.

```suggestion
    let expected_wdl = std::path::Path::new(path.file_name()?).with_extension("wdl");
```

crates/wdl-lsp/src/handlers/code_lens.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let in_test_dir =
+        parent.is_dir() && parent.file_name().and_then(|s| s.to_str()) == Some("test");
+    let wdl_dir = if in_test_dir {
+        parent.parent()?
+    } else {
+        parent
+    };
+
+    Some(wdl_dir.join(expected_wdl))
```
claymcleod, line 35: I think this resolver should return `None` unless the matching WDL file exists; when `test/` contains `foo.yml` and its parent contains an unrelated `bar.wdl`, `is_sprocket_test_dir` accepts the YAML and this handler emits commands targeting nonexistent `foo.wdl`. Check the derived path with `is_file()`, and preferably share one resolver with the CLI so the sibling and nested layout rules cannot drift.

crates/wdl-lsp/src/server.rs
```diff
@@ -1418,7 +1501,16 @@ impl<S: 'static> Server<S> {
 
     /// `textDocument/didChange` notification handler.
     async fn did_change(mut params: DidChangeTextDocumentParams, state: &ServerState<S>) {
-        normalize_uri_path(&mut params.text_document.uri);
+        if is_sprocket_test_file(&params.text_document.uri) {
```
claymcleod, line 1504: It might be worth routing changes by cache membership established during `didOpen` instead of reclassifying the URI from disk; when a companion WDL is created, removed, or renamed while the YAML is open, this predicate can disagree with the original routing decision and silently drop edits. Cache membership also avoids the synchronous `read_dir` and file-stat work that currently runs for every keystroke.

crates/wdl-lsp/src/server.rs
```diff
@@ -648,6 +654,77 @@ enum Message {
...
+        /// The test entrypoint to run.
+        target: String,
+    },
+}
+
+impl Command {
+    /// Get the title of the command, shown to the user by the client.
+    fn title(&self) -> String {
+        match self {
+            Command::TestIndividual { filter, .. } => format!("Run test '{filter}'"),
```
claymcleod, line 681:
```suggestion
            Command::TestIndividual { filter, .. } => format!("Run test `{filter}`"),
```

crates/wdl-lsp/src/server.rs
```diff
@@ -648,6 +654,77 @@ enum Message {
...
+            Command::TestEntrypoint { target, .. } => format!("Run '{target}' tests"),
```
claymcleod, line 682:
```suggestion
            Command::TestEntrypoint { target, .. } => format!("Run `{target}` tests"),
```

crates/wdl-lsp/src/server.rs
```diff
@@ -1596,6 +1688,18 @@ impl<S: 'static> Server<S> {
     }
 }
 
+/// Converts a URI into a WDL file path.
+fn to_wdl_file_path(uri: &Url) -> Option<PathBuf> {
+    if let Ok(path) = uri.to_file_path()
+        && path.is_file()
```
claymcleod, line 1694: One option is to classify request URIs by their `.wdl` extension without requiring the backing file to exist; when an open WDL is deleted, renamed, or temporarily absent during a Git operation, the analyzer still has the client’s in-memory text, but this new check makes hover, completion, diagnostics, symbols, rename, formatting, and other features return empty results. Dropping `is_file()` also avoids a synchronous stat on every request.

crates/wdl-lsp/src/server.rs
```diff
@@ -1105,15 +1167,40 @@ impl<S: 'static> Server<S> {
...
+        if is_sprocket_test_file(&params.text_document.uri) {
+            let _ = tx.send(
+                handlers::code_lens::code_lens(params.text_document.uri, state)
+                    .await
+                    .map_err(|e| ResponseError::new(ErrorCode::INTERNAL_ERROR, e)),
+            );
+            return;
+        }
+
+        let result = state
```
claymcleod, line 1188: Return `Ok(None)` before calling the WDL analyzer when this URI is neither a cached test YAML nor a WDL document; now that the server advertises code-lens support to clients that open YAML, an unrelated YAML request reaches `Analyzer::code_lens` and becomes an `INTERNAL_ERROR` such as `document not found in graph`. Gate this fallback by the WDL URI classification, as the other document handlers do.

crates/wdl-lsp/tests/code_lens.rs
```diff
@@ -54,6 +56,219 @@ fn assert_code_lenses(response: Vec<CodeLens>, mut expected: Vec<CodeLens>) {
     );
 }
 
+#[tokio::test]
```
claymcleod, line 59: Let’s add lifecycle coverage for the new cache before merging; the current tests only send `didOpen` with disk-identical ASCII text, so they cannot catch stale `didOpen` content, incremental-edit corruption, UTF-16 range mistakes, or `didClose` eviction. Add a test that opens buffer text differing from disk, applies sequential ranged changes including a non-ASCII character, verifies the updated lens, and then closes the document.

crates/wdl-lsp/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* Added support for the `textDocument/codeLens` request on `sprocket test` YAML files
+  ([#1129](https://github.com/stjude-rust-labs/sprocket/pull/1129)).
```
claymcleod, line 13:
```suggestion
  ([#1130](https://github.com/stjude-rust-labs/sprocket/pull/1130)).
```

claymcleod (changes requested):

crates/wdl-lsp/src/server.rs
Serial-ATA: I chose quote intentionally, I think it looks better in the editor. See the screenshot above

crates/wdl-lsp/src/server.rs
Serial-ATA: ditto

crates/wdl-lsp/src/handlers/code_lens.rs
Serial-ATA: > preferably share one resolver with the CLI so the sibling and nested layout rules cannot drift

The `dev test` command tries to find the associated YAML for a WDL file (it also has the advantage of getting its paths from `Analyzer`). We do the inverse here, so there's nothing to share

crates/wdl-lsp/src/server.rs
```diff
@@ -1105,15 +1167,44 @@ impl<S: 'static> Server<S> {
...
+                handlers::code_lens::code_lens(params.text_document.uri, state)
+                    .await
+                    .map_err(|e| ResponseError::new(ErrorCode::INTERNAL_ERROR, e)),
+            );
+            return;
+        }
+
+        if to_wdl_file_path(&params.text_document.uri).is_none() {
+            return; // Not a file we care about
+        }
```
claymcleod, line 1193: Can we send an empty response here before returning? For an unrelated YAML or another non-WDL document, dropping `tx` makes `Server::request` return `INTERNAL_ERROR` with `Background task dropped the request without replying` instead of `null`.

```suggestion
        if to_wdl_file_path(&params.text_document.uri).is_none() {
            // Not a file we care about
            let _ = tx.send(Ok(None));
            return;
        }
```

crates/wdl-lsp/src/test.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    let Some(parent) = path.parent() else {
+        return false;
+    };
+
+    if is_sprocket_test_dir(parent) {
+        return true;
+    }
+
+    let Some(base_name) = path.file_stem() else {
```
claymcleod: We should use the complete YAML filename here so this classifier matches `associated_wdl_file_path`. With `file_stem()`, `align.bwa.yaml` is checked against `align.wdl`, while the handler now correctly derives `align.bwa.wdl`, so dotted sibling test files are still never cached or given lenses.

```suggestion
    let Some(base_name) = path.file_name() else {
```

crates/wdl-lsp/src/test.rs
```diff
@@ -0,0 +1,184 @@
...
+            return Ok(());
+        };
+
+        let test_yaml = Arc::make_mut(entry.get_mut());
+        if change.start.is_some() {
+            let (new_source, new_lines) = change.apply()?;
+            test_yaml.source = new_source;
+            test_yaml.lines = new_lines;
+        } else {
+            change.apply_to(&mut test_yaml.source, &mut test_yaml.lines)?;
```
claymcleod: I think we need to reject reversed edit ranges before using this shared path for YAML. `SourceEdit::apply` checks UTF-8 boundaries but not `start <= end`, so a malformed `didChange` with two valid positions in reverse order reaches `String::replace_range` and panics the LSP message loop. Can we validate the computed range ordering in `SourceEdit::apply` and return an error instead?

crates/wdl-lsp/src/test.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        test_yaml.document = None;
```
claymcleod, line 96: We should apply the whole edit batch to temporary values and only committing it after every edit succeeds. If a later ranged edit errors, the earlier edits have already mutated `source`, the final `LineIndex` rebuild is skipped, and `document` still contains the old parse, so later lenses can use stale commands and mismatched ranges.

```suggestion
        let test_yaml = Arc::make_mut(entry.get_mut());
        let (new_source, new_lines) = if change.start.is_some() {
            change.apply()?
        } else {
            let mut source = test_yaml.source.clone();
            let mut lines = test_yaml.lines.clone();
            change.apply_to(&mut source, &mut lines)?;
            (source, lines)
        };

        test_yaml.source = new_source;
        test_yaml.lines = new_lines;
        test_yaml.document = None;
```

crates/wdl-lsp/tests/requests/code_lens.rs
```diff
@@ -0,0 +1,498 @@
...
+    ctx.server
+        .notify::<DidChangeTextDocument>(DidChangeTextDocumentParams {
+            text_document: VersionedTextDocumentIdentifier {
+                uri: ctx.doc_uri("example.yaml"),
+                version: 1,
+            },
+            content_changes: vec![TextDocumentContentChangeEvent {
+                range: Some(test_single_range),
+                range_length: None,
+                text: "switching_up_again〰".to_string(),
```
claymcleod: Use a non-BMP character such as `🚀` here and assert its two-code-unit range. `〰` is multibyte in UTF-8 but still one UTF-16 code unit, so the current test does not cover surrogate-pair positions where LSP character offsets commonly break.

claymcleod (changes requested):

crates/wdl-lsp/src/test.rs
Serial-ATA: I made the constructor for `SourceEdit`s fallible and now we log the error. Though a panic in that situation would be ideal. It's not our job to work around broken LSP clients (nor can we really do anything, `didChange` is a notification).

Serial-ATA:
@claymcleod are these label assignments automated? this isn't waiting for revisions

claymcleod:
Yes, these were automated. Thank you!

crates/wdl-lsp/src/handlers/code_lens.rs
```diff
@@ -0,0 +1,129 @@
...
+///
+/// See [`is_sprocket_test_file()`](crate::test::is_sprocket_test_file)
+fn associated_wdl_file_path(path: &std::path::Path) -> Option<std::path::PathBuf> {
+    let base_name = path.file_name()?;
+    let expected_wdl = std::path::Path::new(base_name).with_extension("wdl");
+    let parent = path.parent()?;
+
+    let in_test_dir =
+        parent.is_dir() && parent.file_name().and_then(|s| s.to_str()) == Some("test");
+    let wdl_dir = if in_test_dir {
```
claymcleod: Could we check for a sibling WDL before assuming that a directory named `test` contains nested test definitions? With `project/test/foo.wdl` and `project/test/foo.yaml`, the CLI correctly finds the sibling YAML, but this resolver looks for `project/foo.wdl`. That produces no lenses—or commands targeting the wrong WDL if that parent file exists. Please cover this sibling layout in a regression test and use `is_file()` when checking candidates.

crates/wdl-lsp/tests/requests/code_lens.rs
```diff
@@ -0,0 +1,498 @@
...
+
+    assert_code_lenses(response, expected);
+
+    ctx.server
+        .notify::<DidChangeTextDocument>(DidChangeTextDocumentParams {
+            text_document: VersionedTextDocumentIdentifier {
+                uri: ctx.doc_uri("example.yaml"),
+                version: 1,
+            },
+            content_changes: vec![TextDocumentContentChangeEvent {
```
claymcleod, line 159: Thanks for adding the buffer-content, emoji, and close coverage. Could we also exercise multiple sequential ranged edits in a single `didChange` notification, with the second range depending on the first edit’s updated text? The current test sends only one edit, so it doesn’t verify the original stale-line-index scenario. A separate case where a later edit fails should assert that the entire batch leaves the previous content and lenses intact.

Serial-ATA: I have more tests coming in my next PR adding `textDocument/diagnostic` support

claymcleod (approved):

PR #1143, opened 2026-08-24
Inline array literals within `max_line_length`

### LINKED ISSUE #1142: Inline array literals within `max_line_length`

`wdl-format` emits every array literal one element per line, regardless of size, with no consideration of width or arity. Trivially short arrays therefore cost three or four lines. From the `seaseq-case` test fixture:

```wdl
Array[String] string_sra = [
    1,
]  #buffer to allow for sra_id optionality
Array[String] s_sraid = select_first([
    sample_sraid,
    string_sra,
])
```

Hand-written WDL tends to go the other way: short arrays inline, real lists exploded. Across 307 production WDL files (the parseable WDL in [broadinstitute/warp](https://github.com/broadinstitute/warp) `master` and [broadinstitute/gatk](https://github.com/broadinstitute/gatk) `ah_var_store`), 1,150 of the 1,301 array literals (88.4%) are written on one line by their authors, and the ones that aren't are genuine multi-item lists.

An array literal is emitted inline (`[a, b]`) when the formatted line fits within `max_line_length`, and keeps today's one-element-per-line explosion otherwise. Inline arrays omit the trailing comma even when `trailing_commas` is enabled since the purpose of a trailing comma is multi-line diff hygiene.

Scope is array literals only. Map, object, and struct literals stay unconditionally vertical, and arrays containing them or containing comments continue to explode.

A width budget for the array alone, e.g. inline only if it renders narrower than half of `max_line_length`. This is certainly straightforward, and it's what I looked at first. But unfortunately for the corpora cited above the array's own width proved to be a poor predictor of whether the line fits: of the 1,154 array literals that would be eligible for writing on one line, the median already has **38 columns** of line in front of it out of a 90-column budget. Formatting the corpus under a half-budget rule inlines 111 arrays that a true fit test doesn't, and dozens of lines end up over the width budget as a result. It isn't more generous in exchange, either: it inlines 656 arrays to the fit test's 709.

An arity rule (e.g. two elements inline regardless of width). This is also simple but
conflicts with the guarantee that no emitted line exceeds `max_line_length`, and there are many
examples in the corpora where this would inline arrays inappropriately.

**Additional context**

mcovarr:
Closes #1142.

`format_literal_array` unconditionally emitted a hard line break after every element, so even a one or two element array would break across three or four lines:

```wdl
Array[String] string_sra = [
    1,
]  #buffer to allow for sra_id optionality
reference = select_first([
    spikein_reference,
    string_spikein,
]),
```

This PR inlines an array literal when the finished line fits within `max_line_length`, and leaves everything else exactly as it is today. Arrays formatted inline drop the trailing comma even when `trailing_commas` is enabled, since the purpose of a trailing comma is multi-line diff hygiene.

## The rule

```rust
let multiline =
    contains_element_requiring_multiline(element) || overflows_line(element, stream, config);
```

The baseline code in `main` contains no such condition and explodes all nonempty arrays. `contains_element_requiring_multiline` retains the current behavior for arrays that can't be inlined regardless of width: a direct child that is a comment, or a map, object, or struct literal, each of which should be written vertically. `overflows_line` is the fit test:

```rust
let (column, open) = line_position(stream, config);
let (width, wraps) = first_line_width(element, config);

wraps || column + width + array_suffix_width(element, open, config) > max
```

This performs a true fit test with three terms: where the line already is, how wide the array renders flat, and what still has to be written after the array close bracket before the line ends.

## Corpus results

Checked against the parseable WDLs in [broadinstitute/warp](https://github.com/broadinstitute/warp) (`master`) and [broadinstitute/gatk](https://github.com/broadinstitute/gatk) (`ah_var_store`): Every file was formatted with the code from both `main` and this branch.

|                       |   main |    this PR |
|-----------------------|-------:|-----------:|
| arrays inlined        |     26 |    **709** |
| arrays exploded       |  1,275 |        592 |
| lines over 90 columns |  5,207 |      5,207 |

The 26 inlined arrays on `main` are the 26 empty `[]` array literals; there is no inline path for anything else. 683 arrays collapse onto one line, for 1,746 fewer lines (−2.0%), and no lines anywhere in the corpus newly exceed the line length limit. For reference, the authors of these corpora wrote 1,150 of the 1,301 arrays (88%) on one line themselves.

Formatting the corpora costs a little more: parsing and formatting all 307 files in memory runs to a median of 0.162s on main and 0.170s on this branch over nine runs each after a warm-up, about 5%, or 8ms across the corpus. `array_suffix_width` renders siblings, but only the ones sharing the array's line. 
## Scope

Array literals only. Map, object, and struct literals remain unconditionally vertical, and arrays containing any of those elements or containing comments continue to explode, via `contains_element_requiring_multiline`.

## Tests

A new `literal-array-inlining` fixture covers the inlining logic directly. Its input is written inverted: arrays that should collapse start exploded and arrays that must stay exploded start inline, so every case has to flip. It pins the boundary (an array ending at exactly column 90 inlines; the same array one character wider explodes; a byte-identical copy four columns deeper explodes), the suffix cases (a trailing comment that fits, and one that doesn't), and the trivia cases (a comment on a comma, above an element, and in a block above the declaration).

---

For external contributors:

For all contributors:


a-frantz:
Hi @mcovarr !

I have started a fix for #278 that addresses "fit or split" logic, and my incoming PR will handle literal arrays as well as a variety of other node types (closing #666 #721 and the new #1142 ). We appreciate the effort that went into this, but I am going to close your PR as a different implementation is going to land soon.

For any future work on the Sprocket project, I would encourage you to communicate with the maintainers on any planned work before setting out with an implementation; I could have let you know a fix for this issue was in progress.

If you'd like, I can ping you when I open that PR for review so that you can give it a test drive and see if it addresses your issue :)

mcovarr:
HI @a-frantz, thanks for letting me know about the forthcoming PR, that sounds great! I'd definitely like to be pinged when that's out for review, and I may have some comments on those other issues as well.

PR #1145, opened 2026-08-26
fix: clean up Docker task work directories after cancellation

### LINKED ISSUE #1020: wdl-engine: leftover root-owned files from cancelled tests
**Describe the bug**
If you cancel a docker test in `wdl-engine` at the right time, you can end up with leftover temp files that are still owned by root.

They show up when trying to run `cargo clean`:

```shell
$ cargo clean
error: failed to remove file `/home/alex/Projects/Contributions/sprocket/target/tmp/.tmpiQK9sx/attempts/0/work/testdir/hello.txt`

Caused by:
  Permission denied (os error 13)
```

Since the cleanup task only runs if the task completes: https://github.com/stjude-rust-labs/sprocket/blob/5b79a1a1e4eeb36c77dda787221d967673bb08de/crates/wdl-engine/src/backend/docker.rs#L678

**Expected behavior**
The files should be cleaned up, regardless if the task is cancelled.

 - OS: Arch Linux
 - Version: `wdl-engine 0.17.0`

claymcleod:
Closes #1020.

Canceling a Docker task left its work directory owned by `root`, which the user running evaluation cannot delete, because the `chown` cleanup ran only for a task that completed. It now runs whatever the outcome was, with a cancellation token of its own, awaited outside the task manager, and with no event sender. A second Ctrl-C also warns that terminating Sprocket leaves running containers behind.

For all contributors:

### DIFF
--- a/crates/wdl-engine/src/backend/docker.rs
+++ b/crates/wdl-engine/src/backend/docker.rs
@@ -2,6 +2,8 @@
 
 use std::fs;
 use std::path::Path;
+#[cfg(unix)]
+use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::Mutex;
 
@@ -26,6 +28,8 @@ use crankshaft::engine::task::output::Type as OutputType;
 use futures::FutureExt;
 use futures::future::BoxFuture;
 use nonempty::NonEmpty;
+#[cfg(unix)]
+use tokio_util::sync::CancellationToken;
 use tracing::debug;
 use tracing::info;
 use tracing::warn;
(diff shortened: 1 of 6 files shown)


claymcleod:
Note to self, this becomes trivial after Peter's recent changes to Crankshaft where we can simply provide `None` for `events` when we don't want them.

crates/wdl-engine/src/backend/docker.rs
```diff
@@ -493,6 +497,13 @@ pub struct DockerBackend {
     max_memory: u64,
     /// The task manager for the backend.
     manager: TaskManager,
+    /// The token that cancels work directory cleanup.
+    ///
+    /// Cleanup deliberately does not observe the evaluation's cancellation, as
+    /// a canceled task is exactly when its work directory needs handing back.
+    /// This token is canceled only by an explicit request to stop cleaning up.
```
peterhuene: As this token doesn't ever appear to be canceled, where would the "explicit request to stop" originate from?

It seems more accurate to say that we're using a single `CancellationToken` for every cleanup task and that token is never canceled.

The only way to stop waiting for the task is to drop the future.

crates/wdl-engine/src/backend/docker.rs
```diff
@@ -338,40 +342,40 @@ impl ManagedTask for DockerTask<'_> {
...
-/// Represents a cleanup task that is run upon successful completion of a Docker
-/// task.
+/// Represents a cleanup task that is run after a Docker task, whatever the
+/// outcome of that task was.
 ///
 /// On Unix systems, this is used to recursively run `chown` on the work
 /// directory so that files created by a container user (e.g. `root`) are
 /// changed to be owned by the user performing evaluation.
 #[cfg(unix)]
 struct CleanupTask<'a> {
```
peterhuene: Given that this type no longer needs to impl `ManagedTask`, perhaps `run` can just be a free function?

src/commands.rs
```diff
@@ -30,6 +32,30 @@ pub mod submit;
...
+/// Warns that terminating Sprocket leaves Docker containers, and the files
+/// those containers created, behind.
+///
+/// Call this only when [`uses_docker_backend`] holds and Sprocket is about to
+/// exit without waiting for executing tasks to cancel.
+pub fn warn_docker_termination() {
+    warn!(
+        "terminating Sprocket does not remove Docker containers that are still running: files \
+         those containers created may be owned by another user (e.g. `root`) and require elevated \
+         privileges to remove"
```
peterhuene: Nit (suggestion):

```suggestion
        "terminating Sprocket does not remove Docker containers that are still running; files \
         that were created by containers may remain owned by another user (e.g. `root`) \
         and require elevated privileges to remove"
```

peterhuene (commented):
Just some minor feedback, but otherwise looks good 👍

crates/wdl-engine/src/backend/docker.rs
claymcleod: > It seems more accurate to say that we're using a single `CancellationToken` for every cleanup task and that token is never canceled.

Dropped the field instead, since nothing could ever cancel it. `chown_work_dir` now constructs the token it hands to `Backend::run`, and the comment there says nothing cancels it and that dropping the future is the only way to stop waiting.

crates/wdl-engine/src/backend/docker.rs
claymcleod: Done.

peterhuene (approved):

PR #1148, opened 2026-08-26
refactor: share backend and transferer for evaluations.

peterhuene:
This PR introduces the `Engine` type to `wdl-engine`.

An `Engine` keeps the backend, file transferer, and call cache. Each
constructed `Evaluator` is associated with an `Engine`.

This allows concurrent evaluation of different WDL workflows/tasks, commonly
done via the `server` and `test` commands to share the backend resource
management, limits, and monitors.

The events and cancellation context are associated only at the `Evaluator`,
allowing concurrent evaluations to source their events and cancellation
independently.

This PR also contains several important fixes for enums in WDL:

* Fix the enum choice value cache to include the URI of the document as the enum index in the existing cache key is per-document. Additionally, type name reference types and values must store the name by which they were referenced as they might be for imported types that were aliased. These fixes together prevent evaluation from reusing the wrong enum value type from the cache.
* Aliasing an imported enum caused a "not a struct" diagnostic from analysis.
* Static analysis wasn't properly checking for unknown enum choice, which previously was only being caught at runtime.

The `crankshaft` dependency has been updated to latest.

For external contributors:

For all contributors:

      is implemented.


crates/wdl-analysis/src/types.rs
```diff
@@ -1499,6 +1505,68 @@ impl PartialEq for CallType {
...
+struct TypeNameRefTypeInner {
+    /// The name used to refer to the type.
+    name: String,
+    /// The custom type that was referred to.
+    ty: CustomType,
+}
+
+/// Represents a reference to a custom type (struct or enum).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TypeNameRefType(Arc<TypeNameRefTypeInner>);
```
peterhuene: Nit: possibly rename to `TypeNameRef`? The extra `Type` suffix feels redundant.

crates/wdl-engine/src/backend/docker.rs
```diff
@@ -324,10 +269,13 @@ impl<'a> DockerTask<'a> {
         }
 
         let task = Task::builder()
-            .name(&self.name)
+            .name(self.request.name)
             .executions(NonEmpty::new(
                 Execution::builder()
-                    .images(candidate_images)?
+                    .images(
+                        filter_docker_sources(&self.request.constraints.sources)
```
Serial-ATA: btw: https://github.com/stjude-rust-labs/sprocket/pull/1117#discussion_r3823090974

peterhuene: Ah, so this filtering is entirely unnecessary now? I'll remove.

peterhuene: Oh I see what was being said.

I'll revert to what you had prior.

crates/wdl-engine/src/backend/manager.rs
```diff
@@ -312,7 +310,7 @@ impl TaskManager {
 
                 // If a cancellation has occurred, clear the parked tasks; otherwise, unpark
                 // what tasks we can
-                if limits.cancellation.state() != CancellationContextState::NotCanceled {
+                if cancellation.state() != CancellationContextState::NotCanceled {
```
claymcleod: Remove only the canceled request and keep scheduling the shared queue; when evaluators share this `TaskManager`, clearing `state.parked` drops every other run's sender and fails unrelated work. Those receivers then take the error path and credit CPU and memory they never reserved, so the shared limits can also exceed their configured totals.

crates/wdl-engine/src/engine.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::v1;
+
+/// The inner workings of [`Engine`].
+struct EngineInner {
+    /// The configuration for evaluation.
+    config: Arc<Config>,
+    /// The task execution backend for evaluation.
+    backend: Box<dyn TaskExecutionBackend>,
+    /// The transferer for evaluation.
+    transferer: Box<dyn Transferer>,
```
claymcleod, line 30: In a future PR, let's bound or expire the shared transfer result cache; in a long-lived server, each distinct URL and successful `TempPath` remains until shutdown, so repeated unique URLs cause memory and disk growth after evaluations finish.

crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -269,64 +276,48 @@ impl ApptainerRuntime {
...
         };
 
-        let pull = once.get_or_try_init(|| async move {
+        let pull_cancellation = cancellation.clone();
+        let result = once.get_or_try_init(|| async move {
             // SAFETY: the next two `unwrap` calls are safe because the source can't be a
             // file or an unknown source at this point
-            let mut path = self.cache_dir.join(container.scheme().unwrap());
-            let name = container.name().unwrap();
+            let mut path = self.cache_dir.join(image.scheme().unwrap());
```
claymcleod: Derive this shared cache key from an injective digest of the complete `ImageSource`, reject traversal components, and publish the SIF atomically; splitting both `/` and `:` maps distinct references to the same path and accepts `..`. Because `path.exists()` trusts any existing file, one run can poison or escape the persistent cache and a later run can execute the wrong image.

crates/wdl-engine/src/http.rs
```diff
@@ -260,8 +265,8 @@ impl Transferer for HttpTransferer {
                             self.0.client.clone(),
                             source,
                             &*temp_path,
-                            self.0.cancel.clone(),
-                            self.0.events.clone(),
+                            cancellation.first(),
```
claymcleod, line 268: Decouple each caller's cancellation and events from the shared `OnceCell` initializer; when evaluators request the same URL concurrently, only the initializing caller's token and sender are polled, so another run cannot cancel its wait and receives no transfer events for the transfer it depends on.

crates/wdl-engine/src/backend/apptainer.rs
```diff
@@ -269,64 +276,48 @@ impl ApptainerRuntime {
...
             let mut map = self.images.lock().unwrap();
-            map.entry(container.clone())
+            map.entry(image.clone())
                 .or_insert_with(|| Arc::new(OnceCell::new()))
                 .clone()
         };
 
-        let pull = once.get_or_try_init(|| async move {
+        let pull_cancellation = cancellation.clone();
+        let result = once.get_or_try_init(|| async move {
```
claymcleod, line 317: Apply cancellation and events per waiter rather than inside the shared image initializer; when two evaluators need the same image, only the initializing run controls and observes the pull, while a canceled waiter cannot return until that shared pull finishes.

crates/wdl-engine/src/engine.rs
```diff
@@ -0,0 +1,167 @@
...
+/// transferer, and various caches.
+///
+/// Typically there will be one WDL evaluation engine per process.
+///
+/// See the [`create_v1_evaluator`](Self::create_v1_evaluator) method for
+/// creating new [`Evaluator`](crate::v1::Evaluator) from the engine.
+///
+/// This type is cheaply cloned.
+///
+/// Note: as the engine holds a reference to the provided [`Events`], the engine
```
claymcleod: Remove or correct this note because `Engine` does not hold `Events`; only each `Evaluator` does. As written, the new public API documentation gives callers the wrong lifetime requirement.

claymcleod (approved):
I'm going to preemptively approve, as I think most things could go into future PRs. Up to you.

crates/wdl-engine/src/engine.rs
peterhuene: Yep, that's definitely part of my follow-up work to LRU the caches and keep them out of statics.

crates/wdl-engine/src/http.rs
peterhuene: This is a larger problem of various internal mutexes not supporting cancellation. I'll see if I can address it as part of the work to redo caches.

crates/wdl-engine/src/engine.rs
peterhuene: This comment belongs on the evaluator, not engine. I'll move it there.

crates/wdl-engine/src/backend/apptainer.rs
peterhuene: Same as above re: fixing this. The fix will be along the lines of selecting both the `OnceCell` future along with the passed cancellation token so that the caller can cancel the call.

I'll create a general "improve cancellation for operations for various caches" which I'll fix with the work I'm currently doing to improve long-lived caches for `sprocket server`.

crates/wdl-engine/src/engine.rs
peterhuene: Filed as https://github.com/stjude-rust-labs/sprocket/issues/1150.

crates/wdl-engine/src/http.rs
peterhuene: Filed as https://github.com/stjude-rust-labs/sprocket/issues/1150.

crates/wdl-engine/src/backend/apptainer.rs
peterhuene: Filed as https://github.com/stjude-rust-labs/sprocket/issues/1150.

PR #1151, opened 2026-08-26
fix: honor `sprocket run` output directory for database

### LINKED ISSUE #1149: sprocket.db doesn't honor specified output location
**Describe the bug**
When using `sprocket run -o /nondefaultOut` the sprocket.db still gets created in `./out`

**Expected behavior**
I expect the sprocket.db to be created in `/nondefaultOut`

 - OS: ubuntu 24
 - Toolchain version: sprocket v29 - installed with cargo and downloaded GitHub binary 

**Additional context**
Looking at src/commands/run.rs:941 , it looks like it grabs the db location from the server config, which might not be intended for `sprocket run`

claymcleod:
Closes #1149.

`sprocket run` selected run outputs from `-o` or `run.output_dir`, but it created the default database from the server output configuration. I derived the default database path from the effective run output directory while preserving explicit database URLs and existing server behavior.

For all contributors:


src/commands/run.rs
```diff
@@ -902,6 +902,12 @@ pub async fn run(
     }
 
     let report_mode = args.report_mode.unwrap_or(config.common.report_mode);
+    if let Some(output_dir) = &args.output_dir {
+        config.run.output_dir.clone_from(output_dir);
+    }
+    // The database URL uses the server output directory, so keep the local
+    // server configuration aligned with the effective run output directory.
+    config.server.output_dir.clone_from(&config.run.output_dir);
```
peterhuene: This is a little weird to me given they're separate settings in the configuration.

I'd expect them to be separately configurable, but both overridden by the CLI args.

src/commands/run.rs
```diff
@@ -1191,11 +1197,7 @@ async fn setup_run_context(
     inputs: &Inputs,
 ) -> Result<(RunContext, RunDirectory, Arc<dyn Database>, HeartbeatGuard)> {
     // Set up output directory structure
-    let output_dir = OutputDirectory::new(
-        args.output_dir
-            .clone()
-            .unwrap_or_else(|| config.run.output_dir.clone()),
-    );
+    let output_dir = OutputDirectory::new(config.run.output_dir.clone());
```
peterhuene, line 1197: This feels like it should be an argument rather than based on the configuration so that the `server` can use the server output directory and `run` can use the run output directory.

peterhuene: Oh never mind, we're not sharing this between `server` and `run`. This is just the path for `run`.

src/commands/run.rs
peterhuene: Never mind. Looking at the code closer I see that this is for `sprocket run` and forcing the server configuration to what was specified in `run` makes sense if that's the only thing we're creating the database path from.

peterhuene (approved):

peterhuene (approved):

PR #1154, opened 2026-08-27
feat(format): fit or split logic

### LINKED ISSUE #278: more sophisticated line breaking in `wdl-format`

### LINKED ISSUE #666: Configuration for call input formatting in `wdl-format`

### LINKED ISSUE #721: Allow short if/then/else expressions to remain on a single line

### LINKED ISSUE #1142: Inline array literals within `max_line_length`

a-frantz:

closes #278 #721 #1142 as completed

closes #666 as a won't do. See the issue for a brief discussion of the utility of this configuration option (tldr; not useful in real world WDL; although contrived test WDL may appear better if treated with this "fit or split" logic)

For external contributors:

For all contributors:

      is implemented.


crates/wdl-format/src/v1/import.rs
```diff
@@ -79,22 +79,30 @@ pub fn format_import_members(
...
         {
             (comma).write(stream, config);
             if items.peek().is_some() {
-                stream.end_line();
+                stream.potential_split();
+            } else {
+                trailing_comma_inserted = true;
             }
-        } else if config.trailing_commas {
-            stream.push_literal(",".into(), SyntaxKind::Comma);
```
adthrasher, line 92: Are we dropping this config option entirely?

crates/wdl-format/tests/format/trailing-commas-drop/source.formatted.wdl
```diff
@@ -46,7 +42,7 @@ workflow test {
         Person {
             name: "John",
             age: 26
-        }
+        },
```
adthrasher: This is related to the config change above.

crates/wdl-format/tests/format/symbolic_imports_comments/source.formatted.wdl
```diff
@@ -4,9 +4,12 @@ import {
     sort,  # explanation for sort
     cut,
     grep,  # explanation for grep
-} from openwdl/csvkit
+        } from openwdl/csvkit
```
adthrasher: This is a bug right? The extra indentation.

crates/wdl-format/src/v1/expr.rs
```diff
@@ -1001,7 +1009,10 @@ pub fn format_if_expr(
             }
             // only match on `else`; `then` could be considered for "chaining" but that
             // makes it harder to read IMO (a-frantz).
-            result = matches!(cur.kind(), SyntaxKind::ElseKeyword);
+            result = matches!(
+                cur.kind(),
+                SyntaxKind::ElseKeyword // | SyntaxKind::ThenKeyword
```
adthrasher: Should the commented out code be deleted?

crates/wdl-format/src/token/post.rs
```diff
@@ -523,21 +563,241 @@ impl Postprocessor {
...
+        *self = init_self;
+
+        break_stack.clear();
+        prev_kind = None;
+
+        pre_buffer = in_stream.iter().enumerate().peekable();
+        post_buffer.clear();
+
+        self.fit_potential_splits = false;
+        let mut fit_spans = fit_spans.iter();
```
adthrasher: I'm not sure we should retain the variable name when shadowing the prior `Vec`.

crates/wdl-format/src/token/post.rs
```diff
@@ -508,6 +511,43 @@ impl Postprocessor {
...
+                }
+            }
+            PreToken::FitOrSplitEnd {
+                fit_end,
+                split_end,
+                split_end_line,
+            } => {
+                if self.fit_potential_splits {
+                    stream.push(PostToken::Literal(fit_end));
+                } else {
```
adthrasher, line 546:
```suggestion
                 if self.fit_potential_splits {
                    if fit_end.is_empty() {
                        self.trim_last_line(stream);
                    }
                     stream.push(PostToken::Literal(fit_end));
                 } else {
                    if split_end.is_empty() {
                        self.trim_last_line(stream);
                    }
```

crates/wdl-format/src/token/post.rs
adthrasher, line 505: I think this call to `end_line` in the comment is why the extra indent happens.

crates/wdl-format/src/token/post.rs
a-frantz: 🙌  🙏 🙌 

this is the fix I've been missing!

crates/wdl-format/src/v1/import.rs
a-frantz: good catch, this was a mistake. Fixed in https://github.com/stjude-rust-labs/sprocket/pull/1154/commits/032c41a417d4e73aea720d970134f9ca37ebb8b3

crates/wdl-format/src/token/post.rs
a-frantz: I think that's relatively standard issue rust convention. But I updated it to `into_iter()` to more clearly indicate the `Vec` isn't needed after

crates/wdl-format/src/token/pre.rs
```diff
@@ -62,6 +62,33 @@ pub enum PreToken {
...
+    /// The start of a fit or split block.
+    FitOrSplitStart {
+        /// If the block will be "fit", insert this literal string at the
+        /// beginning.
+        fit_start: Rc<String>,
+        /// If the block will be "fit", insert this literal string between each
+        /// potential split.
+        fit_delimiter: Rc<String>,
+        /// If the block will be split, end the line after inserting
+        /// `fit_start`.
```
a-frantz:
```suggestion
        /// If the block will be split, end the line immediately.
```

crates/wdl-format/src/token/post.rs
```diff
@@ -523,21 +569,241 @@ impl Postprocessor {
...
         let mut post_buffer = TokenStream::<PostToken>::default();
 
-        while let Some(token) = pre_buffer.next() {
+        // If we encounter any potential splits, we use the first iteration to find any
+        // spans that will fit. If we don't find any potential splits, we can
+        // just push the result of the first iteration to the out stream and be
+        // done.
+        self.fit_potential_splits = true;
+        let mut fit_spans = Vec::new();
+        let mut fit_start = None;
```
adthrasher, line 588: It looks like this will format nested entries like the following:
```wdl
Array[Array[Int]] matrix = [[1, 2], [3, 4]]
```
as
```wdl
Array[Array[Int]] matrix = [
    [1, 2],
    [3, 4],
]
```

Is that what we want? If `fit_start` became a `Vec`, then evaluate each level of the nesting.

a-frantz: I played with that locally, but couldn't get it right and didn't include it here. Nested fit or split blocks get messy and there are improvements we could make to this algorithm.

The reason its not as simple as tracking with a stack is because of line `post.rs:L586`: `self.fit_potential_splits = true;`. As the outermost blocks are not linebroken this iteration, the width of the line extends and then the nested blocks end up tripping the `last_line_width > max` clause and do **not** get line broken.
Then in the second iteration, the outermost blocks **are** linebroken, but as the inner blocks didn't "fit" in the prior iteration they are _also_ linebroken. That looks "more incorrect" than only allowing the innermost blocks to be broken.

The workaround would involve changing how "does it fit?" is determined in regard to the max width (e.g. changing how "last line width" is calculated to account for nested fit or split blocks).

IMO this implementation is "good enough", but I can create a tracking issue pointing at this comment thread explanaining the future changes.

adthrasher: That's fine with me and I think a totally reasonable boundary for this PR.

adthrasher (approved):
With a follow-up issue for the one comment, LGTM. This is a nice improvement on the formatting of these elements.

crates/wdl-format/src/token/post.rs
a-frantz: #1191

PR #1163, opened 2026-09-01
fix: make standard library LSP prose version-aware

### LINKED ISSUE #1006: LSP stdlib hover prose not version-specific
**Describe the bug**

For standard library functions whose overloads span multiple WDL versions, LSP hover shows documentation prose that is not specific to the document's declared version. Even when the signature list is filtered to the overloads available at the document's version (see #1004), the accompanying prose still describes the newest version's behavior: parameters, return types, and examples that may not apply at the declared version.

In `crates/wdl-analysis/src/stdlib.rs`, each `FunctionSignature` has its own documentation field `definition`, but currently all signatures for a polymorphic function are initialized with a single shared `*_DEFINITION` constant. So regardless of which overload the hover logic chooses as the source of prose, the shared constant with prose for the latest WDL version is displayed.

For example, `BASENAME_DEFINITION` is shared by all three `basename` overloads. Its text describes the "basename of a file **or directory**" and includes a `version 1.2` example. In WDL 1.0, `basename` accepts only `File`/`String`, but hover in a `version 1.0` document still renders the 1.2-worded prose. Note that the `Directory` *type* does not even exist in WDL 1.0 as it was introduced in WDL 1.2.

7 of the 12 polymorphic stdlib functions have overloads spanning more than one WDL version, so that in an older document version the shared prose describes behavior from a newer overload. These are `basename`, `size`, `read_tsv`, `write_tsv` (each 1.0 → 1.2), `write_object`, `write_objects` (each 1.0 → 1.1), and `keys` (1.1 → 1.2). The remaining 5 (`min`, `max`, `join_paths`, `contains_key`, `length`) share a single minimum version across all overloads, so their prose is not mismatched in practice.

**Expected behavior**

Hover prose for a polymorphic function should describe the function as it behaves at the document's declared version: parameters, return types, and examples that match the overloads actually available there.

**Terminal Output**

N/A

**Screenshots**

This is tricky to visualize from screenshots until #1004 is addressed since version-inappropriate signatures are currently being returned in hover text in addition to the version-inappropriate prose described here. 

 - OS: macOS 26.5.2
 - Version: `sprocket v0.27`
 - Shell: zsh
 - Toolchain version: `rustc 1.96.0`

**Additional context**

N/A

claymcleod:
Standard-library hover filtered overloads by the document's WDL version but still described newer behavior. Signature help also exposed overloads introduced after the declared version.

I added documentation tiers for the seven affected polymorphic functions and selected the newest eligible tier consistently across hover, completion, and signature help. Signature help now filters overloads without disappearing during incremental edits, and the shared `size` overload description now says "a file system path" so it reads correctly in every version instead of naming the WDL 1.2-only `Directory` type.

Closes #1006.

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/CHANGELOG.md
+++ b/crates/wdl-analysis/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Fixed
+
+* Standard-library hover, completion, and signature help now use documentation
+  appropriate for the document's WDL version, and signature help no longer
+  shows overloads introduced in later versions
+  ([#1006](https://github.com/stjude-rust-labs/sprocket/issues/1006)).
+
 ## 0.25.0 - 2026-08-26
 
 #### Added
--- a/crates/wdl-analysis/src/handlers/completions.rs
+++ b/crates/wdl-analysis/src/handlers/completions.rs
@@ -792,16 +792,16 @@ fn add_stdlib_completions(version: Option<SupportedVersion>, items: &mut Vec<Com
                 })
             }
             Function::Polymorphic(p) => {
+                let docs = p.definition(v).and_then(|d| make_md_docs(d.to_string()));
                 for sig in p.signatures().iter().filter(|s| s.minimum_version() <= v) {
                     let params = TypeParameters::new(sig.type_parameters());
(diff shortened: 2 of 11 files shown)


claymcleod:
On second thought, I updated this to use the WDL 1.0 specification text directly for all six historical definition blocks, with only formatting, version markers, and typo corrections. Let me know if anyone has issues with that approach.

PR #1164, opened 2026-09-02
refactor: standardize CLI output with `CommandOutput`

### LINKED ISSUE #1133: Standardize remaining CLI output with `CommandOutput`
The `sprocket dev module` commands introduced `CommandOutput` and a consistent style for action lines, details, diagnostics, color, and prompts. The rest of the CLI still uses ad hoc output in several places.

Update the remaining human-facing CLI output to use the same presentation conventions and shared helper, extending it only where needed for existing output shapes.

- Preserve machine-readable and generated payloads as undecorated output.
- Preserve existing stdout and stderr contracts.
- Update CLI snapshots for deliberate presentation changes.

claymcleod:
Several commands still handled stream selection and presentation directly, which left human-facing output inconsistent with the module commands.

I routed command output through `CommandOutput`, preserved undecorated machine payloads and existing stdout/stderr contracts, and standardized action-oriented results for cancellation, baseline generation, format checks, and test summaries.

Closes #1133.

For external contributors:

For all contributors:

      is implemented.


src/lib.rs
```diff
@@ -190,18 +189,13 @@ async fn real_main() -> CommandResult<()> {
             commands::lock::lock(args, config, colorize).await
         }
```
a-frantz, line 190: is there a reason `validate`, `doc` and `lock` didn't get updated to use `Output`?

src/commands/test.rs
a-frantz, line 1: is there a reason the emojis were replaced with plain text? `test` can create a lot of text to parse through, the emojis make scanning that easier

a-frantz (commented):
This looks to me like its only a partial standardization, but the logic for what is left as it is and what is standardized is not clearly communicated. There are still commands untouched by this PR (`doc`, `validate`, `lock`) and some of the stream/color logic is still handled ad hoc through `if colorize {...} else {...}` blocks within the business logic of commands.

The PR description mentions "stdout/stderr contracts" which we do not have at this point in time. I'd prefer to break existing conventions in favor of consistent behavior and polish throughout Sprocket.

Can we explicitly log what needs to be standardized and what this PR completes? We don't need to handle every piece of it in one PR, but I would like to see a clear path for completing the standardization of all commands.

src/commands/test.rs
Serial-ATA: agreed, I think the original style (e.g., "✅ \`foo::bar::baz\`") is much easier to read. If you want to drop emojis, I think the `libtest` style (e.g., "foo::bar::baz ... ok") would be more readable

PR #1165, opened 2026-09-02
docs: define Sprocket 1.x compatibility policy

### LINKED ISSUE #1071: Define the Sprocket 1.x compatibility and deprecation policy
Sprocket cannot freeze a 1.0 surface until maintainers and users know which changes count as breaking. The CLI, `sprocket.toml`, module manifests, `/api/v1`, and machine-readable JSON currently evolve without one shared compatibility and deprecation policy.

Publish a Sprocket 1.x compatibility policy that defines the stable interfaces, additive and breaking changes, deprecation periods, experimental namespaces, documented exit-code and stream behavior, and the process for approving exceptions. Link the policy from contributor and release documentation.

Acceptance criteria:
- The policy covers CLI commands and options, `sprocket.toml`, `module.json`, `module-lock.json`, `/api/v1`, and documented JSON output.
- The policy distinguishes human-readable text from machine-readable output.
- The policy defines how WDL 1.4 and independently versioned Rust/Python APIs remain experimental.
- Maintainers approve the policy before public schemas freeze.

Relying on Semantic Versioning alone does not identify which surfaces Sprocket considers public. Freezing each subsystem independently would produce conflicting guarantees.

**Additional context**
This issue is the first gate for Sprocket 1.0 and blocks the `Surface Freeze` checkpoint.

claymcleod:
Sprocket 1.0 needs one contract that defines which executable interfaces users can rely on and how maintainers may evolve them. Semantic Versioning alone cannot describe those guarantees because the executable uses version numbers to mark product milestones rather than compatibility boundaries.

This publishes the normative 1.x policy, connects it to contributor and release workflows, and pins generated `sprocket.toml` schema links to each release tag. The coordinated user-facing summary is in stjude-rust-labs/sprocket.bio#56 and should merge after or with this PR so its policy link resolves.

Closes #1071.

For external contributors:

N/A—submitted by a maintainer.

For all contributors:

- [x] You have classified changes to stable interfaces under the [compatibility policy](https://github.com/stjude-rust-labs/sprocket/blob/main/COMPATIBILITY.md) (when appropriate).

N/A—this PR does not change lint rules.

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -1,0 +1,225 @@
+# Sprocket Compatibility Policy
+
+This document defines the compatibility guarantees for the `sprocket` command
+line tool beginning with Sprocket 1.0. It applies to every Sprocket 1.x release.
+Before 1.0, any interface may change without the notice required by this
+policy.
+
+## Versioning
+
+The `sprocket` executable uses `MAJOR.MINOR.PATCH` version numbers, but it does
+not use Semantic Versioning to communicate compatibility.
+
+- Patch releases contain bug fixes and compatible behavioral corrections.
+- Minor releases may add compatible features and remove features whose full
+  deprecation period has elapsed.
+- Major versions identify broader changes in Sprocket's functionality or
+  purpose. A major version does not necessarily contain a breaking change.
+
+This policy governs the 1.x release line. A later major release may publish a
+successor policy, but changing the major version does not by itself waive the
+review and notice requirements in this document.
+
+The `wdl` and `wdl-*` Rust crates and any separately released Python bindings
+follow [Semantic Versioning](https://semver.org/) according to their own package
+versions. Sprocket 1.0 does not stabilize those APIs. In particular, a pre-1.0
+package may make a SemVer-permitted breaking change in a minor release.
+
+The Rust library in the root `sprocket` package is an experimental
+implementation interface. It shares the executable's version and does not have
+a separate SemVer compatibility guarantee. Users should not depend on the
+`sprocket` crate as a library; the project neither recommends nor supports that
+use.
+
+## Stable Interfaces
+
+Beginning with Sprocket 1.0, the following interfaces are stable unless their
+documentation explicitly marks them experimental:
+
+| Interface | Stable contract | Documentation of record |
+| --- | --- | --- |
+| Command line | Non-`dev` commands, options, arguments, and documented behavior | `sprocket --help`, subcommand help, and [Sprocket documentation](https://sprocket.bio/) |
+| Process behavior | Documented exit statuses and placement of output on standard output or standard error | This policy, command help, and [Sprocket documentation](https://sprocket.bio/) |
+| Configuration | Documented `sprocket.toml` keys, value types, defaults, and meanings | The version of `jsonschemas/sprocket.toml.json` shipped with the release for structure and types, and [configuration documentation](https://sprocket.bio/configuration/overview.html) for behavior |
+| HTTP API | `/api/v1` paths, methods, parameters, request and response bodies, status codes, and documented meanings | The server's `/api/v1/openapi.json` document and [server documentation](https://sprocket.bio/subcommands/server.html) |
+| Machine-readable output | Documented JSON fields, types, meanings, exit status, and output stream, including documented generated JSON files | Command help and [Sprocket documentation](https://sprocket.bio/) |
+| Run artifacts | Documented `inputs.json` and `outputs.json` formats and documented `--index-on` behavior | [Provenance documentation](https://sprocket.bio/concepts/provenance.html) |
+
+An interface that is not documented is outside this contract until its
+documentation explicitly adds it. Internal Rust types and implementation
+details are not made stable merely because they are public in generated API
+documentation.
+
+### Exit Status and Stream Conventions
+
+Sprocket uses these command-wide exit statuses:
+
+| Status | Meaning |
+| --- | --- |
+| `0` | The command completed successfully. |
+| `1` | The command could not complete successfully, including configuration, validation, analysis, execution, or operational failures. |
+| `2` | The command line could not be parsed, or an argument value failed argument-level validation, such as when a source path does not exist. |
+
+Unless a command documents a different contract, its requested result or
+machine-readable payload goes to standard output. Diagnostics, warnings,
+progress, prompts, and logs go to standard error. Moving documented output to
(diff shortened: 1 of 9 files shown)


src/commands/config.rs
```diff
@@ -11,7 +11,11 @@ use crate::config::Config;
 /// The [Taplo schema directive] for `sprocket.toml`.
 ///
 /// [Taplo schema directive]: https://taplo.tamasfe.dev/configuration/directives.html#the-schema-directive
-const SCHEMA_DIRECTIVE: &str = "#:schema https://raw.githubusercontent.com/stjude-rust-labs/sprocket/refs/heads/main/jsonschemas/sprocket.toml.json";
+const SCHEMA_DIRECTIVE: &str = concat!(
```
Serial-ATA, line 14: Since the config is gonna be stable going forward, we should just move the schema to `jsonschemas/v1/sprocket.toml.json` rather than pinning to specific versions

Serial-ATA (changes requested):

src/commands/config.rs
claymcleod: I think, for this one, we probably still want to pin the version: though the config is stable, we can add new optional configuration items in future versions of Sprocket, and, without pinning the version, every version of sprocket will always suggest all keys (even the ones unsupported by that version).

Serial-ATA: Pinning the version here puts the burden of updates on the user. I think it's much more valuable to have it reflect the latest version (especially for doc updates), and if someone truly can't update their binary they can pin it manually.

> we can add new optional configuration items in future versions of Sprocket

I think 2 things can improve that experience:

1. We could add version numbers in the docs of the fields (e.g., `Added in 1.2.3`)
2. We can make the `sprocket.toml` parser forward compatible by warning on unknown fields instead of erroring. That'd make it clear that this version of `sprocket` doesn't support that field.

For example, `cargo` does this:

```toml
# Cargo.toml
[package]
foo = "bar"
```
```shell
$ cargo b
warning: Cargo.toml: unused manifest key: package.foo
```

While we just fail:

```toml
# sprocket.toml
foo = "bar"
```
```shell
$ sprocket check
error: failed to deserialize configuration file `sprocket.toml`
  ┌─ sprocket.toml:1:1
  │
1 │ foo = "bar"
  │ ^^^ unexpected key
```

COMPATIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+- Human-readable output details excluded under
+  [Exit Status and Stream Conventions](#exit-status-and-stream-conventions).
+
+The `sprocket analyzer` command's documented invocation is stable, but LSP
+capabilities and protocol behavior are stable only when Sprocket documentation
+explicitly describes them.
+
+A pull request that moves a command or feature out of `dev` must list every
+command, option, file format, schema, and behavior that becomes stable. The
+release notes must announce that transition.
```
peterhuene, line 109: Although `dev` commands are inherently unstable, the promotion of a `dev` command to stable would immediately be a breaking change to anyone that was helping us test the command in its early stages.

Perhaps we should consider that the promotion of a command to stable should deprecate the corresponding `dev` command (i.e. we alias the `dev` subcommand and its invocation causes a deprecation warning to remove the use of `dev` for running the command)?

peterhuene (approved):
This all looks good to me.

I just have the one comment.

COMPATIBILITY.md
claymcleod: Yeah good idea.

COMPATIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+- Commands, options, formats, and behavior under `sprocket dev`.
+- `module.json`, `module-lock.json`, module signatures, and related module
+  formats while module commands remain under `sprocket dev module`.
+- WDL test definitions and their schema while testing remains under
+  `sprocket dev test`.
+- WDL 1.4 support.
+- Configuration keys and features explicitly marked experimental.
+- Configuration keys that exclusively configure `sprocket dev` commands until
+  those commands graduate.
+- The internal run directory layout, including `runs/`, `_latest`, task attempt
```
a-frantz, line 91: I started commenting that I'd like to see `_latest` covered by a contract, as its useful for scripting, but while typing I realized that any scripting involving `_latest` would be better served through the provenance index and/or HTTP API.

Nothing to resolve or change here, but thought worth pointing out 👍

src/commands/config.rs
a-frantz: I like Alex's suggestion as it will function as a nudge towards keeping Sprocket binaries up to date with latest.

>  every version of sprocket will always suggest all keys (even the ones unsupported by that version).

I view this as a friendly reminder "there's new stuff!" that can always be ignored if the user has reason to remain pinned to something earlier.

COMPATIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+newer Sprocket release is not guaranteed to work with an older release.
+
+## Compatible and Breaking Changes
+
+Compatible changes preserve documented behavior for existing users. Examples
+include:
+
+- Adding an optional command-line option without changing existing defaults.
+- Accepting an input that an earlier release rejected.
+- Adding an optional `sprocket.toml` key.
```
a-frantz, line 138: all of our config keys are optional
```suggestion
- Adding a new `sprocket.toml` key, so long as its default value is consistent with existing behavior
```

a-frantz:
A few weeks ago we discussed how `sprocket format` should be handled under this contract. As this team often does, I find myself looking at Rust tooling to inform that decision: The St Jude Rust Labs projects tend to use the stable toolchain for all things, _except formatting_ where we use the experimental nightly toolchain. I understand this is common in Rust projects, where many teams prefer to use the "bleeding edge" formatting tools while relying on stable tooling for all else.

This suggests that including `format` output (and `format check` results) under a backwards compatibility contract is not ideal. Format output is not easily feature gated when the underlying algorithm for calculating line width/linebreaking changes. See the planned change #1191: `The workaround would involve changing how "does it fit?" is determined in regard to the max width (e.g. changing how "last line width" is calculated to account for nested fit or split blocks).`

The cost for "breaking" someone's `format --check` CI is about as minimal as it can be. Just re-run `format` and commit the automated change. I think including this under a contract is overkill and that formatting should have flexibility to improve the underlying algorithm without violating a contract.

If the rest of the team thinks that is wrong for whatever reason, I'd acquiesce. Anyone else have thoughts here?

PR #1178, opened 2026-09-04
refactor(wdl-engine): make various caches specific to evaluation.

peterhuene:
Currently in `wdl-engine` there are process-wide (or engine-wide) caches that should be per evaluation.

These caches include:

* The local digest cache.
* The remote digest cache.
* The caches for HTTP client responses.

Additionally, these caches were unbounded and would continue to grow over time, which is not ideal for `sprocket server` use.

Therefore, the caches have been updated to bounded LRU caches with configurable capacity limits.

The in-memory image cache for the Apptainer backends has been removed and now it sources from the file system always (the in-memory cache was only preventing a file metadata read operation to check if the image is in the cache).

For external contributors:

For all contributors:

      is implemented.


claymcleod (approved):
This looks great. No notes. Good job

crates/wdl-engine/src/lib.rs
```diff
@@ -124,3 +132,127 @@ impl From<ContentKind> for crankshaft::engine::task::input::Type {
...
+        Q: Hash + Eq + ?Sized + ToOwned<Owned = K>,
+        V: Clone,
+        F: AsyncFnOnce() -> Result<V, E>,
+    {
+        let value = {
+            let mut cache = self.0.lock().expect("failed to lock cache");
+            cache.get_or_insert_ref(key, Default::default).clone()
+        };
+
+        let token = cancellation.first();
```
claymcleod: Please preserve slow-cancellation completion here by allowing cache operations needed by running tasks to continue until the second cancellation signal. 

With `FailureMode::Slow`, the first signal leaves a TES task running, but after it succeeds, evaluate_output calls resolve_paths and the HTTP exists check reaches this cache with the first token already canceled. The biased select immediately returns `None`, even for a cached result, so output evaluation fails and `outputs.json` is not written. 

Previously, the remote existence check was not canceled this way. Consider passing the appropriate cancellation token into the cache helper so input transfers can still stop on the first signal while output validation and final digest calculation can finish.

claymcleod:
The task-output fix addresses the reported failure, but please consider whether these remaining post-execution paths should also use the second token:

- **Task private declarations after execution:** `add_backend_inputs` still uses the first token for `resolve_paths` and remote downloads. These declarations are part of completing a task that was allowed to keep running, so canceling them on the first signal can discard an otherwise successful result.
- **Workflow output and downstream declaration expressions:** `WorkflowEvaluationContext::http` always uses the first token. After slow cancellation lets active calls finish, HTTP-backed expressions such as `read_string` can therefore prevent the workflow from producing its final outputs.
- **Workflow output and downstream declaration path validation:** these paths pass the first token to `resolve_paths`. A remote `File` or `Directory` can therefore fail its `exists` check immediately after its producing call completes, reproducing the task-level issue at the workflow level.

Also, `first_cancel_completes_task` uses `read_string`, so it does not exercise the original remote `File`/`Directory` → `resolve_paths` → `exists` path. Could we add a remote `File` output to cover that case directly?

peterhuene:
> **Task private declarations after execution:** `add_backend_inputs` still uses the first token for `resolve_paths` and remote downloads. These declarations are part of completing a task that was allowed to keep running, so canceling them on the first signal can discard an otherwise successful result.

I'll fix this. In fact, this could generally be improved by just having a command section implicitly depend on every private declaration and it should in case they are "unused" private decls containing `File` or `Directory` values that might be needed for backend inputs.

Currently we only add the implicit dependency edge from the command to a private decl if it is declared with the `env` modifier. This restriction should be lifted and all private declarations should be implicitly depended upon by the command section and the corresponding logic for detecting unused private declarations updated.

By extension, it will mean that there won't be any more private declarations evaluated _after_ the task has executed.

>  **Workflow output and downstream declaration expressions:** `WorkflowEvaluationContext::http` always uses the first token. After slow cancellation lets active calls finish, HTTP-backed expressions such as `read_string` can therefore prevent the workflow from producing its final outputs.
> **Workflow output and downstream declaration path validation:** these paths pass the first token to `resolve_paths`. A remote `File` or `Directory` can therefore fail its `exists` check immediately after its producing call completes, reproducing the task-level issue at the workflow level.

I'll fix this. Originally I envisioned the "slow cancellation" to be specific to _task_ execution, where there is a specific "point of no return" in terms of a tasks starting to execute.

Having "slow cancellation" also apply to the output section of a workflow makes sense.

However, this is a bit tricker to implement vs. a task where this is a specific point of execution. We'll need to detect the first output being evaluated and treat that as a sign that the second cancellation should be used for further evaluation. 

> Also, `first_cancel_completes_task` uses `read_string`, so it does not exercise the original remote `File`/`Directory` → `resolve_paths` → `exists` path. Could we add a remote `File` output to cover that case directly?

I'll update the test.

crates/wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -1854,6 +1717,171 @@ impl State {
     }
 }
 
+impl Evaluator {
```
peterhuene, line 1720: FYI, this section was just moved down so that `impl State` was closer to its declaration.

claymcleod (approved):

Issue #557, opened 2026-01-23
Lint for mixed `meta` sections and doc comments


Serial-ATA:
**Is your feature request related to a problem? Please describe.**
`wdl-doc` (after #551) will have experimental support for doc comments. While the implementation supports mixing `meta`/`parameter_meta` sections with doc comments, there should be a lint to discourage doing so.

**Describe the solution you'd like**
A new `wdl-lint` rule that checks for the presence of doc comments with overlapping `meta`/`parameter_meta` sections. For example:

```wdl
struct Person {
    ## The name of the person.
    String name

    parameter_meta {
        name: "This description will not be rendered, as doc comments take precedence."
    }
}
```

Since doc comments take precedence, the diagnostic should suggest removing the `meta`/`parameter_meta` section.

**Additional context**
#224

kris70lesgo:
@a-frantz I’d be happy to pick this up if no one is actively implementing it.

From the linked docs pr, it looks like doc comments are intentionally allowed to coexist with meta/parameter_meta today, but the consensus is that we should discourage or regulate that via linting rather than documentation alone.
Before I start working, I want to confirm expectations around:
- whether this rule should be configurable from the start (e.g. defaulting to `Conflicting`), and
- how it should interact with existing lints like `MetaDescription`, `ParameterMetaMatched`, and `MetaSections`.

a-frantz:
@kris70lesgo you are asking good questions that I don't have an answer for yet! I just added a blocked label to this issue as I think we need to have some conversations internally about how we want to handle the adoption of doc comments. Once we make some big picture decisions about how we want to roll-out the new doc comments, we can unblock this issue!

Serial-ATA:
> I'm tempted to make a hard switch to encouraging all doc comments instead of trying to keep both around

After the meeting yesterday, it looks like doc comments are going to stay a Sprocket-specific feature. So we'll have to keep support for both. I guess that also makes this fall under `SprocketCompatibility`

> how it should interact with existing lints like MetaDescription, ParameterMetaMatched, and MetaSections.

Getting a node's doc comments was made a lot easier since this issue was made, so we can update these rules to be doc comment aware:

* `MetaDescription` - Also check the `task`/`workflow`/`struct` for top-level doc comments
* `ParameterMetaMatched` - Also check if the decl has doc comments
* `MetaSections`:
	* No top-level doc comment - missing `meta` section
	* No doc comments on *any* input - missing `parameter_meta` section
[...]

PR #1184, opened 2026-09-07
feat(wdl-lint): support doc comments in `ParameterMetaMatched`

Serial-ATA:
part of #557

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lint/src/rules/parameter_meta_matched.rs
```diff
@@ -191,50 +214,101 @@ task say_hello {
...
+                Some(name.to_string())
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    let actual_order: Vec<_> = parameter_meta_map
+        .keys()
+        .filter(|name| decls_map.contains_key(&**name))
```
claymcleod: A declaration documented by both sources always triggers this ordering check, so filter `actual_order` with the same `!has_doc_comments` condition used for `expected_order`. For example, when `first` has both a doc comment and a `parameter_meta` entry followed by a normal `second` entry, `expected_order` is `["second"]` while `actual_order` is `["first", "second"]`; the rule reports an ordering problem and suggests deleting `first` even though the entries already follow declaration order. Mixed sources are meant to be handled by `MixedDocSources`, so both order lists should ignore the same declarations.

crates/wdl-lint/src/rules/parameter_meta_matched.rs
```diff
@@ -191,50 +214,101 @@ task say_hello {
...
-        if !actual_map.contains_key(name) {
+    for (name, (span, decl_node, has_doc_comments)) in &decls_map {
+        if !has_doc_comments && !parameter_meta_map.contains_key(name) {
             diagnostics.exceptable_add(
-                missing_param_meta(parent, name, *span),
-                param_meta.inner(),
+                missing_param_meta(parent, name, *span, suggest_doc_comments),
+                param_meta
+                    .as_ref()
+                    .map_or(*decl_node, |param_meta| param_meta.inner()),
```
claymcleod, line 288: Diagnostics in the no-`parameter_meta` path cannot be suppressed locally, so add the relevant declaration or task/workflow/struct node kinds to `exceptable_nodes`. This branch anchors the diagnostic on `decl_node`, but the rule currently permits exceptions only above a version statement or `parameter_meta` section; neither is an ancestor here, and a nearby `#@ except: ParameterMetaMatched` is rejected as having no effect.

crates/wdl-lint/tests/lints/matching-param-meta/source.wdl
```diff
@@ -98,6 +84,51 @@ task baz {
...
+
+    parameter_meta {
+        third: "`third` also gets a parameter_meta entry"
+        first: "`first` is documented with parameter_meta"
+    }
+
+    command <<<>>>
+}
+
+task corge {
```
claymcleod, line 121: Add a fixture with partial doc comments and no `parameter_meta` section so the new `param_meta == None` behavior is covered. Every new task here still has a `parameter_meta` section, which leaves the early return, declaration-node diagnostic anchor, and `let Some(param_meta) = param_meta else { return; }` branch untested; the added file-wide exception in `unused-doc-comments/source.wdl` suppresses the only existing case that would exercise this path.

crates/wdl-ast/src/lib.rs
```diff
@@ -67,7 +67,7 @@ pub(crate) mod python;
 pub mod v1;
 
 /// An [`AstNode`] that may have documentation comments attached to it.
-pub trait Documented<N: TreeNode>: AstNode<N> {
+pub trait Documented<N: TreeNode> {
```
claymcleod, line 70: Please preserve `Documented`’s `AstNode<N>` supertrait and implement `AstNode<N>` for `Decl<N>` instead. `Expr<N>` already demonstrates how an enum of AST node variants can delegate an `AstNode` implementation, which keeps `Documented`’s existing type contract intact rather than weakening it for one enum.

crates/wdl-lint/src/rules/parameter_meta_matched.rs
```diff
@@ -29,26 +29,49 @@ use crate::TagSet;
...
+
+    let mut diagnostic = Diagnostic::warning(format!(
+        "{context} `{parent}` is missing a {suggestion} for {decl_type} `{missing}`",
         parent = parent.text(),
     ))
     .with_rule(ID)
     .with_label(
-        "this input does not have an entry in the parameter metadata section",
+        format!(
+            "this input does not have {}",
```
claymcleod: Use `decl_type` in this label as well so struct diagnostics consistently call members “fields.” The headline now says `field`, but the regenerated struct fixture still labels the same declaration as “this input.”

crates/wdl-lint/src/rules/parameter_meta_matched.rs
```diff
@@ -29,26 +29,49 @@ use crate::TagSet;
...
+        ),
         span,
-    )
-    .with_fix(format!(
-        "add a `{missing}` key to the `parameter_meta` section with a detailed description of the \
-         input.",
-    ))
+    );
+
+    if !suggest_doc_comments {
```
claymcleod: Add remediation text for the doc-comment branch so both forms of this diagnostic tell the user how to resolve it. The new `corge` warning is the only `ParameterMetaMatched` diagnostic in the fixture without an `= fix:` line; it could suggest adding a `##` comment with a detailed description above the declaration.

claymcleod (dismissed):
Update the rule’s user-facing description, explanation, example, and `RULES.md` entry so they explain that doc comments can satisfy `ParameterMetaMatched`. Also record the `Documented`/`Decl` API change in `crates/wdl-ast/CHANGELOG.md`.

crates/wdl-lint/src/rules/parameter_meta_matched.rs
```diff
@@ -191,50 +222,105 @@ task say_hello {
...
+
+    let actual_order: Vec<_> = parameter_meta_map
+        .keys()
+        .filter(|name| {
+            decls_map
+                .get(&**name)
+                .is_some_and(|(_, _, has_doc_comments)| !has_doc_comments)
+        })
+        .cloned()
+        .collect();
```
adthrasher, line 320: What happens if a user has both a doc comment and parameter meta, e.g.
```wdl
task foo {
    input {
        ## `a` has a doc comment
        String a
        String b
        String c
    }
    parameter_meta {
        b: "b desc"
        a: "a desc, out of place"   # out of order
        c: "c desc"
    }
    ...
}
```

Serial-ATA: nothing, the doc comments always take precedence. that case is meant to be caught by `MixedDocSources` (see https://github.com/stjude-rust-labs/sprocket/issues/557#issuecomment-5530403004). the idea is that `ParameterMetaMatched` allows this, then `MixedDocSources` flags it, so you either:

1. remove the `parameter_meta` section entirely and all is well
2. remove the doc comment, then on a re-run you get the out of order warning

a-frantz (approved):

PR #1185, opened 2026-09-07
feat(wdl-lint): support doc comments in `MetaSections`

Serial-ATA:
part of #557

For external contributors:

For all contributors:

      is implemented.


crates/wdl-ast/src/lib.rs
```diff
@@ -78,13 +78,29 @@ pub trait Documented<N: TreeNode>: AstNode<N> {
 }
 
 /// Shared doc comment extraction logic.
+#[allow(clippy::needless_bool)] // For clarity
 pub fn doc_comments<N: TreeNode>(
-    preceding_trivia: impl IntoIterator<Item = N::Token>,
+    preceding_trivia: impl DoubleEndedIterator<Item = N::Token>,
+    allow_floating: bool,
```
a-frantz, line 84: I think this should be consistent instead of decided ad hoc for each node type. I'm not sure which is preferable (always include "floating" doc comments or never), but I think we should pick one and apply that everywhere.

crates/wdl-lint/src/rules/meta_sections.rs
a-frantz, line 315: OOS for this PR, but should we start linting <1.2 structs for doc comment presence? They can't have meta sections within spec, but no reason they can't have doc comments.

crates/wdl-ast/src/lib.rs
Serial-ATA: floating comments are only allowed for preambles, which I think makes sense:

```wdl
## This is a preamble

version 1.3

## This is a comment for `foo`
task foo {}
```

In that example, the comment on `foo` is clearly attached to the task definition. The preamble, though, is logically attached to the document and not the `version` statement, despite us internally treating the `version` statement as the documented node.

crates/wdl-lint/src/rules/meta_sections.rs
Serial-ATA: sure, I'll make an issue after this is merged

crates/wdl-lint/src/rules/meta_sections.rs
```diff
@@ -181,6 +189,58 @@ task say_hello {
...
+        let self_documented = node.first_token().is_some_and(|t| {
+            doc_comments::<SyntaxNode>(t.preceding_trivia(), false)
+                .next()
+                .is_some()
+        });
+
+        let inputs_present = inputs.is_some();
+        let inputs_documented = inputs.is_some_and(|i| {
+            i.declarations()
+                .any(|d| d.doc_comments().is_some_and(|c| !c.is_empty()))
```
claymcleod, line 208: Please require all inputs to have doc comments here, or update `ParameterMetaMatched` in this PR, because `.any()` lets one documented input suppress the `parameter_meta` requirement for every undocumented input. `ParameterMetaMatched` explicitly does nothing when the section is absent, so a task with ten inputs and one `##` comment currently passes both rules.

crates/wdl-lint/src/rules/meta_sections.rs
```diff
@@ -284,24 +317,14 @@ impl Visitor for MetaSectionsRule {
...
-                missing_section(def.name(), Section::ParameterMeta, Context::Struct),
-                def.inner(),
-                &self.exceptable_nodes(),
-            );
-        }
+        self.check_meta_sections(
+            diagnostics,
+            def.name(),
+            def.inner(),
+            None,
```
claymcleod: Please pass the struct members through the same doc-comment check instead of `None`; this makes a fully doc-commented WDL 1.2 struct still require `parameter_meta`. `UnboundDecl::doc_comments` already supports members of `StructDefinitionNode`, and `wdl-doc` gives those comments precedence over `parameter_meta`, so the diagnostic currently asks users to add documentation that rendering will ignore.

crates/wdl-lint/src/rules/meta_sections.rs
```diff
@@ -181,6 +189,58 @@ task say_hello {
...
+                .is_some()
+        });
+
+        let inputs_present = inputs.is_some();
+        let inputs_documented = inputs.is_some_and(|i| {
+            i.declarations()
+                .any(|d| d.doc_comments().is_some_and(|c| !c.is_empty()))
+        });
+
+        let needs_meta = meta.is_none() && !self_documented;
```
claymcleod, line 211: Please update `description()`, `explanation()`, and the generated `RULES.md` entry to describe doc comments as an accepted alternative. They still say tasks and workflows require `meta` and `parameter_meta` sections, so `sprocket explain MetaSections` now contradicts this new behavior.

crates/wdl-ast/src/lib.rs
```diff
@@ -67,7 +67,7 @@ pub(crate) mod python;
 pub mod v1;
 
 /// An [`AstNode`] that may have documentation comments attached to it.
-pub trait Documented<N: TreeNode>: AstNode<N> {
+pub trait Documented<N: TreeNode> {
```
claymcleod: Could we implement `AstNode` for `Decl` instead of removing this supertrait? `Decl` already provides the required `can_cast`, `cast`, and `inner` methods, and `Expr` follows the same enum-over-node-types pattern with an `AstNode` implementation. Keeping the bound would also let `check_meta_sections` use `Documented` directly instead of rediscovering comments from a raw `SyntaxNode`.

crates/wdl-ast/src/lib.rs
```diff
@@ -78,13 +78,29 @@ pub trait Documented<N: TreeNode>: AstNode<N> {
 }
 
 /// Shared doc comment extraction logic.
+#[allow(clippy::needless_bool)] // For clarity
 pub fn doc_comments<N: TreeNode>(
```
claymcleod, line 106: Please add changelog entries for the public API changes in `wdl-ast` and `wdl-grammar`; only the `wdl-lint` changelog currently mentions this PR. `doc_comments` gains a required parameter and iterator bound, `Documented` changes its bound, and `SyntaxTokenExt::preceding_trivia` changes its return contract; `wdl-grammar` already records a comparable return-type change under `Unreleased`.

crates/wdl-ast/src/lib.rs
```diff
@@ -78,13 +78,29 @@ pub trait Documented<N: TreeNode>: AstNode<N> {
...
+            if token.kind() == SyntaxKind::Comment {
+                return true;
+            }
+
+            if token.kind() != SyntaxKind::Whitespace {
+                return false;
+            }
+
+            let lines = token.text().chars().filter(|c| *c == '\n').count();
+            if lines > 1 && !allow_floating {
```
claymcleod, line 122: Please make `UnusedDocComments` use this same blank-line boundary. A `##` block separated from a definition by a blank line is no longer returned by `doc_comments`, but `search_siblings_for_doc_comment_target` skips all trivia and still treats that definition as a valid target, so the detached documentation is never reported as unused.

crates/wdl-lint/tests/lints/missing-meta/source.wdl
```diff
@@ -8,11 +8,11 @@ task test {
     input {}
 
     command <<<>>>
+}
 
-    output {}
-
-    #@ except: ExpectedRuntimeKeys
-    runtime {}
+## This doc comment suppresses the lint
```
claymcleod, line 13: Please add negative regression cases for the new boundaries, not only examples where doc comments suppress diagnostics. In particular, a blank-line-separated `##` block must not satisfy `meta`, and one documented input among undocumented inputs must not make the task pass; neither case is covered by the current `MetaSections` fixtures.

claymcleod (dismissed):

crates/wdl-lint/src/rules/meta_sections.rs
Serial-ATA: https://github.com/stjude-rust-labs/sprocket/pull/1184

crates/wdl-ast/src/lib.rs
Serial-ATA: that can be a separate PR

crates/wdl-ast/src/lib.rs
a-frantz: That makes sense to me, but can we update the `allow_floating` name to something like `preamble` and update the doc comment to make that clearer?

crates/wdl-lint/tests/lints/missing-meta/source.wdl
Serial-ATA: > a blank-line-separated ## block must not satisfy meta

That's better suited for an update to `UnusedDocComments`: https://github.com/stjude-rust-labs/sprocket/pull/1185#discussion_r3991804391

> one documented input among undocumented inputs must not make the task pass

The case of undocumented inputs is covered by `ParameterMetaMatched`: https://github.com/stjude-rust-labs/sprocket/issues/557#issuecomment-5530403004

crates/wdl-ast/src/v1/decls.rs
```diff
@@ -811,8 +811,11 @@ impl Documented<SyntaxNode> for UnboundDecl<SyntaxNode> {
         }
 
         Some(
-            crate::doc_comments::<SyntaxNode>(self.inner().first_token()?.preceding_trivia())
-                .collect(),
+            crate::doc_comments::<SyntaxNode>(
+                self.inner().first_token()?.preceding_trivia(),
```
a-frantz, line 818: this is collecting _all_ trivia, not just doc comments

crates/wdl-ast/src/v1/decls.rs
```diff
@@ -876,8 +879,11 @@ impl Documented<SyntaxNode> for BoundDecl<SyntaxNode> {
...
```
a-frantz, line 887: same here, collecting all trivia instead of only doc comments

crates/wdl-ast/src/v1/decls.rs
```diff
@@ -811,8 +811,11 @@ impl Documented<SyntaxNode> for UnboundDecl<SyntaxNode> {
         }
```
a-frantz, line 814: this isn't currently touched by the PR, but it looks incorrect?

crates/wdl-ast/src/v1/enum.rs
```diff
@@ -136,7 +136,10 @@ impl<N: TreeNode> AstNode<N> for EnumDefinition<N> {
 
 impl Documented<SyntaxNode> for EnumDefinition<SyntaxNode> {
     fn doc_comments(&self) -> Option<Vec<Comment<<SyntaxNode as TreeNode>::Token>>> {
-        Some(crate::doc_comments::<SyntaxNode>(self.keyword().inner().preceding_trivia()).collect())
+        Some(
+            crate::doc_comments::<SyntaxNode>(self.keyword().inner().preceding_trivia(), false)
```
a-frantz, line 140: same about all trivia

crates/wdl-ast/src/v1/enum.rs
```diff
@@ -210,7 +213,10 @@ impl<N: TreeNode> AstNode<N> for EnumChoice<N> {
 
 impl Documented<SyntaxNode> for EnumChoice<SyntaxNode> {
     fn doc_comments(&self) -> Option<Vec<Comment<<SyntaxNode as TreeNode>::Token>>> {
-        Some(crate::doc_comments::<SyntaxNode>(self.name().inner().preceding_trivia()).collect())
+        Some(
+            crate::doc_comments::<SyntaxNode>(self.name().inner().preceding_trivia(), false)
```
a-frantz, line 217: same

crates/wdl-ast/src/lib.rs
a-frantz: > that can be a separate PR

agree, but can you make sure to track that with an issue so we don't forget?

a-frantz (changes requested):
I haven't verified with a local test, but I believe this will have false negatives for non-doc-comment preceding trivia. Can you please verify with tests?

I also noticed what looks like an unrelated bug in the `Documented` trait; can you verify that it works correctly for non-struct and non-input nodes?

crates/wdl-ast/src/v1/decls.rs
Serial-ATA: not sure what you mean? `doc_comments` expects all trivia, it does filtering itself

crates/wdl-ast/src/v1/decls.rs
Serial-ATA: can unbound decls be used anywhere else? `doc_comments` is supposed to return `None` for nodes that don't support them: https://docs.rs/wdl-ast/latest/wdl_ast/trait.Documented.html#tymethod.doc_comments

crates/wdl-ast/src/v1/decls.rs
a-frantz: Ah, I totally misread this 🤦‍♀️ sorry that's my mistake. But can we still add the test case that non-doc comments aren't returned erroneously? If that test was here I wouldn't have made the reading mistake resulting in these comments.

crates/wdl-ast/src/v1/decls.rs
a-frantz: Again, my mistake 🤦‍♀️ I should have double checked the documentation before making this comment. This all looks correct to me now that you corrected my misunderstanding, apologies for being too quick with my review!

crates/wdl-ast/src/lib.rs
Serial-ATA: https://github.com/stjude-rust-labs/sprocket/issues/1210

a-frantz (approved):
Looks great!

PR #1189, opened 2026-09-09
fix(wdl-lsp): ensure document analysis

Serial-ATA:
Depending on the order of requests, documents may not be parsed/analyzed in time. That can cause clients to get empty responses to important requests like `textDocument/semanticTokens`.

Split this out of a larger LSP PR awhile ago and forgot about it

For external contributors:

For all contributors:

      is implemented.

### DIFF
--- a/crates/wdl-analysis/src/queue.rs
+++ b/crates/wdl-analysis/src/queue.rs
@@ -87,9 +87,9 @@ pub enum Request<Context> {
     /// A request to analyze documents.
     Analyze(AnalyzeRequest<Context>),
     /// A request to get all callers of a symbol.
-    CallHierarchy(CallHierarchyRequest),
+    CallHierarchy(CallHierarchyRequest<Context>),
     /// A request to get all code lenses in a document.
-    CodeLens(CodeLensRequest),
+    CodeLens(CodeLensRequest<Context>),
     /// A request to unroot documents in the graph.
     UnrootDocuments(UnrootDocumentsRequest),
     /// A request to delete documents from the graph.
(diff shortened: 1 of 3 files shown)


crates/wdl-analysis/src/queue.rs
```diff
@@ -914,6 +966,11 @@ where
                         char = position.character
                     );
 
+                    if !self.ensure_analyzed(document.clone(), context) {
```
claymcleod, line 969: This guard only analyzes the requested document and its dependency closure, so analyze the definition's dependent search scope before running rename. On a cold workspace, importer documents may still be unanalyzed; `handlers::find_all_references` silently skips nodes without `document()`, which lets this return a partial `WorkspaceEdit` and leave call sites unchanged. A full workspace analysis or targeted transitive-dependent analysis would prevent an incomplete rename.

crates/wdl-analysis/src/queue.rs
```diff
@@ -1179,6 +1236,79 @@ where
...
+    /// the document either doesn't exist or fails analysis.
+    fn ensure_analyzed(&self, document: Url, context: Context) -> bool {
+        {
+            let graph = self.graph.read();
+
+            match graph.get_index(&document) {
+                Some(index) => {
+                    let node = graph.get(index);
+                    if node.document().is_some() {
+                        // Already analyzed
```
claymcleod, line 1287: `node.document().is_some()` does not prove the analysis is current, so include dependency freshness in this fast path or run analysis before returning. If document A imports B and B receives an incremental change, only B's `document` is cleared; A stays populated until `analyze` parses B and propagates invalidation to A. A request in A can therefore serve stale completion, hover, tokens, or hints, and this regresses completion because its old path always called `analyze`.

crates/wdl-analysis/src/queue.rs
```diff
@@ -1179,6 +1236,79 @@ where
...
+                }
+                None => {
+                    debug!("document `{document}` not found in graph");
+                    return false;
+                }
+            }
+        }
+
+        trace!("document not yet analyzed");
+        match self.analyze(Some(document), context, None) {
```
claymcleod, line 1299: Passing `None` disables both cancellation and progress, so either carry a real cancellation/progress signal through on-demand requests or remove `Context` from the new public APIs. On a cold document, this can run parse, import download, and analysis work on the serial queue after the client no longer needs the response, while every new LSP call supplies the no-op `ProgressToken::default()`. The added `Context` parameters are therefore source-breaking without producing progress.

crates/wdl-analysis/src/queue.rs
```diff
@@ -1179,6 +1236,79 @@ where
...
+            Cancelable::Completed(Ok(_)) => true,
```
claymcleod, line 1300: `Ok(_)` does not guarantee this document was analyzed, so re-read the node and require `document().is_some()` before returning `true`. `analyze(Some(document))` returns `Ok(Vec::new())` for a graph node that is not rooted, and it also completes after an individual node analysis panic leaves `document` unset. A follow-up handler then runs against missing analysis data and returns `None`, preserving the empty-result behavior this helper is meant to prevent.

crates/wdl-lsp/src/server.rs
```diff
@@ -1263,7 +1270,11 @@ impl<S: 'static> Server<S> {
         let result = state
             .config
             .analyzer
-            .inlay_hints(params.text_document.uri, params.range)
+            .inlay_hints(
+                ProgressToken::default(),
```
claymcleod, line 1274: Remove the full-workspace `analyze` call above and rely on the new per-document readiness check behind this call. As written, every inlay-hint request analyzes all rooted documents first, so `ensure_analyzed` immediately takes its fast path and the new guard adds no protection here. Keeping both paths preserves unnecessary workspace-wide work for a request editors send frequently.

claymcleod (dismissed):
`workspace/symbol` still reads the graph without ensuring analysis, so it can return an empty list on a cold workspace; please include it in this fix. Please also add regression tests for cold requests, dependency changes, and cross-document rename, and record the user-visible fix—plus any retained public API break—in the Unreleased changelogs.

Serial-ATA:
> Please also add regression tests for cold requests, dependency changes, and cross-document rename

This doesn't need tests of its own, `TestContext::initialize()` was doing an implicit `workspace/diagnostic` request to make up for this. I removed that request to prove the tests still work as expected with these new guards

peterhuene (approved):

PR #1203, opened 2026-09-15
fix(format): preserve section ordering of tasks and workflows

### LINKED ISSUE #845: wdl-format: declaration position isn't preserved
**Describe the bug**
Passing the following through `wdl-format`:

```wdl
version 1.2

task greet {
    String greeting = "Hello"

    command <<<
        echo "~{greeting}, ~{name}!"
    >>>

    String name = "World"
}
```

results in:

```wdl
version 1.2

task greet {
    String greeting = "Hello"

    String name = "World"

    command <<<
        echo "~{greeting}, ~{name}!"
    >>>
}
```

**Expected behavior**
`name` should stay under the `command` section. The declaration order doesn't matter here for evaluation, but it feels like `wdl-format` is overstepping here by shifting things around like this.

**Additional context**
#573

a-frantz:

closes #845 

For external contributors:

For all contributors:

      is implemented.


crates/wdl-lint/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Removed
 
+* Removed `SectionOrder` lint rule. Use `format check` with `reorder_sections = true` to satisfy any ordering checks ([#1203](https://github.com/stjude-rust-labs/sprocket/pull/1203)).
```
adthrasher, line 16: This should probably note the behavior change explicitly.

crates/wdl-format/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
 
+* Formatting no longer forces users to adopt Sprocket's opinionated section ordering of tasks and workflows. Set `reorder_sections = true` to retain prior behavior ([#1203](https://github.com/stjude-rust-labs/sprocket/pull/1203)).
```
adthrasher, line 17: This probably needs something to indicate the breaking change.

adthrasher (approved):
This is a nice addition! We need to make sure sprocket.bio gets updated as well.

Two small comments on the CHANGELOGs as I think we want to be semi-loud about this change.

crates/wdl-lint/CHANGELOG.md
a-frantz: This is in the `wdl-lint` changelog, so I don't think commenting on `format` behavior is right

crates/wdl-format/CHANGELOG.md
a-frantz: Like what? We don't include a `# Breaking change` section in the changelog like some projects do.
At least in how I read it, `<do thing> to retain prior behavior` is a clear indication that _the default behavior_ has changed (and is thus a breaking change)

crates/wdl-lint/CHANGELOG.md
adthrasher: I didn't necessarily mean we should comment on `format`'s behavior, but rather how to preserve the `lint` functionality.

I don't recall the default before, but I assume SectionOrder was on by default. If so, I'd update this to something like:

```suggestion
* Removed `SectionOrder` lint rule. To preserve prior functionality, use `format check` with `reorder_sections = true` to satisfy any ordering checks ([#1203](https://github.com/stjude-rust-labs/sprocket/pull/1203)).
```

I'm OK with other wording, but I think it's important that this clearly states that the behavior has changed and the user now needs a second check to get the prior behavior.

adthrasher: Essentially if a `lint` user is reading the changelog prior to updating, I want them to understand clearly that they now need a second check with the `format` command, otherwise their results will change.

crates/wdl-format/CHANGELOG.md
adthrasher: I'm not 100% on this. Maybe just shifting this from `Fixed` to `Changed`.

adthrasher: > Like what? We don't include a `# Breaking change` section in the changelog like some projects do. At least in how I read it, `<do thing> to retain prior behavior` is a clear indication that _the default behavior_ has changed (and is thus a breaking change)

I also now realize that I suggested very similar language in my comment on the `wdl-lint` changelog. Maybe just shift the order of the second sentence.

```suggestion
* Formatting no longer forces users to adopt Sprocket's opinionated section ordering of tasks and workflows. To retain prior behavior , set `reorder_sections = true` ([#1203](https://github.com/stjude-rust-labs/sprocket/pull/1203)).
```

crates/wdl-format/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
```
adthrasher:
```suggestion
#### Changed
```

PR #1222, opened 2026-09-17
feat(wdl-lint): support doc comments in `MetaDescription`

Serial-ATA:
part of #557

For external contributors:

For all contributors:

      is implemented.


a-frantz (commented):
Code all looks fine, but I think we want a more radical change here. Why are we requiring an _empty_ meta section? Does having a doc comment eliminate the need for a meta section entirely? Is there a consequence to that for other rules?

Serial-ATA:
@a-frantz 

> Does having a doc comment eliminate the need for a meta section entirely

yeah, the tests only have empty `meta` sections since this lint only fires if the task/workflow/struct actually _contains_ one. we're just testing that the doc comment actually suppresses it.

#1185 made it so `MetaSections` also doesn't fire if you have a doc comment

the flow before was:

```wdl
workflow foo {}
```

```shell
$ sprocket lint
note[MetaSections]: workflow `foo` is missing a `meta` section                                                                                                                                                                                                                                                      
  ┌─ foo.wdl:3:10
  │
3 │ workflow foo {
  │          ^^^ this workflow is missing a `meta` section
  │
  = fix: add the missing section
```

Ok I need to add a `meta` section...

```wdl
workflow foo {
    meta {}
}
```

```shell
$ sprocket lint
note[MetaDescription]: workflow `foo` is missing a description key                                                                                                                                                                                                                                                  
  ┌─ foo.wdl:4:5
  │
4 │     meta {}
  │     ^^^^
  │
  = fix: add a `description` key to the meta section
```

Ok my `meta` section needs a description...

```wdl
workflow foo {
    meta {
        description: "does foo and such"
    }
}
```

```shell
$ sprocket lint
```

But if you just use doc comments, neither fires:

```wdl
## does foo and such
workflow foo {}
```

```shell
$ sprocket lint
```

a-frantz:
nice, you are ahead of me then!

I still wonder if this is right though 🤔 The help text reads as confused to me:
```
        "Ensures that items with a `meta` section either contain a `description` key or a doc \
         comment."
```
why should something with a meta section have a doc comment? I think maybe the right behavior is: a `meta` section without a `description` guarantees firing this rule, but omitting a `meta` section entirely is fine if doc comments are present. Maybe the rules need to be reshaped a bit and consolidated.

The double edit workflow you describe is _bad_, but I'm not sure this impl is the right fix.

Granted, you've spent more time with this part of the codebase recently, so I might be missing something obvious.

Serial-ATA:
> why should something with a meta section have a doc comment?

`meta` can be useful outside of `description`/`help` keys. If it helps, I plan to have `MixedDocSources` behave like this:

```wdl
## Does foo and such
workflow foo {
    input {
        ## Some input
        String bar
        ## Some other input
        String baz
    }

    meta {
        description: "Does foo and such"
        author: "Serial-ATA"
    }

    parameter_meta {
        bar: {
            important: true
        }
        baz: "Some other input"
    }
}
```

`foo` has both a doc comment and a `meta.description`. Those would be flagged, while `meta.author` is perfectly fine to keep.

`bar` and `baz` both have doc comments and `parameter_meta` entries, but only `baz` would trigger `MixedDocSources`, they're both just text descriptions. `bar` has some other structured meta, so it passes

a-frantz (approved):
I'm trying to see the forest through the trees on this one. @Serial-ATA what else is going to be part of this (integrating doc comments in `lint`)? I think it might be helpful for you to write up your plan for how all the rules and pieces fit together into a cohesive UX. I'm sure whatever you write could also be repurposed as sprocket.bio documentation, but I'd also find it helpful for reviewing these component PRs.
Maybe you could write a mini RFC on https://github.com/stjude-rust-labs/rfcs ?

That said, you've got a few PRs already merged that interact with this, so I'm approving with the rationale that the train is already moving 😅 However I don't have a clear view of the destination, so to speak.

Serial-ATA:
@a-frantz these PRs are just implementing these behaviors: https://github.com/stjude-rust-labs/sprocket/issues/557#issuecomment-5530403004

this process hasn't been fundamentally _changing_ any lints, it's just fitting doc comments into them. the problem is having these two sources of docs, with one of them being specific to `sprocket`.

I've kept the existing lints (with their `meta`/`parameter_meta` suggestions) to, by default, push people to use the spec-defined doc sources to keep things portable. it's just that now, any project that only cares about `sprocket` and sticks strictly to doc comments won't get tripped up on these `meta` lints.

> write up your plan for how all the rules and pieces fit together into a cohesive UX

I think the best way to describe it is: `wdl-doc` already treats doc comments and `meta.{description,help}` identically, and this just makes `wdl-lint` do the same. you can always sub one out for the other and nothing breaks, and whenever they're mixed, doc comments get priority

a-frantz:
> I've kept the existing lints (with their `meta`/`parameter_meta` suggestions) to, by default, push people to use the spec-defined doc sources to keep things portable. it's just that now, any project that only cares about `sprocket` and sticks strictly to doc comments won't get tripped up on these `meta` lints.

this is what I missing. I was unclear exactly how hard we were "selling" doc comments, but what you've outlined here clarifies everything for me, thanks! :D

Serial-ATA:
yeah, I don't know if we're necessarily "selling" doc comments anymore since `meta`/`parameter_meta` are here to stay. they kinda just exist as another (easier) option if you happen to exclusively use `sprocket`. that does unfortunately make doc comments something that users have to discover through the docs, since we only ever suggest them if you *already* use them

a-frantz:
Yea, I wouldn't call the issue totally settled, but it also doesn't seem to be making any more progress. No one is driving the issue forward at the moment. What you've implemented in these PRs feel appropriate for the current state of the WDL world.

----

It seemed to me that there was general excitement about the prospect, but that got dampened by the acknowledgement that the structure of meta sections has sunk in deep around the ecosystem. The path forward I see is to introduce that structure into doc comments, via some annotation system, but there's no consensus on what that system would be.

Someone has to own the issue and implement something (could be one of us!). I'm concerned about further fracturing if the "annotation system" isn't satisfactory enough to all parties to get adopted. Something for the future to deal with though, I don't see anyone leading the charge in the near term.

PR #1223, opened 2026-09-17
chore: make `report_mode` a global CL option

a-frantz:
this removes the `report_mode` from _a bunch_ of subcommands that all repeated the logic (sometimes erroneously)

as the diff will show, this also _adds_ the `-m, --report_mode` flag to some commands, even though it will be ignored by that command. That seems like a fine trade to me rather than tracking this arg a dozen different times (and inconsistently applying its value to the config)

For external contributors:

For all contributors:

      is implemented.


Serial-ATA (approved):
> this also adds the -m, --report_mode flag to some commands, even though it will be ignored by that command

I think that's fine, no different than `--color` which doesn't apply to every subcommand

PR #1225, opened 2026-09-17
feat!: rename `config init` to `config default`

a-frantz:

breaking obviously. This was mentioned a few times both on Slack and (I believe) in a comment thread or two on prior PRs (apologies for not digging up the links).

For posterity, the rationale is that this command isn't actually "initializing" anything. Sprocket is happy to run without a config file, and the contents here are not needed for anything. However being able to print out the _default_ configuration is useful for a few different workflows (e.g. seeing how your `config resolve` diffs with the default)

For external contributors:

For all contributors:

      is implemented.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* renamed the `sprocket config init` command to `sprocket config default` ([#1225](https://github.com/stjude-rust-labs/sprocket/pull/1225)).
```
Serial-ATA:
```suggestion
* Renamed the `sprocket config init` command to `sprocket config default` ([#1225](https://github.com/stjude-rust-labs/sprocket/pull/1225)).
```

src/commands/config.rs
```diff
@@ -35,7 +35,7 @@ pub enum ConfigSubcommand {
     Schema,
 
     /// Generates a default configuration file.
```
Serial-ATA, line 37:
```suggestion
    /// Generates a full default configuration file.
```

Maybe? Don't know if this necessary to specify, feel free to ignore

Serial-ATA (approved):

src/commands/config.rs
a-frantz: Not sure this would be 100% true. I think some of the engine internal config options don't serialize out under this command (that's an unverified hunch from prior experience poking around this command, could be wrong)

